PR #4, opened 2016-05-25
Py3: use bytes for Message payload and key

edenhill:


README.md
```diff
@@ -1,40 +1,94 @@
 Confluent's Apache Kafka client for Python
 ==========================================
 
+Confluent's Kafka client for Python wraps the librdkafka C library, providing
+full Kafka protocol support at great performance and reliability.
```
ewencp: support at great -> support with great

README.md
```diff
@@ -1,40 +1,94 @@
...
 
-Prerequisites
-===============
+The Python bindings provides a high-level Producer and Consumer with support
+for the balanced consumer groups of Apache Kafka 0.9.
 
-    librdkafka >=0.9.1 (or master>=2016-04-13)
-    py.test  (pip install pytest)
+See the [API documentation](http://docs.confluent.io/3.0.0/clients/confluent-kafka-python/index.html) for more info.
```
ewencp: If you want this link versioned, that's fine. You can also use http://docs.confluent.io/current/clients/confluent-kafka-python/index.html if you want something that doesn't need updating w/ each version

ewencp:
@edenhill LGTM. I only managed to test the integration test in python3. Trying to run the tests in a py3 virtualenv was running some completely different set of tests that I have no idea where they came from. What's your process for doing testing under both py2 and py3?

edenhill:
So far I've manually switched between two virtualenvs (2.7 and 3.5) and then doing:
```
 py.test
 examples/integration_test.py myrealbroker
```

ewencp:
ok, i must have just screwed something up when setting up the py3 virtualenv

PR #5, opened 2016-05-26
Removed hard coded c extentions lib/include paths

jofusa:
With packaging systems like conda users need to be able to specify the location of librdkafka. 

If users have compiled librdkafka to a non standard location you can build with:

    CPLUS_INCLUDE_PATH=/path/to/include LIBRARY_PATH=/path/to/lib $PYTHON setup.py install

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,9 @@ Build
 
     python setup.by build
 
+If librdkafka is installed in a non-standard location provide the include and library directories with:
 
+    CPLUS_INCLUDE_PATH=/path/to/include LIBRARY_PATH=/path/to/lib python setup.py ...
 
 
 Tests
--- a/setup.py
+++ b/setup.py
@@ -5,9 +5,7 @@ from distutils.core import Extension
 
 
 module = Extension('confluent_kafka.cimpl',
-                    include_dirs = ['/usr/local/include'],
                     libraries= ['rdkafka'],
-                    library_dirs = ['/usr/local/lib'],
                     sources=['confluent_kafka/src/confluent_kafka.c',
                              'confluent_kafka/src/Producer.c',
                              'confluent_kafka/src/Consumer.c'])


jofusa:
[clabot:check]

ewencp:
@edenhill This seems ok to me. Any reason we included them explicitly?

edenhill:
The main reason is that manual installs of librdkafka will be placed in /usr/local, so defaulting to that location allows seemless building both for package-installed librdkafka (e, g, deb or rpm, which resides in /usr and is automatically picked up by compiler and linker) as well as manual installers of librdkafka.

@jofusa Are these default paths conflicting with conda?

jofusa:
It does. Conda places conda [version](https://anaconda.org/ActivisionGameScience/librdkafka) of librdkafka into you personal env directory. Which is located in your home directory. The build script looks like:

    CPLUS_INCLUDE_PATH=$PREFIX/include LIBRARY_PATH=$PREFIX/lib $PYTHON setup.py install

With $PREFIX being set by conda.

jofusa:
I should add that with these two lines removed I have this package built with conda.

ewencp:
@edenhill On Mac this works fine w/ this patch -- it still picks up the `/usr/local` version.

jofusa:
thoughts on this? This is forcing me to build from a private fork instead of the release tar.

edenhill:
@ewencp That wont be true on Linux though.

Lets go ahead with removing the explicit include and lib dirs, but add a line like this to the build section of the README:
> If librdkafka is installed in a non-standard location provide the include and library directories by `CPLUS_INCLUDE_PATH=/path/to/include LIBRARY_PATH=/path/to/lib python setup.py ...`

jofusa:
Sounds good will add to the pr.

jofusa:
added docs

edenhill:
Looks good to me.
@ewencp ?

ewencp:
Yup, LGTM

edenhill:
Thanks @jofusa!

Issue #14, opened 2016-07-05
Expose error_cb to Python


paulcavallaro:
Proposed Enhancement:

For both the Consumer and Producer, it seems necessary to expose the error_cb in the Python API so that Consumer and Producer code can deal with errors, especially w.r.t. connections with Kafka.

For example, consider the Consumer pseudo code:

```
while True:
    msg = consumer.poll(1.0)
    if not msg:
        logger.info('No messages found in polling window')
        continue
    if msg.error():
        logger.exception('Received error', error=msg.error())
        raise confluent_kafka.KafkaException(msg.error())
```

This code will run forever, even if the consumer loses connection to the brokers, which isn't always what you want. In my case, I'd like to fail if the error has persisted for the past 5 minutes.

It seems like it wouldn't be too much work to do something similar to what is done for the `on_revoke` and `on_assign` callbacks, but not sure about the intricacies of the librdkafka's error_cb.

This is blocking us from adopting confluent-kafka-python more broadly (or at all).

Thanks for your time.

paulcavallaro:
@edenhill Awesome, thanks for such a fast turnaround on this!

edenhill:
This is now merged.

Thanks for your bug report

PR #15, opened 2016-07-06
Added error callback for propagating generic errors from librdkafka to app

edenhill:


docs/index.rst
```diff
@@ -78,6 +78,9 @@ The Python bindings also provide some additional configuration properties:
 * ``default.topic.config``: value is a dict of topic-level configuration
   properties that are applied to all used topics for the instance.
 
+* ``error_cb``: Callback for generic/global error events. This callback is served by
```
ewencp: Should we be including the argument lists for these callbacks?

ewencp:
@edenhill LGTM, but it has very limited testing, just that the setup works. For example, we don't actually evaluate that the expected set of arguments matches. Anything we can do after creating the producer/consumer in the unit tests to evaluate that invoking the callback actually works?

Also, should we start tracking changelogs like we do for other projects (e.g. http://docs.confluent.io/3.0.0/schema-registry/docs/changelog.html). We normally update these just before release since its easier than updating with each patch, but given that clients need to release independently and we're doing a new tag + release for this asap....

edenhill:
Updated with unit test and callback arg documentation.

confluent_kafka/src/Consumer.c
```diff
@@ -55,12 +58,15 @@ static void Consumer_dealloc (Consumer *self) {
...
-	if (self->on_assign)
-		Py_VISIT(self->on_assign);
-	if (self->on_revoke)
-		Py_VISIT(self->on_revoke);
+static int Consumer_traverse (Handle *self,
+			      visitproc visit, void *arg) {
+	if (self->u.Consumer.on_assign)
+		Py_VISIT(self->u.Consumer.on_assign);
+	if (self->u.Consumer.on_revoke)
+		Py_VISIT(self->u.Consumer.on_revoke);
```
ewencp: Are we missing visiting `on_commit`?

ewencp:
@edenhill One last question, but LGTM otherwise.

paulcavallaro:
Wanted to ping on this pull request because isssue #14 is blocking our adoption of this library. Thanks,

-Paul

ewencp:
@paulcavallaro Should see some follow up soon, sorry for the delay. My last comment was right before @edenhill went on vacation, but he is returning imminently!

ewencp:
@edenhill I'm seeing this:

```
$ examples/integration_test.py localhost:9092
Using confluent_kafka module version 0.9.1 (0x90100)
Using librdkafka version 0.9.1 (0x901ff)
('==============================', 'Verifying Producer', '==============================')
  File "examples/integration_test.py", line 366, in <module>
    verify_producer()
  File "examples/integration_test.py", line 86, in verify_producer
    p = confluent_kafka.Producer(**conf)
cimpl.KafkaException: KafkaError{code=_INVALID_ARG,val=-186,str="Property "error_cb" must be set through dedicated .._set_..() function"}
```

ewencp:
@edenhill Ok, nm, must have had something stale somewhere. Reran in a virtualenv and its fine.

edenhill:
That looks like you used an old confluent_kafka python module version without error_cb support (thus the fallthru to librdkafka)

PR #21, opened 2016-07-25
using async in names is deprecated on python 3.6

johnnoone:


ewencp:
@johnnoone Good catch, we would like to avoid abusing the keyword if possible. @edenhill I think the main concern here is probably compatibility. Are we comfortable breaking the existing API given how new this is? Or should we make the patch support both `async` and `asynchronous` for now, deprecate `async`, and remove it in a later release?

edenhill:
This is still a young project so I vote for not being backwards compatible, let's go with asynchronous in the next release.

Is asynchronous user friendly enough, or should it be something easier to spell, like do_async?

ewencp:
@edenhill Doesn't look like the PEP suggests a specific alternative. Maybe check a couple of other libraries that have async options to see if there is a de facto standard?

edenhill:
Sort of missed the train with this one, so we'll add `asynchronous=bool` kwarg but keep the old `async=bool` one for backwards compatibility (possibly with a warning), e.g. something like:
```
 def foo (.., asynchronous=False, async=None):
    if async is not None:
         warning('Use asynchronous instead, kthx')
         asynchronous=async
...
```

ah-:
I think it's time to implement this now, since async has started to error on Python 3.7 nightly:
```
E       self.consumer.commit(offsets=offsets_to_commit, async=False)
E                                                           ^
E   SyntaxError: invalid syntax
```

Renaming async to asynchronous sounds like a decent solution.

edenhill:
@ah- I think we should provide an "async" alias for <Py3.7 for a while to not break existing applications. What do you think?

We have an upcoming 1.0.0 release bump coming up in a couple of months at which time we can break the API.

ah-:
Yes, keeping backwards compatibility would be great. Looking at https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/src/Consumer.c#L378 it should be as trivial as adding "asynchronous" to the list of kwargs and then having ```async = PyObject_IsTrue(async_o) || PyObject_IsTrue(asynchronous_o)```.

We can probably even keep the old `async` afterwrads, as far as I can tell the only thing that breaks is writing `async=False`, you could still pass kwargs manually setting {'async': True}.

renstrom:
Something like this ought to do it (not very pretty though!):

```python
def foo(..., asynchronous=None, **kwargs):
    asynch = asynchronous or kwargs.get('async')
```

edenhill:
@renstrom That will allow any (unknown) kw arg to be set though. Say Hi to Löfling from me :)

A PR is on the way.

edenhill:
Updated PR in https://github.com/confluentinc/confluent-kafka-python/pull/319

PR #27, opened 2016-08-08
Thread safety fixes

edenhill:


tests/test_threads.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+
+from confluent_kafka import Producer, KafkaError, KafkaException
+import threading, time, queue
```
ewencp: Hmm, `queue` is the py3 name, whereas `Queue` is used in py2. Currently this fails on py2 for me.

tests/test_threads.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class IntendedException (Exception):
+    pass
+
+def thread_run (myid,p,q):
```
ewencp: these parameter lists might fail pep8 checks, normally there are spaces after commas.

tests/test_threads.py
```diff
@@ -0,0 +1,67 @@
...
+            cb = do_crash
+        p.produce('mytopic', value='hi', callback=cb)
+        t = time.time()
+        try:
+            p.flush()
+            print(myid, 'Flush took %.3f' % (time.time() - t))
+        except IntendedException:
+            print(myid, "Intentional callback crash: ok")
+            continue
+        except:
```
ewencp: This final `except` block isn't needed

tests/test_threads.py
```diff
@@ -0,0 +1,67 @@
...
+def test_thread_safety():
+    """ Basic thread safety tests. """
+
+    q = queue.Queue()
+    p = Producer({'socket.timeout.ms':10,
+                  'socket.blocking.max.ms': 10,
+                  'default.topic.config': {'message.timeout.ms': 10}})
+
+    threads = list()
+    for i in range(1, 5):
```
ewencp: Are the settings you have here (5 threads, 3 messages each) enough to reliably trigger an error before the patch?

ewencp:
@edenhill Patch looks good, just had questions re: test

tests/test_threads.py
edenhill: Yup, fails immediately and repeatedly on master:
```
$ python tests/t.py 
Exception in thread 1:
NoneType

Fatal Python error: PyEval_RestoreThread: NULL tstate
```

Issue #30, opened 2016-08-11
After calling close, consumers issue periodic metadata requests


aaronhays:
If you create a consumer, connect, read some data, and then close but don't destroy every reference to the consumer, it will continue issuing metadata requests every five minutes.

I'm assuming this is because you are calling `rd_kafka_consumer_close` in the python `close()` method, but not calling `rd_kafka_destroy` until the consumer object is deallocated. It seems like once you call `close()`, the `Consumer` instance cannot be used for anything - is there a reason not to call `rd_kafka_destroy` in `close()`?

edenhill:
Your analysis is correct.
close() "only" closes the high-level consumer functionality (namely leaving the group, committing offsets, etc), but the underlying librdkafka consumer handle remains functional for legacy consumer and basic operations (such as metadata requests!).

Now, there's probably no real use to close the consumer and then do something else on the same handle, so calling rd_kafka_destroy() from close() in the Python API sounds like a good idea.

edenhill:
Fixed on master.

PR #33, opened 2016-08-19
Destroy rd_kafka_t handle on consumer.close() (#30)

edenhill:
Solves issue #30

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -391,12 +391,18 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
 
 static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
 	CallState cs;
+	int raise = 0;
 
 	CallState_begin(self, &cs);
 
 	rd_kafka_consumer_close(self->rk);
 
-	if (!CallState_end(self, &cs))
+	raise = !CallState_end(self, &cs);
+
+	rd_kafka_destroy(self->rk);
+	self->rk = NULL;
+
+	if (raise)
 		return NULL;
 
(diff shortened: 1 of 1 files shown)


confluent_kafka/src/Consumer.c
```diff
@@ -391,12 +391,18 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
 static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
 	CallState cs;
+	int raise = 0;
 
 	CallState_begin(self, &cs);
 
 	rd_kafka_consumer_close(self->rk);
 
-	if (!CallState_end(self, &cs))
+	raise = !CallState_end(self, &cs);
```
ewencp: My understanding of the behavior here is that we essentially want to do the equivalent of `Consumer_dealloc`, but as far as I can tell, the `rd_kafka_destroy` is only a fraction of what should actually be done? Why isn't the full `Consumer_clear` (or more completely, `Consumer_dealloc`) performed? In particular, since the full `Consumer_clear` is not performed, doesn't this mean we'll leave some callbacks un-decref'd?

edenhill: Yes, almost, dealloc() is eventually triggered by the GC and I left it for that to clean up any lingering objects (such as callbacks), but they will be cleaned up on the next GC spin.

We could call cleanup from close() as an optimization to speed up GC, apart from that there shouldnt be any functional differences.
What do you think?

ewencp: @edenhill Hmm, I guess it doesn't matter much then. Seems fine as is and this LGTM.

Issue #31, opened 2016-08-12
Access to watermark offsets API in python


aaronhays:
I just ran across https://github.com/edenhill/librdkafka/issues/503 after spending some time hacking in a solution using the available Python APIs that I'm too embarrassed to discuss publicly. It works fine for initial offset queries, but is not sufficient for keeping track of the current highwater mark during consumption. It would be great to get access to these APIs in Python. Is this on the roadmap anywhere?

I would prefer not to hack this in myself because I'm not quite sure how to correctly manage the memory for those `int64_t *` output parameters (`low` and `high`) between C and Python, as I'm not proficient with the Python C-API. I would probably adapt the interface to return a tuple of `low` and `high` and raise an exception in the case of error, but I don't know if that is the best way to proceed.

edenhill:
Merged!

PR #34, opened 2016-08-19
Add assignment and watermark offsets APIs

edenhill:
For issue #31

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -231,6 +231,28 @@ static PyObject *Consumer_unassign (Handle *self, PyObject *ignore) {
 	Py_RETURN_NONE;
 }
 
+static PyObject *Consumer_assignment (Handle *self, PyObject *args,
+				      PyObject *kwargs) {
+
+	PyObject *plist;
+	rd_kafka_topic_partition_list_t *c_parts;
+	rd_kafka_resp_err_t err;
+
+	err = rd_kafka_assignment(self->rk, &c_parts);
+	if (err) {
+		cfl_PyErr_Format(err,
+				 "Failed to get assignment: %s",
+				 rd_kafka_err2str(err));
+		return NULL;
+	}
+
+
+	plist = c_parts_to_py(c_parts);
+	rd_kafka_topic_partition_list_destroy(c_parts);
+
+	return plist;
+}
+
 
 
 static PyObject *Consumer_commit (Handle *self, PyObject *args,
@@ -489,6 +511,17 @@ static PyMethodDef Consumer_methods[] = {
           "  :raises: KafkaException\n"
           "\n"
         },
+	{ "assignment", (PyCFunction)Consumer_assignment,
+	  METH_VARARGS|METH_KEYWORDS,
+	  ".. py:function:: assignment()\n"
+	  "\n"
+	  "  Returns the current partition assignment.\n"
+	  "\n"
+	  "  :returns: List of assigned topic+partitions.\n"
+	  "  :rtype: list(TopicPartition)\n"
+	  "  :raises: KafkaException\n"
+	  "\n"
+	},
 	{ "commit", (PyCFunction)Consumer_commit, METH_VARARGS|METH_KEYWORDS,
 	  ".. py:function:: commit([message=None], [offsets=None], [async=True])\n"
 	  "\n"
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -232,25 +232,25 @@ static PyObject *Consumer_unassign (Handle *self, PyObject *ignore) {
 }
 
 static PyObject *Consumer_assignment (Handle *self, PyObject *args,
-				      PyObject *kwargs) {
+                                      PyObject *kwargs) {
 
-	PyObject *plist;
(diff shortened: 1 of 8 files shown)


confluent_kafka/src/Consumer.c
```diff
@@ -508,6 +589,22 @@ static PyMethodDef Consumer_methods[] = {
 	  "  :raises: KafkaException\n"
 	  "\n"
 	},
+	{ "get_watermark_offsets", (PyCFunction)Consumer_get_watermark_offsets,
```
ewencp: Assignment and watermark offsets seem quite different re: the PR. The assignment seems pretty uncontroversial -- we know what we want exposed here.

Is the watermark the same? If we're consistent everywhere, that's fine, just asking because I think this is something we want to be consistent about across languages and I'm not sure we've fully specified the semantics of. Even the cached versions seem like they may need clarification (e.g., what is the cache duration?). The high/low offsets seem clearer, but I think in the Java client there may be some desire to make any request for offsets a bit more on-demand compared to the high offset provided with each fetch request.

/cc @hachikuji since I think we may one day (hopefully in the near future!) want to provide this for the Java clients as `KafkaBasedLog` and similar classes could greatly benefit from a real LEO API...

edenhill: I needed assignment() to test get_watermark_offsets() :)

confluent_kafka/src/Consumer.c
```diff
@@ -359,6 +381,54 @@ static PyObject *Consumer_position (Handle *self, PyObject *args,
...
+	}
+
+	if (err) {
+		cfl_PyErr_Format(err,
+				 "Failed to get watermark offsets: %s",
+				 rd_kafka_err2str(err));
+		return NULL;
+	}
+
+	rlist = PyList_New(2);
```
ewencp: Since it is fixed length anyway, this would normally be a tuple. (And docs below would need update too.)

examples/integration_test.py
```diff
@@ -258,6 +258,18 @@ def verify_consumer():
             print('max_msgcnt %d reached' % msgcnt)
             break
 
+    # Get current assignment
+    assignment = c.assignment()
+
+    # Get cached watermark offsets
+    # Since we're not making use of statistics the low offset is not known so ignore it.
```
ewencp: I get that we don't have stats on, but does't that mean we can at least validate that it is invalid? We don't seem to validate the high offset either?

tests/test_Consumer.py
```diff
@@ -39,6 +39,20 @@ def dummy_assign_revoke (consumer, partitions):
     partitions = list(map(lambda p: TopicPartition("test", p), range(0,100,3)))
     kc.assign(partitions)
 
+    # Verify assignment
+    assignment = kc.assignment()
+    assert partitions == assignment
+
+    # Get cached watermarks, should all be invalid.
+    lo, hi = kc.get_watermark_offsets(partitions[0], cached=True)
+    assert lo == -1001 and hi == -1001
```
ewencp: Should we be exposing `RD_KAFKA_OFFSET_INVALID`? If nothing else, I guess at least document it in the docstring?

confluent_kafka/src/Consumer.c
edenhill: re watermark consistency:
I briefly discussed such an API with @hachikuji about six months ago, but we didnt really decide on anything.
Prior to librdkafka 0.9.0 I added two APIs, query_watermark_offsets(ask broker) and get_watermark_offsets(cached) which the Python client abstracts in one call.
In that very limited sense it makes the API consistent.

But I agree with you that a more generic API that allows querying for any offset (logical, time based, absolute) makes more sense.

re caching: the main use of the cached API is to allow an application to check the currently known high watermark (and possibly somewhat outdated low watermark if stats is enabled) in its fast path with low cost. Typical use is an updated consumer lag (even though the stats already exposes that..).

tests/test_Consumer.py
edenhill: Yeah, we absolutely should, there's an old FIXME for that somewhere.
What's a good name:  confluent_kafka.OffsetInvalid, confluent_kafka.InvalidOffset, ..?

confluent_kafka/src/Consumer.c
ewencp: Oh, for consistency I even just meant consistency across all the clients we're building based on librdkafka, but agreed that there could be others.

I'm fine with exposing it, just wanted to make sure we're thinking about compatibility and designing good, user friendly APIs along the way.

tests/test_Consumer.py
ewencp: Should be capitalized to be pep8 style. `confluent_kafka.OFFSET_INVALID`? `INVALID_OFFSET` reads a bit better, but it might make more sense to just stay consistent with the underlying name.

wangxiaoyu:
This is a very desirable feature. But it seems not merged yet. Maybe it could become available in the next release then?

edenhill:
@ewencp with the dust settled, should we merge this as-is or rename the function get_watermark_offsets?

ewencp:
@edenhill I'm fine to merge it as is (or close to as is, needs rebase and I think there's still that comment about tuple instead of list) as long as we stay consistent with our other clients that wrap librdkafka.

Issue #36, opened 2016-08-23
Integrate client with Confluent schema registry and Avro


criccomini:
We are looking at integrating this client with the [Confluent schema registry](https://github.com/confluentinc/schema-registry) and Avro with this client. Proposal is to add a new submodule in this repo that has:
- Caching schema registry client
- Message serializer
- Wrapper around base producer/consumer

The wrapper would override send/receive methods on base producer/consumer, and pass the objects/bytes through the message serializer. The message serializer would, in turn, use the caching schema client to fetch the schema, and encode/decode the message. Wrapper would be pluggable so other serde implementations could be used as well (e.g. Thrift/Protobuf), though we're not planning to implement any others.

Does this sound OK?

/cc @roopahc

ewencp:
@criccomini All the components you mention sound fine, this is essentially what I would do. Given how small the overrides are for send/receive, I might not even make it pluggable in the same way the Java ones are. Python's duck typing makes this a lot easier to just wrap the two locations that need serialization with normal overloads, so a simple `AvroProducer` and `AvroConsumer` might just be simpler for the user than having to separately configure/instantiate the serializer.

The one question I have is whether we want it in this repo. Were you thinking about moving `confluent_kafka` to a subdirectory and still publishing it separately from this new package, or including it directly. I just wonder if requiring extra dependencies on `confluent_kafka` will be acceptable to folks not using Avro (and if we add Thrift/Protobufs eventually the dependencies would grow even more).

criccomini:
@ewencp 

> Were you thinking about moving confluent_kafka to a subdirectory and still publishing it separately from this new package, or including it directly

Not sure. I'm open to suggestions. One thing that I've seen work with Airflow is keeping everything in one package, but using [setup.py `extras_requires`](https://github.com/apache/incubator-airflow/blob/master/setup.py#L207). This would mean you'd:

    pip install confluent-kafka
    pip install confluent-kafka[avro]

The second one would include the Avro dependencies. We could have `confluent-kafka[thrift]`, etc. Thoughts?

ewencp:
@criccomini Ah, cool, `extras_require` seems like a pretty nice solution. Crazy that people have encountered this issue before and created a reasonable solution for it :) Do you just implement things normally and people who don't want to use the optional features just need to avoid the relevant imports? If so, this seems like a pretty nice solution to the problem without requiring much effort on our part -- compared to requiring completely separate libraries in many other languages, this might be a really easy solution.

[...]

PR #40, opened 2016-09-07
Integrate client with Confluent schema registry and Avro 

roopahc:
Fixes : #36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,14 @@
 language: python
+sudo: required
 python:
   - "2.7"
   - "3.4"
 before_install:
   - bash tools/bootstrap-librdkafka.sh v0.9.2 tmp-build
+  - pip install --upgrade pip
   - pip install pytest-timeout
 install:
-  - pip install -v --global-option=build_ext --global-option="-Itmp-build/include/" --global-option="-Ltmp-build/lib" .
+  - pip install -v --global-option=build_ext --global-option="-Itmp-build/include/" --global-option="-Ltmp-build/lib" . .[avro]
 env:
   - LD_LIBRARY_PATH=$PWD/tmp-build/lib
-script: py.test -v --timeout 20 --ignore=tmp-build
+script: py.test -v --timeout 20 --ignore=tmp-build --import-mode append
--- a/README.md
+++ b/README.md
@@ -46,6 +46,20 @@ while running:
 c.close()
 ```
 
+**AvroProducer**
+```
+from confluent_kafka import avro 
+from confluent_kafka.avro import AvroProducer
+
+value_schema = avro.load('ValueSchema.avsc')
+key_schema = avro.load('KeySchema.avsc')
+value = {"name": "Value"}
+key = {"name": "Key"}
+
+avroProducer = AvroProducer({'bootstrap.servers': 'mybroker,mybroker2', 'schema.registry.url': 'http://schem_registry_host:port'}, default_key_schema=key_schema, default_value_schema=value_schema)
+avroProducer.produce(topic='my_topic', value=value, key=key)
+```
+
 See [examples](examples) for more examples.
 
 
@@ -85,11 +99,17 @@ Install
 **Install from PyPi:**
 
     $ pip install confluent-kafka
+    
+    # for AvroProducer
+    $ pip install confluent-kafka[avro]
 
 
 **Install from source / tarball:**
 
     $ pip install .
+
+    # for AvroProducer
+    $ pip install .[avro]
 
 
 Build
--- a/confluent_kafka/__init__.py
+++ b/confluent_kafka/__init__.py
@@ -1,2 +1,2 @@
-__all__ = ['cimpl','kafkatest']
+__all__ = ['cimpl', 'avro', 'kafkatest']
 from .cimpl import *
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -1,0 +1,114 @@
+"""
+    Avro schema registry module: Deals with encoding and decoding of messages with avro schemas
+
+"""
+import sys
+
+from confluent_kafka import Producer
+
+VALID_LEVELS = ['NONE', 'FULL', 'FORWARD', 'BACKWARD']
+
+
+def loads(schema_str):
+    """ Parse a schema given a schema string """
+    if sys.version_info[0] < 3:
+        return schema.parse(schema_str)
+    else:
+        return schema.Parse(schema_str)
+
+
+def load(fp):
+    """ Parse a schema from a file path """
+    with open(fp) as f:
+        return loads(f.read())
+
+
+# avro.schema.RecordSchema and avro.schema.PrimitiveSchema classes are not hashable. Hence defining them explicitely as a quick fix
+def _hash_func(self):
+    return hash(str(self))
+
+
+try:
+    from avro import schema
+
+    schema.RecordSchema.__hash__ = _hash_func
+    schema.PrimitiveSchema.__hash__ = _hash_func
+except ImportError:
+    pass
+
+
+class ClientError(Exception):
+    """ Error thrown by Schema Registry clients """
+
+    def __init__(self, message, http_code=None):
+        self.message = message
+        self.http_code = http_code
+        super(ClientError, self).__init__(self.__str__())
+
+    def __repr__(self):
+        return "ClientError(error={error})".format(error=self.message)
+
+    def __str__(self):
+        return self.message
+
+
+from confluent_kafka.avro.cached_schema_registry_client import CachedSchemaRegistryClient
+from confluent_kafka.avro.serializer import SerializerError
+from confluent_kafka.avro.serializer.message_serializer import MessageSerializer
+
+
+class AvroProducer(Producer):
+    """
+        Kafka Producer client which does avro schema encoding to messages.
+        Handles schema registration, Message serialization.
+
+        Constructor takes below parameters
+
+        @:param: config: dict object with config parameters containing url for schema registry (schema.registry.url).
+        @:param: default_key_schema: Optional avro schema for key
+        @:param: default_value_schema: Optional avro schema for value
+    """
+
+    def __init__(self, config, default_key_schema=None,
+                 default_value_schema=None):
+        if ('schema.registry.url' not in config.keys()):
+            raise ValueError("Missing parameter: schema.registry.url")
+        schem_registry_url = config["schema.registry.url"]
+        del config["schema.registry.url"]
+
+        super(AvroProducer, self).__init__(config)
+        self._serializer = MessageSerializer(CachedSchemaRegistryClient(url=schem_registry_url))
+        self._key_schema = default_key_schema
+        self._value_schema = default_value_schema
+
+    def produce(self, **kwargs):
+        """
+            Sends message to kafka by encoding with specified avro schema
+            @:param: topic: topic name
+            @:param: value: A dictionary object
+            @:param: value_schema : Avro schema for value
+            @:param: key: A dictionary object
+            @:param: key_schema : Avro schema for key
+            @:exception: SerializerError
+        """
+        # get schemas from  kwargs if defined
+        key_schema = kwargs.pop('key_schema', self._key_schema)
+        value_schema = kwargs.pop('value_schema', self._value_schema)
+        topic = kwargs.pop('topic', None)
+        if not topic:
+            raise ClientError("Topic name not specified.")
+        value = kwargs.pop('value', None)
+        key = kwargs.pop('key', None)
+        if value:
+            if value_schema:
+                value = self._serializer.encode_record_with_schema(topic, value_schema, value)
+            else:
+                raise SerializerError("Avro schema required for value")
+
+        if key:
+            if key_schema:
+                key = self._serializer.encode_record_with_schema(topic, key_schema, key, True)
+            else:
+                raise SerializerError("Avro schema required for key")
+
+        super(AvroProducer, self).produce(topic, value, key, **kwargs)
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -1,0 +1,313 @@
+#!/usr/bin/env python
+#
+# Copyright 2016 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+#
+# derived from https://github.com/verisign/python-confluent-schemaregistry.git
+#
+import json
+import logging
+from collections import defaultdict
+
+import requests
+
+from . import ClientError, VALID_LEVELS
+
+# Common accept header sent
+ACCEPT_HDR = "application/vnd.schemaregistry.v1+json, application/vnd.schemaregistry+json, application/json"
+log = logging.getLogger(__name__)
+
+
+class CachedSchemaRegistryClient(object):
+    """
+    A client that talks to a Schema Registry over HTTP
+
+    See http://confluent.io/docs/current/schema-registry/docs/intro.html
+
+    Errors communicating to the server will result in a ClientError being raised.
+
+    @:param: url: url to schema registry
+    """
+
+    def __init__(self, url, max_schemas_per_subject=1000):
+        """Construct a client by passing in the base URL of the schema registry server"""
+
+        self.url = url.rstrip('/')
+
+        self.max_schemas_per_subject = max_schemas_per_subject
+        # subj => { schema => id }
+        self.subject_to_schema_ids = defaultdict(dict)
+        # id => avro_schema
+        self.id_to_schema = defaultdict(dict)
+        # subj => { schema => version }
+        self.subject_to_schema_versions = defaultdict(dict)
+
+    def _send_request(self, url, method='GET', body=None, headers=None):
+        if body:
+            body = json.dumps(body)
+            body = body.encode('utf8')
+        _headers = dict()
+        _headers["Accept"] = ACCEPT_HDR
+        if body:
+            _headers["Content-Length"] = str(len(body))
+            _headers["Content-Type"] = "application/vnd.schemaregistry.v1+json"
+
+        if headers:
+            for header_name in headers:
+                _headers[header_name] = headers[header_name]
+        if method == 'GET':
+            response = requests.get(url, headers=_headers)
+        elif method == 'POST':
+            response = requests.post(url, body, headers=_headers)
+        elif method == 'PUT':
+            response = requests.put(url, body, headers=_headers)
+        elif method == 'DELETE':
+            response = requests.delete(url, headers=_headers)
+        else:
+            raise ClientError("Invalid HTTP request type")
+
+        result = json.loads(response.text)
+        return (result, response.status_code)
+
+    def _add_to_cache(self, cache, subject, schema, value):
+        sub_cache = cache[subject]
+        sub_cache[schema] = value
+
+    def _cache_schema(self, schema, schema_id, subject=None, version=None):
+        # don't overwrite anything
+        if schema_id in self.id_to_schema:
+            schema = self.id_to_schema[schema_id]
+        else:
+            self.id_to_schema[schema_id] = schema
+
+        if subject:
+            self._add_to_cache(self.subject_to_schema_ids,
+                               subject, schema, schema_id)
+            if version:
+                self._add_to_cache(self.subject_to_schema_versions,
+                                   subject, schema, version)
+
+    def register(self, subject, avro_schema):
+        """
+        POST /subjects/(string: subject)/versions
+        Register a schema with the registry under the given subject
+        and receive a schema id.
+
+        avro_schema must be a parsed schema from the python avro library
+
+        Multiple instances of the same schema will result in cache misses.
+
+        @:param: subject: subject name
+        @:param: avro_schema: Avro schema to be registered
+        @:returns: schema_id: int value
+        """
+
+        schemas_to_id = self.subject_to_schema_ids[subject]
+        schema_id = schemas_to_id.get(avro_schema, None)
+        if schema_id != None:
+            return schema_id
+        # send it up
+        url = '/'.join([self.url, 'subjects', subject, 'versions'])
+        # body is { schema : json_string }
+
+        body = {'schema': json.dumps(avro_schema.to_json())}
+        result, code = self._send_request(url, method='POST', body=body)
+        if code == 409:
+            raise ClientError("Incompatible Avro schema:" + str(code))
+        elif code == 422:
+            raise ClientError("Invalid Avro schema:" + str(code))
+        elif not (code >= 200 and code <= 299):
+            raise ClientError("Unable to register schema. Error code:" + str(code))
+        # result is a dict
+        schema_id = result['id']
+        # cache it
+        self._cache_schema(avro_schema, schema_id, subject)
+        return schema_id
+
+    def get_by_id(self, schema_id):
+        """
+        GET /schemas/ids/{int: id}
+        Retrieve a parsed avro schema by id or None if not found
+        @:param: schema_id: int value
+        @:returns: Avro schema
+        """
+        if schema_id in self.id_to_schema:
+            return self.id_to_schema[schema_id]
+        # fetch from the registry
+        url = '/'.join([self.url, 'schemas', 'ids', str(schema_id)])
+
+        result, code = self._send_request(url)
+        if code == 404:
+            log.error("Schema not found:" + str(code))
+            return None
+        elif not (code >= 200 and code <= 299):
+            log.error("Unable to get schema for the specific ID:" + str(code))
+            return None
+        else:
+            # need to parse the schema
+            schema_str = result.get("schema")
+            try:
+                result = avro.loads(schema_str)
+                # cache it
+                self._cache_schema(result, schema_id)
+                return result
+            except:
+                # bad schema - should not happen
+                raise ClientError("Received bad schema from registry.")
+
+    def get_latest_schema(self, subject):
+        """
+        GET /subjects/(string: subject)/versions/(versionId: version)
+
+        Return the latest 3-tuple of:
+        (the schema id, the parsed avro schema, the schema version)
+        for a particular subject.
+
+        This call always contacts the registry.
+
+        If the subject is not found, (None,None,None) is returned.
+        @:param: subject: subject name
+        @:returns: (schema_id, schema, version)
+        """
+        url = '/'.join([self.url, 'subjects', subject, 'versions', 'latest'])
+
+        result, code = self._send_request(url)
+        if code == 404:
+            log.error("Schema not found:" + str(code))
+            return (None, None, None)
+        elif code == 422:
+            log.error("Invalid version:" + str(code))
+            return (None, None, None)
+        elif not (code >= 200 and code <= 299):
+            return (None, None, None)
+        schema_id = result['id']
+        version = result['version']
+        if schema_id in self.id_to_schema:
+            schema = self.id_to_schema[schema_id]
+        else:
+            try:
+                schema = avro.loads(result['schema'])
+            except:
+                # bad schema - should not happen
+                raise ClientError("Received bad schema from registry.")
+
+        self._cache_schema(schema, schema_id, subject, version)
+        return (schema_id, schema, version)
+
+    def get_version(self, subject, avro_schema):
+        """
+        POST /subjects/(string: subject)
+
+        Get the version of a schema for a given subject.
+
+        Returns None if not found.
+        @:param: subject: subject name
+        @:param: avro_schema: Avro schema
+        @:returns: version
+        """
+        schemas_to_version = self.subject_to_schema_versions[subject]
+        version = schemas_to_version.get(avro_schema, None)
+        if version != None:
+            return version
+
+        url = '/'.join([self.url, 'subjects', subject])
+        body = {'schema': json.dumps(avro_schema.to_json())}
+
+        result, code = self._send_request(url, method='POST', body=body)
+        if code == 404:
+            log.error("Not found:" + str(code))
+            return None
+        elif not (code >= 200 and code <= 299):
+            log.error("Unable to get version of a schema:" + str(code))
+            return None
+        schema_id = result['id']
+        version = result['version']
+        self._cache_schema(avro_schema, schema_id, subject, version)
+        return version
+
+    def test_compatibility(self, subject, avro_schema, version='latest'):
+        """
+        POST /compatibility/subjects/(string: subject)/versions/(versionId: version)
+
+        Test the compatibility of a candidate parsed schema for a given subject.
+
+        By default the latest version is checked against.
+        @:param: subject: subject name
+        @:param: avro_schema: Avro schema
+        @:return: True if compatible, False if not compatible
+        """
+        url = '/'.join([self.url, 'compatibility', 'subjects', subject,
+                        'versions', str(version)])
+        body = {'schema': json.dumps(avro_schema.to_json())}
+        try:
+            result, code = self._send_request(url, method='POST', body=body)
+            if code == 404:
+                log.error(("Subject or version not found:" + str(code)))
+                return False
+            elif code == 422:
+                log.error(("Invalid subject or schema:" + str(code)))
+                return False
+            elif code >= 200 and code <= 299:
+                return result.get('is_compatible')
+            else:
+                log.error("Unable to check the compatibility")
+                False
+        except:
+            return False
+
+    def update_compatibility(self, level, subject=None):
+        """
+        PUT /config/(string: subject)
+
+        Update the compatibility level for a subject.  Level must be one of:
+
+        @:param: level: ex: 'NONE','FULL','FORWARD', or 'BACKWARD'
+        """
+        if level not in VALID_LEVELS:
+            raise ClientError("Invalid level specified: %s" % (str(level)))
+
+        url = '/'.join([self.url, 'config'])
+        if subject:
+            url += '/' + subject
+
+        body = {"compatibility": level}
+        result, code = self._send_request(url, method='PUT', body=body)
+        if code >= 200 and code <= 299:
+            return result['compatibility']
+        else:
+            raise ClientError("Unable to update level: %s. Error code: %d" % (str(level)), code)
+
+    def get_compatibility(self, subject=None):
+        """
+        GET /config
+        Get the current compatibility level for a subject.  Result will be one of:
+
+        @:param: subject: subject name
+        @:return: 'NONE','FULL','FORWARD', or 'BACKWARD'
+        """
+        url = '/'.join([self.url, 'config'])
+        if subject:
+            url += '/' + subject
+
+        result, code = self._send_request(url)
+        if code >= 200 and code <= 299:
+            compatibility = result.get('compatibility', None)
+
+        if not compatibility:
+            compatibility = result.get('compatibilityLevel')
+
+        return compatibility
--- a/confluent_kafka/avro/serializer/__init__.py
+++ b/confluent_kafka/avro/serializer/__init__.py
@@ -1,0 +1,11 @@
+class SerializerError(Exception):
+    """Generic error from serializer package"""
+
+    def __init__(self, message):
+        self.message = message
+
+        def __repr__(self):
+            return 'SerializerError(error={error})'.format(error=self.message)
+
+        def __str__(self):
+            return self.message
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -1,0 +1,208 @@
+#!/usr/bin/env python
+#
+# Copyright 2016 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+#
+# derived from https://github.com/verisign/python-confluent-schemaregistry.git
+#
+import io
+import logging
+import struct
+import sys
+import traceback
+
+import avro
+import avro.io
+
+from confluent_kafka.avro import ClientError
+from . import SerializerError
+
+log = logging.getLogger(__name__)
+
+MAGIC_BYTE = 0
+
+HAS_FAST = False
+try:
+    from fastavro.reader import read_data
+
+    HAS_FAST = True
+except:
+    pass
+
+
+class ContextStringIO(io.BytesIO):
+    """
+    Wrapper to allow use of StringIO via 'with' constructs.
+    """
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+        return False
+
+
+class MessageSerializer(object):
+    """
+    A helper class that can serialize and deserialize messages
+    that need to be encoded or decoded using the schema registry.
+
+    All encode_* methods return a buffer that can be sent to kafka.
+    All decode_* methods expect a buffer received from kafka.
+    """
+
+    def __init__(self, registry_client):
+        self.registry_client = registry_client
+        self.id_to_decoder_func = {}
+        self.id_to_writers = {}
+
+    '''
+
+    '''
+
+    def encode_record_with_schema(self, topic, schema, record, is_key=False):
+        """
+        Given a parsed avro schema, encode a record for the given topic.  The
+        record is expected to be a dictionary.
+
+        The schema is registered with the subject of 'topic-value'
+        @:param topic : Topic name
+        @:param schema : Avro Schema
+        @:param record : A dictionary object
+        @:returns : Encoded record with schema ID as bytes
+        """
+        if not isinstance(record, dict):
+            raise SerializerError("record must be a dictionary")
+        subject_suffix = ('-key' if is_key else '-value')
+        # get the latest schema for the subject
+        subject = topic + subject_suffix
+        # register it
+        schema_id = self.registry_client.register(subject, schema)
+        if not schema_id:
+            message = "Unable to retrieve schema id for subject %s" % (subject)
+            raise SerializerError(message)
+
+        # cache writer
+        self.id_to_writers[schema_id] = avro.io.DatumWriter(schema)
+
+        return self.encode_record_with_schema_id(schema_id, record)
+
+    def encode_record_with_schema_id(self, schema_id, record):
+        """
+        Encode a record with a given schema id.  The record must
+        be a python dictionary.
+        @:param: schema_id : integer ID
+        @:param: record : A dictionary object
+        @:returns: decoder function
+        """
+        if not isinstance(record, dict):
+            raise SerializerError("record must be a dictionary")
+        # use slow avro
+        if schema_id not in self.id_to_writers:
+            # get the writer + schema
+
+            try:
+                schema = self.registry_client.get_by_id(schema_id)
+                if not schema:
+                    raise SerializerError("Schema does not exist")
+                self.id_to_writers[schema_id] = avro.io.DatumWriter(schema)
+            except ClientError as e:
+                exc_type, exc_value, exc_traceback = sys.exc_info()
+                raise SerializerError("Error fetching schema from registry:" + repr(
+                    traceback.format_exception(exc_type, exc_value, exc_traceback)))
+
+        # get the writer
+        writer = self.id_to_writers[schema_id]
+        with ContextStringIO() as outf:
+            # write the header
+            # magic byte
+
+            outf.write(struct.pack('b', MAGIC_BYTE))
+
+            # write the schema ID in network byte order (big end)
+
+            outf.write(struct.pack('>I', schema_id))
+
+            # write the record to the rest of it
+            # Create an encoder that we'll write to
+            encoder = avro.io.BinaryEncoder(outf)
+            # write the magic byte
+            # write the object in 'obj' as Avro to the fake file...
+            writer.write(record, encoder)
+
+            return outf.getvalue()
+
+    # Decoder support
+    def _get_decoder_func(self, schema_id, payload):
+        if schema_id in self.id_to_decoder_func:
+            return self.id_to_decoder_func[schema_id]
+
+        # fetch from schema reg
+        try:
+            schema = self.registry_client.get_by_id(schema_id)
+        except:
+            schema = None
+
+        if not schema:
+            err = "unable to fetch schema with id %d" % (schema_id)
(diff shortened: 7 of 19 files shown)


criccomini:
@edenhill @ewencp Would appreciate a review when you've got the cycles. :)

roopahc:
@ConfluentCLABot Have signed CLA

confluent_kafka/avro/serializer/MessageSerializer.py
```diff
@@ -0,0 +1,241 @@
...
+HAS_FAST = False
+try:
+    from fastavro.reader import read_data
+
+    HAS_FAST = True
+except:
+    pass
+
+
+class ContextStingIO(io.BytesIO):
```
wkiser: nit: should be String

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # below classes were not hashable. hence defining them explicitely as a quick fix
+        schema.RecordSchema.__hash__ = self.hash_func
+        schema.PrimitiveSchema.__hash__ = self.hash_func
+
+    def _send_request(self, url, method='GET', body=None, headers=None):
+        if body:
+            body = json.dumps(body)
+            body = body.encode('utf8')
+            print(body)
```
matthayes: Shouldn't use `print` because this will spam console output and there's no way to turn it off.  Instead use something like: `self.log.debug("Sending request: %s", body)`.

confluent_kafka/avro/serializer/MessageSerializer.py
```diff
@@ -0,0 +1,241 @@
...
+
+            try:
+                schema = self.registry_client.get_by_id(schema_id)
+                if not schema:
+                    self.log.error("Schema does not exist")
+                    raise SerializerError("Schema does not exist")
+                self.id_to_writers[schema_id] = avro.io.DatumWriter(schema)
+            except ClientError as e:
+                exc_type, exc_value, exc_traceback = sys.exc_info()
+                print(repr(traceback.format_exception(exc_type, exc_value, exc_traceback)))
```
matthayes: Don't `print`

tests/mock_registry.py
```diff
@@ -0,0 +1,189 @@
...
+        schema_str = json.dumps(avro_schema.to_json())
+        if schema_str in self.schema_cache:
+            return self.schema_cache[schema_str]
+        self.schema_cache[schema_str] = avro_schema
+        return avro_schema
+
+    def _get_schema_from_body(self, req):
+        length = int(req.headers['content-length'])
+        data = req.rfile.read(length)
+        print(data.decode("utf-8"))
```
matthayes: Don't `print`

matthayes:
FYI I tried testing this and was able to consume message using `kafka-avro-console-consumer`.  Thanks for working on this 👍

criccomini:
Hi Matt! :)

matthayes:
Hey! :)

confluent_kafka/avro/AvroProducer.py
```diff
@@ -0,0 +1,62 @@
...
+                raise SerializerError("Avro schema required for value")
+
+        if key is not None:
+            if key_avro_schema is not None:
+                key = self.serializer.encode_record_with_schema(topic, key_avro_schema, key, True)
+            else:
+                self.log.error("Schema required for key serialization")
+                raise SerializerError("Avro schema required for key")
+
+        self.producer.produce(topic, value, key)
```
matthayes: kwargs is being ignored, so if you pass `callback`/`on_delivery` or `partition` they aren't used.

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+    See http://confluent.io/docs/current/schema-registry/docs/intro.html
+
+    Errors communicating to the server will result in a ClientError being raised.
+
+    @:param: url: url to schema registry
+    """
+
+    def __init__(self, url, max_schemas_per_subject=1000):
+        """Construct a client by passing in the base URL of the schema registry server"""
+        self.log = logging.getLogger(__name__)
```
matthayes: nit: If you're going to have a logger in the class, you may consider using the class's fully qualified name.  Otherwise I don't know if there is an advantage with declaring it here as opposed to outside the class (e.g. `logger = logging.getLogger(__name__)`) because any class declared in this file would have the same logger name.

confluent_kafka/avro/AvroProducer.py
```diff
@@ -0,0 +1,62 @@
...
+
+        @:param: producer: confluent_kafka.Producer object
+        @:param: message_serializer: Message Serializer object
+    '''
+    def __init__(self, producer, message_serializer):  # real signature unknown; restored from __doc__
+        self.producer = producer
+        self.serializer = message_serializer
+        self.log = logging.getLogger(__name__)
+
+    def produce(self, topic, value=None, value_avro_schema=None, key=None, key_avro_schema=None, *args, **kwargs):
```
matthayes: I don't see any reason to have `*args` if it isn't used.  Also I would prefer a definition like this:

```
def produce(self, topic, *, value=None, value_avro_schema=None, key=None, key_avro_schema=None, **kwargs):
```

Otherwise I think it could be easy to confuse the order of arguments.

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        @:param: schema_id: int value
+        @:returns: Avro schema
+        """
+        if schema_id in self.id_to_schema:
+            return self.id_to_schema[schema_id]
+        # fetch from the registry
+        url = '/'.join([self.url, 'schemas', 'ids', str(schema_id)])
+
+        result, code = self._send_request(url)
+        if (code != 200):
```
matthayes: nit: May want to explicitly check for 404 vs other status codes and log an error where appropriate.

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        This call always contacts the registry.
+
+        If the subject is not found, (None,None,None) is returned.
+        @:param: subject: subject name
+        @:returns: (schema_id, schema, version)
+        """
+        url = '/'.join([self.url, 'subjects', subject, 'versions', 'latest'])
+
+        result, code = self._send_request(url)
+        if (code != 200):
```
matthayes: nit: check status code for 404 vs others

criccomini:
@edenhill @ewencp ping

confluent_kafka/avro/AvroProducer.py
```diff
@@ -0,0 +1,62 @@
...
+        '''
+            Sends message to kafka by encoding with specified avro schema
+            @:param: topic: topic name
+            @:param: value: A dictionary object
+            @:param: value_avro_schema : Avro schema for value
```
ewencp: I'd shorten these names to `value_schema` and `key_schema` -- we're in an Avro-specific class, so they couldn't really be anything else.

I don't know if people might also find it useful to be able to set the schema as a default in the constructor? In languages like Java where the schema is attached anyway and is handled automatically if you use code generation, this isn't as annoying. It might be a bit more annoying to have to pass the same schema (the normal case) into every produce call. Not sure how valuable this will be since usually you only have a small number of call sites to `produce()` anyway, but seems annoying to have to carry the schema around separately everywhere.

confluent_kafka/avro/AvroProducer.py
```diff
@@ -0,0 +1,62 @@
...
+            @:param: key: A dictionary object
+            @:param: key_avro_schema : Avro schema for key
+            @:exception: SerializerError
+        '''
+        if value is not None:
+            if value_avro_schema is not None:
+                value = self.serializer.encode_record_with_schema(topic, value_avro_schema, value)
+            else:
+                self.log.error("Schema required for value serialization")
+                raise SerializerError("Avro schema required for value")
```
ewencp: In the Java serializer we auto-generate schemas for primitive types. Should we consider doing that here to make it easier to use? I think this is especially common for keys.

confluent_kafka/avro/AvroProducer.py
ewencp: Hmm, this argument order is kind of interesting. With only one signature, it's annoying to support everything cleanly. The main thing that's unusual here is that `key` comes after `value`. I get the reason for this is so you can produce values without keyword arguments, but would it make sense to try parsing a `*args` list so you don't have to use keyword arguments to get the natural ordering when using both keys and values? Or does that get too messy (or maybe not possible to always distinguish the right mapping)?

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        self.max_schemas_per_subject = max_schemas_per_subject
+        # subj => { schema => id }
+        self.subject_to_schema_ids = {}
+        # id => avro_schema
+        self.id_to_schema = {}
+        # subj => { schema => version }
+        self.subject_to_schema_versions = {}
+
+        # below classes were not hashable. hence defining them explicitely as a quick fix
+        schema.RecordSchema.__hash__ = self.hash_func
```
ewencp: Do this once at the module level so it only happens on import?

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+    def _send_request(self, url, method='GET', body=None, headers=None):
+        if body:
+            body = json.dumps(body)
+            body = body.encode('utf8')
+            print(body)
+        _headers = dict()
+        _headers["Accept"] = ACCEPT_HDR
+        if body:
+            _headers["Content-Length"] = str(len(body))
+            _headers["Content-Type"] = "application/json"
```
ewencp: This should really use the content type `application/vnd.schemaregistry.v1+json` like you're using in the accept header (which ensures it continues to work as expected even if we add a new version of the API).

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        elif method == 'PUT':
+            response = requests.put(url, body, headers=_headers)
+        elif method == 'DELETE':
+            response = requests.delete(url, headers=_headers)
+        result = json.loads(response.text)
+        return (result, response.status_code)
+
+    def _add_to_cache(self, cache, subject, schema, value):
+        if subject not in cache:
+            cache[subject] = {}
```
ewencp: nit: you can also use `collections.defaultdict` so this is automatic

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        avro_schema must be a parsed schema from the python avro library
+
+        Multiple instances of the same schema will result in cache misses.
+
+        @:param: subject: subject name
+        @:param: avro_schema: Avro schema to be registered
+        @:returns: schema_id: int value
+        """
+
+        schemas_to_id = self.subject_to_schema_ids.get(subject, {})
```
ewencp: `defaultdict` is probably better here too since it uses a function to generate the default value only if the entry doesn't exist, whereas this creates the default value on every call to `register()`

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        schema_id = schemas_to_id.get(avro_schema, None)
+        if schema_id != None:
+            return schema_id
+        # send it up
+        url = '/'.join([self.url, 'subjects', subject, 'versions'])
+        # body is { schema : json_string }
+
+        body = {'schema': json.dumps(avro_schema.to_json())}
+        result, code = self._send_request(url, method='POST', body=body)
+        if (code != 200):
```
ewencp: This probably isn't as robust as it should be -- other 2xx level codes would also indicate success. I don't think we use any today, but, for example, `201 CREATED` would be a reasonable response here.

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        Test the compatibility of a candidate parsed schema for a given subject.
+
+        By default the latest version is checked against.
+        @:param: subject: subject name
+        @:param: avro_schema: Avro schema
+        @:return: True if compatible, False if not compatible
+        """
+        url = '/'.join([self.url, 'compatibility', 'subjects', subject,
+                        'versions', str(version)])
+        avro_schema = body = {'schema': json.dumps(avro_schema.to_json())}
```
ewencp: Why is `avro_schema` being overwritten here? It doesn't look like it gets used again?

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+        if level not in VALID_LEVELS:
+            raise ClientError("Invalid level specified: %s" % (str(level)))
+
+        url = '/'.join([self.url, 'config'])
+        if subject:
+            url += '/' + subject
+
+        body = {"compatibility": level}
+        result, code = self._send_request(url, method='PUT', body=body)
+        if (code == 200):
```
ewencp: Other status codes should be valid here too

confluent_kafka/avro/CachedSchemaRegistryClient.py
```diff
@@ -0,0 +1,288 @@
...
+
+        @:param: subject: subject name
+        @:return: 'NONE','FULL','FORWARD', or 'BACKWARD'
+        """
+        url = '/'.join([self.url, 'config'])
+        if subject:
+            url += '/' + subject
+
+        result, code = self._send_request(url)
+        if (code == 200):
```
ewencp: Other status codes should be valid here too

confluent_kafka/avro/serializer/MessageSerializer.py
```diff
@@ -0,0 +1,241 @@
...
+            self.log.error(message)
+            raise SerializerError(message)
+
+        # cache writer
+        self.id_to_writers[schema_id] = avro.io.DatumWriter(schema)
+
+        return self.encode_record_with_schema_id(schema_id, record)
+
+    # subject = topic + suffix
+    def encode_record_for_topic(self, topic, record, is_key=False):
```
ewencp: This seems like an odd use case -- it's basically assuming that the latest schema will be compatible with the data you pass in? Are you using this in practice?

confluent_kafka/avro/serializer/MessageSerializer.py
```diff
@@ -0,0 +1,241 @@
...
+import avro.io
+
+from confluent_kafka.avro import ClientError
+from . import SerializerError
+
+MAGIC_BYTE = 0
+
+HAS_FAST = False
+try:
+    from fastavro.reader import read_data
```
ewencp: Should we also get the fast writer?

confluent_kafka/avro/serializer/Util.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
ewencp: Module names shouldn't be capitalized (i.e. the file name should just be `util.py`, not `Util.py`. In fact, this is true for all the filenames under this module (CachedSchemaRegistryClient, AvroProducer, etc).

README.md
```diff
@@ -46,6 +46,29 @@ while running:
...
+from confluent_kafka.avro.serializer.MessageSerializer import MessageSerializer
+
+value_avro_schema = Util.parse_schema_from_file('ValueSchema.avsc')
+key_avro_schema = Util.parse_schema_from_file('KeySchema.avsc')
+value = {"name": "Value"}
+key = {"name": "Key"}
+
+producer = Producer({'bootstrap.servers': 'mybroker,mybroker2'})
+
+schemaRegistryUrl = 'https://<host>:<port>'
```
ewencp: nit: these variables should use underscores instead of camel case

confluent_kafka/avro/AvroProducer.py
matthayes: Personally I prefer required keyword arguments in this case (i.e. `def produce(self, topic, *, value=None, value_avro_schema=None, etc.`)  It's hard to remember the argument order and easy to get it wrong (e.g. people may differ on whether they think key or value first is more natural).

README.md
```diff
@@ -46,6 +46,29 @@ while running:
...
+schemaRegistryClient = CachedSchemaRegistryClient(url=schemaRegistryUrl)
+serializer = MessageSerializer(schemaRegistryClient)
+
+avroProducer = AvroProducer(producer, serializer)
```
roopahc: @ewencp I started with inheritance, but it did not let me inherit class Producer. 
```
  File "test.py", line 4, in <module>
    class AvroProducer(Producer):
TypeError: Error when calling the metaclass bases
    type 'cimpl.Producer' is not an acceptable base type
```

confluent_kafka/avro/serializer/MessageSerializer.py
roopahc: This method is not being used anywhere. We can take it off .

criccomini:
>  I think the main thing that would need to be decided is the composition/delegation approach to the API vs inheriting from the Producer class

@roopahc initially had the class try to extend Producer. This didn't work because it's a C class, not a Python class underneath. I'm open to suggestions on how to fix this, but Google hasn't yielded anything. It appears you can't extend it. It gives an error (which I no longer have, sigh).

confluent_kafka/avro/AvroProducer.py
criccomini: Would it be OK to leave this as a separate Github issue/PR? I agree it's useful, but I don't think this PR needs to be exhaustive, and I think you should be able to bolt it on without backwards incompatible changes right?

confluent_kafka/avro/serializer/MessageSerializer.py
roopahc: @ewencp May be this can be considered as performance improvement and be part of separate PR?

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if key is not None:
+            if key_schema is not None:
+                key = self._serializer.encode_record_with_schema(topic, key_schema, key, True)
+            else:
+                log.error("Schema required for key serialization")
+                raise SerializerError("Avro schema required for key")
+
+        self._producer.produce(topic, value, key, *args, **kwargs)
+    def poll(self, timeout):
```
criccomini: Nit: newline  above `def poll`

confluent_kafka/avro/AvroProducer.py
criccomini: I agree having key after value is weird. kw args seems best

confluent_kafka/avro/AvroProducer.py
criccomini: Do you think a topic->schema mapping in the constructor would suffice?

criccomini: We'd have to do two: one for keys and one for values

ewencp: @criccomini I like that, but there are definitely use cases where people intentionally publish data with different schemas to the same topic. So I guess it kind of depends on how opinionated we want to be...

confluent_kafka/avro/AvroProducer.py
ewencp: Yeah, sure that's fine. Just a gap I saw.

confluent_kafka/avro/serializer/MessageSerializer.py
ewencp: Sure, that's fine. Was just surprised to see reader but not writer :)

confluent_kafka/avro/AvroProducer.py
criccomini: > but there are definitely use cases where people intentionally publish data with different schemas to the same topic

Won't the schema registry reject that (if compatibility is turned on)?

criccomini: I guess I agree it's possible. It's just inconvenient, lol. :)

confluent_kafka/avro/AvroProducer.py
criccomini: @roopahc just pointed out that the parent (wrapped) class' produce() method has nearly an identical order:

http://docs.confluent.io/3.0.1/clients/confluent-kafka-python/index.html#confluent_kafka.Producer.produce

Seems like keeping it as-is actually matches best with what's already there...

confluent_kafka/avro/AvroProducer.py
criccomini: Another annoying option would be to provide a wrapper class for key/value to emulate Java. This would allow us to exactly mirror the parent's produce() method. The wrapper would have like payload and schema fields. AvroProducer would take the schema field out and use it, and pass the serialized payload into the parent producer.

confluent_kafka/avro/AvroProducer.py
matthayes: I still think keyword args would be clearer.  The order is similar except with schema params inserted.  It's also not obvious whether the `callback`/`on_delivery` and `partitioner` args from producer will work as positional arguments after key_avro_schema and in what order they need to be or if they need to be keyword args.  It's probably best to have these explicitly declared as keyword args to reduce confusion and mistakes.

confluent_kafka/avro/AvroProducer.py
ewencp: Yeah, people turn off compatibility for those cases. I find the pattern questionable, but people do it...

The most convenient way for users might be to just use `produce(*args, **kwargs)` and sort out the details in the header of the method (i.e. 1 positional arg = value, 2 position args = (key, value), kwargs just work by name). The obvious drawback is lack of documentation in the method signature.

confluent_kafka/avro/AvroProducer.py
ewencp: Hah, it's possible we had this exact same discussion when reviewing the original Python API proposal and I have just completely forgotten about it...

roopahc: I have tried aligning the order of arguments with confluent_kafka.Producer().produce() method documentation.
http://docs.confluent.io/3.0.1/clients/confluent-kafka-python/index.html. It might be misleading if we change the order in AvroProducer class

criccomini: At this point, I yield. :) I don't have a super strong opinion. Can you guys give an *exact* method header for what you think is best? My Python foo is weak. Is this it?

    def produce(self, *args, **kwargs):

criccomini: @matthayes mentioned required keyword args. It looks like this is only 3.0 compatible [PEP-3102](http://legacy.python.org/dev/peps/pep-3102/). Based on @ewencp's comment below, I still think:

    produce(*args, **kwargs)

Is the best we can do right now.

README.md
```diff
@@ -46,6 +46,27 @@ while running:
...
+value_avro_schema = util.parse_schema_from_file('ValueSchema.avsc')
+key_avro_schema = util.parse_schema_from_file('KeySchema.avsc')
+value = {"name": "Value"}
+key = {"name": "Key"}
+
+producer = Producer({'bootstrap.servers': 'mybroker,mybroker2'})
+
+schema_registry_url = 'https://<host>:<port>'
+
+avroProducer = AvroProducer(producer, schema_registry_url, value_schema=value_avro_schema, key_schema=key_avro_schema)
```
criccomini: Either make the names `default_value_schema` and `default_key_schema` or switch to maps from topic to schema:

    default_key_schemas, default_value_schemas

confluent_kafka/avro/avro_producer.py
```diff
@@ -30,11 +32,14 @@ class AvroProducer(object):
         @:param: producer: confluent_kafka.Producer object
         @:param: message_serializer: Message Serializer object
     '''
-    def __init__(self, producer, message_serializer):  # real signature unknown; restored from __doc__
+    def __init__(self, producer, schema_registry_url, key_schema = None, value_schema = None):  # real signature unknown; restored from __doc__
         self._producer = producer
-        self._serializer = message_serializer
+        _cash_client= CachedSchemaRegistryClient(url=schema_registry_url)
```
criccomini: nit: _cache_client

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+        Handles schema registration, Message serialization.
+
+        Constructor takes below parameters
+
+        @:param: producer: confluent_kafka.Producer object
+        @:param: message_serializer: Message Serializer object
+    '''
+    def __init__(self, producer, schema_registry_url, key_schema = None, value_schema = None):  # real signature unknown; restored from __doc__
+        self._producer = producer
+        _cash_client= CachedSchemaRegistryClient(url=schema_registry_url)
```
criccomini: nit: _cache_client

criccomini: Also add a space between t=

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            @:param: value: A dictionary object
+            @:param: value_schema : Avro schema for value
+            @:param: key: A dictionary object
+            @:param: key_schema : Avro schema for key
+            @:exception: SerializerError
+        '''
+        # get schemas from  kwargs if defined
+        key_schema = kwargs.pop('key_schema', None)
+        value_schema = kwargs.pop('value_schema', None)
+        topic= kwargs.pop('topic', None)
```
criccomini: nit: space

criccomini: Giving up on commenting on this, but yo uhave a bunch of differen space styles:

    style= one
    style = two
    style=three

Please just use `style = two`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ClientError("Topic name not specified.")
+        value= kwargs.pop('value', None)
+        key= kwargs.pop('key', None)
+
+        # if key_schema is not initialized, fall back on default key_schema passed as construction param.
+        if key_schema is None:
+            key_schema=self.key_schema
+
+        # if value_schema is not initialized, fall back on default value_schema passed as construction param.
+        if value_schema is None:
```
criccomini: `if not value_schema:`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            value_schema=self.value_schema
+
+        if value is not None:
```
criccomini: `if value:`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            if value_schema is not None:
```
criccomini: `if value_schema`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+                value = self._serializer.encode_record_with_schema(topic, value_schema, value)
+            else:
+                log.error("Schema required for value serialization")
+                raise SerializerError("Avro schema required for value")
+
+        if key is not None:
```
criccomini: `if key:`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            if key_schema is not None:
```
criccomini: `if key_schema:`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            @:param: value_schema : Avro schema for value
+            @:param: key: A dictionary object
+            @:param: key_schema : Avro schema for key
+            @:exception: SerializerError
+        '''
+        # get schemas from  kwargs if defined
+        key_schema = kwargs.pop('key_schema', None)
+        value_schema = kwargs.pop('value_schema', None)
+        topic= kwargs.pop('topic', None)
+        if topic is None:
```
criccomini: `if not topic:`

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,91 @@
...
+            log.error("Topic name not specified.")
+            raise ClientError("Topic name not specified.")
+        value= kwargs.pop('value', None)
+        key= kwargs.pop('key', None)
+
+        # if key_schema is not initialized, fall back on default key_schema passed as construction param.
+        if key_schema is None:
```
criccomini: `if not key_schema:`

README.md
```diff
@@ -85,6 +106,9 @@ Install
 **Install from PyPi:**
 
     $ pip install confluent-kafka
+    
+    #for Avroproducer
```
criccomini: Nit: space between # and f

README.md
```diff
@@ -54,16 +54,16 @@ from confluent_kafka.avro.cached_schema_registry_client import CachedSchemaRegis
...
+default_key_schema = util.parse_schema_from_file('KeySchema.avsc')
 value = {"name": "Value"}
 key = {"name": "Key"}
 
 producer = Producer({'bootstrap.servers': 'mybroker,mybroker2'})
 
 schema_registry_url = 'https://<host>:<port>'
 
-avroProducer = AvroProducer(producer, schema_registry_url, value_schema=value_avro_schema, key_schema=key_avro_schema)
+avroProducer = AvroProducer(producer, schema_registry_url, value_schema=default_value_schema, key_schema=default_key_schema)
```
criccomini: sorry, I meant name the pram `default_value_schema` and `default_key_schema`, not the values you pass in.

    avroProducer = AvroProducer(producer, schema_registry_url, default_value_schema=value_schema, default_key_schema=key_schema)

criccomini:
@ewencp @matthayes look good?

matthayes:
LGTM.  I also tested the new code and it's working in my codebase, including parameters passed to `produce`.

criccomini:
Awesome, thanks Matt! @ewencp we've also moved forward internally with this code. Could you or @edenhill please merge? When that's done, we can open follow-up issues, and start on consumer.

criccomini:
ping

confluent_kafka/avro/serializer/Util.py
blbradley: Indeed, try to use `from confluent_kafka.avro.serializer import util` leads to `ImportError: cannot import name 'util'`.

confluent_kafka/avro/AvroProducer.py
blbradley: Personally, I think this PR should go ahead with supplying schemas for primitive types. Right now, this code only accepts records.

blbradley:
Thoughts:

* Schema generation for primitives should be included in the first patch. It would reduce the design changes required to implement it.
* `AvroProducer.produce` requires keyword arguments and `Producer.produce` does not. These should be consistent.

Anything we can do to get this moving? I'd gladly do the work.

criccomini:
@blbradley no one from Confluent is responding. We've been going back and forth on the produce() method for over a week. At the end of the day, we need to make forward progress.

> Anything we can do to get this moving? I'd gladly do the work.

Go for it! :) We're starting to look at the consumer side. If you want to forge ahead with the producer, I say 👍 .

criccomini:
Could someone with merge permissions please chime in here? @ewencp @edenhill

blbradley:
@criccomini I've actually decided to go with Kafka REST Proxy API. It wasn't any harder than trying to use this PR. So, I probably won't be doing the work. Sorry for getting your hopes up! Good Luck

criccomini:
K, np.

README.md
edenhill: I will look into the sub-classing problem, and I agree with @ewencp that it would be preferred to wrapping it.

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,96 @@
...
+        Constructor takes below parameters
+
+        @:param: producer: confluent_kafka.Producer object
+        @:param: schema_registry_url: Schema registry URL
+        @:param: default_key_schema: Optional avro schema for key
+        @:param: default_value_schema: Optional avro schema for value
+    '''
+
+    def __init__(self, producer, schema_registry_url, default_key_schema=None,
+                 default_value_schema=None):  # real signature unknown; restored from __doc__
```
ewencp: what is this comment referring to?

confluent_kafka/avro/avro_producer.py
```diff
@@ -0,0 +1,96 @@
...
+            Sends message to kafka by encoding with specified avro schema
+            @:param: topic: topic name
+            @:param: value: A dictionary object
+            @:param: value_schema : Avro schema for value
+            @:param: key: A dictionary object
+            @:param: key_schema : Avro schema for key
+            @:exception: SerializerError
+        '''
+        # get schemas from  kwargs if defined
+        key_schema = kwargs.pop('key_schema', None)
```
ewencp: You can avoid the extra logic below for defaults by just doing `key_schema = kwargs.pop('key_schema', self.key_schema)`

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -0,0 +1,326 @@
...
+
+        body = {'schema': json.dumps(avro_schema.to_json())}
+        result, code = self._send_request(url, method='POST', body=body)
+        if code == 409:
+            log.error("Incompatible Avro schema:" + str(code))
+            raise ClientError("Incompatible Avro schema:" + str(code))
+        elif code == 422:
+            log.error("Invalid Avro schema:" + str(code))
+            raise ClientError("Invalid Avro schema:" + str(code))
+        elif code != 200:
```
ewencp: I think I mentioned this elsewhere, but every status in the 200s is valid, so this should probably just be `elif code >= 300` (and maybe < 200).

tests/test_avro_producer.py
```diff
@@ -0,0 +1,73 @@
...
+    import unittest2 as unittest
+from confluent_kafka.avro.avro_producer import AvroProducer
+
+
+class TestAvroProducer(unittest.TestCase):
+    def setUp(self):
+        pass
+
+    def test_instantiation(self):
+        obj = AvroProducer(None, 'http://127.0.0.1:9002')
```
ewencp: If possible, it would be better to not depend on a specific port. This is a really common way to generate confusion when developers have something else running on that port or to cause random failures of tests on shared infrastructure like Jenkins. We had lots of these problems in Kafka until I converted all the places we bind to ports to use `0` (to select a random available port) and then constructing URLs to match whatever port ends up being bound.

roopahc:
@ewencp Accommodated the comments. Would you please merge it  :)

confluent_kafka/avro/serializer/Util.py
roopahc: @blbradley Fixed this.

roopahc:
@ewencp @edenhill ping :)

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -0,0 +1,326 @@
...
+            for header_name in headers:
+                _headers[header_name] = headers[header_name]
+        if method == 'GET':
+            response = requests.get(url, headers=_headers)
+        elif method == 'POST':
+            response = requests.post(url, body, headers=_headers)
+        elif method == 'PUT':
+            response = requests.put(url, body, headers=_headers)
+        elif method == 'DELETE':
+            response = requests.delete(url, headers=_headers)
```
ewencp: Should we have an exception thrown in an `else` clause here?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -0,0 +1,326 @@
...
+        url = '/'.join([self.url, 'config'])
+        if subject:
+            url += '/' + subject
+
+        result, code = self._send_request(url)
+        if code >= 200 and code <= 299:
+            compatibility = result.get('compatibility', None)
+
+        if not compatibility:
+            compatbility = result.get('compatibilityLevel')
```
ewencp: typo: compatbility -> compatibility

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -0,0 +1,326 @@
...
+
+        return compatbility
```
ewencp: Same typo as above

tests/test_avro_producer.py
```diff
@@ -0,0 +1,73 @@
...
+        obj = AvroProducer(None, 'http://127.0.0.1:0')
+        self.assertTrue(isinstance(obj, AvroProducer))
+        self.assertNotEqual(obj, None)
+
+    def test_Produce(self):
+        producer = AvroProducer(None, 'http://127.0.0.1:0')
+        valueSchema = util.parse_schema_from_file("basic_schema.avsc")
+        try:
+            producer.produce(topic='test', value={"name": 'abc"'}, value_schema=valueSchema, key='mykey')
+            self.fail("Should expect value_schema")
```
ewencp: value_schema is provided, I assume this should be "Should expect key_schema"?

tests/test_avro_producer.py
```diff
@@ -0,0 +1,73 @@
...
+        producer = AvroProducer(None, 'http://127.0.0.1:0', default_value_schema=value_schema)
+
+        try:
+            producer.produce(topic='test', value={"name": 'abc"'})
+        except Exception as e:
+            exc_type, exc_obj, exc_tb = sys.exc_info()
+            if exc_type.__name__ == 'SerializerError':
+                self.fail()
+
+    def test_Produce_arguments(self):
```
ewencp: This test overwrites the previous one since they have the same name

tests/test_avro_producer.py
```diff
@@ -0,0 +1,73 @@
...
+        except Exception as e:
+            pass
+
+    def test_Produce_arguments(self):
+        value_schema = util.parse_schema_from_file("basic_schema.avsc")
+        producer = AvroProducer(None, 'http://127.0.0.1:0', default_value_schema=value_schema)
+
+        try:
+            producer.produce(topic='test', value={"name": 'abc"'})
+        except Exception as e:
```
ewencp: Seems like this should succeed, does it fail sometime after the point that the SerializerError would occur if you didn't have the default value schema?

tests/setup_test_path.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
ewencp: Is this supposed to be imported somewhere? Unless it's guaranteed imported first it doesn't seem like it will ensure anything about import order?

tests/test_cached_client.py
```diff
@@ -0,0 +1,145 @@
...
+    import unittest2 as unittest
+
+from confluent_kafka.avro.cached_schema_registry_client import CachedSchemaRegistryClient
+from confluent_kafka.avro.serializer import util
+
+
+class TestCacheSchemaRegistryClient(unittest.TestCase):
+    def setUp(self):
+        # Make RecordSchema and PrimitiveSchema hashable
+        schema.RecordSchema.__hash__ = self.hash_func
```
ewencp: This seems to be getting duplicated a few different places. Any reason not to move it into, e.g., one of the `__init__.py` files that would guarantee it gets imported if you use anything from the `confluent_kafka.avro` module?

confluent_kafka/avro/__init__.py
```diff
@@ -0,0 +1,16 @@
+VALID_LEVELS = ['NONE', 'FULL', 'FORWARD', 'BACKWARD']
+
+
+class ClientError(Exception, object):
+    """Error thrown by Schema Registry clients"""
+
+    def __init__(self, message, http_code=-1):
```
roopahc: http_code is integer number in all the places. So -1 was considered. I shall change it to None.

tests/test_cached_client.py
roopahc: Moved to __init__.py

ewencp:
@edenhill How do you feel about merging this so we can get this (very large) patch landed, then deal with fixing the inheritance part once the C code supports subclassing?

edenhill:
Let's merge this to a WIP branch (e.g., schema_registry) and let work continue there, we want to avoid having incomplete features on master branch.

criccomini:
@edenhill what work remains? Are you just referring to eliminating the need for a wrap vs extend?

ewencp:
@criccomini Mainly the wrap vs extend -- the concern is that folks pick this stuff up from master and, whether we think it is reasonable or not given an unreleased version, start depending on it and complain loudly when you change things. The only other concern I would have is if anything pops up in how we have to handle the deserialization for consumers (though I don't really envision that being an issue).

criccomini:
I guess my counter-concern is just that this goes into a branch, diverges from master, and ends up in the dust bin after two months of work. We also have a consumer PR incoming today.

Do you or @edenhill have a suggestion about how to fix the wrap/extend issue?

edenhill:
Oh that certainly wont happen, despite the fact that we've been extremely slow on this we really want your Avro clients to be part of this project.

I'm currently working on making the C classes sub-classable and I will have a branch ready within a day that I'd like you try out.

criccomini:
👍

edenhill:
PR #63 provides sub-classable Producer and Consumer classes, please try it out.

confluent_kafka/avro/__init__.py
```diff
@@ -44,23 +46,28 @@ def __str__(self):
...
+class AvroProducer(Producer):
     """
         Kafka Producer client which does avro schema encoding to messages.
         Handles schema registration, Message serialization.
 
         Constructor takes below parameters
 
-        @:param: producer: confluent_kafka.Producer object
-        @:param: schema_registry_url: Schema registry URL
+        @:param: config: dict object with config parameters.
```
edenhill: Mention `schema.registry.url` and that it is required.

confluent_kafka/avro/__init__.py
```diff
@@ -1,6 +1,8 @@
 import logging
 import sys
 
+from confluent_kafka.cimpl import Producer
```
edenhill: Are you sure it should import from .cimpl directly and not from confluent.kafka?

confluent_kafka/avro/__init__.py
```diff
@@ -44,23 +46,28 @@ def __str__(self):
...
         @:param: default_value_schema: Optional avro schema for value
     """
 
-    def __init__(self, producer, schema_registry_url, default_key_schema=None,
+    def __init__(self, config, default_key_schema=None,
                  default_value_schema=None):
-        self._producer = producer
-        self._serializer = MessageSerializer(CachedSchemaRegistryClient(url=schema_registry_url))
+        if ('schema.registry.url' not in config.keys()):
+            raise ClientError("Missing parameter: schema.registry.url")
```
edenhill: Might want to use a standard ValueError

confluent_kafka/avro/__init__.py
edenhill: Possibly related to this:
https://travis-ci.org/confluentinc/confluent-kafka-python/jobs/175863152#L584

roopahc:
@edenhill  I am getting ` ImportError: dlopen(/Users/roopac/development/confluent-kafka-python/confluent_kafka/cimpl.so, 2): Symbol not found: __Py_ZeroStruct` Possibly not related to my code change

ewencp:
@roopahc Are you possibly building against a different python devel package than you are running it with? `Py_ZeroStruct` exists in 2.7 but not in python 3+.

roopahc:
@ewencp Thanks.. tests pass in my system... Errors seem unrelated.

edenhill:
Since the tests on master are okay I think this failure is related to your code changes.
I see that `avro` is included unconditionally in `__init__.py`, is this correct?

edenhill:
There's a bunch of errors like these:
`E       IOError: [Errno 2] No such file or directory: 'basic_schema.avsc'`

The py.test (or tox) runs from the top directory, so you'll need to reference the .avsc either by tests/ prefix or preferably by dirname()ing the script to get its directory.

I also feel it would be cleaner if the avro tests were in its own sub-directory, e.g., tests/avro, or avro/tests
py.test et.al are pretty good (too good) at finding tests so this shouldnt be a problem.

edenhill:
There's no docstring for the avro module

edenhill:
I updated docstring verification tests on master, so please rebase on latest master again as you fix the other issues. Thanks

edenhill:
@roopahc Check `pip install --help`, `pip install`  just takes a single module, you need to split that up into multiple lines

edenhill:
Also, please fixup/squash some commits, e.g., the review-fix stuff commits should be joined with their parent commits:
```
 $ git rebase -i master
 # Mark "fixes-of-fixes" as "f" and move them to the line below their logical parent commit,
 # repeat as necessary

 # When happy with the results (no wip/fix/temporary-tests commit), do a force push:
 $ git push --dry-run --force <remote> <branch>

  # Looks okay? Drop --dry-run
 ````

File: .travis.yml [hunk 55]
```diff
@@ -7,6 +7,7 @@ before_install:
   - pip install pytest-timeout
 install:
   - pip install -v --global-option=build_ext --global-option="-Itmp-build/include/" --global-option="-Ltmp-build/lib" .
+  - pip install --upgrade pip
```
edenhill: Is the .avro install requiring a newer pip version?
If so, we probably need to document what version is required.
And also move this to before_install prior to any other pip commands.

criccomini: I've found that a lot of older pips don't properly support extra_requires

criccomini: (We're just shooting in the dark)

criccomini:
Tests are passing now! Please have a look and merge this.

confluent_kafka/__init__.py
```diff
@@ -1,2 +1,2 @@
-__all__ = ['cimpl','kafkatest']
+__all__ = ['cimpl', 'avro', 'kafkatest']
```
edenhill: Should this be conditional based on install type? 
What happens if people dont install the avro submodule?

confluent_kafka/avro/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+            @:param: key: A dictionary object
+            @:param: key_schema : Avro schema for key
+            @:exception: SerializerError
+        """
+        # get schemas from  kwargs if defined
+        key_schema = kwargs.pop('key_schema', self._key_schema)
+        value_schema = kwargs.pop('value_schema', self._value_schema)
+        topic = kwargs.pop('topic', None)
+        if not topic:
+            log.error("Topic name not specified.")
```
edenhill: Why both log and exception?

confluent_kafka/__init__.py
roopahc: @edenhill This should not affect using the client normally, just with basic producer and consumer.
The only catch is you would not be able to use AvroProducer unless the dependencies are installed.

confluent_kafka/avro/__init__.py
```diff
@@ -98,22 +98,21 @@ def produce(self, **kwargs):
...
         topic = kwargs.pop('topic', None)
         if not topic:
-            log.error("Topic name not specified.")
             raise ClientError("Topic name not specified.")
         value = kwargs.pop('value', None)
         key = kwargs.pop('key', None)
         if value:
             if value_schema:
                 value = self._serializer.encode_record_with_schema(topic, value_schema, value)
+                log.debug("Value encoded successfully!")
```
edenhill: I dont think there is much value (pun!) to these log messages, they dont include any actual information, and there is an exception to counter-represent their coverage, so I think you should remove all the just-before-exception logs, and logs like this.

roopahc:
@edenhill @ewencp  is it good for merging ? :)

confluent_kafka/avro/__init__.py
edenhill: I just commented on one of these log occurences in the previous review, but all of them should be removed.

edenhill:
Yeah, I think we should just get rid of the excessive/redundant log messages.
after that we're good to go!

confluent_kafka/avro/__init__.py
roopahc: Done!

roopahc:
@edenhill done :)

edenhill:
LGTM!

edenhill:
Finally!

PR #44, opened 2016-09-14
Fix 32-bit arch build warnings

edenhill:
Fixes the compilation warnings when doing fat builds (32+64 bit) on OSX.


ewencp:
@edenhill The patches look fine, but what's the process for verifying, i.e. how do I build a fat build?

PR #50, opened 2016-09-30
Add timestamp() to librdkafka messages

qix:
`tstype` still needs to be exposed somehow, not sure of the best way to do that.

The whitespace in `confluent_kafka.c` is a mix of tabs/spaces as well as spaces in blank lines which my editor automatically removed.

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -327,10 +327,15 @@ static PyObject *Message_offset (Message *self, PyObject *ignore) {
 
 
 static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
-	if (self->timestamp > 0)
+	if (self->tstype != RD_KAFKA_TIMESTAMP_NOT_AVAILABLE)
 		return PyLong_FromLong(self->timestamp);
 	else
 		Py_RETURN_NONE;
+}
+
+
+static PyObject *Message_tstype (Message *self, PyObject *ignore) {
+  return PyLong_FromLong(self->tstype);
 }
 
 
@@ -373,6 +378,11 @@ static PyMethodDef Message_methods[] = {
 	{ "timestamp", (PyCFunction)Message_timestamp, METH_NOARGS,
 	  "  :returns: message timestamp or None if not available.\n"
 	  "  :rtype: int or None\n"
+	  "\n"
+	},
+	{ "tstype", (PyCFunction)Message_tstype, METH_NOARGS,
+	  "  :returns: message timestamp type.\n"
+	  "  :rtype: int\n"
 	  "\n"
 	},
 	{ NULL }
@@ -508,11 +518,7 @@ PyObject *Message_new0 (const rd_kafka_message_t *rkm) {
 	self->partition = rkm->partition;
 	self->offset = rkm->offset;
 
-	rd_kafka_timestamp_type_t tstype;
-	self->timestamp = rd_kafka_message_timestamp(rkm, &tstype);
-	if (tstype != RD_KAFKA_TIMESTAMP_NOT_AVAILABLE) {
-		// todo: make tstype available to python api
-	}
+	self->timestamp = rd_kafka_message_timestamp(rkm, &self->tstype);
 
 	return (PyObject *)self;
 }
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -326,6 +326,14 @@ static PyObject *Message_offset (Message *self, PyObject *ignore) {
 }
 
 
+static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
+	if (self->timestamp > 0)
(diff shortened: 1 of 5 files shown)


qix:
CLA signed.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -326,6 +326,14 @@ static PyObject *Message_offset (Message *self, PyObject *ignore) {
 }
 
 
+static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
+	if (self->timestamp > 0)
```
ewencp: Should this be `>=`?

confluent_kafka/src/confluent_kafka.c
```diff
@@ -362,6 +370,11 @@ static PyMethodDef Message_methods[] = {
 	  "  :rtype: int or None\n"
 	  "\n"
 	},
+	{ "timestamp", (PyCFunction)Message_timestamp, METH_NOARGS,
```
edenhill: It needs to return the timestamp type as well (none, append, create).
Im not sure whether it should be some class level constants or strings, probably the former. @ewencp?

confluent_kafka/src/confluent_kafka.c
```diff
@@ -495,6 +508,12 @@ PyObject *Message_new0 (const rd_kafka_message_t *rkm) {
 	self->partition = rkm->partition;
 	self->offset = rkm->offset;
 
+	rd_kafka_timestamp_type_t tstype;
```
edenhill: Put variable declarations at the beginning of the scope to be C89 safe

confluent_kafka/src/confluent_kafka.c
edenhill: It should depend on the tstype, not the value of the timestamp.
E.g., even a seemingly invalid timestamp with a proper tstype is valid (because it might mean something in the future)

qix:
Sorry about the delay, this was blocked on a librdkafka bug: https://github.com/edenhill/librdkafka/pull/858

* I added a `tstype()` function but didn't add the class level constants. I feel like that is the correct approach, but didn't know what to name them and that can be done in a separate commit.

* Message object wasn't available in test_Consumer so couldn't add a test there without large changes.

* The integration "test" doesn't have any assertions and isn't written as a test. I added tstype / timestamp to the `print` statement and manually verified.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -327,13 +327,18 @@ static PyObject *Message_offset (Message *self, PyObject *ignore) {
 
 
 static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
```
edenhill: I think it would be better to have a single timestamp() method that returns a tuple: tstype, timestamp

confluent_kafka/src/confluent_kafka.c
edenhill: Please add the timestamp type as a class level constant, e.g.:
TIMESTAMP_NOT_AVAILABLE, TIMESTAMP_CREATE_TIME, TIMESTAMP_APPEND_TIME
Map it directly to the librdkafka counterparts.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -362,6 +370,11 @@ static PyMethodDef Message_methods[] = {
 	  "  :rtype: int or None\n"
 	  "\n"
 	},
+	{ "timestamp", (PyCFunction)Message_timestamp, METH_NOARGS,
+	  "  :returns: message timestamp or None if not available.\n"
+	  "  :rtype: int or None\n"
```
edenhill: It should probably return a local timestamp in Python format (e.g float), and the doc string should be updated to reflect what is being returned.
The timestamp() method should also return the tstype (tstype,timestamp) tuple

qix:
Hi @edenhill, I've made most of the changes you requested, but don't have too much time to work on the very specifics especially if there's going to be a lot of back and forth. Feel free to make the remaining changes yourself if you want.

Some thoughts/issues:

* You do not have any unit tests that actually get a `Message` object, so adding that is a much larger change. I got `_Consumer` working but that meant commenting out the `.unsubscribe` line and tweaking it **a lot**.

* This package has minimal documentation already. I had to browse the source to figure out how to use it, I don't know where you're expecting me to add documentation.

* I don't believe converting it to a float is future safe. If you're expecting other uses of the timestamp field based on `tstype` they might not all be milliseconds so `/ 1000` isn't a safe assumption. I decided to leave as is with the tuple.

* I made the api return the tuple even if `tstype=0`. Might want to consider returning `None` in that case, but that will cause `(tstype, timestamp) = msg.timestamp()` to fail.

qix:
p.s. We released https://github.com/smyte/kafka_store today which uses a fork based on my original commit. As soon as we get this merged in and a new release is out I'll remove the fork and move it back over.

edenhill:
Thanks a lot for your effort!

PR #60, opened 2016-11-08
Travis CI integration

edenhill:


ewencp:
LGTM. The one caveat is that we may eventually need to use a prerelease version of librdkafka as we've done previously. The `bootstrap-librdkafka` script looks like it might eventually need to be upgraded. (In particular, if we want to test against the combo of stuff that will ultimately be released.) Then again, maybe that'll just turn into a larger matrix of librdkafka x confluent-kafka-python versions anyway, and extra branches from git for librdkafka may just be a special case.

edenhill:
You are right, but let's tackle that when we get there.
I wanted to keep it off master to avoid PR runs to fail just because of librdkafka master regression.

PR #63, opened 2016-11-09
Consumer, Producer, TopicPartition classes are now sub-classable

edenhill:


confluent_kafka/src/Consumer.c
```diff
@@ -55,7 +55,7 @@ static void Consumer_dealloc (Handle *self) {
 	if (self->rk)
 		rd_kafka_destroy(self->rk);
 
-	Py_TYPE(self)->tp_free((PyObject *)self);
+        Py_TYPE(self)->tp_free((PyObject *)self);
```
ewencp: We seem to be getting mixed tabs & spaces...

edenhill: The file originally has tabs for indention, which sucks but I probably coded it with the wrong .emacs file in place.
New changes use space for indent instead.
We can do a forklift untabify at some point.

edenhill:
@roopahc and @criccomini  I'd like you to try this out before I merge it.

roopahc:
@edenhill @ewencp I tested with AvroProducer class. It works good.. Thanks for this update.

edenhill:
@roopahc Thanks!

Issue #66, opened 2016-11-11
segfault after KeyboardInterrupt during consumer.poll(n)


fillest:
confluent-kafka==0.9.2, librdkafka 0.9.2
```
from confluent_kafka import Consumer, KafkaError

print 'consuming'
c = Consumer(**{
	'client.id': 'test-client',
	'bootstrap.servers': 'localhost',
	'group.id': 'test-group',
	'enable.auto.commit': False,
	'enable.auto.offset.store': True,
	'offset.store.method': 'broker',
	'enable.partition.eof': True,
	'default.topic.config': {'auto.offset.reset': 'earliest'},
})
c.subscribe(['test'])
while True:
    msg = c.poll(1)
```
I start the script, press ctrl-c, and get
```
$ python ktest.py
consuming
^CTraceback (most recent call last):
  File "ktest.py", line 84, in <module>
    msg = c.poll(1)
KeyboardInterrupt
Segmentation fault (core dumped)
```

Without a timeout the script doesn't interrupt on ctrl-c at all (this probably needs a separate issue, not sure yet)

edenhill:
Which Python version?

fillest:
Python 2.7.6

edenhill:
If you exit your script (such as with an unhandled exception) without calling `consumer.close()` then that will be done automatically by the Consumer class deallocator.
The cause of the crash is that this might trigger callbacks to be called, and the code was not set up to expect callbacks from the deallocator, thus the crash.
I'll add support for this, but as a workaround for now I suggest you catch the exception and call close()

PR #67, opened 2016-11-11
Consumer: set up callstate on dealloc to allow callbacks (#66)

edenhill:
If the application doesn't close() the consumer prior to
object destruction (such as on exit) then rd_kafka_destroy()
will be called by the deallocator, which in turn calls consumer_close()
which may trigger final callbacks to be called (rebalance_cb, offset_commit_cb, etc)


ewencp:
LGTM

PR #80, opened 2016-11-22
AvroConsumer for handling schema registry

### LINKED ISSUE #36: Integrate client with Confluent schema registry and Avro

roopahc:
Fixes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,33 @@ avroProducer = AvroProducer({'bootstrap.servers': 'mybroker,mybroker2', 'schema.
 avroProducer.produce(topic='my_topic', value=value, key=key)
 ```
 
+**AvroConsumer**
+```
+import sys
+import traceback
+from confluent_kafka import KafkaError
+from confluent_kafka.avro import AvroConsumer
+from confluent_kafka.avro.serializer import SerializerError
+
+c = AvroConsumer({'bootstrap.servers': 'mybroker,mybroker2', 'group.id': 'groupid', 'schema.registry.url': 'http://127.0.0.1:8081'})
+c.subscribe(['my_topic'])
+running = True
+while running:
+    try:
+        msg = c.poll(10)
+        if msg:
+            if not msg.error():
+                print(msg.value())
+            elif msg.error().code() != KafkaError._PARTITION_EOF:
+                print(msg.error())
+                running = False
+    except SerializerError as e:
+        print("Message deserialization failed for %s: %s" % (msg, e))
+        running = False
+        
+c.close()
+```
+
 See [examples](examples) for more examples.
 
 
@@ -100,7 +127,7 @@ Install
 
     $ pip install confluent-kafka
     
-    # for AvroProducer
+    # for AvroProducer or AvroConsumer
     $ pip install confluent-kafka[avro]
 
 
@@ -108,7 +135,7 @@ Install
 
     $ pip install .
 
-    # for AvroProducer
+    # for AvroProducer or AvroConsumer
     $ pip install .[avro]
 
 
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -4,7 +4,7 @@
 """
 import sys
 
-from confluent_kafka import Producer
+from confluent_kafka import Producer, Consumer
 
 VALID_LEVELS = ['NONE', 'FULL', 'FORWARD', 'BACKWARD']
 
@@ -112,3 +112,43 @@ class AvroProducer(Producer):
                 raise SerializerError("Avro schema required for key")
 
         super(AvroProducer, self).produce(topic, value, key, **kwargs)
+
+
+class AvroConsumer(Consumer):
+    """
+    Kafka Consumer client which does avro schema decoding of messages.
+    Handles message deserialization.
+
+    Constructor takes below parameters
+
+    @:param: config: dict object with config parameters containing url for schema registry (schema.registry.url).
+    """
+    def __init__(self, config):
+
+        if ('schema.registry.url' not in config.keys()):
+            raise ValueError("Missing parameter: schema.registry.url")
+        schem_registry_url = config["schema.registry.url"]
+        del config["schema.registry.url"]
+
+        super(AvroConsumer, self).__init__(config)
+        self._serializer = MessageSerializer(CachedSchemaRegistryClient(url=schem_registry_url))
+
+    def poll(self, timeout):
+        """
+        This is an overriden method from confluent_kafka.Consumer class. This handles message
+        deserialization using avro schema
+
+        @:param timeout
+        @:return message object with deserialized key and value as dict objects
+        """
+        message = super(AvroConsumer, self).poll(timeout)
+        if not message:
+            return message
+        if not message.error():
+            if message.value() is not None:
+                decoded_value = self._serializer.decode_message(message.value())
+                message.set_value(decoded_value)
+            if message.key() is not None:
+                decoded_key = self._serializer.decode_message(message.key())
+                message.set_key(decoded_key)
+        return message
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -332,6 +332,23 @@ static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
 			     self->timestamp);
 }
 
+static PyObject *Message_set_value (Message *self, PyObject *new_val) {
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -60,6 +60,24 @@ avroProducer = AvroProducer({'bootstrap.servers': 'mybroker,mybroker2', 'schema.
 avroProducer.produce(topic='my_topic', value=value, key=key)
 ```
 
+**AvroConsumer**
+```
+from confluent_kafka.avro import AvroConsumer
+
+c = AvroConsumer({'bootstrap.servers': 'mybroker,mybroker2', 'group.id': 'greoupid', 'schema.registry.url': 'http://127.0.0.1:9002'})
```
edenhill: typo: greoupid

the schema registry url port almost looks like 9092 (kafka port), while the default port is 8081, so might want to use that instead

README.md
```diff
@@ -60,6 +60,24 @@ avroProducer = AvroProducer({'bootstrap.servers': 'mybroker,mybroker2', 'schema.
...
+c.subscribe(['my_topic'])
+running = True
+while running:
+    msg = c.poll(10)
+    if msg:
+        if not msg.error():
+            print(msg.value())
```
edenhill: Do we want an example that actually uses the deserialized type? This looks just like a standard non-avro consumer.

confluent_kafka/avro/__init__.py
```diff
@@ -112,3 +112,44 @@ def produce(self, **kwargs):
...
+        This is an overriden method from confluent_kafka.Consumer class. This handles message
+        deserialization using avro schema
+
+        @:param timeout
+        @:return message object with deserialized key and value as dict objects
+        """
+        message = super(AvroConsumer, self).poll(timeout)
+        if not message:
+            return message
+        print(message)
```
edenhill: remove print

confluent_kafka/src/confluent_kafka.c
```diff
@@ -332,6 +332,19 @@ static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
 			     self->timestamp);
 }
 
+static PyObject *Message_set_value (Message *self, PyObject *new_val) {
+   Py_DECREF(self->value);
```
edenhill: I believe both value and key can be NULL, do:
````
if (self->value)
  Py_DECREF(self->value);
```

and same for key

confluent_kafka/src/confluent_kafka.c
```diff
@@ -391,6 +404,16 @@ static PyMethodDef Message_methods[] = {
 	  "  :rtype: (int, int)\n"
 	  "\n"
 	},
+	{ "set_value", (PyCFunction)Message_set_value, METH_O,
+	  "  :returns: None.\n"
```
edenhill: Missing docstring saying what the method actually does

README.md
roopahc: @edenhill How do we show the type of msg? msg value is going to be a dict object.

edenhill: No idea, but it is boring if the example is no different than the vanilla example, right? :)

README.md
```diff
@@ -64,13 +64,15 @@ avroProducer.produce(topic='my_topic', value=value, key=key)
...
 
-c = AvroConsumer({'bootstrap.servers': 'mybroker,mybroker2', 'group.id': 'greoupid', 'schema.registry.url': 'http://127.0.0.1:9002'})
+c = AvroConsumer({'bootstrap.servers': 'mybroker,mybroker2', 'group.id': 'groupid', 'schema.registry.url': 'http://127.0.0.1:8081'})
 c.subscribe(['my_topic'])
 running = True
 while running:
     msg = c.poll(10)
     if msg:
         if not msg.error():
+            if (msg.value() && isinstance(msg.value(), dict)):
```
edenhill: Is this ever not true?
From what I understand the deserializer will raise an exception if it can't deserialize a message.
Speaking of which, this example should probably have a proper try:..except block catching such errors, right?

confluent_kafka/src/confluent_kafka.c
```diff
@@ -333,16 +333,20 @@ static PyObject *Message_timestamp (Message *self, PyObject *ignore) {
 }
 
 static PyObject *Message_set_value (Message *self, PyObject *new_val) {
-   Py_DECREF(self->value);
-   self->value = new_val;
-   Py_INCREF(self->value);
+   if (self->value) {
+        Py_DECREF(self->value);
+        self->value = new_val;
```
edenhill: This only sets a new value if there was an old one, which isnt right.
Look at my initial comment.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -405,11 +409,13 @@ static PyMethodDef Message_methods[] = {
 	  "\n"
 	},
 	{ "set_value", (PyCFunction)Message_set_value, METH_O,
+	  "  Set the field 'Message.value' with new value.\n"
```
edenhill: Document param and type, see an example here:
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/src/confluent_kafka.c#L741

README.md
```diff
@@ -68,15 +68,17 @@ c = AvroConsumer({'bootstrap.servers': 'mybroker,mybroker2', 'group.id': 'groupi
...
-            running = False
+    try:
+        msg = c.poll(10)
+        if msg:
+            if not msg.error():
+                print(msg.value())
+            elif msg.error().code() != KafkaError._PARTITION_EOF:
+                print(msg.error())
+                running = False
+    except SerializerError:
```
edenhill: where is SerializerError defined? I can't seem to find it

README.md
```diff
@@ -68,15 +68,17 @@ c = AvroConsumer({'bootstrap.servers': 'mybroker,mybroker2', 'group.id': 'groupi
...
+        print("Unable to decode the message")
```
edenhill: print the exception message

confluent_kafka/src/confluent_kafka.c
```diff
@@ -410,12 +410,14 @@ static PyMethodDef Message_methods[] = {
 	},
 	{ "set_value", (PyCFunction)Message_set_value, METH_O,
 	  "  Set the field 'Message.value' with new value.\n"
+	  "  :param: PyObject value: Message.value.\n"
```
edenhill: PyObject is a C thing, right? Maybe you 'object' or 'anyobject' or such

confluent_kafka/avro/__init__.py
```diff
@@ -112,3 +112,43 @@ def produce(self, **kwargs):
...
+        deserialization using avro schema
+
+        @:param timeout
+        @:return message object with deserialized key and value as dict objects
+        """
+        message = super(AvroConsumer, self).poll(timeout)
+        if not message:
+            return message
+        if not message.error():
+            if message.value():
```
ewencp: I think you want to check `if message.value() is not None` here. An empty `bytes` can be a valid message but wouldn't be deserialized here. (In particular this can be an issue if the serialization format does not encode fields that contain their default values, resulting in an empty, but non-null, value.) Same for the check of the key below.

roopahc:
@edenhill @ewencp  Let me know if it looks OK now.

README.md
```diff
@@ -76,9 +80,11 @@ while running:
             elif msg.error().code() != KafkaError._PARTITION_EOF:
                 print(msg.error())
                 running = False
-    except SerializerError:
-        print("Unable to decode the message")
+    except SerializerError as e:
+        exc_type, exc_value, exc_traceback = sys.exc_info()
```
edenhill: This is nice, but maybe a bit excessive for an example, I just ment to print the exception string, something like:
` print("Message deserialization failed for %s: %s" % (msg, e))`

edenhill:
I'm happy with this. @ewencp ?

ewencp:
LGTM

edenhill:
Thanks again, great work!

PR #87, opened 2016-11-25
replace from .cimpl import * with explicit names.

co-dh:
pylint complains there is no Consumer in confluent_kafka
better explicit on what's imported than *.


ewencp:
@co-dh LGTM. Sorry for the confusion w/ CLABot, you can sign [here](http://clabot.confluent.io/cla).

edenhill:
Looks good, but seems a bit problematic to maintain, would it be possible to auto-generate the import list by dir(cimpl) somehow?

ewencp:
@edenhill I checked elsewhere and this seems to be common practice. It seems it's not that hard to maintain in practice since the number of exports doesn't change frequently (and losing one of them implies we'd simply need to add tests that exercise the use of that export).

co-dh:
I've signed, are we good to merge?

edenhill:
Thanks!

PR #97, opened 2016-12-07
Add timestamp= arg to produce()

edenhill:
Requires librdkafka >= 0.9.3 at build- and run-time

### DIFF
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -242,7 +242,49 @@ int32_t Producer_partitioner_cb (const rd_kafka_topic_t *rkt,
 }
 
 
+#if HAVE_PRODUCEV
+static rd_kafka_resp_err_t
+Producer_producev (Handle *self,
+                   const char *topic, int32_t partition,
+                   const void *value, size_t value_len,
+                   const void *key, size_t key_len,
+                   void *opaque, int64_t timestamp) {
 
+        return rd_kafka_producev(self->rk,
+                                 RD_KAFKA_V_MSGFLAGS(RD_KAFKA_MSG_F_COPY),
+                                 RD_KAFKA_V_TOPIC(topic),
+                                 RD_KAFKA_V_PARTITION(partition),
+                                 RD_KAFKA_V_KEY(key, (size_t)key_len),
+                                 RD_KAFKA_V_VALUE((void *)value,
+                                                  (size_t)value_len),
+                                 RD_KAFKA_V_TIMESTAMP(timestamp),
+                                 RD_KAFKA_V_OPAQUE(opaque),
+                                 RD_KAFKA_V_END);
+}
+#else
+
+static rd_kafka_resp_err_t
+Producer_produce0 (Handle *self,
+                   const char *topic, int32_t partition,
+                   const void *value, size_t value_len,
+                   const void *key, size_t key_len,
+                   void *opaque) {
+        rd_kafka_topic_t *rkt;
+        rd_kafka_resp_err_t err = RD_KAFKA_RESP_ERR_NO_ERROR;
+
+        if (!(rkt = rd_kafka_topic_new(self->rk, topic, NULL)))
+                return RD_KAFKA_RESP_ERR__INVALID_ARG;
+
+	if (rd_kafka_produce(rkt, partition, RD_KAFKA_MSG_F_COPY,
+			     (void *)value, value_len,
+			     (void *)key, key_len, opaque) == -1)
+                err = rd_kafka_last_error();
+
+        rd_kafka_topic_destroy(rkt);
+
+        return err;
+}
+#endif
 
 
 static PyObject *Producer_produce (Handle *self, PyObject *args,
@@ -251,7 +293,8 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 	int value_len = 0, key_len = 0;
 	int partition = RD_KAFKA_PARTITION_UA;
 	PyObject *dr_cb = NULL, *dr_cb2 = NULL, *partitioner_cb = NULL;
-	rd_kafka_topic_t *rkt;
+        long long timestamp = 0;
+        rd_kafka_resp_err_t err;
 	struct Producer_msgstate *msgstate;
 	static char *kws[] = { "topic",
 			       "value",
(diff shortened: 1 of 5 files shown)


confluent_kafka/src/Producer.c
```diff
@@ -242,7 +242,49 @@ int32_t Producer_partitioner_cb (const rd_kafka_topic_t *rkt,
...
+                   const void *value, size_t value_len,
+                   const void *key, size_t key_len,
+                   void *opaque, int64_t timestamp) {
+
+        return rd_kafka_producev(self->rk,
+                                 RD_KAFKA_V_MSGFLAGS(RD_KAFKA_MSG_F_COPY),
+                                 RD_KAFKA_V_TOPIC(topic),
+                                 RD_KAFKA_V_PARTITION(partition),
+                                 RD_KAFKA_V_KEY(key, (size_t)key_len),
+                                 RD_KAFKA_V_VALUE((void *)value,
```
ewencp: This is in librdkafka, but why is the key `const void *` but the value is `void *`?

confluent_kafka/src/Producer.c
```diff
@@ -260,25 +302,32 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 			       "callback",
 			       "on_delivery", /* Alias */
 			       "partitioner",
+#if HAVE_PRODUCEV
+                               "timestamp",
```
ewencp: Is there any concern that users might write code but then if they run it against a build of the correct version of this library but an old version of librdkafka, they'll get weird errors about unexpected arguments? (I am, in particular, concerned that this could become a FAQ.) Also, we need to generate a single set of API docs, so users may become confused when they can't use a parameter that is listed in the docs.

Another option would be to always parse the argument, but throw an exception with a useful message (including listing what version of librdkafka they have).

confluent_kafka/src/Producer.c
edenhill: Because librdkafka will free the value pointer (when done) if `msgflags & RD_KAFKA_MSG_F_FREE`, in which case it can't be const, while in all other cases it is.

confluent_kafka/src/Producer.c
edenhill: Aye, I like the second approach, throwing an exception.

edenhill:
re 0.9.2 & 0.9.3 on travis: I'm not sure we need to, the next version of this client will probably be released at the same time as the next librdkafka version, if not we can fix it then.

ewencp:
re 0.9.2 & 0.9.3: i'm not sure i understand why releasing together matters? Are you saying we'll remove 0.9.2 librdkafka before releasing 0.9.3 confluent-kafka-python and require 0.9.3 librdkafka? My concern is that we'll have branches that won't get automated testing, which seems dangerous.

edenhill:
Updated after review comments

confluent_kafka/src/Producer.c
```diff
@@ -251,7 +293,8 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 	int value_len = 0, key_len = 0;
 	int partition = RD_KAFKA_PARTITION_UA;
 	PyObject *dr_cb = NULL, *dr_cb2 = NULL, *partitioner_cb = NULL;
-	rd_kafka_topic_t *rkt;
+        long long timestamp = 0;
```
ewencp: nit: indentation is messed up

confluent_kafka/src/Producer.c
```diff
@@ -260,25 +303,31 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
 
 	if (!PyArg_ParseTupleAndKeywords(args, kwargs,
-					 "s|z#z#iOOO", kws,
+					 "s|z#z#iOOOL"
+                                         , kws,
 					 &topic, &value, &value_len,
 					 &key, &key_len, &partition,
-					 &dr_cb, &dr_cb2, &partitioner_cb))
+					 &dr_cb, &dr_cb2, &partitioner_cb,
+                                         &timestamp))
```
ewencp: nit: indentation

confluent_kafka/src/Producer.c
ewencp: Also, is setting this to 0 by default the correct behavior? If we are always going to force use of CreateTime, -1 seems more consistent with KIP-32 (since that's what's filled in for earlier produce request versions). Alternatively, should we be including support for LogAppendTime vs CreateTime?

confluent_kafka/src/Producer.c
```diff
@@ -288,27 +337,34 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
+
+        if (err) {
 		if (msgstate)
 			Producer_msgstate_destroy(msgstate);
 
 		if (err == RD_KAFKA_RESP_ERR__QUEUE_FULL)
 			PyErr_Format(PyExc_BufferError,
 				     "%s", rd_kafka_err2str(err));
-		else
+                else
```
ewencp: nit: whitespace

confluent_kafka/src/Producer.c
```diff
@@ -380,11 +436,13 @@ static PyMethodDef Producer_methods[] = {
 	  "  :param func on_delivery(err,msg): Delivery report callback to call "
 	  "(from :py:func:`poll()` or :py:func:`flush()`) on successful or "
 	  "failed delivery\n"
+          "  :param int timestamp: Message timestamp (CreateTime) in microseconds since epoch UTC (requires librdkafka >= v0.9.3, api.version.request=true, and broker >= 0.10.0.0). Default value is current time.\n"
```
ewencp: Coming back to what the default value should be, is librdkafka using 0 as a sentinel for current time?

confluent_kafka/src/Producer.c
edenhill: re indent: I'm trying to move away from tabs to spaces.
I should probably just do a forklift update to untabify everything in one commit.

edenhill: librdkafka uses 0 as a sentinel in which case it will use the current wall clock time.

I believe LogAppend vs LogCreate time should be enforced on the broker only (as is currently the case), and let supporting clients always send their LogCreateTime.
It seems to me to be the most forgiving scenario.

ewencp: re: indent: I see, on GH I guess it becomes obvious due to the extra `+` in the diff that the tabs on the previous line don't account for. I hate losing git history, but there's really no good time to do a mass reformatting, so we should probably just bite the bullet and do it once we hit a quiet period with PRs.

re: timestamps, I think we end up diverging just mildly from Java behavior. For some reason I had it in my head that if you continued to pass null to the Java producer it would fallback to LogAppendTime, but apparently newer clients will default to the current time on the client if the timestamp is null. Since we can't do that with a `long long`, the `0` sentinel value works fine -- I suppose not many folks will be trying to use the epoch as a real timestamp...

ewencp:
@edenhill I think this got lost in the shuffle, if you're happy with this I think it's ready to merge.

PR #105, opened 2017-01-10
Add optional timeout argument to Producer.flush()

edenhill:
Requires librdkafka >= v0.9.2

### DIFF
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -408,12 +408,31 @@ static PyObject *Producer_poll (Handle *self, PyObject *args,
 }
 
 
-static PyObject *Producer_flush (Handle *self, PyObject *ignore) {
-	while (rd_kafka_outq_len(self->rk) > 0) {
-		if (Producer_poll0(self, 500) == -1)
-			return NULL;
-	}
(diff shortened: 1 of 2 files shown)


confluent_kafka/src/Producer.c
```diff
@@ -408,12 +408,31 @@ static PyObject *Producer_poll (Handle *self, PyObject *args,
...
-			return NULL;
-	}
-	Py_RETURN_NONE;
+static PyObject *Producer_flush (Handle *self, PyObject *args,
+                                 PyObject *kwargs) {
+        double tmout = -1;
+        int qlen;
+        static char *kws[] = { "timeout", NULL };
+#if RD_KAFKA_VERSION >= 0x00090300
+        CallState cs;
```
ewencp: Hmm, so now that we have this alternate version it occurs to me -- shouldn't we always to the `CallState` steps, even in the old version? In fact, it's probably more important in the old version since there isn't even a timeout, so we really want the GIL unlocked.

edenhill: You are absolutely right!

edenhill: Producer_poll0(); already does the CallState dance, so we're good here.

PR #109, opened 2017-01-19
Fixes in AvroConsumer

t0mk:
poll() of Consumer returns None on timeout expiration. This PR is just to fix the handling of poll() == None in AvroConsumer.

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -53,8 +53,7 @@ class ClientError(Exception):
 
 
 from confluent_kafka.avro.cached_schema_registry_client import CachedSchemaRegistryClient
-from confluent_kafka.avro.serializer import (SerializerError,
-                                             KeySerializerError,
+from confluent_kafka.avro.serializer import (KeySerializerError,
                                              ValueSerializerError)
 from confluent_kafka.avro.serializer.message_serializer import MessageSerializer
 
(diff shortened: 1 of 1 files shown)


t0mk:
@ConfluentCLABot I have no problem signing it, just don't feel like searching for it. It  would be cool if the bot posted a link to where the CLA can be signed.

ewencp:
@t0mk It's at http://clabot.confluent.io/cla

confluent_kafka/avro/__init__.py
```diff
@@ -53,8 +53,7 @@ def __str__(self):
 
 
 from confluent_kafka.avro.cached_schema_registry_client import CachedSchemaRegistryClient
-from confluent_kafka.avro.serializer import (SerializerError,
-                                             KeySerializerError,
+from confluent_kafka.avro.serializer import (KeySerializerError,
```
ewencp: I think we want the `SerializerError` import so users only have to `from confluent_kafka.avro import (X, Y, Z)` and don't have to dig all the way into `confluent_kafka.avro.serializer`.

t0mk: hi, OK, I will revert this. I only removed it because the SerializerError is not used in the __init__.py script. I saw it as unnecessary import..

t0mk:
I did the change as requested in @edenhill's review.

ewencp:
LGTM, thanks for the fix @t0mk!

Issue #45, opened 2016-09-16
build librdkafka when installing confluent-kafka


felixthecat:
currently when i do a `pip install confluent-kafka`, it fails because i haven't built librdkafka yet. it'd be nice if your setup.py could also build librdkafka to make the installation process easier. there are multiple internal services using our client that's a wrapper around confluent-kafka, and we'd like it if they could just do a simple `pip install our-client`. this is a little more difficult for us since we then have to automate the build process on our end. 

basically i'm hoping you can do something similar to what `librabbitmq` (https://pypi.python.org/pypi/librabbitmq) is doing.

ZmeiGorynych:
I run into apparently the same error - can it be fixed manually?

```
   x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fno-strict-aliasing -Wdate-time -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security -fPIC -I/usr/include/python2.7 -c confluent_kafka/src/confluent_kafka.c -o build/temp.linux-x86_64-2.7/confluent_kafka/src/confluent_kafka.o
    In file included from confluent_kafka/src/confluent_kafka.c:17:0:
    confluent_kafka/src/confluent_kafka.h:20:32: fatal error: librdkafka/rdkafka.h: No such file or directory
    compilation terminated.
    error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```

edenhill:
@ZmeiGorynych Download, build and install librdkafka manually:
```
curl -L https://github.com/edenhill/librdkafka/archive/v0.9.2-RC1.tar.gz | tar xzf -
cd librdkafka-0.9.2-RC1/
./configure --prefix=/usr
make -j
sudo make install
```

Then try to build/install confluent-kafka-python again.

edenhill:
Re issue in general: we'll investigate what bootstrapping librdkafka through setup.py might look like.

ZmeiGorynych:
That seems to have worked, thanks!
The only strangeness is that I had to do  `sudo pip install confluent-kafka` as installing as me gave 
`OSError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/dist-packages/confluent_kafka-0.9.1.2.dist-info'`

ZmeiGorynych:
Please, what's the equivalent for Windows?

[...]

PR #121, opened 2017-02-07
Bundle librdkafka when building python package

wooparadog:
This is a proposal for #45 .

Methods:

1. Use git subtree to import librdkafka@v0.9.3 into this repo
2. Modify header include location
3. Build command will first run `configure` then `make libs`, then update linking libraries for c extension.
4. c extension will use librdkafka.a for building.


edenhill:
Thank you for your effort on this PR.

I'm afraid this risks not being merged for the following reasons:
 * this is not the path we're aiming to go down, we're looking into bundling prebuilt librdkafka binaries on platforms where dependencies are a chore, such as OSX (when brew is not available) and Windows. For Linux there are already available RPMs and Debs that the user should install, which also takes care of librdkafka's own dependencies. (and there are brew packages on OSX).
 * While bootstrapping and building librdkafka with the Python client makes sense in some scenarios it is error prone and I'm afraid the end-user experience will suffer. E.g., there is no way to enforce librdkafka dependencies.
 * this will ignore an already installed librdkafka, thus making it hard to upgrade only librdkafka.
 * did you see the existing librdkafka bootstrap script in tools/ that does this? See .travis.yml for an example how to use it.

wooparadog:
I'm suggesting a method for easier installation. I fully agree this pull request should not be merged. 
Bootstrapping librdkafka does pose a lot of problems, but as a python user, it's a little frustrating that I need to install a system package before running `pip install`. 

Maybe `confluent-kafka-python` can provide an alternative installation method (`pip install confluent-kafka-python[librdkafka]`) to bootstrap librdkafka? And verbosely tell users the problems with this method?

edenhill:
That's not a bad idea, will look into it.
Thanks!

PR #123, opened 2017-02-08
Test py35 and librdkafka 0.9.3

kwilcox:
py35 is failing with a threading issue on both librdkafka 0.9.3 and the current master. I can reproduce it locally but haven't dug any further. Travis will explain more!


edenhill:
Thanks Kyle, will look into this.

PR #142, opened 2017-03-01
updated readme

mhowlett:


README.md
```diff
@@ -1,8 +1,27 @@
-Confluent's Apache Kafka client for Python
-==========================================
+Confluent's Python Client for Apache Kafka<sup>TM</sup>
```
edenhill: client, not Client.

mhowlett: in fact this should be capitalized (it's a heading) and words in other headings on this page should also be capitalized.

mhowlett: rule of thumb from http://grammar.yourdictionary.com/capitalization/rules-for-capitalization-in-titles.html "Capitalize all words in titles of publications and documents, except a, an, the, at, by, for, in, of, on, to, up, and, as, but, or, and nor."

edenhill:
LGTM

PR #154, opened 2017-03-13
Add 'get_metadata' to Producer and Consumer.

stephan-hof:
Hi,

in my application the need came up to get some metadata from the broker.
As far as I can see, `confluent-kafka-python` has no API to retrieve this easily.
This pull request wraps the `rd_kafka_metadata` and returns the corresponding structures in python.

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -643,6 +643,15 @@ static PyMethodDef Consumer_methods[] = {
 	  "  :rtype: None\n"
 	  "\n"
 	},
+
+    {
+        "get_metadata",
+        (PyCFunction)Kafka_generic_get_metadata,
+        METH_VARARGS|METH_KEYWORDS,
+        Kafka_generic_get_metadata_doc
+    },
+
+
 	{ NULL }
 };
 
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -445,7 +445,6 @@ static PyObject *Producer_flush (Handle *self, PyObject *args,
         return PyLong_FromLong(qlen);
 }
 
-
 static PyMethodDef Producer_methods[] = {
 	{ "produce", (PyCFunction)Producer_produce,
 	  METH_VARARGS|METH_KEYWORDS,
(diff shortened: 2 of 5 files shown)


stephan-hof:
Anything you need from me to get it accepted ? 
I thought I have signed the CLA. [clabot:check]

edenhill:
Thank you for this PR, @stephan-hof !

The plan is to add a richer metadata API to librdkafka (also supporting KIP-4) and the Python bindings should make use of this new API rather than the old one, so I'm going to need to put this PR on hold for a while.

jeffwidman:
> The plan is to add a richer metadata API to librdkafka (also supporting KIP-4)

Is there an associated issue or PR to track progress on this?

edenhill:
Added issue to track it:
https://github.com/edenhill/librdkafka/issues/1158

stephan-hof:
All of what has been said so far makes much sense to me.
So I guess the pull request in its current form will never be accepted, because he is using a deprecated API. When the new API is there I'm pretty sure the code to get the metadata will look differently.
=> I'll close this pull request and may add a new one when the new API is there.
Until the new API is there I'll use my branch where this pull request originates from.

edenhill:
Thank you!

tbsaunde:
I'm not sure it really makes a lot of sense to hold off here.  Looking at the librdkafka issue and the kafka wire protocol changes to the metadata API it seems like it should be pretty easy to keep compatability with python users while changing to the newer librdkafka API when implemented, and exposing more information from the brokers to python clients.  In any case the old librdkafka API can't go away until a soname bump if that ever happens, and  there's plenty of C users of this API already, so having one more consumer that eventually needs to change to a better API doesn't seem like a big deal.  its also worth considering that there's already 8 months of librdkafka versions out there that only provide the rd_kafka_metadata() API, so if someone can only change the python wrapper for some reason falling back to the older API there would be useful for them.  So it seems to me it would be good to merge something like this pr and then work on adding new API to librdkafka and changing consumers to use it.

edenhill:
You are right, sorry for the long delay.
We'll revisit this PR shortly.

edenhill:
We should revamp this and change it to mimic the java ListTopics() API

tbsaunde:
I did that in #309 (I don't know how to update someone else's PR)

PR #155, opened 2017-03-13
Fix memory leak on errors.

stephan-hof:
The reference count handling in `cfl_PyErr_Format` is not done right.
The count of `_eo` must be decreased after calling `PyErr_SetObject`, otherwise the object is leaking.


edenhill:
Thanks for fixing this!
As soon as you sign the CLA I'll merge it.

stephan-hof:
[clabot:check]

edenhill:
Thank you!

PR #160, opened 2017-03-23
ENH Allow to pass custom schema registry instance.

rolando:
This allows, for example, to use a static schema registry.

### DIFF
--- a/examples/integration_test.py
+++ b/examples/integration_test.py
@@ -38,7 +38,7 @@ except ImportError as e:
 bootstrap_servers = None
 
 # Confluent schema-registry
-schema_registry = None
+schema_registry_url = None
 
 # Topic to use
 topic = 'test'
@@ -54,6 +54,20 @@ good_stats_cb_result = False
 
 def error_cb (err):
     print('Error: %s' % err)
+
+
+class InMemorySchemaRegistry(object):
+
+    schemas = {}
+    next_idx = 0
+
+    def register(self, subject, schema):
+        self.next_idx += 1
+        self.schemas[self.next_idx] = schema
+        return self.next_idx
+
+    def get_by_id(self, idx):
+        return self.schemas.get(idx)
 
 
 class MyTestDr(object):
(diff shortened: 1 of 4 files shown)


rolando:
[clabot:check]

examples/integration_test.py
```diff
@@ -56,6 +56,22 @@ def error_cb (err):
     print('Error: %s' % err)
 
 
+class SchemaRegistry(object):
```
edenhill: Can we call this something else since looking at the sub-sequent code doesn't make it clear if the SchemaRegistry is a real thing or a mock just looking at the name. It also risks colliding with a future real schema registry class.

E.g., "MySchemaRegistry"

tests/avro/test_avro_producer.py
```diff
@@ -82,3 +86,15 @@ def test_produce_primitive_key_and_value(self):
         producer = AvroProducer({'schema.registry.url': 'http://127.0.0.1:9001'})
         with self.assertRaises(ConnectionError):  # Unexistent schema-registry
             producer.produce(topic='test', value=32., value_schema=value_schema, key='mykey', key_schema=key_schema)
+
+    def test_product_with_custom_registry(self):
```
edenhill: Should probably be produce, not product

tests/avro/test_avro_producer.py
```diff
@@ -82,3 +86,15 @@ def test_produce_primitive_key_and_value(self):
...
+        schema_registry = MockSchemaRegistryClient()
+        value_schema = avro.load(os.path.join(avsc_dir, "basic_schema.avsc"))
+        key_schema = avro.load(os.path.join(avsc_dir, "primitive_string.avsc"))
+        producer = AvroProducer({}, schema_registry=schema_registry)
+        producer.produce(topic='test', value={"name": 'abc"'}, value_schema=value_schema, key='mykey', key_schema=key_schema)
+
+    def test_product_with_custom_registry_and_registry_url(self):
```
edenhill: dito

examples/integration_test.py
rolando: At first I was thinking a global ``schema_registry = SchemaRegistry()`` or ``schema_registry = CustomSchemaRegistry()`` but went with class methods. The latter looks good to me. What do you think?

rolando: Opted for `InMemorySchemaRegistry` and no global instance.

tests/avro/test_avro_producer.py
rolando: Oh, typo.

rolando:
@edenhill PR updated, please check again.

edenhill:
Thanks for this, very neat!

rolando:
Great!

PR #175, opened 2017-04-10
None conf values are now converted to NULL (closes #133)

### LINKED ISSUE #133: The on_assign callback for subscribe provides an empty partitions list
We do custom partition/offset tracking, so we run our consumers with `enable.auto.commit=False` and `group.id=None` and we call `Consumer.subscribe` with a callback set for `on_assign`. Usually, our initial call to `subscribe` results in an empty `partitions` parameter to the `on_assign` callback. We added retry logic around our `subscribe` calls, and we generally get a valid `partitions` list in `on_assign` within a minute or so.

From this, it seems like initially the `Consumer` isn't really ready to field `subscribe` calls with `on_assign` but eventually becomes ready (perhaps related to initial metadata requests running in the background?). Is there some way for users to determine this readiness and delay calls to subscribe? Our retries are a bit hacky and I'd like to remove them.

Or, is our usage horribly wrong, requiring a new approach to our offset management?

edenhill:
They were previously converted to the string "None"

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -1302,16 +1302,21 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
 		/*
 		 * Pass configuration property through to librdkafka.
 		 */
-		if (!(vs = cfl_PyObject_Unistr(vo))) {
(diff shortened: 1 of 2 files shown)


edenhill:
@hqin @mhowlett please review

PR #182, opened 2017-04-28
Implement pause/resume support

### LINKED ISSUE #120: Feature request: pause/resume support
Hi all,
I've got a question about the difference in behaviour of consumer.poll(0).
If you check the java doc for the consumer ( [https://kafka.apache.org/0100/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html](url)) it states the following for the poll(0) function:

> If 0, returns immediately with any records that are available currently in the buffer, else returns empty

I tried to do the same in python but when executing the code it already gives me the next record. 
Sample of the code I am using:
`

                        future_response= executor.submit(do_processing,msg.value(),self.logger, kafka_key.generated_uuid,False,False)

                        is_running = True
                        while is_running == True:
                            is_running=future_response.running()
                            self.logger.info("Still waiting for the future to finish. Going into sleep")
                            msgx=consumer.poll(0)
                            time.sleep(0.5)
`
If I do a print statement of msgx.value() in the while loop it gives me the next record. 
Why is there a difference in implementation ? 
Is there a way to do manual heartbeating in python?

Since the future might take a while to get the result I want to avoid to set session.timeout.ms to a very high value.

dangra:
Adds support for pause()/resume()

Closes #120


dangra:
[clabot:check]

dangra:
hello @edenhill, I think this PR is ready to review.

confluent_kafka/src/Consumer.c
```diff
@@ -611,6 +660,30 @@ static PyMethodDef Consumer_methods[] = {
...
 	  "\n"
 	},
+	{ "pause", (PyCFunction)Consumer_pause,
+	  METH_VARARGS|METH_KEYWORDS,
+	  ".. py:function:: pause(partitions)\n"
+	  "\n"
+	  "  Pause consumption for the provided list of partitions.\n"
+	  "\n"
+	  "  :param list(TopicPartition) partitions: List of topic+partitions "
+	  "to pause. \m"
```
chekunkov: @dangra `\m` looks like a typo, maybe `s/ \m/\n/`?

edenhill:
Thank you @dangra, will review this shortly

confluent_kafka/src/Consumer.c
dangra: @chekunkov true, thanks.

confluent_kafka/src/Consumer.c
```diff
@@ -391,6 +391,55 @@ static PyObject *Consumer_position (Handle *self, PyObject *args,
...
+    static char *kws[] = {"partitions", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", kws, &plist))
+        return NULL;
+
+    if (!(c_parts = py_to_c_parts(plist)))
+        return NULL;
+
+    err = rd_kafka_pause_partitions(self->rk, c_parts);
+    if (err) {
```
edenhill: move list_destroy(c_parts) up here so it won't leak when !err

confluent_kafka/src/Consumer.c
```diff
@@ -391,6 +391,55 @@ static PyObject *Consumer_position (Handle *self, PyObject *args,
...
+    static char *kws[] = {"partitions", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", kws, &plist))
+        return NULL;
+
+    if (!(c_parts = py_to_c_parts(plist)))
+        return NULL;
+
+    err = rd_kafka_resume_partitions(self->rk, c_parts);
+    if (err) {
```
edenhill: c_parts will leak in the success case.
Move list_destroy(c_parts) up here so it won't leak when !err

edenhill:
Overall LGTM.
Need to fix a memory leak and add an integration test to examples/integration_test.py.
Suggest the integration test is something like:
 * assign() two partitions with offset set to beginning (or auto.offset.reset=earliest)
 * consume a single message
 * pause both partitions
 * try to consume, make sure nothing is received for 1000ms
 * resume partitions
 * try to consume, make sure message is received.

Use the topic used by the other integration tests and make sure this test runs after the other tests so the topic is primed with messages by producer tests.

dangra:
> Need to fix a memory leak and add an integration test to examples/integration_test.py.

Fixed. thanks

rdonnelly3:
Is there an update on this feature?  We have a long running process between consumes that could definitely utilize the pause/resume feature.  Curious on when this will get released.  If it won't be for awhile are there any recommendations on alternative ways of handling long processing between polls?

edenhill:
@rdonnelly3 You don't need to call poll() to retain your group membership (which was the case in older versions of the Java consumer), a background thread is doing all the protocol level heartbeats, etc.
But you will need to call poll() to find out when the group rebalances (revoke + assign) so that your consumer is not processing messages for partitions it is no longer the assigned member for.
Pause/resume does not really change that so I don't believe it will be a magic bullet.

After a long-time processing of a message you find that the given partition has been revoked you should typically discard that message and not commit the offset, but start over with the next message returned from poll, coming from the new assignment.

For long time processing you should also disable automatic commits, since message offsets are stored for auto commit when the message is passed to the application which means that it is likely that the offset will be committed while you process the message.

edenhill:
Thank you for this!

PR #191, opened 2017-05-14
Make build flake8 clean and add flake8 precheck to .travis.yml

ewencp:


ewencp:
This should get us `flake8` clean and adds `flake8` to the travis config so we'll stay so. Most of the changes are straightforward, but a bit of refactoring was needed to handle some weird import order dependencies. The few `# noqa` annotations are for `__init__.py` files where unused imports are intentional.

confluent_kafka/kafkatest/verifiable_consumer.py
```diff
@@ -131,13 +126,10 @@ def on_commit (self, err, partitions):
...
 
-
-    def do_commit (self, immediate=False, async=None):
+    def do_commit(self, immediate=False, async=None):
         """ Commit every 1000 messages or whenever there is a consume timeout
             or immediate. """
-        if (self.use_auto_commit or
-            self.consumed_msgs_at_last_commit + (0 if immediate else 1000) >
-            self.consumed_msgs):
+        if self.use_auto_commit or self.consumed_msgs_at_last_commit + (0 if immediate else 1000) > self.consumed_msgs:
```
edenhill: I dont see how this is easier to read, but I'm not fighting the flake8ers today

confluent_kafka/kafkatest/verifiable_consumer.py
```diff
@@ -178,11 +169,11 @@ def msg_consume (self, msg):
             return
 
         if False:
-            self.dbg('Read msg from %s [%d] @ %d' % \
+            self.dbg('Read msg from %s [%d] @ %d' %
                      (msg.topic(), msg.partition(), msg.offset()))
 
         if self.max_msgs >= 0 and self.consumed_msgs >= self.max_msgs:
-            return # ignore extra messages
+            return  # ignore extra messages
```
edenhill: double space intentional?

tests/test_Consumer.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
-from confluent_kafka import Consumer, TopicPartition, KafkaError, KafkaException, TIMESTAMP_NOT_AVAILABLE, OFFSET_INVALID, libversion
+from confluent_kafka import (Consumer, TopicPartition, KafkaError, KafkaException, TIMESTAMP_NOT_AVAILABLE,
+                             OFFSET_INVALID, libversion)
```
edenhill: ..but split over multiple lines here?
This inconsistency is at least consistent with Python's consistent inconsistency

confluent_kafka/kafkatest/verifiable_consumer.py
ewencp: It actually just didn't like the `>` comparison being split across multiple lines, I've tweaked the style a bit. flake8 doesn't complain about it, but using parentheses when they aren't necessary isn't normal python style, but here it does avoid the `\` if the condition is split across multiple lines.

confluent_kafka/kafkatest/verifiable_consumer.py
ewencp: yeah, inline comments are supposed to get 2 spaces. We can choose some warnings to disable if they get too annoying.

tests/test_Consumer.py
ewencp: This one hits the line length limit, requiring it to be split over multiple lines.

PR #193, opened 2017-05-16
Fastavro Producer + Better Consumer Support

lewis-grimmett-zefr:
Adding support to the AvroProducer to use the `schemaless_writer` from fastavro - we've seen write speed for large records improve by a factor of 4.

The key difference between `fastavro` and `avro` is that fastavro expects the schema to be a python dict, so I've added an `id_to_json_schema` cache to the CachedSchemaRegistryClient.

Trying to use `schema.to_json()` for large and complex schemas (unions, lists of record objects etc) gave the following error:

```
  File "confluent-kafka-python-1/avro-load-test.py", line 97, in test_fastavro
    fastavro.schemaless_writer(b, schema.to_json(), record)
  File "fastavro/_writer.py", line 498, in fastavro._writer.schemaless_writer (fastavro/_writer.c:13154)
  File "fastavro/_writer.py", line 511, in fastavro._writer.schemaless_writer (fastavro/_writer.c:13012)
  File "fastavro/_writer.py", line 366, in fastavro._writer.acquaint_schema (fastavro/_writer.c:10916)
  File "fastavro/_schema.py", line 57, in fastavro._schema.extract_named_schemas_into_repo (fastavro/_schema.c:2862)
  File "fastavro/_schema.py", line 109, in fastavro._schema.extract_named_schemas_into_repo (fastavro/_schema.c:2696)
  File "fastavro/_schema.py", line 60, in fastavro._schema.extract_named_schemas_into_repo (fastavro/_schema.c:2032)
  File "fastavro/_schema.py", line 109, in fastavro._schema.extract_named_schemas_into_repo (fastavro/_schema.c:2696)
  File "fastavro/_schema.py", line 60, in fastavro._schema.extract_named_schemas_into_repo (fastavro/_schema.c:2032)
  File "fastavro/_schema.py", line 74, in fastavro._schema.extract_named_schemas_into_repo (fastavro/_schema.c:2117)
TypeError: unhashable type: 'ImmutableDict'
```
This issue occurred with the reading records as well - using fastavro was consistently failing in the schema and falling back to the default avro implementation.

Calling `schema.to_json()` was also significantly slower than  using a stored json schema, hence also using the `id_to_json_schema` cache in the fastavro consumer. In one test, calling `schem.to_json()` every time cancelled out the performance improvements of using fastavro.

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -36,7 +36,7 @@ log = logging.getLogger(__name__)
 
 HAS_FAST = False
 try:
-    from fastavro.reader import read_data
+    import fastavro  # noqa: F401
 
     HAS_FAST = True
 except:
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -32,6 +32,14 @@ from . import loads
 ACCEPT_HDR = "application/vnd.schemaregistry.v1+json, application/vnd.schemaregistry+json, application/json"
 log = logging.getLogger(__name__)
 
+HAS_FAST = False
+try:
+    from fastavro.reader import read_data
+
+    HAS_FAST = True
+except:
+    pass
+
 
 class CachedSchemaRegistryClient(object):
     """
@@ -50,11 +58,13 @@ class CachedSchemaRegistryClient(object):
         self.url = url.rstrip('/')
 
         self.max_schemas_per_subject = max_schemas_per_subject
-        # subj => { schema => id }
+        # subj => { avro_schema => id }
         self.subject_to_schema_ids = defaultdict(dict)
         # id => avro_schema
         self.id_to_schema = defaultdict(dict)
-        # subj => { schema => version }
+        # id => json_schema
+        self.id_to_json_schema = defaultdict(dict)
+        # subj => { avro_schema => version }
         self.subject_to_schema_versions = defaultdict(dict)
 
     def _send_request(self, url, method='GET', body=None, headers=None):
@@ -102,6 +112,10 @@ class CachedSchemaRegistryClient(object):
                 self._add_to_cache(self.subject_to_schema_versions,
                                    subject, schema, version)
 
+    def _cache_json_schema(self, schema, schema_id):
+        if schema_id not in self.id_to_schema:
+            self.id_to_json_schema[schema_id] = schema.to_json()
+
     def register(self, subject, avro_schema):
         """
         POST /subjects/(string: subject)/versions
@@ -137,17 +151,22 @@ class CachedSchemaRegistryClient(object):
         schema_id = result['id']
         # cache it
         self._cache_schema(avro_schema, schema_id, subject)
+
         return schema_id
 
-    def get_by_id(self, schema_id):
+    def get_by_id(self, schema_id, json_format=False):
         """
         GET /schemas/ids/{int: id}
         Retrieve a parsed avro schema by id or None if not found
         @:param: schema_id: int value
         @:returns: Avro schema
         """
-        if schema_id in self.id_to_schema:
-            return self.id_to_schema[schema_id]
+        if json_format:
+            if schema_id in self.id_to_json_schema:
+                return self.id_to_json_schema[schema_id]
+        else:
+            if schema_id in self.id_to_schema:
+                return self.id_to_schema[schema_id]
         # fetch from the registry
         url = '/'.join([self.url, 'schemas', 'ids', str(schema_id)])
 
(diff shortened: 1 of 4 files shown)


lewis-grimmett-zefr:
[clabot:check]

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -139,17 +153,22 @@ def register(self, subject, avro_schema):
         schema_id = result['id']
         # cache it
         self._cache_schema(avro_schema, schema_id, subject)
+
         return schema_id
 
-    def get_by_id(self, schema_id):
+    def get_by_id(self, schema_id, json_format=False):
```
edenhill: is json_format really necessary? It seems that HAS_FAST would be sufficient and already covers all the code paths where they are both true.

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -116,21 +117,22 @@ def encode_record_with_schema_id(self, schema_id, record, is_key=False):
...
+        if HAS_FAST:
+            json_schema = self.registry_client.get_by_id(schema_id, json_format=True)
+        else:
+            if schema_id not in self.id_to_writers:
+                # get the writer + schema
+
+                try:
+                    schema = self.registry_client.get_by_id(schema_id)
+                    if not schema:
+                        raise serialize_err("Schema does not exist")
```
edenhill: "Schema %d does not exist"..

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -155,34 +161,36 @@ def _get_decoder_func(self, schema_id, payload):
...
                 pass
 
+        # fetch from schema reg
+        try:
+            schema = self.registry_client.get_by_id(schema_id)
+        except:
+            schema = None
+
+        if not schema:
+            err = "unable to fetch schema with id %d" % (schema_id)
```
edenhill: capital "Unable .."

no need for err variable

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -155,34 +161,36 @@ def _get_decoder_func(self, schema_id, payload):
...
             except:
                 pass
 
+        # fetch from schema reg
+        try:
+            schema = self.registry_client.get_by_id(schema_id)
+        except:
+            schema = None
+
+        if not schema:
```
edenhill: if schema is None

lewis-grimmett-zefr: this was just moved - used to be before attempting to use fastavro, but I can change it

confluent_kafka/avro/serializer/message_serializer.py
lewis-grimmett-zefr: this was just moved - used to be before attempting to use fastavro, but I can change it

confluent_kafka/avro/cached_schema_registry_client.py
lewis-grimmett-zefr: ideally it would be sufficient, but if fastavro fails for some reason, i wanted to preserve the current consumer functionality where it falls back to to the standard avro library - in which case the schema still needs to be available somehow as a an avro Schema object.

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -104,6 +114,10 @@ def _cache_schema(self, schema, schema_id, subject=None, version=None):
                 self._add_to_cache(self.subject_to_schema_versions,
                                    subject, schema, version)
 
+    def _cache_json_schema(self, schema, schema_id):
+        if schema_id not in self.id_to_schema:
```
AndreiLieb: shouldn't this be: if schema_id not in self.id_to_json_schema

I don't see how json schema is getting cached as _cache_json_schema gets called after _cache_schema in register

AndreiLieb:
Did some more testing with this. Performance drops quite a bit from 2K msg/sec to 300 msg/sec.
This is due to heavy acquaint_schema call inside schemaless_writer on every message as well as:

[https://github.com/lewis-grimmett-zefr/confluent-kafka-python/blob/fastavro-optimizations/confluent_kafka/avro/__init__.py#L26-L28](url)

Hashing avro.schema.RecordSchema using its str is pretty heavy for large schemas (our schema is 10KB), this slows down cached schema lookups.

Bypassing both of these code paths brought performance back up to around 2K msg/sec.
Unfortunately I don't have the bandwidth to put together a proper PR in the near future.

lewis-grimmett-zefr:
@AndreiLieb yeah the hashing of the schema was by far the slowest part for us (11KB schema). Unfortunately I've gotten side tracked from this PR since we needed a feature breaking fork to bypass this (and to reduce `acquaint_schema` calls).. We're getting up to 4K msg/sec by removing the feature that allows message-level schemas to be supplied (i.e. always using the `default_key_schema` and `default_value_schema` with corresponding ids fetched on start up) which obviously avoids the need for hashing and caching.

Issue #198, opened 2017-05-24
Memory leak in cfl_PyUnistr_AsUTF8


stephan-hof:
In `confluent_kafka.h` this function has the following definition: `#define cfl_PyUnistr_AsUTF8(X) PyBytes_AsString(PyUnicode_AsUTF8String(X))`

The problems is that `PyUnicode_AsUTF8String` returns a new reference which is never released via `Py_DECREF` later on. Which means the return value of `PyUnicode_AsUTF8String` leaks memory.

I know a proper fix is hard, because the `Py_DECREF` must be deferred to the point where the result of `PyUnicode_AsUTF8String` is not used any more (considering also error cases ..).

I did not have problems with this leak yet, because the places where it happens are not triggered often in my code. However as soon as this is called frequently, for example `TopicPartition_str0` with an `errstr` set, it will have an impact.

edenhill:
Ooh, good catch, thanks!

PR #239, opened 2017-09-01
Fix utf8 string conversion memory leak on Python 2 (#198)

edenhill:
We need to keep and destroy the intermediary utf8 object on Python 2.

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -664,19 +664,27 @@ static PyMemberDef TopicPartition_members[] = {
 
 
 static PyObject *TopicPartition_str0 (TopicPartition *self) {
-	PyObject *errstr = self->error == Py_None ? NULL :
-		cfl_PyObject_Unistr(self->error);
+        PyObject *errstr = NULL;
+        PyObject *errstr8 = NULL;
+        const char *c_errstr = NULL;
 	PyObject *ret;
 	char offset_str[40];
+
 	snprintf(offset_str, sizeof(offset_str), "%"PRId64"", self->offset);
+
+        if (self->error != Py_None) {
+                errstr = cfl_PyObject_Unistr(self->error);
+                c_errstr = cfl_PyUnistr_AsUTF8(errstr, &errstr8);
+        }
+
 	ret = cfl_PyUnistr(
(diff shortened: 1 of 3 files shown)


confluent_kafka/src/confluent_kafka.c
```diff
@@ -1245,6 +1260,7 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
 						"as a callable function");
 				rd_kafka_topic_conf_destroy(tconf);
 				rd_kafka_conf_destroy(conf);
+                                Py_XDECREF(ks8);
```
ewencp: I know it becomes inconsistent with the way `ks` is handled, but couldn't we just `Py_XDECREF` immediately so we only have it once and don't have to scatter it on all of these paths? In fact, it looks like we could do that for `ks` in some of the cases that have been converted here.

confluent_kafka/src/confluent_kafka.h
```diff
@@ -64,8 +64,15 @@
 
 /**
  * @returns Unicode Python object as char * in UTF-8 encoding
+ * @param uobjp might be set to NULL or a new object reference (depending
```
ewencp: Along the same lines as my other comment -- why not `Py_XDECREF` immediately inside these methods? Doesn't seem like there would ever be a case where we'd want to save this intermediate data, especially since it is only an issue in Py2..

confluent_kafka/src/confluent_kafka.c
edenhill: On  Python 2.x ks8 will point to another object than ks, namely the utf8 representation of ks, and since `k` is a char pointer to ks's underlying memory we must retain the reference until all usage of `k` is over.

Ony Python 3 ks8 will always be NULL, thus causing XDECREF to be a no-op.

confluent_kafka/src/confluent_kafka.h
edenhill: I think this comment is misleading, it is not a new object reference to the original object, but a new object reference to an utf8 reprsentation object of the original object, do you object?

confluent_kafka/src/confluent_kafka.c
ewencp: ack, makes sense. Kinda sucks that the API naming doesn't make it really obvious when you're getting internal data vs a completely new object.

PR #243, opened 2017-09-04
Build autonomous binary wheels for Linux and OSX

edenhill:
There's a bunch of stuff going on and not going on here:
 * Uses [cibuildwheel](https://github.com/joerick/cibuildwheel) to build manylinux and osx wheels on Travis
 * The wheels are uploaded from Travis to S3.
 * Disabled support for the same on AppVeyor due to messed up Python/C toolchain, will be tended to in the future.
 * Some scripts to build and test manylinux locally using docker, these can be used for later use in packaging.
 * The .travis.yml file is rewritten to only cater to wheel building, before the merge it will be adjusted to do what it does on master (mainly code verification across Python versions) and the new wheel building stuff when a tag is available.
 * There's a script in tools/download-s3.py that fetches the wheels from S3 by tag or git sha. This is used to collect the wheels to publish to PyPi.
 * Some corresponding README changes.

Proof-of-concept packages are published to test.pypiy.org as confluent-kafka version 0.11.0.1rc1. Please try them out using the following little snippet:
```
virtualenv /tmp/pyenv      # Create a throw-away virtualenv
source /tmp/pyenv/bin/activate
pip install -i https://test.pypi.org/simple confluent-kafka==0.11.0.1rc1
python -c 'from confluent_kafka import libversion; assert libversion() == ("0.11.0", 0xb00ff), libversion()'
deactivate
 ```

### DIFF
--- a/.appveyor-disabled.yml
+++ b/.appveyor-disabled.yml
@@ -1,0 +1,25 @@
+environment:
+  global:
+    CIBW_TEST_REQUIRES: pytest requests avro
+    CIBW_TEST_COMMAND: "pytest --import-mode=append --ignore={project}\\tests\\avro {project}\\tests"
+    # SDK v7.0 MSVC Express 2008's SetEnv.cmd script will fail if the
+    # /E:ON and /V:ON options are not enabled in the batch script intepreter
+    # See: http://stackoverflow.com/a/13751649/163740
+    CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\tools\\appveyor\\run_with_env.cmd"
+
+#install:
+
+build_script:
+  - dir c:\python27
+  - dir c:\python27\include
+  - "curl https://raw.githubusercontent.com/chemeris/msinttypes/master/inttypes.h -o c:\\python27\\include\\inttypes.h"
+  - "curl https://raw.githubusercontent.com/chemeris/msinttypes/master/stdint.h -o c:\\python27\\include\\stdint.h"
+  - nuget install librdkafka.redist -OutputDirectory dest
+  - pip install https://github.com/joerick/cibuildwheel/archive/master.zip
+  - set INCLUDE=C:\\projects\\confluent-kafka-python\\dest\\librdkafka.redist.0.11.0\\build\\native\\include\\;%INCLUDE%
+  - set LIB=C:\\projects\\confluent-kafka-python\\dest\\librdkafka.redist.0.11.0\\runtimes\\win7-x64\\native;%LIB%
+  - cibuildwheel --output-dir wheelhouse
+
+artifacts:
+  - path: "wheelhouse\\*.whl"
+    name: Wheels
--- a/tools/test-manylinux.sh
+++ b/tools/test-manylinux.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+#
+#
+# Tests the manylinux wheels on a plethora of bare-bone Linux docker images.
+
+set -ex
+
+
+function setup_centos {
+    # CentOS container setup
+    yum install -q -y python epel-release
+    yum install -q -y python-pip
+}
+
+function setup_ubuntu {
+    # Ubuntu container setup
+    apt-get update
+    apt-get install -y python python-pip
+}
+
+
+function run_single_in_docker {
+    # Run single test inside docker container
+
+    # Detect OS
+    if grep -qi centos /etc/system-release ; then
+        setup_centos
+    elif grep -qi ubuntu /etc/os-release ; then
+        setup_ubuntu
+    else
+        echo "WARNING: Don't know what platform I'm on: $(uname -a)"
+    fi
+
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: tools/download-s3.py [hunk 1]
```diff
@@ -0,0 +1,154 @@
...
+
+        return None
+
+    def collect_s3(self):
+        """ Collect and download build-artifacts from S3 based on git reference """
+        print('Collecting artifacts matching tag/sha %s from S3 bucket %s' % (self.gitref, s3_bucket))
+        self.s3 = boto3.resource('s3')
+        self.s3_bucket = self.s3.Bucket(s3_bucket)
+        self.s3.meta.client.head_bucket(Bucket=s3_bucket)
+        for key in self.s3_bucket.objects.all():
```
ewencp: not a huge deal right now, but this will get increasingly slow over time as S3 listing isn't always particularly fast and it seems like this bucket will get uploads for all our CI builds as well. Would be good to think about if there is a layout of the paths that will allow you to restrict the set of objects that need to be listed.

tools/test-manylinux.sh
```diff
@@ -0,0 +1,96 @@
...
+    popd
+
+}
+
+function run_all_with_docker {
+    # Run tests in all listed docker containers.
+    # This is executed on the host.
+
+    [[ ! -z $DOCKER_IMAGES ]] || \
+        DOCKER_IMAGES="ubuntu:14.04 ubuntu:devel centos:6.6 centos:latest"
```
ewencp: probably all the ubuntu LTS versions? a centos 7 version that's explicitly 7 instead of relying on `latest` (which will eventually change to 8)? do we want to use pure debian as well since these are so easy to add/test here?

tools/download-s3.py
edenhill: It will be limited to just tagged builds, but with 3-4 client repositories your point still stands (and it is already kind of slow..).

From a matching perspective we look for two things: the project (confluent-kafka-python, which is already a top-level directory) and the git ref (tag or sha) which is currently included in the sub-directory.
Multiple CI services may provide artifacts for the same git ref, and even multiple artifacts from the same service (e.g., travis-osx, travis-linux, windows-x64, windows-x86, ..).

It is not possible to conditionally change the deploy path based on whether a tag is present or not, so the best we can do is:
`confluent-kafka-python/<TAG>-<SHA>/<ci>-<variant>...`

And since we might not be confident that we know both the sha and tag at the same time we would still need to enumerate on confluent-kafka-python/*.

We could use just the tag, `confluent-kafka-python/<TAG>/`, but there is no guarantee that it will be unique if the tag is later moved and artifacts rebuilt, which would result in conflicting artifacts under that tag using different SHAs.

I don't know what's best.

tools/test-manylinux.sh
edenhill: Sure.

Turns out that centos 6.6 is stuck at Python 2.6, which we don't support (and pip is deprecating).

tools/download-s3.py
edenhill: Kept the current format for now.

.appveyor-disabled.yml
```diff
@@ -0,0 +1,25 @@
...
+    # /E:ON and /V:ON options are not enabled in the batch script intepreter
+    # See: http://stackoverflow.com/a/13751649/163740
+    CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\tools\\appveyor\\run_with_env.cmd"
+
+#install:
+
+build_script:
+  - dir c:\python27
+  - dir c:\python27\include
+  - "curl https://raw.githubusercontent.com/chemeris/msinttypes/master/inttypes.h -o c:\\python27\\include\\inttypes.h"
```
ewencp: I know this is disabled, but should we really be referring to a branch rather than a specific gitref? Won't be reproducible if that file is ever changed on master.

tools/download-s3.py
ewencp: Isn't the sha the thing you ultimately care about? Why doesn't organizing it as `confluent-kafka-python/<sha>/<anything else you want>` reduce it to basically the minimal set of files you'd need to scan through? Seems like in this script you can just convert tag -> sha if needed.

.appveyor-disabled.yml
edenhill: This is in no way complete, but serves a starting point for this future endeavour.
But your point is good, for that future PR :)

edenhill:
@ewencp a build off a sha and a build off a tag with that same sha do not necessarily result in the same  binary since the tag itself may be used to alter the build, such as set the version string.
While the python client currently doesn't do this, it is quite common.

ewencp:
huh, i guess i would have just assumed those adjustments would be made in the commit that ends up getting tagged. building things like deb packages is much more painful if this isn't the case since you need to customize those adjustments to work with each build system, though i admit in the case of the python client we probably don't care about packages besides pypi.

PR #245, opened 2017-09-14
Added Consumer.store_offsets() API

ctrochalakis:
Hello all,

This PR implements the `Cosumer.store_offsets()` API. The signature is similar to the `.commit()` method.

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -379,15 +379,13 @@ static PyObject *Consumer_store_offsets (Handle *self, PyObject *args,
 		c_offsets = rd_kafka_topic_partition_list_new(1);
 		rd_kafka_topic_partition_list_add(
 			c_offsets, cfl_PyUnistr_AsUTF8(m->topic, &uo8),
-			m->partition)->offset =m->offset + 1;
+			m->partition)->offset = m->offset + 1;
 		Py_XDECREF(uo8);
 	}
 
 
 	err = rd_kafka_offsets_store(self->rk, c_offsets);
-
-	if (c_offsets)
-		rd_kafka_topic_partition_list_destroy(c_offsets);
+	rd_kafka_topic_partition_list_destroy(c_offsets);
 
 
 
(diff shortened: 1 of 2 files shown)


confluent_kafka/src/Consumer.c
```diff
@@ -325,6 +325,83 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
...
+				     "expected %s", MessageType.tp_name);
+			return NULL;
+		}
+
+		m = (Message *)msg;
+
+		c_offsets = rd_kafka_topic_partition_list_new(1);
+		rd_kafka_topic_partition_list_add(
+			c_offsets, cfl_PyUnistr_AsUTF8(m->topic, &uo8),
+			m->partition)->offset =m->offset + 1;
```
edenhill: nit: space after =

confluent_kafka/src/Consumer.c
```diff
@@ -325,6 +325,83 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
...
+		Py_XDECREF(uo8);
+	}
+
+
+	err = rd_kafka_offsets_store(self->rk, c_offsets);
+
+	if (c_offsets)
```
edenhill: this is always true

ctrochalakis: That's true, fixed!

confluent_kafka/src/Consumer.c
ctrochalakis: fixed

confluent_kafka/src/Consumer.c
```diff
@@ -341,7 +341,7 @@ static PyObject *Consumer_store_offsets (Handle *self, PyObject *args,
 		     "buildtime 0x%x)",
 		     rd_kafka_version(), RD_KAFKA_VERSION);
 	return NULL;
-#endif
+#else
```
edenhill: good catch

confluent_kafka/src/Consumer.c
```diff
@@ -341,7 +341,7 @@ static PyObject *Consumer_store_offsets (Handle *self, PyObject *args,
 		     "buildtime 0x%x)",
 		     rd_kafka_version(), RD_KAFKA_VERSION);
 	return NULL;
-#endif
```
edenhill: With this #else guard the variables defined above will be unused and produce warnings.
I suggest moving the variable decls down to the #else block.

edenhill:
The thread test failure is not your fault, ignore it.
Waiting for clogged up travis osx workers for final verification.

edenhill:
Thanks for this!

PR #246, opened 2017-09-14
Pass py.test arguments from tox

ctrochalakis:
You can now run a single test by running:

$ tox -e py27 -- tests/test_Consumer.py::test_basic_api


edenhill:
Thanks for this!

Issue #248, opened 2017-09-14
AvroProducer: Attempt to pull schema from schema-registry if not cached during publish


mfe5003:
It is currently only possible to publish to a topic if the schema is specified.  It would be nice to be able to have an `AvroProducer` attempt to pull the latest schema from the schema-registry if there is a schema cache miss.
The schema could be referenced by subject i.e. 'topic-value' or by schema id.
The workaround is to do a GET request to the schema-registry and then pass the returned schema into the producer call, but it seems cleaner to have the registry cache handle it.

If this is of interest, I have modified the behavior of `AvroProducer.produce` to also accept a schema id parameter such that if it is specified the registry cache will fetch the schema if available.
It is a simple change since the behavior was already written into the registry cache class.
I did not modify `AvroProducer.produce` to try to fetch a missing schema by subject, since this seemed to conflict with a test case.

PR #249, opened 2017-09-14
AvroProducer: fetch schema to cache by id during produce

mfe5003:
This pull request allows an `AvroProducer` to fetch a schema from the schema registry by schema id, if missing from the local registry cache.
For issue #248

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -27,9 +27,9 @@ class AvroProducer(Producer):
 
     def __init__(self, config, default_key_schema=None,
                  default_value_schema=None,
-                 default_value_schema_id=None,
+                 schema_registry=None,
                  default_key_schema_id=None,
-                 schema_registry=None):
+                 default_value_schema_id=None):
         schema_registry_url = config.pop("schema.registry.url", None)
         if schema_registry is None:
             if schema_registry_url is None:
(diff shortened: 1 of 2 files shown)


mfe5003:
[clabot:check]

confluent_kafka/avro/__init__.py
```diff
@@ -26,7 +26,10 @@ class AvroProducer(Producer):
     """
 
     def __init__(self, config, default_key_schema=None,
-                 default_value_schema=None, schema_registry=None):
+                 default_value_schema=None,
```
ewencp: may not be ideal, but it's better to not rearrange the order of existing arguments since users could already be calling this without using keyword arguments

mfe5003:
I am migrating an existing IoT system that uses a format very similar to Avro + schema-registry.
The device registers a schema and receives unique identifier which it prepends to all outgoing binary data.
My producers then need to do the translation from the old format to the new Avro before appending to the topic, and don't necessarily have the schema cached, so they may need to check the registry.

Basically [this situation](https://gist.github.com/mfe5003/16140f8255f2a9472e1e762d49b9ee15) is what I am looking for, where the schema_id would be known to a client that interacts with the producers.

mfe5003:
I have overlooked the fact that for my application I need to look up the topic by the associated schema id.  This implicitly assumes that the schema id and topic have a 1-to-1 mapping which is true for my application, but not in general.  I am closing this PR, sorry for the confusion.

PR #254, opened 2017-10-03
Makes calling consumer.close() multiple times safe

johnistan:
Currently if you call `consumer.close()` more then once you get a segfault.

This PR just makes calling `close` a second time a noop.

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -550,17 +550,24 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
 
 
 static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
-	CallState cs;
 
-	CallState_begin(self, &cs);
+	if (!self->rk) {
+            PyErr_SetString(PyExc_RuntimeError,
+                            "Consumer already closed");
+            return NULL;
+    }
 
-	rd_kafka_consumer_close(self->rk);
+    CallState cs;
 
-	rd_kafka_destroy(self->rk);
-	self->rk = NULL;
+    CallState_begin(self, &cs);
 
-        if (!CallState_end(self, &cs))
-                return NULL;
+    rd_kafka_consumer_close(self->rk);
+
+    rd_kafka_destroy(self->rk);
+    self->rk = NULL;
+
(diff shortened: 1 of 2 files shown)


johnistan:
After some reading of the code. I appears most consumer functions as unsafe after `close` is called.

I am happy to contribute some code to raise exceptions in this cases. However there are no examples of python client specific errors. They all have to be in the `rd_kafka_resp_err_t` enum. We could use one of the existing error codes but none seem to map quite right.

johnistan:
Updated to throw a RuntimeError as described in #255

confluent_kafka/src/Consumer.c
```diff
@@ -550,17 +550,24 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
 
 
 static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
-	CallState cs;
```
edenhill: To be compatible with older compilers the variable definition needs to be at the beginning of the scope, so keep this line here.

confluent_kafka/src/Consumer.c
```diff
@@ -550,17 +550,24 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
 
-	CallState_begin(self, &cs);
+	if (!self->rk) {
+            PyErr_SetString(PyExc_RuntimeError,
+                            "Consumer already closed");
+            return NULL;
+    }
```
edenhill: indentation is messed up here.
Should be tabs as 8 whitespaces.

tests/test_Consumer.py
```diff
@@ -158,3 +158,22 @@ def poll(self, somearg):
...
+                  'session.timeout.ms': 100})
+
+    c.subscribe(["test"])
+
+    c.unsubscribe()
+    c.close()
+
+    with pytest.raises(RuntimeError) as ex:
+        c.close()
+        assert 'Consumer already closed' == str(ex.value)
```
edenhill: :+1:

confluent_kafka/src/Consumer.c
johnistan: fixed

edenhill:
Great stuff, thanks!

PR #260, opened 2017-10-06
Kafkatest fixes

edenhill:


edenhill:
I'd like to get a review of first two commits:
 * Make Consumer.commit(..,async=False) return offset commit results
 * Messages could be leaked&lost if exception raised from callback trigg…

The kafkatest stuff can be ignored.

confluent_kafka/src/Consumer.c
```diff
@@ -673,8 +799,10 @@ static PyMethodDef Consumer_methods[] = {
 	  "\n"
 	  "  :param confluent_kafka.Message message: Commit message's offset+1.\n"
 	  "  :param list(TopicPartition) offsets: List of topic+partitions+offsets to commit.\n"
-	  "  :param bool async: Asynchronous commit, return immediately.\n"
-	  "  :rtype: None\n"
+	  "  :param bool async: Asynchronous commit, return None immediately. "
+          "If False the commit() call will block until the commit succeeds or "
```
mhowlett: Things I'm wondering: under what circumstances does commit() fail? how long can this method potentially block (is there/what is the relevant config parameter controlling this?). It'd be good to document this here I think. If this call can potentially block indefinitely, I think a timeout parameter should be added (i.e. there should be no variant that can potentially block indefinitely).

edenhill: The non-timeout behaviour is in sync (hehe) with the Java client's commitSync():
https://kafka.apache.org/0100/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html

Typical errors are REQUEST_TIMED_OUT, NOT_COORDINATOR_FOR_GROUP, _WAIT_COORD, or any error returned by the broker. But yeah, it should be documented, however that goes for all APIs and that is a bigger task which we'll do for all clients (starting with librdkafka) some time in the future. Bit of a mess right now.

edenhill: The timeout is session.timeout.ms + socket.timeout.ms

mhowlett:
LGTM

PR #262, opened 2017-10-10
Raise RuntimeError for all consumer methods after close

### LINKED ISSUE #255: Segfault after calling a closed consumer
Description
===========

Once a Consumer has been closed, calling any subsequent method references the inter handle `self->rk` which is now `NULL` and causes a segfault.

The consumer should instead raise an exception for the application code to handle. 

How to reproduce
================

```
c = Consumer()
c.close()
c.assignment() # SEGFAULT
```

johnistan:
After much copy and paste this closes #255 

The code formatting appears inconsistent between methods. tabs and spaces mixed up. I tried to match each method's indenting. If i missed one let me know.


edenhill:
Great stuff!
[clabot:check]

PR #263, opened 2017-10-10
Fix consumer close RuntimeError tests

johnistan:
My exception contents tests for #255 are noops. here are fixes.


johnistan:
Broken tests in first commit. Fix in second. 

I am usually in love with pytest conventions, this one is a real foot gun.

johnistan:
already tracked here https://github.com/pytest-dev/pytest/issues/2795

edenhill:
Thanks!

PR #279, opened 2017-11-14
Fix bare except

edenhill:
flake8 now warns on bare exception, this PR attempts to fix that.

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -168,9 +168,9 @@ class CachedSchemaRegistryClient(object):
                 # cache it
                 self._cache_schema(result, schema_id)
                 return result
-            except:
+            except ClientError as e:
                 # bad schema - should not happen
-                raise ClientError("Received bad schema from registry.")
+                raise ClientError("Received bad schema (id %s) from registry: %s" % (schema_id, e))
 
     def get_latest_schema(self, subject):
         """
@@ -204,9 +204,9 @@ class CachedSchemaRegistryClient(object):
         else:
             try:
                 schema = loads(result['schema'])
-            except:
+            except ClientError:
                 # bad schema - should not happen
-                raise ClientError("Received bad schema from registry.")
+                raise
 
         self._cache_schema(schema, schema_id, subject, version)
         return (schema_id, schema, version)
@@ -269,7 +269,8 @@ class CachedSchemaRegistryClient(object):
             else:
                 log.error("Unable to check the compatibility")
                 False
-        except:
+        except Exception as e:
+            log.error("_send_request() failed: %s", e)
             return False
 
     def update_compatibility(self, level, subject=None):
--- a/confluent_kafka/avro/load.py
+++ b/confluent_kafka/avro/load.py
@@ -17,13 +17,18 @@
 
 import sys
 
+from confluent_kafka.avro.error import ClientError
+
 
 def loads(schema_str):
     """ Parse a schema given a schema string """
-    if sys.version_info[0] < 3:
-        return schema.parse(schema_str)
-    else:
-        return schema.Parse(schema_str)
+    try:
+        if sys.version_info[0] < 3:
+            return schema.parse(schema_str)
+        else:
+            return schema.Parse(schema_str)
+    except schema.AvroException.SchemaParseException as e:
+        raise ClientError("Schema parse failed: %s" % (str(e)))
 
 
 def load(fp):
@@ -44,5 +49,6 @@ try:
     schema.RecordSchema.__hash__ = _hash_func
     schema.PrimitiveSchema.__hash__ = _hash_func
     schema.UnionSchema.__hash__ = _hash_func
+
 except ImportError:
     schema = None
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -42,7 +42,7 @@ try:
     from fastavro.reader import read_data
 
     HAS_FAST = True
-except:
+except ImportError:
     pass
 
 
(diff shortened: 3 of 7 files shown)


confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -168,9 +169,9 @@ def get_by_id(self, schema_id):
                 # cache it
                 self._cache_schema(result, schema_id)
                 return result
-            except:
+            except ClientError as e:
                 # bad schema - should not happen
-                raise ClientError("Received bad schema from registry.")
+                raise ClientError("Received bad schema (id %s) from registry: %s" % (schema_id, str(e)))
```
jeffwidman: Why cast to `e` to `str`? Why not rely on the implicit conversion?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -269,7 +270,8 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
             else:
                 log.error("Unable to check the compatibility")
                 False
-        except:
+        except: # noqa
```
jeffwidman: Shouldn't you at least restrict this to `except Exception:`?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -269,7 +270,8 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
...
+            log.error("_send_request() failed: %s" % sys.exc_info()[0])
```
jeffwidman: Best practice is to use the `logging` framework's built-in string interpolation:

    log.level("string %s", arg1)

That way the string isn't built unless the log level is hit. For error-level, this is simply a style thing, but for debug-level logging this can occasionally have a non-trivial performance impact.

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -180,7 +179,7 @@ def _get_decoder_func(self, schema_id, payload):
 
                 self.id_to_decoder_func[schema_id] = lambda p: read_data(p, schema_dict)
                 return self.id_to_decoder_func[schema_id]
-            except:
+            except: # noqa
```
jeffwidman: `except Exception`?

confluent_kafka/avro/cached_schema_registry_client.py
edenhill: To favour __str__ over __repr__, but I guess __str__ kicks in here thanks to %s anyway

confluent_kafka/avro/cached_schema_registry_client.py
edenhill: That's what I thought too, until I found this:

> In situations where you need to catch all “normal” errors, such as in a framework that runs callbacks, you can catch the base class for all normal exceptions, Exception. Unfortunately in Python 2.x it is possible for third-party code to raise exceptions that do not inherit from Exception, so in Python 2.x there are some cases where you may have to use a bare except: and manually re-raise the exceptions you don’t want to catch.

from https://docs.python.org/2/howto/doanddont.html#except

confluent_kafka/avro/cached_schema_registry_client.py
edenhill: Good input, thanks!

confluent_kafka/avro/serializer/message_serializer.py
edenhill: Thanks for the review!

confluent_kafka/avro/cached_schema_registry_client.py
jeffwidman: Are the exceptions being caught here from the app code or from this library (and other open-source libraries)?

If app code, that makes sense, but if from libraries, than I don't think it makes sense as generally most open source libs are well-behaved and if it is an issue, maintainers are generally very open to cleaning this up.

edenhill:
This should address all of Jeff's review points.

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -269,7 +269,8 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
             else:
                 log.error("Unable to check the compatibility")
                 False
-        except:
+        except Exception as e:
+            log.error("_send_request() failed: %s", e)
```
jeffwidman: use `log.exception()`... it is the same as `log.error()` but it will automatically append the exception:
 
    log.exception("_send_request() failed.")

confluent_kafka/avro/load.py
```diff
@@ -17,13 +17,18 @@
...
-        return schema.parse(schema_str)
-    else:
-        return schema.Parse(schema_str)
+    try:
+        if sys.version_info[0] < 3:
+            return schema.parse(schema_str)
+        else:
+            return schema.Parse(schema_str)
+    except schema.AvroException.SchemaParseException as e:
+        raise ClientError("Schema parse failed: %s" % (str(e)))
```
jeffwidman: no need to cast to string?

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -158,12 +158,11 @@ def _get_decoder_func(self, schema_id, payload):
         # fetch from schema reg
         try:
             schema = self.registry_client.get_by_id(schema_id)
-        except:
-            schema = None
+        except ClientError as e:
+            raise SerializerError("unable to fetch schema with id %d: %s" % (schema_id, str(e)))
```
jeffwidman: no need to cast the exception to a string?

tests/avro/mock_registry.py
```diff
@@ -174,11 +176,18 @@ def __init__(self, port):
...
         self.daemon = True
+        self.started = Event()
 
     def run(self):
         self.server = MockServer(('127.0.0.1', self.port), ReqHandler)
+        self.started.set()
         self.server.serve_forever()
 
+    def start(self):
+        """ Start, and wait for server to be fully started, before returning. """
```
jeffwidman: nit: Per [PEP-257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings), docstrings don't need space before/after the sentence.

edenhill:
Thanks for your review, @jeffwidman !

jeffwidman:
No problem. Curious why you didn't implement the other points in my review of the fixes though? Was that purposeful decision or accidental oversight?

edenhill:
@jeffwidman I thought I covered them all, what did I miss?

jeffwidman:
Scroll here and you will see a few unaddressed comments: https://github.com/confluentinc/confluent-kafka-python/pull/279/files

No problem if you choose not to incorporate them, I mentioned only because I had a hunch you never even saw them.

PR #281, opened 2017-11-26
Fix librdkafka install command for macOS

vkroz:
Do not use sudo when running brew install


edenhill:
[clabot:check]

vkroz:
[clabot:check]

edenhill:
Thank you!

Issue #252, opened 2017-09-28
Batch consumer ? 


gowtham-sai:
Is there a way where I can get batch of messages instead of a single message. 

What I'm trying to achieve is to get batch of messages and convert them to Python dataframe using from_json, and then do some operations in the dataframe and store the parquet files to s3. 

The problem here is I can't just wait make the dataframe from only few messages. It has to be atleast 2L messages. 

For this I can't iterate over 2L times, and poll the consume to get one by one message and store it and then process further. 

Instead if there is a way where we can tell the consumer poll to return next 100 messages, it will be good. So it should return 100 messages if 100 messages are available or the n number of messages that are available.  

I think this is general use case where most of the time we don't process message by message but a batch of messages.

edenhill:
Feature request noted.

In the meantime you could make your own batching wrapper for poll().

gowtham-sai:
@edenhill when can we expect this? Or can you guide me with some resources to implement that myself in C? 
I need this approach for one of my production optimisation. Current way is taking too much time like 2000s, with this one, I can bought this time under 20s - 50s.

mrocklin:
Does anyone have a sense for what the performance would be for a batching solution?  Do we know what the current bottleneck is within `confluent-kafka-python` or `librdkafka`?

[This blogpost](http://activisiongamescience.github.io/2016/06/15/Kafka-Client-Benchmarking/) gives performance numbers for `confluent-kafka-python` at around 200k msgs/sec or 25MB/s.  

This is certainly better than other Python Kafka libraries, but also not yet saturating typical ethernet bandwidth within datacenters.

mrocklin:
@gowtham-sai JSON parsing speeds are only slightly better than this rate (~50MB/s?).  It's not clear to me how much benefit you're likely to get from speeding up kafka bandwidth here.

gowtham-sai:
@mrocklin 
>This is certainly better than other Python Kafka libraries, but also not yet saturating typical ethernet bandwidth within datacenters.

Yep, you are correct. Confluent Kafka speed is better than any other python library now. 

>  Do we know what the current bottleneck is within confluent-kafka-python or librdkafka?

As per my knowledge, poll is implemented in librdkafka, if this is the case, calling poll every time is worst when you want to fetch say 2,00,000 msgs. This means, we need to iterate over 2,00,000 like below..

```python
from confluent_kafka import Consumer, KafkaError

[...]

PR #282, opened 2017-12-04
first attempt at batch consumption of messages

tburmeister:
This is a first attempt at adding batch consumption, a la https://github.com/confluentinc/confluent-kafka-python/issues/252

Have not done a ton of testing yet, but I am able to retrieve messages. Let me know what you think.

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -41,6 +41,10 @@ static int Consumer_clear (Handle *self) {
 		Py_DECREF(self->u.Consumer.on_commit);
 		self->u.Consumer.on_commit = NULL;
 	}
+	if (self->u.Consumer.rkqu) {
+	        rd_kafka_queue_destroy(self->u.Consumer.rkqu);
+	        self->u.Consumer.rkqu = NULL;
+	}
 
 	Handle_clear(self);
 
@@ -372,7 +376,7 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
 	rd_kafka_topic_partition_list_t *c_offsets;
 	int async = 1;
 	static char *kws[] = { "message", "offsets", "async",NULL };
-        rd_kafka_queue_t *rkqu = NULL;
+        rd_kafka_queue_t *rkqu = self->u.Consumer.rkqu;
         struct commit_return commit_return;
         PyThreadState *thread_state;
 
@@ -426,7 +430,8 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
         if (async) {
                 /* Async mode: Use consumer queue for offset commit callback,
                  *             served by consumer_poll() */
-                rkqu = rd_kafka_queue_get_consumer(self->rk);
+                if (!rkqu)
+                        rkqu = self->u.Consumer.rkqu = rd_kafka_queue_get_consumer(self->rk);
 
         } else {
                 /* Sync mode: Let commit_queue() trigger the callback. */
@@ -446,11 +451,7 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
         if (c_offsets)
                 rd_kafka_topic_partition_list_destroy(c_offsets);
 
-        if (async) {
-                /* Loose reference to consumer queue */
-                rd_kafka_queue_destroy(rkqu);
-
-        } else {
+        if (!async) {
                 /* Re-lock GIL */
                 PyEval_RestoreThread(thread_state);
 
@@ -750,7 +751,7 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
         static char *kws[] = { "num_messages", "timeout", NULL };
         rd_kafka_message_t **rkmessages;
         PyObject *msglist;
-        rd_kafka_queue_t *rkqu = NULL;
+        rd_kafka_queue_t *rkqu = self->u.Consumer.rkqu;
         CallState cs;
 
         if (!self->rk) {
@@ -771,7 +772,8 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
 
         CallState_begin(self, &cs);
 
-        rkqu = rd_kafka_queue_get_consumer(self->rk);
+        if (!rkqu)
+                rkqu = self->u.Consumer.rkqu = rd_kafka_queue_get_consumer(self->rk);
 
         rkmessages = malloc(num_messages * sizeof(rd_kafka_message_t *));
 
@@ -779,9 +781,6 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
                 tmout >= 0 ? (int)(tmout * 1000.0f) : -1,
                 rkmessages,
                 num_messages);
-
-        /* Loose reference to consumer queue */
-        rd_kafka_queue_destroy(rkqu);
 
         if (!CallState_end(self, &cs)) {
                 for (Py_ssize_t i = 0; i < n; i++) {
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -907,7 +907,7 @@ static PyMethodDef Consumer_methods[] = {
 	  "\n"
 	  "  :param int num_messages: Maximum number of messages to return (default: 1).\n"
 	  "  :param float timeout: Maximum time to block waiting for message, event or callback (default: infinite (-1)).\n"
-	  "  :returns: A list of Message objects\n"
+	  "  :returns: A list of Message objects (possibly empty on timeout)\n"
 	  "  :rtype: list(Message)\n"
           "  :raises: RuntimeError if called on a closed consumer, KafkaError "
           "in case of internal error, or ValueError if num_messages > 1M.\n"
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -376,7 +376,7 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
 	rd_kafka_topic_partition_list_t *c_offsets;
 	int async = 1;
 	static char *kws[] = { "message", "offsets", "async",NULL };
-        rd_kafka_queue_t *rkqu = self->u.Consumer.rkqu;
+        rd_kafka_queue_t *rkqu = NULL;
         struct commit_return commit_return;
         PyThreadState *thread_state;
 
@@ -427,7 +427,12 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
 		c_offsets = NULL;
 	}
 
-        if (!async) {
+        if (async) {
+                /* Async mode: Use consumer queue for offset commit
+                 *             served by consumer_poll() */
+                rkqu = self->u.Consumer.rkqu;
+
+        } else {
                 /* Sync mode: Let commit_queue() trigger the callback. */
                 memset(&commit_return, 0, sizeof(commit_return));
 
@@ -747,6 +752,7 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
         PyObject *msglist;
         rd_kafka_queue_t *rkqu = self->u.Consumer.rkqu;
         CallState cs;
+        Py_ssize_t i;
 
         if (!self->rk) {
                 PyErr_SetString(PyExc_RuntimeError,
@@ -774,7 +780,7 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
                 num_messages);
 
         if (!CallState_end(self, &cs)) {
-                for (Py_ssize_t i = 0; i < n; i++) {
+                for (i = 0; i < n; i++) {
                         rd_kafka_message_destroy(rkmessages[i]);
                 }
                 free(rkmessages);
@@ -790,7 +796,7 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
 
         msglist = PyList_New(n);
 
-        for (Py_ssize_t i = 0; i < n; i++) {
+        for (i = 0; i < n; i++) {
                 PyObject *msgobj = Message_new0(self, rkmessages[i]);
                 PyList_SET_ITEM(msglist, i, msgobj);
                 rd_kafka_message_destroy(rkmessages[i]);
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -760,7 +760,7 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
 
 	if (num_messages > 1000000) {
 	        PyErr_SetString(PyExc_ValueError,
-	                        "num_messages must be <= 1000000 (1M)");
+	                        "num_messages must be between 0 and 1000000 (1M)");
 	        return NULL;
 	}
 
@@ -817,8 +817,8 @@ static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
 
         if (self->u.Consumer.rkqu) {
                 rd_kafka_queue_destroy(self->u.Consumer.rkqu);
-	        self->u.Consumer.rkqu = NULL;
-	}
+                self->u.Consumer.rkqu = NULL;
+        }
 
         rd_kafka_destroy(self->rk);
         self->rk = NULL;
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -15,6 +15,7 @@
  */
 
 #include "confluent_kafka.h"
+#include <stdio.h>
 
 
 /****************************************************************************
@@ -743,6 +744,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
 }
 
 
+static PyObject *Consumer_consume (Handle *self, PyObject *args,
+                                        PyObject *kwargs) {
+        size_t num_messages = 100;
+        double tmout = -1.0f;
+        static char *kws[] = { "timeout", "num_messages", NULL };
+        rd_kafka_message_t **rkmessages;
+        PyObject *msglist;
+        rd_kafka_queue_t *rkqu = NULL;
+        CallState cs;
+
+        if (!self->rk) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Consumer closed");
+                return NULL;
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|nd", kws,
+					 (Py_ssize_t *)&num_messages, &tmout))
+		return NULL;
+
+        CallState_begin(self, &cs); // This unlocks GIL
+
+        rkqu = rd_kafka_queue_get_consumer(self->rk);
+
+        rkmessages = malloc(num_messages * sizeof(rd_kafka_message_t *));
+
+        Py_ssize_t n = (Py_ssize_t)rd_kafka_consume_batch_queue(rkqu,
+                tmout >= 0 ? (int)(tmout * 1000.0f) : -1,
+                rkmessages,
+                num_messages > 0 ? (size_t)num_messages : 1);
+
+        /* Loose reference to consumer queue */
+        rd_kafka_queue_destroy(rkqu);
+
+        if (!CallState_end(self, &cs)) {
+                for (Py_ssize_t i = 0; i < n; i++) {
+                        rd_kafka_message_destroy(rkmessages[i]);
+                }
+                free(rkmessages);
+                return NULL;
+        }
+
+        if (n < 0) {
+                free(rkmessages);
+                Py_RETURN_NONE;
+        }
+
+        msglist = PyList_New(n);
+
+        for (Py_ssize_t i = 0; i < n; i++) {
+                PyObject *msgobj = Message_new0(self, rkmessages[i]);
+                PyList_SET_ITEM(msglist, i, msgobj);
+                rd_kafka_message_destroy(rkmessages[i]);
+        }
+
+        free(rkmessages);
+
+        return msglist;
+}
+
+
 static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
         CallState cs;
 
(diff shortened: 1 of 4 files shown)


confluent_kafka/src/Consumer.c
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "confluent_kafka.h"
+#include <stdio.h>
```
tburmeister: Had some debug `printf`s in earlier, will remove this.

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +744,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Consumer closed");
+                return NULL;
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|nd", kws,
+					 (Py_ssize_t *)&num_messages, &tmout))
+		return NULL;
+
+        CallState_begin(self, &cs); // This unlocks GIL
```
tburmeister: The comment was for me; will remove.

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +744,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+
+        rkqu = rd_kafka_queue_get_consumer(self->rk);
```
tburmeister: Think this returns `NULL` if consumer was not properly initialized, which leads to a segfault further down; maybe just need to check this and exit if `NULL`?

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +744,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
 }
 
 
+static PyObject *Consumer_consume (Handle *self, PyObject *args,
+                                        PyObject *kwargs) {
+        size_t num_messages = 100;
+        double tmout = -1.0f;
+        static char *kws[] = { "timeout", "num_messages", NULL };
```
tburmeister: This is backwards, should be `{ "num_messages", "timeout", NULL }`.

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
 }
 
 
+static PyObject *Consumer_consume (Handle *self, PyObject *args,
+                                        PyObject *kwargs) {
+        size_t num_messages = 100;
```
edenhill: 1 seems like a better default.

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+                return NULL;
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|nd", kws,
+					 (Py_ssize_t *)&num_messages, &tmout))
+		return NULL;
+
+        CallState_begin(self, &cs);
+
+        rkqu = rd_kafka_queue_get_consumer(self->rk);
```
edenhill: optional future improvement: cache this on `self`

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+        rd_kafka_queue_t *rkqu = NULL;
+        CallState cs;
+
+        if (!self->rk) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Consumer closed");
+                return NULL;
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|nd", kws,
```
edenhill: Use `I` for num_messages (`unsigned int`) to avoid checking for < 0.

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+        if (!CallState_end(self, &cs)) {
+                for (Py_ssize_t i = 0; i < n; i++) {
+                        rd_kafka_message_destroy(rkmessages[i]);
+                }
+                free(rkmessages);
+                return NULL;
+        }
+
+        if (n < 0) {
+                free(rkmessages);
```
edenhill: Should raise exception from rd_kafka_last_error() here, something like:
`cfl_PyErr_Format(rd_kafka_last_error(), "%s", rd_kafka_err2str(rd_kafka_last_error());`

and return NULL (to raise the exception)

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+        if (!self->rk) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Consumer closed");
+                return NULL;
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|nd", kws,
+					 (Py_ssize_t *)&num_messages, &tmout))
+		return NULL;
+
```
edenhill: Might want to make sure num_messages is not completely out of bounds since we're using the value directly to allocate memory.
If it is higher than 1M messages, raise an ValueError with a proper error message

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,67 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+        CallState_begin(self, &cs);
+
+        rkqu = rd_kafka_queue_get_consumer(self->rk);
+
+        rkmessages = malloc(num_messages * sizeof(rd_kafka_message_t *));
+
+        Py_ssize_t n = (Py_ssize_t)rd_kafka_consume_batch_queue(rkqu,
+                tmout >= 0 ? (int)(tmout * 1000.0f) : -1,
+                rkmessages,
+                num_messages > 0 ? (size_t)num_messages : 1);
```
edenhill: can drop this check if using unsigned int

confluent_kafka/src/Consumer.c
```diff
@@ -825,6 +886,28 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "  The application must check the returned :py:class:`Message` "
+	  "object's :py:func:`Message.error()` method to distinguish "
+	  "between proper messages (error() returns None), or an event or "
+	  "error for each :py:class:`Message` in the list (see error().code() "
+	  "for specifics).\n"
+	  "\n"
+	  "  .. note: Callbacks may be called from this method, "
+	  "such as ``on_assign``, ``on_revoke``, et.al.\n"
+	  "\n"
+	  "  :param int num_messages: Maximum number of messages to return.\n"
```
edenhill: + ` (default: 1)`

confluent_kafka/src/Consumer.c
```diff
@@ -825,6 +886,28 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "  :param float timeout: Maximum time to block waiting for message, event or callback.\n"
+	  "  :returns: A list of Message objects or None on timeout\n"
```
edenhill: Do we really need to use a different return type for timeout? An empty list seems easier to handle from the application's perspective.

confluent_kafka/src/Consumer.c
```diff
@@ -825,6 +886,28 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "object's :py:func:`Message.error()` method to distinguish "
+	  "between proper messages (error() returns None), or an event or "
+	  "error for each :py:class:`Message` in the list (see error().code() "
+	  "for specifics).\n"
+	  "\n"
+	  "  .. note: Callbacks may be called from this method, "
+	  "such as ``on_assign``, ``on_revoke``, et.al.\n"
+	  "\n"
+	  "  :param int num_messages: Maximum number of messages to return.\n"
+	  "  :param float timeout: Maximum time to block waiting for message, event or callback.\n"
```
edenhill: `(default: infinite (-1))`

confluent_kafka/src/Consumer.c
edenhill: That should not happen, so no need to check

tburmeister:
[clabot:check]

tburmeister:
Made suggested changes. Will work on unit tests and integration tests next, and might take a stab at caching the queue.

edenhill:
Great! For caching the queue, to avoid locking issues, I suggest extracting the queue in Consumer_init() right after poll_set_consumer() and destryoing the queue in Consumer_close() after the rd_kafka_consumer_close() call, and in Consumer_dealloc() prior to rd_kafka_destroy() (and outside the Callstate )

confluent_kafka/src/Consumer.c
```diff
@@ -743,6 +743,75 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+        CallState cs;
+
+        if (!self->rk) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Consumer closed");
+                return NULL;
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|Id", kws,
+					 (unsigned int *)&num_messages, &tmout))
```
edenhill: It is not safe to cast to integer pointers of possibly other size.
Use `unsigned int` for num_message's type directly to avoid the case.

confluent_kafka/src/Consumer.c
```diff
@@ -825,6 +894,28 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "error for each :py:class:`Message` in the list (see error().code() "
+	  "for specifics).\n"
+	  "\n"
+	  "  .. note: Callbacks may be called from this method, "
+	  "such as ``on_assign``, ``on_revoke``, et.al.\n"
+	  "\n"
+	  "  :param int num_messages: Maximum number of messages to return (default: 1).\n"
+	  "  :param float timeout: Maximum time to block waiting for message, event or callback (default: infinite (-1)).\n"
+	  "  :returns: A list of Message objects\n"
+	  "  :rtype: list(Message) or None\n"
```
edenhill: Does not return None anymore.

confluent_kafka/src/Consumer.c
```diff
@@ -825,6 +894,28 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "between proper messages (error() returns None), or an event or "
+	  "error for each :py:class:`Message` in the list (see error().code() "
+	  "for specifics).\n"
+	  "\n"
+	  "  .. note: Callbacks may be called from this method, "
+	  "such as ``on_assign``, ``on_revoke``, et.al.\n"
+	  "\n"
+	  "  :param int num_messages: Maximum number of messages to return (default: 1).\n"
+	  "  :param float timeout: Maximum time to block waiting for message, event or callback (default: infinite (-1)).\n"
+	  "  :returns: A list of Message objects\n"
```
edenhill: + `(possibly empty on timeout)`

confluent_kafka/src/Consumer.c
```diff
@@ -825,6 +894,28 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "  :rtype: list(Message) or None\n"
+          "  :raises: RuntimeError if called on a closed consumer\n"
```
edenhill: or KafkaError in case of internal error.

tburmeister:
Missed your comment before pushing my last commit with caching, but that all makes sense.

tburmeister:
I added some tests, but I'm having trouble running the tests because synchronous commits crash on my machine; `commit_return.c_parts` appears to be `NULL` in `Consumer_commit`.

tburmeister:
Version 0.11.0.

edenhill:
Ack, do you have a gdb backtrace to share?
`list` and `bt`

examples/integration_test.py
```diff
@@ -565,6 +565,175 @@ def my_on_revoke(consumer, partitions):
...
+
+            bytecnt += len(msg)
+            msgcnt += 1
+
+            if bar is not None and (msgcnt % 10000) == 0:
+                bar.next(n=10000)
+
+            if msgcnt == 1:
+                t_first_msg = time.time()
+            if msgcnt >= max_msgcnt:
```
edenhill: elif

examples/integration_test.py
```diff
@@ -565,6 +565,175 @@ def my_on_revoke(consumer, partitions):
...
+                break
+
+        if msgcnt >= max_msgcnt:
```
edenhill: move this to the `while ..`

examples/integration_test.py
```diff
@@ -565,6 +565,175 @@ def my_on_revoke(consumer, partitions):
...
+            break
+
+    if bar is not None:
+        bar.finish()
+
+    if msgcnt > 0:
+        t_spent = time.time() - t_first_msg
+        print('%d messages (%.2fMb) consumed in %.3fs: %d msgs/s, %.2f Mb/s' %
```
edenhill: what thruput are you getting with the batch interface, compared to the standard one?

examples/integration_test.py
```diff
@@ -565,6 +565,175 @@ def my_on_revoke(consumer, partitions):
...
+    c.subscribe([topic])
+
+    max_msgcnt = 100
+    msgcnt = 0
+
+    while True:
+        # Consume until EOF or error
+
+        # Consume message (error()==0) or event (error()!=0)
+        msglist = c.consume(max_msgcnt, 1.0)
```
edenhill: Knowing that there is at least N messages in the topic (thanks to the produce tests), and where N > max_msgcnt (which it should be), we should check that we actually get max_msgcnt back for each call until we've hit our expected message count (which is some arbitrary value <=N).
Relying on EOF is not safe since the topic may feature multiple partitions and we don't know what the message distribution is like.

tests/test_Consumer.py
```diff
@@ -41,6 +41,17 @@ def dummy_assign_revoke(consumer, partitions):
     if msg is not None:
         assert msg.timestamp() == (TIMESTAMP_NOT_AVAILABLE, -1)
 
+    msglist = kc.consume(num_messages=10, timeout=0.001)
+    if len(msglist) == 0:
```
edenhill: aren't we expecting exactly 0 messages here? use `assert len(msglist) == 0, "expected 0 messages, not %d" % len(msglist)`

confluent_kafka/src/Consumer.c
```diff
@@ -756,6 +815,11 @@ static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
 
         rd_kafka_consumer_close(self->rk);
 
+        if (self->u.Consumer.rkqu) {
+                rd_kafka_queue_destroy(self->u.Consumer.rkqu);
```
edenhill: indent looks weird.

tests/test_Consumer.py
```diff
@@ -41,6 +41,17 @@ def dummy_assign_revoke(consumer, partitions):
...
+        print('OK: consume() timeout')
+    else:
+        for msg in msglist:
+            if msg.error():
+                print('OK: consumer error: %s' % msg.error().str())
+                break
+
+        print('OK: consumed messages')
+
```
edenhill: Please add a failing test for invalid number of num_messages (e.g., -100 and > 1M)

tburmeister:
Struggling to get debug symbols working properly on my Mac, so the best I can give you right now is:
```(gdb) bt
#0  0x0000000101be67e0 in c_parts_to_py () from /Users/tburmeister/anaconda/lib/python3.6/site-packages/confluent_kafka/cimpl.cpython-36m-darwin.so
#1  0x0000000101becceb in Consumer_commit () from /Users/tburmeister/anaconda/lib/python3.6/site-packages/confluent_kafka/cimpl.cpython-36m-darwin.so
#2  0x000000010005894a in _PyCFunction_FastCallDict () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#3  0x0000000100058b5b in _PyCFunction_FastCallKeywords () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#4  0x00000001000da787 in call_function () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#5  0x00000001000d7001 in _PyEval_EvalFrameDefault () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#6  0x00000001000db9ed in fast_function () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#7  0x00000001000da761 in call_function () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#8  0x00000001000d6f62 in _PyEval_EvalFrameDefault () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#9  0x00000001000db1ef in _PyEval_EvalCodeWithName () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#10 0x00000001000d03d4 in PyEval_EvalCode () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#11 0x00000001001060f1 in PyRun_FileExFlags () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#12 0x00000001001058a3 in PyRun_SimpleFileExFlags () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#13 0x000000010011e70f in Py_Main () from /Users/tburmeister/anaconda/lib/libpython3.6m.dylib
#14 0x0000000100000df8 in main ()```

I do not have this issue on my Ubuntu dev machine.

**Comment by edenhill** [comment]
> I added some tests, but I'm having trouble running the tests because synchronous commits crash on my machine; commit_return.c_parts appears to be NULL in Consumer_commit.

That is weird since there is assert just above that checks that commit_return.c_parts is not NULL: 
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/src/Consumer.c#L480

Can you add some printfs around that part to figure out what is going on?

File: confluent_kafka/src/Consumer.c (hunk 23 above)
Comment (tburmeister) [reply confluent_kafka/src/Consumer.c]: Will fix. I noticed that there are a bunch of tabs in the code currently, and I gather from the majority of the code and the librdkafka style guide that spaces are preferred - I can clean those up as well.

**Comment by tburmeister** [comment]
The assert seems to not actually be checked. I added:
```
                /* sync commit returns the topic,partition,offset,err list */
                assert(commit_return.c_parts);
                printf("commit_return.c_parts: %p\n", commit_return.c_parts);
```
at line 477, and get:
```============================== Verifying Consumer ==============================
test[0]@193: key=b'0', value=b'Message #0', tstype=1, timestamp=1512499907092
test[0]@194: key=b'1', value=b'Message #1', tstype=1, timestamp=1512499907092
test[0]@195: key=b'2', value=b'Message #2', tstype=1, timestamp=1512499907092
test[0]@196: key=b'3', value=b'Message #3', tstype=1, timestamp=1512499907092
commit_return.c_parts: 0x0
Segmentation fault: 11
```

examples/integration_test.py
```diff
@@ -589,21 +589,17 @@ def verify_batch_consumer():
     max_msgcnt = 100
     msgcnt = 0
 
-    while True:
+    while msgcnt < max_msgcnt:
```
edenhill: let max_msgcnt=1000 and batch_cnt=100

examples/integration_test.py
```diff
@@ -715,12 +706,9 @@ def my_on_revoke(consumer, partitions):
 
             if msgcnt == 1:
                 t_first_msg = time.time()
-            if msgcnt >= max_msgcnt:
+            elif msgcnt >= max_msgcnt:
```
edenhill: can get rid of this one now with the while loop

examples/integration_test.py
tburmeister: Having trouble running the integration test, but running a quick test I wrote on an actual Kafka stream I got:
```
consume took 1.33 seconds
1000000 messages, 662.56 MB, 497.54 MB/sec

poll tool 2.17 seconds
1000000 messages, 662.62 MB, 304.57 MB/sec
```
So definitely an improvement, though I found I had to tune the batch size to get the best results, e.g. 1k messages performs better than 10k messages or 100 messages.

examples/integration_test.py
```diff
@@ -565,6 +565,162 @@ def my_on_revoke(consumer, partitions):
...
+    max_msgcnt = 1000
+    batch_cnt = 100
+    msgcnt = 0
+
+    while msgcnt < max_msgcnt:
+        # Consume until EOF or error
+
+        # Consume messages (error()==0) or event (error()!=0)
+        msglist = c.consume(batch_cnt, 1.0)
+        assert len(msglist) == max_msgcnt
```
edenhill: this should compare to batch_cnt

examples/integration_test.py
```diff
@@ -565,6 +565,162 @@ def my_on_revoke(consumer, partitions):
...
+                offsets = c.commit(msg, async=False)
+                assert len(offsets) == 1, 'expected 1 offset, not %s' % (offsets)
+                assert offsets[0].offset == msg.offset()+1, \
+                    'expected offset %d to be committed, not %s' % \
+                    (msg.offset(), offsets)
+                print('Sync committed offset: %s' % offsets)
+
+            msgcnt += 1
+
+        print('max_msgcnt %d reached' % msgcnt)
```
edenhill: wrong indent, should be outside while loop.

examples/integration_test.py
edenhill: damn!

examples/integration_test.py
```diff
@@ -586,14 +586,15 @@ def verify_batch_consumer():
...
     msgcnt = 0
 
     while msgcnt < max_msgcnt:
         # Consume until EOF or error
 
-        # Consume message (error()==0) or event (error()!=0)
-        msglist = c.consume(max_msgcnt, 1.0)
+        # Consume messages (error()==0) or event (error()!=0)
+        msglist = c.consume(batch_cnt, 1.0)
         assert len(msglist) == max_msgcnt
```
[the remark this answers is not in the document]
tburmeister: Yep, good catch.

tburmeister:
One small thing: installing on Ubuntu I get:
```
confluent_kafka/src/Consumer.c:793:9: error: 'for' loop initial declarations are only allowed in C99 mode
             for (Py_ssize_t i = 0; i < n; i++) {
```
Would you prefer that I declare `i` outside the for loop or add `extra_compile_args = ['-std=gnu99']` to `setup.py`?

tburmeister:
Ohhhh I figured out why commits were crashing and it is my fault - will push a fix soon.

tburmeister:
I can now run the full integration test locally, without errors. I really should have confirmed the tests worked locally on master first.

tburmeister:
So what needs to be done to get this merged? Looking at the CI tests, it looks like it's passing the actual tests but failing because of pep8 violations, which I believe are addressed in https://github.com/confluentinc/confluent-kafka-python/pull/279

edenhill:
Huge thanks for doing this! :heart_eyes:

tburmeister:
Awesome! Thanks! This feature will be super useful to me. By the way, we are big librdkafka users at AppNexus.

isamaru:
Thanks! This is very useful to us. Any timeline for when this becomes part of a pipy release?

edenhill:
@isamaru We're aiming for a feature release end of january.

messense:
@edenhill Maybe you can upload a pre-release version to PyPi so we can try it? Thanks!

https://packaging.python.org/tutorials/distributing-packages/#pre-release-versioning

aswinjoseroy:
Awesome work adding this feature! Any updates as to when the pypi release will happen?

edenhill:
Final release in about 2 weeks, we'll have an rc out hopefully this week

aswinjoseroy:
Is the batch consumer released yet? Any links to usage / any documentation etc are highly appreciated! Thanks!

alannesta:
This batch load api is still undocumented, is it ready?

> Is the batch consumer released yet? Any links to usage / any documentation etc are highly appreciated! Thanks!

@aswinjoseroy  Take a look at the `consume` method in the doc. That's the new function added in this PR 
https://docs.confluent.io/current/clients/confluent-kafka-python/#confluent_kafka.Consumer.consume

Issue #287, opened 2017-12-13
Support for Message headers


johnistan:
Description
===========
[KIP-82](https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers) adds support for message headers. The librdkafka support is in a PR here https://github.com/edenhill/librdkafka/pull/1480

I wanted to track python support here and discuss the python api when it lands upstream.

The headers are an ordered multimap with `string key -> binary value` where a key can be repeated.

There are a couple ways to support this in producer and message api.

List of Tuples/Lists:
```
headers_to_send = [('key1', b'value1'), ('key2', b'value2'), ('key1', b'repeated_key')]
producer.produce('topic_name', b'message_value', headers=headers_to_send)
```

Ordered dict. Python doesn't have a multimap built in, but we could build one or just use an ordered map where values are always nested in lists.
```
headers_to_send = OrderedMultiDict()
headers_to_send['key1'] = [b'value1',  b'repeated_key']
headers_to_send['key2'] = [b'value1']

OR

headers_to_send = OrderedMultiDict()
headers_to_send['key1'] = b'value1'
headers_to_send['key1'].append(b'repeated_key')
headers_to_send['key2'] = b'value1'

producer.produce('topic_name', b'message_value', headers=headers_to_send)
```

What ever the api for producer the consumer `msg.headers()` should return the same format as the producer input. Also for multi-version kafka support mirror the `msg.timestamp()` return flag of `(int, X)` could be used.
```
msg = consumer.poll()
supported_flag, list_of_tuples = msg.headers()
print(list_of_tuples)  # [('key1', b'value1'), ('key2', b'value2'), ('key1', b'repeated_key')]
```

This isn't an exhaustive list of options but i wanted to get the discussion started.

edenhill:
Good input, @johnistan !

The nested list approach does seem like the easiest to work with from a user's perspective.
We can also add support for a standard dict in the produce() call since it is probably rare for an application to add multiple headers with the same name (interceptors might).

Another alternative is to have a richer Headers class with basic CRUD methods, as well as a "supported()" method.

edenhill:
fyi: Message headers have now landed in librdkafka.

PR #299, opened 2018-01-16
message header support

johnistan:
Initial stab at closing #287.

This implements the tuple idea mentioned in #287. Each header is a list of tuples `[(header_key, header_value)...]` This was the simplest to implement and I think will cover most use cases. For unique header_keys and user can just `dict(header_list)` for python 3.6 or `OrderedDict(header_tuple)` for before python 3.5. 

This is still a WIP. Need to:

- throw an error on incorrect version if librdkafka
- better testing around malformed headers and incorrect types

Some questions:

- Should the client perform any validation on the shape of the headers or just trust the user.
- Right now the `producer.produce()` only accepts list. We can expand to support any iter that contains key, value pairs.

### DIFF
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -308,7 +308,10 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
         long long timestamp = 0;
         rd_kafka_resp_err_t err;
 	struct Producer_msgstate *msgstate;
+#ifdef RD_KAFKA_V_HEADERS
     rd_kafka_headers_t *rd_headers = NULL;
+#endif
+
 	static char *kws[] = { "topic",
 			       "value",
 			       "key",
@@ -361,10 +364,12 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 	if (!partitioner_cb || partitioner_cb == Py_None)
 		partitioner_cb = self->u.Producer.partitioner_cb;
 
+#ifdef RD_KAFKA_V_HEADERS
     if (headers) {
         if(!(rd_headers = py_headers_to_c(headers)))
             return NULL;
     }
+#endif
 
 	/* Create msgstate if necessary, may return NULL if no callbacks
 	 * are wanted. */
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -313,23 +313,33 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 			       "value",
 			       "key",
 			       "partition",
-                   "headers",
 			       "callback",
 			       "on_delivery", /* Alias */
 			       "partitioner",
                    "timestamp",
+                   "headers",
 			       NULL };
 
 	if (!PyArg_ParseTupleAndKeywords(args, kwargs,
-					 "s|z#z#iOOOOL"
+					 "s|z#z#iOOOLO"
                                          , kws,
 					 &topic, &value, &value_len,
 					 &key, &key_len, &partition,
-					 &headers, &dr_cb, &dr_cb2, 
-                     &partitioner_cb, &timestamp))
+					 &dr_cb, &dr_cb2, &partitioner_cb,
+                     &timestamp, &headers))
 		return NULL;
 
-//TODO: Add version check for headers
+#ifndef RD_KAFKA_V_HEADERS
+    if (headers) {
+            PyErr_Format(PyExc_NotImplementedError,
+                         "Producer message headers requires "
+                         "confluent-kafka-python built for librdkafka "
+                         "version >=v0.11.3 (librdkafka runtime 0x%x, "
+                         "buildtime 0x%x)",
+                         rd_kafka_version(), RD_KAFKA_VERSION);
+            return NULL;
+    }
+#endif
 
 #if !HAVE_PRODUCEV
         if (timestamp) {
@@ -357,25 +367,9 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 
         /* Produce message */
 #if HAVE_PRODUCEV
-        if (headers){
-            int i, len;
-            len = PyList_Size(headers);
-            rd_headers = rd_kafka_headers_new(len);
-            for (i = 0; i < len; i++) {
-                const char *header_key, *header_value = NULL;
-                int header_key_len = 0, header_value_len = 0;
-
-                PyArg_ParseTuple(PyList_GET_ITEM(headers, i), "s#z#", &header_key,
-                        &header_key_len, &header_value, &header_value_len);
-
-                err = rd_kafka_header_add(rd_headers, header_key, header_key_len, header_value, header_value_len);
-                if (err) {
-                    cfl_PyErr_Format(err,
-                             "Unable to produce message: %s",
-                             rd_kafka_err2str(err));
-                    return NULL;
-                }
-            }
+        if (headers) {
+            if(!(rd_headers = py_headers_to_c(headers)))
+                return NULL;
         }
 
         err = Producer_producev(self, topic, partition,
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -252,14 +252,17 @@ int32_t Producer_partitioner_cb (const rd_kafka_topic_t *rkt,
 }
 
 
+#if HAVE_PRODUCEV
+static rd_kafka_resp_err_t
+Producer_producev (Handle *self,
+                   const char *topic, int32_t partition,
+                   const void *value, size_t value_len,
+                   const void *key, size_t key_len,
+                   void *opaque, int64_t timestamp
 #if RD_KAFKA_V_HEADERS
-static rd_kafka_resp_err_t
-Producer_producev (Handle *self,
-                   const char *topic, int32_t partition,
-                   const void *value, size_t value_len,
-                   const void *key, size_t key_len,
-                   void *opaque, int64_t timestamp,
-                   rd_kafka_headers_t *headers) {
+                   ,rd_kafka_headers_t *headers
+#endif
+                   ) {
 
         return rd_kafka_producev(self->rk,
                                  RD_KAFKA_V_MSGFLAGS(RD_KAFKA_MSG_F_COPY),
@@ -269,26 +272,9 @@ Producer_producev (Handle *self,
                                  RD_KAFKA_V_VALUE((void *)value,
                                                   (size_t)value_len),
                                  RD_KAFKA_V_TIMESTAMP(timestamp),
+#if RD_KAFKA_V_HEADERS
                                  RD_KAFKA_V_HEADERS(headers),
-                                 RD_KAFKA_V_OPAQUE(opaque),
-                                 RD_KAFKA_V_END);
-}
-#elif HAVE_PRODUCEV
-static rd_kafka_resp_err_t
-Producer_producev (Handle *self,
-                   const char *topic, int32_t partition,
-                   const void *value, size_t value_len,
-                   const void *key, size_t key_len,
-                   void *opaque, int64_t timestamp) {
-
-        return rd_kafka_producev(self->rk,
-                                 RD_KAFKA_V_MSGFLAGS(RD_KAFKA_MSG_F_COPY),
-                                 RD_KAFKA_V_TOPIC(topic),
-                                 RD_KAFKA_V_PARTITION(partition),
-                                 RD_KAFKA_V_KEY(key, (size_t)key_len),
-                                 RD_KAFKA_V_VALUE((void *)value,
-                                                  (size_t)value_len),
-                                 RD_KAFKA_V_TIMESTAMP(timestamp),
+#endif
                                  RD_KAFKA_V_OPAQUE(opaque),
                                  RD_KAFKA_V_END);
 }
@@ -351,18 +337,6 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
                      &timestamp, &headers))
 		return NULL;
 
-#ifndef RD_KAFKA_V_HEADERS
-    if (headers) {
-            PyErr_Format(PyExc_NotImplementedError,
-                         "Producer message headers requires "
-                         "confluent-kafka-python built for librdkafka "
-                         "version >=v0.11.4 (librdkafka runtime 0x%x, "
-                         "buildtime 0x%x)",
-                         rd_kafka_version(), RD_KAFKA_VERSION);
-            return NULL;
-    }
-#endif
-
 #if !HAVE_PRODUCEV
         if (timestamp) {
                 PyErr_Format(PyExc_NotImplementedError,
@@ -375,6 +349,24 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
         }
 #endif
 
+#ifndef RD_KAFKA_V_HEADERS
+    if (headers) {
+            PyErr_Format(PyExc_NotImplementedError,
+                         "Producer message headers requires "
+                         "confluent-kafka-python built for librdkafka "
+                         "version >=v0.11.4 (librdkafka runtime 0x%x, "
+                         "buildtime 0x%x)",
+                         rd_kafka_version(), RD_KAFKA_VERSION);
+            return NULL;
+    }
+#else
+    if (headers) {
+        if(!(rd_headers = py_headers_to_c(headers)))
+            return NULL;
+    }
+#endif
+
+
 	if (dr_cb2 && !dr_cb) /* Alias */
 		dr_cb = dr_cb2;
 
@@ -383,12 +375,6 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 	if (!partitioner_cb || partitioner_cb == Py_None)
 		partitioner_cb = self->u.Producer.partitioner_cb;
 
-#ifdef RD_KAFKA_V_HEADERS
-    if (headers) {
-        if(!(rd_headers = py_headers_to_c(headers)))
-            return NULL;
-    }
-#endif
 
 	/* Create msgstate if necessary, may return NULL if no callbacks
 	 * are wanted. */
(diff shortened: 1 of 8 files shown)


johnistan:
Also, do we need a `set_header` on the message object?

confluent_kafka/src/Producer.c
```diff
@@ -302,29 +304,33 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
+	PyObject *headers = NULL, *dr_cb = NULL, *dr_cb2 = NULL, *partitioner_cb = NULL;
         long long timestamp = 0;
         rd_kafka_resp_err_t err;
 	struct Producer_msgstate *msgstate;
+    rd_kafka_headers_t *rd_headers = NULL;
 	static char *kws[] = { "topic",
 			       "value",
 			       "key",
 			       "partition",
+                   "headers",
```
edenhill: Append headers to the end to not change the order of arguments.

confluent_kafka/src/Producer.c
```diff
@@ -302,29 +304,33 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
                                          , kws,
 					 &topic, &value, &value_len,
 					 &key, &key_len, &partition,
-					 &dr_cb, &dr_cb2, &partitioner_cb,
-                                         &timestamp))
+					 &headers, &dr_cb, &dr_cb2, 
+                     &partitioner_cb, &timestamp))
 		return NULL;
 
+//TODO: Add version check for headers
```
edenhill: `#ifdef RD_KAFKA_V_HEADERS` is sufficient

confluent_kafka/src/Producer.c
```diff
@@ -351,10 +357,32 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 
         /* Produce message */
 #if HAVE_PRODUCEV
+        if (headers){
+            int i, len;
```
edenhill: refactor into a separate function   headers_to_c() or something

confluent_kafka/src/confluent_kafka.c
```diff
@@ -561,6 +583,40 @@ PyObject *Message_new0 (const Handle *handle, const rd_kafka_message_t *rkm) {
 
 	self->timestamp = rd_kafka_message_timestamp(rkm, &self->tstype);
 
+    rd_kafka_headers_t *hdrs;
```
edenhill: put variable definitions at the beginning of the scope.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -561,6 +583,40 @@ PyObject *Message_new0 (const Handle *handle, const rd_kafka_message_t *rkm) {
...
+
+    if (!rd_kafka_message_headers(rkm, &hdrs)) {
+
+        size_t idx = 0;
```
edenhill: refactor into c_to_headers()  or something along those lines

examples/integration_test.py
```diff
@@ -117,11 +117,13 @@ def verify_producer():
     p = confluent_kafka.Producer(**conf)
     print('producer at %s' % p)
 
+    headers = [('foo1', 'bar'), ('population_producer_id', b'1')]
```
edenhill: add duplicate header keys and a null header

examples/integration_test.py
```diff
@@ -451,9 +453,11 @@ def print_wmark(consumer, parts):
                 break
 
         tstype, timestamp = msg.timestamp()
-        print('%s[%d]@%d: key=%s, value=%s, tstype=%d, timestamp=%s' %
+        headers = msg.headers()
+
+        print('%s[%d]@%d: key=%s, value=%s, tstype=%d, timestamp=%s headers=%s' %
               (msg.topic(), msg.partition(), msg.offset(),
-               msg.key(), msg.value(), tstype, timestamp))
+               msg.key(), msg.value(), tstype, timestamp, headers))
```
edenhill: should verify the headers here.

tests/test_Producer.py
```diff
@@ -22,6 +22,7 @@ def test_basic_api():
 
     p.produce('mytopic')
     p.produce('mytopic', value='somedata', key='a key')
+    p.produce('mytopic', value='somedata', key='a key', headers=[('headerkey', 'headervalue')])
```
edenhill: Add:
 * dup key,
 * None value
 * empty headers=[]

johnistan: done

confluent_kafka/src/Producer.c
```diff
@@ -302,29 +304,43 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
+    if (headers) {
+            PyErr_Format(PyExc_NotImplementedError,
+                         "Producer message headers requires "
+                         "confluent-kafka-python built for librdkafka "
+                         "version >=v0.11.3 (librdkafka runtime 0x%x, "
+                         "buildtime 0x%x)",
+                         rd_kafka_version(), RD_KAFKA_VERSION);
+            return NULL;
+    }
+#endif
```
johnistan: The library will not compile with an older version of librdkafka so i don't think this is needed.

confluent_kafka/src/Producer.c
```diff
@@ -302,29 +304,43 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
-                                         &timestamp))
+                     &timestamp, &headers))
 		return NULL;
 
+#ifndef RD_KAFKA_V_HEADERS
+    if (headers) {
+            PyErr_Format(PyExc_NotImplementedError,
+                         "Producer message headers requires "
+                         "confluent-kafka-python built for librdkafka "
+                         "version >=v0.11.3 (librdkafka runtime 0x%x, "
```
edenhill: Change to v0.11.4

confluent_kafka/src/Producer.c
edenhill: It should be able to build with older librdkafka versions as well, thus the ifdef guards, and the runtime checks.

confluent_kafka/src/Producer.c
```diff
@@ -351,10 +367,16 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 
         /* Produce message */
 #if HAVE_PRODUCEV
+        if (headers) {
+            if(!(rd_headers = py_headers_to_c(headers)))
+                return NULL;
```
edenhill: move this above the msgstate allocation to avoid memory leak on failure

confluent_kafka/src/confluent_kafka.c
```diff
@@ -561,6 +584,10 @@ PyObject *Message_new0 (const Handle *handle, const rd_kafka_message_t *rkm) {
 
 	self->timestamp = rd_kafka_message_timestamp(rkm, &self->tstype);
 
+    if (!rd_kafka_message_headers(rkm, &hdrs)) {
```
edenhill: Object creation is the most expensive thing in Python so I think we should put this off until headers are actually requested, in the headers() accessor method.

So I suggest we just extract the C headers list here and keep it on the Message object, making sure to call set_headers(NULL) on the message to transfer ownership to the Python code.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -883,6 +910,78 @@ rd_kafka_topic_partition_list_t *py_to_c_parts (PyObject *plist) {
...
+
+    len = PyList_Size(headers_plist);
+    rd_headers = rd_kafka_headers_new(len);
+
+    for (i = 0; i < len; i++) {
+
+        if(!PyArg_ParseTuple(PyList_GET_ITEM(headers_plist, i), "s#z#", &header_key,
+                &header_key_len, &header_value, &header_value_len)){
+            PyErr_SetString(PyExc_TypeError,
+                    "Headers are expected to be a tuple of (key, value)");
```
edenhill: destroy(rd_headers)

confluent_kafka/src/confluent_kafka.h
```diff
@@ -286,6 +287,7 @@ typedef struct {
 	PyObject *topic;
 	PyObject *value;
 	PyObject *key;
+	PyObject *headers;
```
edenhill: also add c_headers here

tests/test_Producer.py
```diff
@@ -22,6 +23,14 @@ def test_basic_api():
...
+    p.produce('mytopic', value='somedata', key='a key', headers=[('dupkey', 'dupvalue'), ('dupkey', 'dupvalue')])
```
edenhill: have the second dupvalue be something different to make sure they are actually treated as different entities

edenhill:
Hey @johnistan , do you think you'll have time to address the final review points in the coming week?
We're looking to release new versions of clients end of february and it would be great to get this in.
If you don't have time I can pick up the work.

johnistan:
Let me see what I can get done.

johnistan:
I am hitting an issue that i suspect is just a noob C problem hopefully you can help.

I added a pointer to the Message struct `rd_kafka_headers_t *c_headers;`

I am writing the c_headers to the Message struct like so:
`rd_kafka_message_headers(rkm, &self->c_headers)`

In that function scope I am able to access and use the headers no issue:

```
    if (!rd_kafka_message_headers(rkm, &self->c_headers)) {
        size_t header_size;
        header_size = rd_kafka_header_cnt(self->c_headers);
        printf("after c_header set size %zu\n", header_size);
   }
```
prints `after c_header set size 2`

However, later when I try to access the same struct in the `Message_headers` class method the c_headers have changed and are no longer usable:

```
    } else if (self->c_headers) {
        size_t header_size;
        header_size = rd_kafka_header_cnt(self->c_headers);
        printf("inside Message_headers c_header size %zu\n", header_size);
```
I get `inside Message_headers c_header size 0`

Any idea what I'm doing wrong?

johnistan:
ok found it I actually need to copy the headers to the struct not just set the pointer.

edenhill:
message_set_headers() makes the message_t (and thus librdkafka) take ownership of the passed headers.
This also means you can use message_set_headers(rkm, NULL); to move ownership of headers from librdkafka to the application.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -585,7 +612,7 @@ PyObject *Message_new0 (const Handle *handle, const rd_kafka_message_t *rkm) {
 	self->timestamp = rd_kafka_message_timestamp(rkm, &self->tstype);
 
     if (!rd_kafka_message_headers(rkm, &hdrs)) {
-        self->headers = c_headers_to_py(hdrs);
+        self->c_headers = rd_kafka_headers_copy(hdrs);
```
edenhill: We can take ownership of the headers by doing `rd_kafka_message_set_headers(rkm, NULL)`, to avoid copying them.

johnistan: Ok i think i follow. We can store the address of the existing c_headers on the python Message struct. Then we disassociate the headers from rkm so that when it's destroyed the underlying headers aren't.

When I make this change I am getting a warning from the compiler that the `rkm` is declared as a const but we are mutating it. Should i change the signature to match?

edenhill: Right on!

Yes, remove const from Message_new0()

johnistan: I am not doing something right. If I ignore the warning the headers are still destroyed.

https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_msg.c#L1061

`rd_kafka_message_set_headers` calls `rd_kafka_headers_destroy` on the existing headers. I am reading that right?

edenhill: You are absolutely right, sorry for the confusion.
Let me add a `rd_kafka_message_detach_headers(rkm, &hdrs)` API for this purpose.

johnistan: There is another issue with removing the const from the `Message_new0`. This is used by the delivery callback and that has a signature of const too. 

So you end up with:

```
  In file included from confluent_kafka/src/confluent_kafka.h:21:0,
                   from confluent_kafka/src/Producer.c:17:
  /home/vagrant/repos/librdkafka/out/include/librdkafka/rdkafka.h:1326:6: note: expected ‘void (*)(struct rd_kafka_t *, const struct rd_kafka_message_t *, void *)’ but argument is of type ‘void (*)(struct rd_kafka_t *, struct rd_kafka_message_t *, void *)’
   void rd_kafka_conf_set_dr_msg_cb(rd_kafka_conf_t *conf,
```

johnistan:
this uses the new detach api.

when i switched out the const message i get this compiler warning:

```confluent_kafka/src/Producer.c: In function ‘Producer_init’:
  confluent_kafka/src/Producer.c:564:9: warning: passing argument 2 of ‘rd_kafka_conf_set_dr_msg_cb’ from incompatible pointer type [enabled by default]
           rd_kafka_conf_set_dr_msg_cb(conf, dr_msg_cb);
           ^
  In file included from confluent_kafka/src/confluent_kafka.h:21:0,
                   from confluent_kafka/src/Producer.c:17:
  /home/vagrant/repos/librdkafka/out/include/librdkafka/rdkafka.h:1341:6: note: expected ‘void (*)(struct rd_kafka_t *, const struct rd_kafka_message_t *, void *)’ but argument is of type ‘void (*)(struct rd_kafka_t *, struct rd_kafka_message_t *, void *)’
   void rd_kafka_conf_set_dr_msg_cb(rd_kafka_conf_t *conf,
        ^```

The callback use the same Message_new0 so it causes issues in the producer callback.

**Comment by edenhill** [comment]
Okay, let's move detach_headers() outside of Message_new0() and only call it from the Consumer code.
We'll add a note to the Producer documentation that message headers are not available in the delivery report callback (we can add it later if there's demand for it).

**Comment by johnistan** [comment]
Added the detach code to `consumer.poll()` and batch `consume.consume()`

**Comment by edenhill** [comment]
That's it, it's done!
Just needs to be rebased on master so it merges cleanly and then we're good to go

**Comment by johnistan** [comment]
pulled master.

**Comment by edenhill** [comment]
Still no luck :/

**Comment by johnistan** [comment]
rebased off master. try now

**Comment by edenhill** [comment]
Travis CI status updates no longer works for whatever reason, they show some builds problems with older librdkafkas, but I can sort that out once this is merged.

**Comment by johnistan** [comment]
ah i see them now. 

I can try to get this building on older versions.

```
    confluent_kafka/src/confluent_kafka.h:270:28: error: unknown type name ‘rd_kafka_headers_t’
     PyObject *c_headers_to_py (rd_kafka_headers_t *headers);
                                ^
    confluent_kafka/src/confluent_kafka.h:291:2: error: unknown type name ‘rd_kafka_headers_t’
      rd_kafka_headers_t *c_headers;
      ^
```

need to wrap these 

```
#ifndef RD_KAFKA_V_HEADERS
```

**Comment by edenhill** [comment]
That'd be great!

**Comment by johnistan** [comment]
I may have to ask for your help. The latest commit should support all librdkafka versions from 0.9.2 but it turns some of the code into a real swamp.

  HAVE_PRODUCEV is defined from RD_KAFKA_V_TIMESTAMP. The 0.11.0 version, for example, has RD_KAFKA_V_TIMESTAMP but not HEADERS. I have had to guard the internal Producer_producev definition. I suspect there is a cleaner way.

File: confluent_kafka/src/Producer.c [hunk 17]
```diff
@@ -361,10 +364,12 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
 	if (!partitioner_cb || partitioner_cb == Py_None)
 		partitioner_cb = self->u.Producer.partitioner_cb;
 
+#ifdef RD_KAFKA_V_HEADERS
```
edenhill: You actually want to bail out with an exception if headers is non-NULL and #ifndef RD_KAFKA_V_HEADERS.
See same thing for timestamp above.

confluent_kafka/src/confluent_kafka.h
```diff
@@ -288,7 +290,9 @@ typedef struct {
 	PyObject *value;
 	PyObject *key;
 	PyObject *headers;
+#ifdef RD_KAFKA_V_HEADERS
```
edenhill: could include `*headers` above too, right?

tests/test_Producer.py
```diff
@@ -65,6 +56,41 @@ def test_produce_timestamp():
     p.flush()
 
 
+# Should be updated to 0.11.4 when it is released
+@pytest.mark.skipif(libversion()[1] < 0x000b0000,
```
edenhill: 0x000b0400

tests/test_Producer.py
```diff
@@ -65,6 +56,41 @@ def test_produce_timestamp():
...
+    p.produce('mytopic', value='somedata', key='a key', headers=[])
+
+    with pytest.raises(TypeError) as ex:
+        p.produce('mytopic', value='somedata', key='a key', headers=[('malformed_header')])
+    assert 'Headers are expected to be a tuple of (key, value)' == str(ex.value)
+
+    p.flush()
+
+# Should be updated to 0.11.4 when it is released
+@pytest.mark.skipif(libversion()[1] >= 0x000b0000,
```
edenhill: dito

confluent_kafka/src/Producer.c
```diff
@@ -252,7 +252,7 @@ int32_t Producer_partitioner_cb (const rd_kafka_topic_t *rkt,
 }
 
 
-#if HAVE_PRODUCEV
+#if RD_KAFKA_V_HEADERS
```
edenhill: I would probably stick this just around the headers argument in the function argument list, to avoid duplication.

confluent_kafka/src/Producer.c
```diff
@@ -377,11 +396,18 @@ static PyObject *Producer_produce (Handle *self, PyObject *args,
...
         err = Producer_producev(self, topic, partition,
                                 value, value_len,
                                 key, key_len,
                                 msgstate, timestamp,
                                 rd_headers);
+#else
+        err = Producer_producev(self, topic, partition,
+                                value, value_len,
+                                key, key_len,
+                                msgstate, timestamp);
```
edenhill: and same thing here

edenhill:
Big thanks for putting all the work into this PR!

PR #303, opened 2018-01-23
Broker loss example consumer failure

johnistan:
This is a reproduction of an error I am having trouble dealing with. If a broker goes down I am unable to successfully consume from any topic that has that broker as a leader. I know this is a core part of the kafka protocol but i am having issues making the python client gracefully handle broker loss. This is causing much pain for my workers. Am I missing something or is this an issue with the client?

added a docker file and compose to help spin up brokers. tox/travis might work. To run:

```
docker-compose build
docker-compose up -d
py.test examples/example_failed_broker.py -sv
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ matrix:
 
 install:
  - if [[ -z $CIBW_BEFORE_BUILD ]]; then rm -rf tmp-build ; tools/bootstrap-librdkafka.sh --require-ssl ${LIBRDKAFKA_VERSION} tmp-build ; fi
- - if [[ -z $CIBW_BEFORE_BUILD ]]; then pip install --upgrade pip && pip install pytest-timeout flake8 ; fi
+ - if [[ -z $CIBW_BEFORE_BUILD ]]; then pip install --upgrade pip && pip install pytest-timeout flake8 && pip install docker-compose; fi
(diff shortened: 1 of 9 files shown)


edenhill:
Wow, this is great! Thanks for taking the time to set this up!

edenhill:
Where do I find the `test-kafka` image?

```
$ docker-compose up -d
Creating network "confluentkafkapython_default" with the default driver
Pulling kafka1 (test-kafka:latest)...
ERROR: pull access denied for test-kafka, repository does not exist or may require 'docker login'
```

johnistan:
sorry try `docker-compose build` first
the image def is in the pr

edenhill:
what librdkafka version are you using?

johnistan:
using conda

`librdkafka                0.9.4                         0    conda-forge`

edenhill:
Okay, tried to reproduce this with librdkafka master and it seems like the real issue is that it takes more than 30s for the paused broker to have its failure detected by the cluster.
When increasing the timeout to 90s it eventually reassigns the paused broker's partitions to the replicas and consumption continues

johnistan:
Thanks for taking a look. I reran on this with librdkafka master I get correct behavior. With librdkafka 0.9.4 no amount of timeout seems to help.

edenhill:
Ok, that is great news. We're looking to release a new version of librdkafka and python client end of february.

PR #306, opened 2018-01-25
Support decoding empty string

qiao-meng-zefr:
Python client throw the `message is too small to decode` error when trying to consume message with an empty string as key in kafka.

If we allow publishing empty key, we should be able to consume it. So this PR is a fix to allow decode empty string.

### DIFF
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -204,6 +204,9 @@ class MessageSerializer(object):
 
         if message is None:
             return None
+          
+        if len(message) == 0:
+            return ""
 
         if len(message) <= 5:
             raise SerializerError("message is too small to decode")


confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -204,6 +204,9 @@ def decode_message(self, message):
 
         if message is None:
             return None
+          
+        if len(message) == 0:
+            return ""
```
edenhill: I don't think this is what the application expects, the deserializer is set up to return a certain deserialized object type - returning an empty string will most likely not be that object.

So I believe the existing behaviour is correct.

qiao-meng-zefr: What I observed is, when the message key is empty string, the input `message` object is just `b''`, so there's no magic byte and schema_id.

So would it be better if check `if message == b''` here?

edenhill: My point is that an AvroConsumer will expect properly serialized messages, any other message should race a deserialization exception, as it does today.

qiao-meng-zefr: ok that's fair.

Issue #164, opened 2017-03-29
Expose seek functionality


davidblewett:
I was looking for a method of rewinding a given consumer group's offset in a given topic by using https://cwiki.apache.org/confluence/display/KAFKA/FAQ#FAQ-HowcanIrewindtheoffsetintheconsumer? , but didn't find it immediately obvious. The Go wrapper exposes the "seek" interface ( https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/consumer.go#L197 ), but I don't see something similar in the Python binding.

davidblewett:
For context, this functionality would be used in a controlled maintenance window. I.e., we stop all members of the consumer group, run a script to rewind their offsets, then restart everything.

edenhill:
Until seek() is added to the Python client you can use assign() by specifying each partition's offset.

davidblewett:
@edenhill I think I tried that, but it didn't seem to "stick". What's the correct method to check the stored value then?

edenhill:
Do you want to rewind a running consumer, or the committed offsets?

davidblewett:
The desired effect is to rewind where a given consumer group is in a topic. The least error prone method I thought would be to stop all consumer processes in the group, run a script to rewind the committed offset for each parttion, then restart all consumer processes.

Our previous environment was using ZooKeeper-based offsets. In that environment, we would stop the group, issue ZooKeeper commands to rewind the offset, then restart the group.

edenhill:
You could do the same using committed() (to get the committed offsets) and commit() (to rewrite the offsets).

craigtmoore:
I was able to do it by unassigning and then reassigning the consumer at a given offset, using the topic partition. Example:

` consumer.commit(offsets=[topic_partition], async=False)`
`consumer.unassign()`
`topic_partition.offset = 0`
`consumer.assign([topic_partition])`

affansyed:
@edenhill when is this feature actually be the "next feature"? if no one is taking up, can you point to where in the code one can look at to help and fix?

edenhill:
PR here: https://github.com/confluentinc/confluent-kafka-python/pull/312

edenhill:
Available in v0.11.4

PR #312, opened 2018-02-12
Add Consumer.seek() (#164)

edenhill:


confluent_kafka/src/Consumer.c
```diff
@@ -988,7 +1037,7 @@ static PyMethodDef Consumer_methods[] = {
 	  "  .. note: Callbacks may be called from this method, "
 	  "such as ``on_assign``, ``on_revoke``, et.al.\n"
 	  "\n"
-	  "  :param float timeout: Maximum time to block waiting for message, event or callback.\n"
+	  "  :param float timeout: Maximum time to block waiting for message, event or callback. (Seconds)\n"
```
mhowlett: I like the idea of including units in the parameter name (even if the convention for timeouts in the language is seconds). `secondsTimeout` is what I'd probably go with, though it may be a bit unusual. Not expecting you to change these, just saying. Much more effective than docs.

examples/integration_test.py
```diff
@@ -402,6 +402,32 @@ def print_commit_result(err, partitions):
...
+    tp = confluent_kafka.TopicPartition(seek_to_msg.topic(),
+                                        seek_to_msg.partition(),
+                                        seek_to_msg.offset())
+    print('seek: Seeking to %s' % tp)
+    c.seek(tp)
+
+    while True:
+        msg = c.poll()
+        assert msg is not None
+        if msg.error():
```
mhowlett: comment is unrelated to this PR, but having non error events exposed via `error()` is not a good API choice imo. it'll confuse people.

examples/integration_test.py
```diff
@@ -402,6 +402,32 @@ def print_commit_result(err, partitions):
...
+            print('seek: Ignoring non-message: %s' % msg)
+            continue
+
+        if msg.topic() != seek_to_msg.topic() or msg.partition() != seek_to_msg.partition():
```
mhowlett: i think this would be better as an assert. in practice, the test'll still fail, but i think more correct as an assert.

mhowlett:
LGTM, aside from a few comments which you can do what you like with :-)

examples/integration_test.py
edenhill: Right, the idea was to avoid exceptions in the fast path and treat consumer errors as non-fatals, which maps to what librdkafka is doing, but yes, it does confuse people and we haven't been great with documentation what errors to handle how.

As for error vs event; the only non-error event is PARTITION_EOF (which can be turned off by config (which should be the default, really))

examples/integration_test.py
edenhill: The topic may have multiple partitions and the subscribe() will most likely have all partitions assigned, but we only care about one single partition here, the one we called seek() on.
So it should just ignore messages for other partitions, an assert would not be correct.

confluent_kafka/src/Consumer.c
edenhill: I believe that is atypical for Python, and we can't change a named parameter anyway, so no luck, Matt :)

PR #315, opened 2018-02-16
WIP: Add support for KIP-98 headers

florean:
Add support for message headers.  This is functionally complete and current tests pass, but I still need to add new tests (and probably poke at it in Python 2.7).
This supports headers as either a dict of str keys and byte values or an iterable of str/byte iterables.  For producing messages, `produce()` will now take a `headers` keyword arg with headers passed as either type.  For receiving messages, the Message object now has a `headers()` method that returns the dict form and a `raw_headers()` method that preserves the KIP-98 requirements for ordering and duplicate values.
Usage looks like:
```Python
p = Producer()
p.produce("topic", "message 1", headers={"k1": b"v1"})
p.produce("topic", "message 2", headers=(("k1", b"v1"), ("k1", b"v2")))

c = Consumer()
c.subscribe(['topic'])
m = c.poll()
m.value()
# b'message 1'
m.headers()
# {'k1': b'v1'}
m.raw_headers
# (('k1', b'v1'),)
m = c.poll()
m.value()
# b'message 2'
m.headers()
# {'k1': b'v2'}
m.raw_headers()
(('k1', b'v1'), ('k1', b'v2'))
```

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -884,7 +884,7 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
         if (!rkm)
                 Py_RETURN_NONE;
 
-        msgobj = Message_new0(self, rkm);
+        msgobj = Message_new0(self, rkm, true);
         rd_kafka_message_destroy(rkm);
 
         return msgobj;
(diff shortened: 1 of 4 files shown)


florean:
[clabot:check]

edenhill:
Thank you for your contribution, it is very much appreciated.
There is however an existing PR for headers support that is just about to get merged:
https://github.com/confluentinc/confluent-kafka-python/pull/299

PR #317, opened 2018-02-17
Fix #if directives in Producer.

florean:
Some checks that should be #ifdef directives were instead #if directives,
causing headers to not be sent with messages.


edenhill:
Thank you!

PR #318, opened 2018-02-19
Add support for logging.logger

edenhill:


tests/test_log.py
```diff
@@ -0,0 +1,77 @@
...
+def test_logging_constructor():
+    """ Verify different forms of constructors """
+
+    for how in ['dict', 'dict+kwarg', 'kwarg']:
+        logger = logging.getLogger('producer: ' + how)
+        logger.setLevel(logging.DEBUG)
+        f = CountingFilter(logger.name)
+        logger.addFilter(f)
+
+        if how is 'dict':
```
ewencp: `is` is for identity. these should be `==`

tests/test_log.py
```diff
@@ -0,0 +1,77 @@
...
+            p = confluent_kafka.Producer({'debug': 'all', 'logger': logger})
+        elif how is 'dict+kwarg':
+            p = confluent_kafka.Producer({'debug': 'all'}, logger=logger)
+        elif how is 'kwarg':
+            conf = {'debug': 'all', 'logger': logger}
+            p = confluent_kafka.Producer(**conf)
+        else:
+            raise('Notreached')
```
jeffwidman: Is `raise()` some kind of custom function? Otherwise typically you would write `raise ExceptionType('message')`... and you'd want to raise something more specific than a generic `Exception`... `RuntimeError` is probably fine here.

tests/test_log.py
```diff
@@ -0,0 +1,77 @@
...
+    f = CountingFilter('consumer')
+    logger.addFilter(f)
+
+    kc = confluent_kafka.Consumer({'group.id': 'test',
+                                   'debug': 'all'},
+                                  logger=logger)
+    while f.cnt == 0:
+        kc.poll(timeout=0.5)
+
+    print('%s: %d log messages seen' % (f.name, f.cnt))
```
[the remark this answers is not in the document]
edenhill: the poll-loop will not exit if f.cnt == 0, the test will eventually time out instead.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1336,6 +1371,7 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
 	rd_kafka_topic_conf_t *tconf;
 	Py_ssize_t pos = 0;
 	PyObject *ko, *vo;
+        PyObject *confdict = NULL;
```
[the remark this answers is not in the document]
edenhill: Yeah, we'll re-indent the entire code base when we have a style checker in place

tests/test_log.py
edenhill: You are right! The ping-ponging between languages has me confused.
Fixing

johnsonyep:
@edenhill how can I poll messages to the topic i set

PR #319, opened 2018-02-19
Rename async -> asynchronous arguments (but keep async for now) #21

edenhill:


confluent_kafka/src/Consumer.c
```diff
@@ -386,8 +386,8 @@ static PyObject *Consumer_commit (Handle *self, PyObject *args,
                 return NULL;
         }
 
-	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|OOO", kws,
-					 &msg, &offsets, &async_o))
+	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|OOOO", kws,
+					 &msg, &offsets, &async_o, &async_o))
```
ah-: That's quite nice, does it just overwrite based on which one is set?

edenhill: Yep, this works since ParseTupleAndKeywords returns borrowed references and does not increase the refcount for each matched object (which wouldn't work here).

mhowlett:
LGTM (with the disclaimer i'm not a Python expert), assuming there is good enough motivation for the name change in the first place.

edenhill:
Shout out to @johnnoone for the initial PR, thank you!

PR #331, opened 2018-03-10
Added kw for poll method in avroconsumer to be able to handle non-avro

fishstamp82:
schema encoded keys

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -95,7 +95,7 @@ class AvroConsumer(Consumer):
         super(AvroConsumer, self).__init__(config)
         self._serializer = MessageSerializer(schema_registry)
 
-    def poll(self, timeout=None):
+    def poll(self, timeout=None, key_schema=True):
         """
         This is an overriden method from confluent_kafka.Consumer class. This handles message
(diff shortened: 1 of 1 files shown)


confluent_kafka/avro/__init__.py
```diff
@@ -114,7 +114,9 @@ def poll(self, timeout=None):
             if message.value() is not None:
                 decoded_value = self._serializer.decode_message(message.value())
                 message.set_value(decoded_value)
-            if message.key() is not None:
+            if message.key() is not None and key_schema:
                 decoded_key = self._serializer.decode_message(message.key())
                 message.set_key(decoded_key)
+            elif message.key() is not None and not key_schema:
+                message.set_key(message.key().decode())
```
edenhill: This seems to make the assumption that keys are UTF-8 strings

fishstamp82: You are right and it was a dirty hack to fix it in my client. A better solution is to add key words for AvroConsumer to set these options, and not add magic keywords for the Poll method.


In any case I don't think I can do this, as I really don't understand the implications of the ACL

rnpridgeon:
closed for inactivity

PR #340, opened 2018-03-23
Restructure code to be more Pythonic and adhere to PEP8 style

maxzheng:
So it looks good and then we look good. :) Otherwise, it's more readable and promotes good practice as many will copy/paste.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,19 +120,19 @@ while True:
     try:
         msg = c.poll(10)
 
-        if msg:
-            if msg.error():
-                if msg.error().code() == KafkaError._PARTITION_EOF:
-                    continue
-                else:
-                    print(msg.error())
-                    break
-
-            print(msg.value())
-
     except SerializerError as e:
         print("Message deserialization failed for {}: {}".format(msg, e))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,21 +108,31 @@ from confluent_kafka import KafkaError
...
-                print(msg.error())
-                running = False
+
     except SerializerError as e:
-        print("Message deserialization failed for %s: %s" % (msg, e))
-        running = False
+        print("Message deserialization failed for {}: {}".format(msg, e))
+        break
+
+    if msg:
```
edenhill: I see that the original code has a bug:
Bool comparison operator calls _msg.__len__, which is implemented to return the number of bytes of the message **value**. Tombstone messages however have a null value (but a non-empty key) and thus a __len__ of 0, which will make these messages ignored by this code.

So this should be changed to:
```
   if msg is None:
       continue
   if msg.error():
...
```

maxzheng:
Ping @edenhill

edenhill:
Thank you for this!

PR #345, opened 2018-03-28
Allow passing headers as both list(tuples) and dict()

edenhill:
This also fixes type checking, the headers were assumed to be a List,
if they were not (such as a dict) it lead to an unhandled exception in CPython that was not propagated, but instead triggered spontaneously for some sub-sequent API call for no apparent reason.

Added test for binary header values.


mhowlett:
LGTM

PR #347, opened 2018-03-29
Check that produce headers are of correct type

edenhill:
This is a minimal short term fix, PR #345 fixes this and adds support for dict headers.


mhowlett:
LGTM

PR #352, opened 2018-04-09
Expose new partitioners

johnistan:
murmur and murmur_random added in librdkafka 0.11.4


edenhill:
These should be directly settable via the standard string-based configuration API now, try:
`Producer({"default.topic.config": {"partitioner":"murmur2_random"}})`

johnistan:
Awesome! will open a doc ticket to make that explicit.

Issue #350, opened 2018-04-04
auto.offset.reset ignored when specified at top level


mhowlett:
Description
===========

Specifying 'auto.offset.reset' as a top level property does not result in an error, but seems to be ignored. 

How to reproduce
================

```
c = Consumer({
    'bootstrap.servers': 'SASL_SSL://<ccloud bootstrap servers>',
    'api.version.request': True,
    'broker.version.fallback': '0.10.0.0',
    'api.version.fallback.ms': 0,
    'sasl.mechanisms': 'PLAIN',
    'security.protocol': 'SASL_SSL',
    'ssl.ca.location': '/usr/local/etc/openssl/cert.pem',
    'sasl.username': '<omitted>',
    'sasl.password': '<omitted>',
    'group.id': str(uuid.uuid1()),
    'auto.offset.reset': 'smallest'
    # 'default.topic.config': {'auto.offset.reset': 'smallest'} - this works
})

c.subscribe(['python-test-topic'])

try:
    while True:
        msg = c.poll(0.1)
        if msg is None:
            continue
        elif not msg.error():
            print('consumed: {0}'.format(msg.value()))
        elif msg.error().code() == KafkaError._PARTITION_EOF:
            print('end of partition: {0}/{1}'.format(msg.topic(), msg.partition()))
        else:
            print('error: {0}'.format(msg.error().str()))

except KeyboardInterrupt:
    pass

finally:
    c.close()
```

Checklist
=========
Please provide the following information:

 - [ ] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):  libversion: ('0.11.4-RC1', 722121), version: ('0.11.4', 721920)
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
We should merge default.topic.config with the global dict and then rely on global-topic fallthru.

knil-sama:
Both didn't work on the AvroConsumer :sob:

rnpridgeon:
This should be fixed with https://github.com/confluentinc/confluent-kafka-python/pull/446 which will be in the next release.  The AvroConsumer passes the value through to librdkafka so that's a bit odd you are finding this does not work. 

[...]

PR #353, opened 2018-04-09
added confluent cloud example

mhowlett:


examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+#
+# Confluent Cloud does not auto-create topics. You will need to use the ccloud
+# cli to create the python-test-topic topic before running this example. The
+# <ccloud bootstrap servers>, <ccloud key> and <ccloud secret> parameters are
+# available via the confluent cloud web interface. For more information,
+# refer to the quick-start:
+#
+# https://docs.confluent.io/current/cloud-quickstart.html
+#
+# to execute using Python 2.x:
```
mhowlett: as a non-python user, I actually appreciate having all this spelled out so I don't need to go look it up, but maybe it doesn't belong here.

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+})
+
+def acked(err, msg):
+    if err is not None:
+        print("failed to deliver message: {0}".format(err.str()))
+    else:
+        print("produced to: {0}/{1}/{2}".format(msg.topic(), msg.partition(), msg.offset()))
+
+p.produce('python-test-topic', value='python test value', callback=acked)
+p.flush(10)
```
mhowlett: i was thinking the `poll` explanation is too long to embed here and we should point to an explanation somewhere else, but feel free to suggest an explanation if you're feeling less lazy than me.

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+        elif msg.error().code() == KafkaError._PARTITION_EOF:
+            print('end of partition: {0}/{1}'.format(msg.topic(), msg.partition()))
+        else:
+            print('error: {0}'.format(msg.error().str()))
+
+except KeyboardInterrupt:
+    pass
+
+finally:
+    c.close()
```
mhowlett: why is there no `producer.close`?

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+from confluent_kafka import Producer, Consumer, KafkaError
+
+p = Producer({
+    'bootstrap.servers': '<bootstrap servers>',
+    'api.version.request': True,
+    'broker.version.fallback': '0.10.0.0',
+    'api.version.fallback.ms': 0,
+    'sasl.mechanisms': 'PLAIN',
+    'security.protocol': 'SASL_SSL',
+    'ssl.ca.location': '/usr/local/etc/openssl/cert.pem',
```
edenhill: Is this really needed with v0.11.4?

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+#   pip install confluent_kafka
+#   python example.py
+#   deactivate
+
+import uuid
+from confluent_kafka import Producer, Consumer, KafkaError
+
+p = Producer({
+    'bootstrap.servers': '<bootstrap servers>',
+    'api.version.request': True,
```
edenhill: not needed

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+#   source ccloud_example/bin/activate
+#   pip install confluent_kafka
+#   python example.py
+#   deactivate
+
+import uuid
+from confluent_kafka import Producer, Consumer, KafkaError
+
+p = Producer({
+    'bootstrap.servers': '<bootstrap servers>',
```
edenhill: `'<ccloud bootstrap servers>'`

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+    'broker.version.fallback': '0.10.0.0',
+    'api.version.fallback.ms': 0,
+    'sasl.mechanisms': 'PLAIN',
+    'security.protocol': 'SASL_SSL',
+    'ssl.ca.location': '/usr/local/etc/openssl/cert.pem',
+    'sasl.username': '<key>',
+    'sasl.password': '<secret>'
+})
+
+def acked(err, msg):
```
edenhill: add docstring what this does.

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+    if err is not None:
+        print("failed to deliver message: {0}".format(err.str()))
+    else:
+        print("produced to: {0}/{1}/{2}".format(msg.topic(), msg.partition(), msg.offset()))
```
edenhill: use our existing format: `topic [part] @ offset`

examples/confluent_cloud.py
edenhill: Add a comment like:
> flush() is typically called when the producer is done sending messages to wait for outstanding messages to be transmitted to the broker and delivery report callbacks to get called.
> For continous producing make sure to call p.poll(0) after each produce() call to trigger delivery report callbacks.

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+
+p.produce('python-test-topic', value='python test value', callback=acked)
+p.flush(10)
+
+c = Consumer({
```
edenhill: ditos

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+    'bootstrap.servers': '<bootstrap servers>',
+    'api.version.request': True,
+    'broker.version.fallback': '0.10.0.0',
+    'api.version.fallback.ms': 0,
+    'sasl.mechanisms': 'PLAIN',
+    'security.protocol': 'SASL_SSL',
+    'ssl.ca.location': '/usr/local/etc/openssl/cert.pem',
+    'sasl.username': '<key>',
+    'sasl.password': '<secret>',
+    'group.id': str(uuid.uuid1()),
```
edenhill: add a comment that this will create a new consumer group on each invocation

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+    'default.topic.config': {'auto.offset.reset': 'smallest'}
```
edenhill: default.topic.config is no longer needed, set it on the global level.

examples/confluent_cloud.py
edenhill: consumer.close() cleanly leaves the group and commits final offsets, since this may trigger callbacks (such as rebalancew and offset commit results) it can't be put in a dtor but an explicit call is needed.

Destructing the producer simply purges all outstanding messages, that's why you call flush() before destructing.
This leaves the application with an option if they care about the outstanding messages or not.

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+# limitations under the License.
+
+
+# This is a simple example demonstrating how to produce a message to
+# Confluent Cloud then read it back again.
+#
+# https://www.confluent.io/confluent-cloud/
+#
+# Confluent Cloud does not auto-create topics. You will need to use the ccloud
+# cli to create the python-test-topic topic before running this example. The
```
edenhill: Example command how to do that?

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+
+p = Producer({
+    'bootstrap.servers': '<bootstrap servers>',
+    'api.version.request': True,
+    'broker.version.fallback': '0.10.0.0',
+    'api.version.fallback.ms': 0,
+    'sasl.mechanisms': 'PLAIN',
+    'security.protocol': 'SASL_SSL',
+    'ssl.ca.location': '/usr/local/etc/openssl/cert.pem',
+    'sasl.username': '<key>',
```
edenhill: `<ccloud key>` and `<ccloud secret>` as stated in the introduction

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+
+try:
+    while True:
+        msg = c.poll(0.1)
+        if msg is None:
+            continue
+        elif not msg.error():
+            print('consumed: {0}'.format(msg.value()))
+        elif msg.error().code() == KafkaError._PARTITION_EOF:
+            print('end of partition: {0}/{1}'.format(msg.topic(), msg.partition()))
```
edenhill: Use our existing format (and include the offset, which is relevant here): `topic [partition] @ offset`

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+    'sasl.password': '<secret>',
+    'group.id': str(uuid.uuid1()),
+    'default.topic.config': {'auto.offset.reset': 'smallest'}
+})
+
+c.subscribe(['python-test-topic'])
+
+try:
+    while True:
+        msg = c.poll(0.1)
```
edenhill: `# Wait for message or event/error`

examples/confluent_cloud.py
edenhill: `# Leave group and commit final offsets`

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+        if msg is None:
+            continue
```
edenhill:
```
# No message available within timeout.
# Initial message consumption may take up to `session.timeout.ms` for the group to rebalance and start consuming
```

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+# Confluent Cloud does not auto-create topics. You will need to use the ccloud
+# cli to create the python-test-topic topic before running this example. The
+# <ccloud bootstrap servers>, <ccloud key> and <ccloud secret> parameters are
+# available via the confluent cloud web interface. For more information,
+# refer to the quick-start:
+#
+# https://docs.confluent.io/current/cloud-quickstart.html
+#
+# to execute using Python 2.x:
+#   virtualenv ccloud_example
```
edenhill: prefix with `$` to indicate shell prompt

examples/confluent_cloud.py
```diff
@@ -0,0 +1,101 @@
...
+#   source ccloud_example/bin/activate
+#   pip install confluent_kafka
+#   python example.py
```
edenhill: the file is called examples/confluent_cloud.py  (please verify that instructions work)

examples/confluent_cloud.py
edenhill: We only support 2.7, so specify that rather than 2.x

examples/confluent_cloud.py
mhowlett: awesome

examples/confluent_cloud.py
mhowlett: ahh. it's called that in my examples folder, which is where I copied this in from...

examples/confluent_cloud.py
mhowlett: given the instructions install the binary wheel now, and that is 0.11.4, yep sounds good.

examples/confluent_cloud.py
mhowlett: I don't know - I'm just doing what the confluent cloud web interface told me to... will test and remove if not.

examples/confluent_cloud.py
mhowlett: oh, i forgot about this. I did that originally, but it didn't work. Will open an issue.

examples/confluent_cloud.py
mhowlett: ahh, excuse me, getting muddled with memory models... it would be worth noting this in the improved documentation we keep talking about.

mhowlett:
thanks for that - lots of little improvements added up to a lot.

examples/confluent_cloud.py
mhowlett: haha, apparently I already did: #350

examples/confluent_cloud.py
```diff
@@ -0,0 +1,113 @@
...
+# to execute using Python 3.x:
+# $ python -m venv ccloud_example
+# $ source ccloud_example/bin/activate
+# $ pip install confluent_kafka
+# $ python confluent_cloud.py
+# $ deactivate
+
+import uuid
+from confluent_kafka import Producer, Consumer, KafkaError
+
```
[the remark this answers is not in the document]
edenhill: We do infact run flake8 in CI (which failed on the initial PR but for other reasons), but you're right that it doesn't complain about this:

```
(venv2) magnus@Magnus-Macbook-Pro:~/src/confluent-kafka-python (master)$ flake8 --version
3.5.0 (mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 2.7.10 on Darwin
(venv2) magnus@Magnus-Macbook-Pro:~/src/confluent-kafka-python (master)$ flake8 examples/confluent_cloud.py
(venv2) magnus@Magnus-Macbook-Pro:~/src/confluent-kafka-python (master)$ echo $?
0
```

PR #355, opened 2018-04-12
Readmefix1

edenhill:


README.md
```diff
@@ -40,11 +40,26 @@ Usage
...
+    """ Called once for each message produced to indicate delivery result.
+        Triggered by poll() """
+    if err is not None:
+        print('Message delivery failed: {}'.format(err))
+    else:
+        print('Message delivered to {} [{}]'.format(msg.topic(), msg.partition()))
 
 for data in some_data_source:
-    p.produce('mytopic', data.encode('utf-8'))
+    # Trigger delivery report callbacks from previous produce() calls
```
mhowlett: maybe 'trigger any available delivery report callbacks from previous produce calls'.

README.md
```diff
@@ -40,11 +40,26 @@ Usage
...
+    p.poll(0)
+
+    # Asynchronously produce a message, the optional but recommended
```
mhowlett: i'd remove 'but recommended'.

Also, I think you should 1. explicitly say under **exactly** what conditions poll is required and 2. explicitly say that if it's not called, messages will pile up indefinitely. My understanding is that errors are also exposed via the same queue, so poll should be called regardless of whether delivery reports are enabled. Is that right?

mhowlett:
Aside from clarifying the comments around poll, LGTM. Happy to re-review them.

Didn't spot anything wrong with the header dict support implementation, and you have tests.

README.md
mhowlett: yep, the pr's a step forward so i'm fine to merge. a point here is that by exposing poll, all of these details are also effectively exposed, whereas if the poll call is managed by the producer, they are effectively encapsulated inside the producer - another reason why not having a manual poll api (if it were possible) would reduce complexity considerably.

PR #362, opened 2018-04-27
add consume() to AvroConsumer

pxLi:
simply override consume() for to allow decode batch of avro messages.


pxLi:
[clabot:check]

syelda:
This would be very useful to have! Any update on when this might be added?

meridional:
Any update on this? Seems to be a simple diff to review. And it would be extremely useful.. (It's called AvroConsumer after all.)

rnpridgeon:
This will be covered with generic serde support, #502

PR #367, opened 2018-05-07
Refactoring example

hrchu:


edenhill:
Looks good but needs to pass flake8

edenhill:
Thank you!

Issue #342, opened 2018-03-26
AvroProducer doesn't serialize the key if it's an empty string


alfiya400:
Description
===========

Based on [the code for AvroProducer](https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/__init__.py#L68)
AvroProducer serializes the key only if `bool(key) == True`, which allows to pass an empty string as a key without the serialization.

This breaks the read using AvroConsumer, because it [tries to deserialize the key if `message.key() is not None`]( https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/__init__.py#L117)
And the serializer raises the following error [at the length check](https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/serializer/message_serializer.py#L208)
```
Traceback (most recent call last):
  File "/Users/alfiya/miniconda3/envs/py3/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 2862, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-20-7566bd90b3a5>", line 1, in <module>
    m=consumer.poll()
  File "/Users/alfiya/miniconda3/envs/py3/lib/python3.6/site-packages/confluent_kafka/avro/__init__.py", line 118, in poll
    decoded_key = self._serializer.decode_message(message.key())
  File "/Users/alfiya/miniconda3/envs/py3/lib/python3.6/site-packages/confluent_kafka/avro/serializer/message_serializer.py", line 209, in decode_message
    raise SerializerError("message is too small to decode")
```

How to reproduce
================

```python
import avro.schema
import confluent_kafka.avro

producer = confluent_kafka.avro.AvroProducer(
    {
        'api.version.request': True,
        'compression.codec': 'gzip',
        'default.topic.config': {'request.required.acks': 1},
        'bootstrap.servers': 'localhost:9093',
        'schema.registry.url': 'http://localhost:8081',
    }, 
    default_key_schema=avro.schema.Parse('{"type": "string"}'), 
    default_value_schema=avro.schema.Parse('{"type": "string"}')
)
producer.produce(topic='bug', value='not empty', key='')

consumer = confluent_kafka.avro.AvroConsumer({
    'group.id': 'mygroup', 
    'api.version.request': True,
    'default.topic.config': {'auto.offset.reset': 'smallest'},
    'bootstrap.servers': 'localhost:9093',
[...]

PR #374, opened 2018-05-15
Fix AvroProducer/AvroConsumer key/value identity check (#342)

rnpridgeon:


rnpridgeon:
Fixes #342, python 3.5 failure appears unrelated

tests/avro/test_avro_producer.py
```diff
@@ -101,3 +101,16 @@ def test_produce_with_custom_registry_and_registry_url(self):
...
+        producer = AvroProducer({}, schema_registry=schema_registry)
+        with self.assertRaises(ValueSerializerError):
+            producer.produce(topic='test', value='', key='not empty')
+
+    def test_produce_with_empty_key_identify(self):
+        schema_registry = MockSchemaRegistryClient()
+        producer = AvroProducer({}, schema_registry=schema_registry,
+                                default_value_schema=avro.loads('{"type": "int"}'))
+        with self.assertRaises(KeySerializerError):
+            producer.produce(topic='test', value=0, key='')
```
edenhill: We might want a positive test where empty strings and zero values are in fact serialized.

rnpridgeon:
@edenhill added a happy test, pre-patch this would have failed as the `Float` would not pass the type checking imposed by `producer.produce()`

tests/avro/test_avro_producer.py
```diff
@@ -101,3 +101,26 @@ def test_produce_with_custom_registry_and_registry_url(self):
...
         with self.assertRaises(ValueError):
             AvroProducer({'schema.registry.url': 'http://127.0.0.1:9001'}, schema_registry=schema_registry)
+
+    def test_produce_with_empty_value_identity(self):
+        schema_registry = MockSchemaRegistryClient()
+        producer = AvroProducer({}, schema_registry=schema_registry)
+        with self.assertRaises(ValueSerializerError):
+            producer.produce(topic='test', value='', key='not empty')
+
+    def test_produce_with_empty_key_identify(self):
```
edenhill: identify -> identity.

Maybe these should be renamed to be more clear why they're supposed to fail:
test_produce_with_empty_value_no_schema
test_produce_with_empty_key_no_schema
test_produce_with_empty_key_value_with_schema

rnpridgeon:
@edenhill, fixed the unit test names, altered the existing Avro integration test to use 'zero' values. This demons the ability to serialize and deserialize these values.

examples/integration_test.py
```diff
@@ -251,8 +251,10 @@ def verify_avro():
     prim_float = avro.load(os.path.join(avsc_dir, "primitive_float.avsc"))
     prim_string = avro.load(os.path.join(avsc_dir, "primitive_string.avsc"))
     basic = avro.load(os.path.join(avsc_dir, "basic_schema.avsc"))
-    str_value = 'abc'
```
edenhill: Don't delete or change existing test cases (unless they are wrong), just add new ones, otherwise we loose test coverage and regression testing.

rnpridgeon:
@edenhill replaced previous values added to 'combinations' to cover the Falsy cases

examples/integration_test.py
```diff
@@ -265,6 +266,8 @@ def verify_avro():
         dict(value=float_value, value_schema=prim_float, key=str_value, key_schema=prim_string),
         dict(value=str_value, value_schema=prim_string, key={'name': 'abc'}, key_schema=basic),
         dict(value=str_value, value_schema=prim_string, key=float_value, key_schema=prim_float),
+        dict(value='', value_schema=prim_string, key=0., key_schema=prim_float),
+        dict(value=0., value_schema=prim_float, key='', key_schema=prim_string),
```
edenhill: It is too bad the consumer doesn't seem to actually verify the consumed messages, but we'll leave that for later.

rnpridgeon: do you want to verify all keys and values or just the falsy ones

rnpridgeon: signed, sealed, delivered. I moved forward with only verifying the `Falsy` values on the consumer end. If this is not sufficient let me know and I'll act accordingly

examples/integration_test.py
```diff
@@ -313,11 +314,20 @@ def verify_avro():
                   (msg.topic(), msg.partition(), msg.offset(),
                    msg.key(), msg.value(), tstype, timestamp))
 
+            # verify Falsy keys and values are consumed
+            if not msg.key() is not None and not msg.key():
```
edenhill: The double negations makes this hard to read.

examples/integration_test.py
```diff
@@ -313,11 +314,20 @@ def verify_avro():
                   (msg.topic(), msg.partition(), msg.offset(),
                    msg.key(), msg.value(), tstype, timestamp))
 
+            # verify Falsy keys and values are consumed
```
edenhill: I think it would be better to verify all messages in a more generic way using the combinations list.
We know what we sent (the `combo`), so we know what to expect.

examples/integration_test.py
rnpridgeon: makes sense, in my mind if further illustrated the fact that you need to check for None not Falsiness

examples/integration_test.py
rnpridgeon: that's fair, I was originally trying avoid post-processing keys and values with empty fields. The latest commit does exactly that and asserts that each combo matches each populated message field

examples/integration_test.py
```diff
@@ -231,6 +231,7 @@ def __del__(self):
     assert DrOnlyTestSuccess_gced == 1
 
 
+# Verify identity check allows Falsy object values(e.g., 0, empty string) to be handled properly (issue #342)
```
edenhill: This comment should probably go in combinations just above the added lines (line 269), since otherwise one might think that this entire function only changes for Falsies.

PR #379, opened 2018-05-19
[Issues/372] Bad code in schema registry client

### LINKED ISSUE #372: Bad code in CachedSchemaRegistryClient
Description
===========
`compatibility` doesn't exist, so this could cause a exception, is that the purpose?
https://github.com/confluentinc/confluent-kafka-python/blob/71217575e65d5652508d80d0532314acd190fd85/confluent_kafka/avro/cached_schema_registry_client.py#L314

This doesn't `return`, seems to be a bug, so the function would actually return None.
https://github.com/confluentinc/confluent-kafka-python/blob/71217575e65d5652508d80d0532314acd190fd85/confluent_kafka/avro/cached_schema_registry_client.py#L271

What's the purpose of this exception catch? if you are going to raise it without logging or don't care?
https://github.com/confluentinc/confluent-kafka-python/blob/71217575e65d5652508d80d0532314acd190fd85/confluent_kafka/avro/cached_schema_registry_client.py#L207

How to reproduce
================
Just looking at code...

Checklist
=========

 - [ ] Operating system:
 - [ ] Critical issue

Enether:
Fixes #372 

- Return `False` instead of `None` when unable to check compatibility
- Fix unassigned variable problem when request inside `get_compatibility()` returns a non-successful status code.

I tried to implement tests for `get_compatibility()` but I could not figure out what would be a proper way to implement `MockSchemaRegistryClient#get_compatibility()`, as I do not grasp the way the schema registry works and would rather not implement some hardcoded mock value like
```
if subject is None: return { 'compatibility': 'NONE', 'compatibilityLevel': -1}
else ...
```

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -301,6 +301,7 @@ class CachedSchemaRegistryClient(object):
         Get the current compatibility level for a subject.  Result will be one of:
 
         @:param: subject: subject name
+        @:raises: ClientError: if the request was unsuccessful or an invalid compatibility level was returned
         @:return: 'NONE','FULL','FORWARD', or 'BACKWARD'
         """
         url = '/'.join([self.url, 'config'])
@@ -308,10 +309,16 @@ class CachedSchemaRegistryClient(object):
             url += '/' + subject
 
         result, code = self._send_request(url)
-        if code >= 200 and code <= 299:
-            compatibility = result.get('compatibility', None)
+        is_successful_request = code >= 200 and code <= 299
+        if not is_successful_request:
+            raise ClientError('Unable to fetch compatibility level. Error code: %d' % code)
 
-        if not compatibility:
-            compatibility = result.get('compatibilityLevel')
+        compatibility = result.get('compatibility', None)
+        if compatibility not in VALID_LEVELS:
+            if compatibility is None:
+                error_msg_suffix = 'No compatibility was returned'
+            else:
+                error_msg_suffix = str(compatibility)
+            raise ClientError('Invalid compatibility level received: %s' % error_msg_suffix)
 
         return compatibility
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -267,8 +267,8 @@ class CachedSchemaRegistryClient(object):
             elif code >= 200 and code <= 299:
                 return result.get('is_compatible')
             else:
-                log.error("Unable to check the compatibility")
-                False
+                log.error("Unable to check the compatibility: " + str(code))
+                return False
         except Exception as e:
             log.error("_send_request() failed: %s", e)
             return False


Enether:
[clabot:check]

Enether:
The only test that failed in the Travis CI build is:
`tests/test_threads.py::test_thread_safety FAILED                         [ 53%]`
I assume this has nothing to do with my changes and maybe requires a re-run or re-configuration?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -259,16 +259,16 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
...
             if code == 404:
-                log.error(("Subject or version not found:" + str(code)))
+                log.error("Subject or version not found:" + str(code))
                 return False
             elif code == 422:
-                log.error(("Invalid subject or schema:" + str(code)))
+                log.error("Invalid subject or schema:" + str(code))
                 return False
-            elif code >= 200 and code <= 299:
+            elif 200 <= code <= 299:
```
edenhill: Please avoid unrelated style changes.

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -259,16 +259,16 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
...
                 return result.get('is_compatible')
             else:
-                log.error("Unable to check the compatibility")
-                False
+                log.error("Unable to check the compatibility:" + str(code))
```
edenhill: Should be a space after ":"

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -308,10 +308,9 @@ def get_compatibility(self, subject=None):
             url += '/' + subject
 
         result, code = self._send_request(url)
-        if code >= 200 and code <= 299:
+        if 200 <= code <= 299:
             compatibility = result.get('compatibility', None)
-
-        if not compatibility:
+        else:
```
edenhill: Don't think this is correct: result.get() may return None, in which case we should fall back to compatibilityLevel here.

confluent_kafka/avro/cached_schema_registry_client.py
Enether: I also thought so but intentionally did not add it so it would be consistent with the two prints above

confluent_kafka/avro/cached_schema_registry_client.py
Enether: 👍

confluent_kafka/avro/cached_schema_registry_client.py
Enether: Yep, absolutely. Great catch, thanks

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -311,7 +312,7 @@ def get_compatibility(self, subject=None):
         if code >= 200 and code <= 299:
             compatibility = result.get('compatibility', None)
 
-        if not compatibility:
+        if compatibility is None:
```
Enether: This changes the behavior if `result.get('compatibility', None)` returns an empty string, but I assume that's not possible

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -311,7 +312,7 @@ def get_compatibility(self, subject=None):
         if code >= 200 and code <= 299:
             compatibility = result.get('compatibility', None)
```
edenhill: Looking at the REST API spec it seems that this value is never returned but `is_compatible`:
https://docs.confluent.io/current/schema-registry/docs/api.html#id1

There is also no mention of `compatibilityLevel` being returned.

So either the documentation is wrong or this function must never have worked.

Enether: The REST API was updated. It did return `compatibility` in version `3.0` - https://docs.confluent.io/3.0.1/schema-registry/docs/api.html#config
And yeah, `compatibilityLevel` is not mentioned anywehere, even in version `2.0`

Enether: Wow haha, nope. We're looking at the wrong API endpoint.
This calls GET /config
https://docs.confluent.io/current/schema-registry/docs/api.html#get--config
Which does return `compatibility`. Still no mention of `compatibilityLevel` though

Enether: The question then becomes - what should we return when the response is a non  2xx status code - 'NONE' ?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -301,17 +301,11 @@ def get_compatibility(self, subject=None):
         Get the current compatibility level for a subject.  Result will be one of:
 
         @:param: subject: subject name
-        @:return: 'NONE','FULL','FORWARD', or 'BACKWARD'
+        @:return: True if compatible, False if not compatible
```
edenhill: Let's follow what the Java code does:
https://github.com/confluentinc/schema-registry/blob/1443a7f8e251c93987099b26b0917bcebd4deca5/client/src/main/java/io/confluent/kafka/schemaregistry/client/CachedSchemaRegistryClient.java#L315

 /config returns a JSON  result containing "compatibility". getCompatibilityLevel() wraps the result and returns that same "compatibility" (and not "compatibilityLevel"), unfortunate naming perhaps.

And I propose raising an exception if code != 2xx or if the returned value is not one of the four strings.

confluent_kafka/avro/cached_schema_registry_client.py
Enether: This method calls '/config':
https://github.com/confluentinc/confluent-kafka-python/blob/66a181459731515dc0bb86a9de69dcc51278797c/confluent_kafka/avro/cached_schema_registry_client.py#L300
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/cached_schema_registry_client.py#L306

edenhill:
Thank you for this! :)

PR #382, opened 2018-05-21
Topic Admin API, list_topics(), Python 3.7, fixes

edenhill:
For an example what the admin api looks like in practice, see examples/adminapi.py


.travis.yml
```diff
@@ -1,31 +1,31 @@
...
-  # Source package verification with Python 2.7 and librdkafka v0.9.5
+    env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=adminapi
+  # Source package verification with Python 2.7 and librdkafka adminapi
   - os: linux
     language: python
     dist: trusty
     python: "2.7"
-    env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=v0.9.5
-  # Source package verification with Python 3.5 and librdkafka master
+    env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=adminapi
```
mhowlett: assume you didn't mean to check these changes in?

confluent_kafka/__init__.py
```diff
@@ -13,7 +16,621 @@
...
+                if error is not None:
+                    # Topic-level exception
+                    fut.set_exception(KafkaException(error))
+                else:
+                    # Topic-level success
+                    fut.set_result(None)
+        except Exception as e:
+            # Request-level exception, raise the same for all topics
+            for topic, fut in futmap.items():
+                fut.set_exception(e)
```
mhowlett: 👍if topic-level results could potentially complete at different times, i think what you're doing here is definitely the right (only?) way to report request level errors and since your api is written for that possibility, this seems good for me.

something i'm wondering about elsewhere is whether or not this is the best approach for an API that only has a request level future (and thinking yes, even in that scenario this is an optimal compromise).

examples/adminapi.py
```diff
@@ -0,0 +1,210 @@
...
+
+#
+# Example Admin clients.
+#
+
+from confluent_kafka import AdminClient, NewTopic, NewPartitions, ConfigResource, ConfigEntry, KafkaException
+import sys
+
+
+def example_create_topics(a, topics):
```
mhowlett: the examples are all synchronous - it would be good to see how this works where you're not blocking on the results to the admin call (for my own Python education as much as anything).

examples/adminapi.py
```diff
@@ -0,0 +1,210 @@
...
+def example_describe_configs(a, args):
+    """ describe configs """
+
+    resources = [ConfigResource(restype, resname) for
+                 restype, resname in zip(args[0::2], args[1::2])]
+
+    fs = a.describe_configs(resources)
+
+    # Wait for operation to finish.
+    for res, f in fs.items():
```
mhowlett: [vaguely related rambling] this for loop is interesting because it's how you'd typically want to handle these results - iterate through them. in Java (and C# traditionally) it's easier/nicer to do this if the return type is a list rather than a map, and for that reason I think a list would be better here in those languages even though a map models the data structure more correctly. but anyway, that doesn't apply in python: a map is perfect.

confluent_kafka/__init__.py
```diff
@@ -13,7 +16,621 @@
...
+        self.synonyms = synonyms
+
+    def __repr__(self):
+        return "ConfigEntry(%s=\"%s\")" % (self.name, self.value)
+
+    def __str__(self):
+        return "%s=\"%s\"" % (self.name, self.value)
+
+    @classmethod
+    def config_source_to_str(cls, source):
```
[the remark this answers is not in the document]
edenhill: Not available in Py2.7

confluent_kafka/src/Admin.c
```diff
@@ -0,0 +1,1621 @@
...
+ */
+#define Admin_options_def_int   (-12345)
+#define Admin_options_def_float ((float)Admin_options_def_int)
+
+struct Admin_options {
+        int   validate_only;      /* needs special bool parsing */
+        float request_timeout;    /* parser: f */
+        float operation_timeout;  /* parser: f */
+        int   broker;             /* parser: i */
+        int   incremental;        /* needs special bool parsing */
```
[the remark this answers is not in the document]
edenhill: This code doesn't know if incremental is supported or not but relies on librdkafka to reject the option if not.

confluent_kafka/src/Admin.c
```diff
@@ -0,0 +1,1621 @@
...
+Admin_options_to_c (Handle *self, const char *forApi,
+                    const struct Admin_options *options,
+                    PyObject *future) {
+        rd_kafka_AdminOptions_t *c_options;
+        rd_kafka_resp_err_t err;
+        char errstr[512];
+
+        c_options = rd_kafka_AdminOptions_new(self->rk, forApi);
+        if (!c_options) {
+                PyErr_Format(PyExc_RuntimeError,
```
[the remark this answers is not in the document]
edenhill: There is no user-supplied parameter here so ValueError wouldn't make sense.
This exception is only raised if the python client version is newer than librdkafka, which shouldnt happen.

confluent_kafka/src/Admin.c
```diff
@@ -0,0 +1,1621 @@
...
+        CallState_end(self, &cs);
+
+        rd_kafka_NewTopic_destroy_array(c_objs, tcnt);
+        rd_kafka_AdminOptions_destroy(c_options);
+        rd_kafka_queue_destroy(rkqu); /* drop reference from get_background */
+
+        /* Increase refcount for the return value since
+         * it is currently a borrowed reference. */
+        Py_INCREF(future);
+        return future;
```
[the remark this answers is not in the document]
edenhill: This was from an earlier incarnation where the future was set up in C.
Changed to return None

examples/adminapi.py
edenhill: Made a contrived delta_alter_configs example

confluent_kafka/__init__.py
```diff
@@ -13,7 +16,621 @@
...
+                    CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG,
+                    CONFIG_SOURCE_STATIC_BROKER_CONFIG,
+                    CONFIG_SOURCE_DEFAULT_CONFIG,
+                    RESOURCE_UNKNOWN,
+                    RESOURCE_ANY,
+                    RESOURCE_TOPIC,
+                    RESOURCE_GROUP,
+                    RESOURCE_BROKER)
 
+import concurrent.futures
```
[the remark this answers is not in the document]
rnpridgeon: I'm also okay with this, cramming it all into __init__.py just seems a bit messy

confluent_kafka/__init__.py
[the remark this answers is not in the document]
mhowlett: the 3.4 implementation has been backported and i'm lead to believe commonly used in 2.7 libraries. can this not be leveraged?

https://pypi.org/project/enum34/

confluent_kafka/__init__.py
```diff
@@ -5,6 +5,9 @@
                     Message,
                     Producer,
                     TopicPartition,
+                    AdminClientImpl,
```
mhowlett: given you don't really want `AdminClientImpl` exposed, I think you probably need to provide a different definition of  `__all__`.

confluent_kafka/__init__.py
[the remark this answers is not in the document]
mhowlett: a sub package seems better here to me, and also consistent with other bindings.

confluent_kafka/__init__.py
```diff
@@ -13,7 +16,633 @@
...
+      * RESOURCE_GROUP - group name
+    """
+    _res_name_by_type = {RESOURCE_UNKNOWN: "unknown",
+                         RESOURCE_ANY: "any",
+                         RESOURCE_TOPIC: "topic",
+                         RESOURCE_GROUP: "group",
+                         RESOURCE_BROKER: "broker"}
+    _res_type_by_name = {v: k for k, v in _res_name_by_type.items()}
+
+    def __init__(self, restype, name,
```
mhowlett: an enum seems better to me for `restype` too, based on my research just now but no actual prior experience guiding me in what's normal. given it's an input rather than output thing though, not as important as `CONFIG_SOURCE` (but they should be consistent).

confluent_kafka/__init__.py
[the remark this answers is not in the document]
edenhill: so `confluent_kafka.admin.`?

confluent_kafka/__init__.py
edenhill: I'm submoduling AdminClient into .admin.

I think AdminClientImpl still needs to implicitly exported since AdminClient is a subclass

confluent_kafka/__init__.py
edenhill: doin' it

confluent_kafka/__init__.py
mhowlett: there's the mistake there - composition is much more appropriate here I think.

confluent_kafka/__init__.py
```diff
@@ -13,7 +16,633 @@
...
+
+        :raises KafkaException: Operation failed locally or on broker.
+        :raises TypeException: Invalid input.
+        :raises ValueException: Invalid input.
+        """
+
+        f, futmap = AdminClient._make_futures(topics, None,
+                                              AdminClient._make_topics_result)
+
+        super(AdminClient, self).delete_topics(topics, f, **kwargs)
```
mhowlett: overriding methods like this seems confusing, better to use composition.

confluent_kafka/__init__.py
edenhill: Not sure what composition buys us in this case.
The only reason there is a Python AdminClient is because it makes it easier to wrap the single-future C AdminClientImpl API into a multiple feature Python API.

confluent_kafka/__init__.py
edenhill: I'm not an OO guy, as you are very well aware of. So please educate me what the practical difference is between composition and inheritance in this case?

confluent_kafka/admin/__init__.py
```diff
@@ -0,0 +1,622 @@
+"""
+Kafka Admin client: create, view, alter, delete topics and resources.
+"""
+from ..cimpl import (KafkaException, # noqa
+                     AdminClientImpl,
```
mhowlett: is it better to call this `_AdminClientImpl`?

confluent_kafka/__init__.py
mhowlett: I'm confused about `__all__` above. Instead I think i meant to suggest calling `AdminClientImpl` it `_AdminClientImpl`.

confluent_kafka/admin/__init__.py
```diff
@@ -0,0 +1,622 @@
...
+        # per topic,future in futmap.
+        f = concurrent.futures.Future()
+        f.add_done_callback(lambda f: make_result_fn(f, futmap))
+
+        if not f.set_running_or_notify_cancel():
+            raise RuntimeError("Future was cancelled prematurely")
+
+        return f, futmap
+
+    def create_topics(self, new_topics, **kwargs):
```
mhowlett: The main thing I was noting about this before is that the signatures of the methods `create_topics` etc. change in the subclass. The new methods conveniently hide the old ones because of `**kwargs`, but it seems like kind of a trick (and my python's not strong enough to know if there are any downsides) and i'm a bit averse to tricks. Generally, the main disadvantage of  inheritance is that coupling is stronger - everything exposed in the base classes flow through to subclasses. Except for the hiding noted above, this would apply here. The other thing that threw me was use of the suffix `Impl` in the super class because it implies to me more of a complete implementation, and I think a bit unusual in this context.

That said, I agree an inheritance relationship would be ok here. My complaint about signature differences + exposing of adminclientimpl details would be mitigated by use of Pythons poor excuse for protected methods - maybe change `AdminClientImpl`'s `create_topics` etc. to `_create_topics_impl` etc. Though, the public interface is worse in that they are visible outside the class (though by convention ignored). Also maybe change the `Impl` suffix to `Base`. I don't particularly like that either, but it's more conventional.

Alternatively, I still think composition is better here. I'd be somewhat better externally. There'd just be one externally visible property `_impl` of type `_AdminClientImpl`. And I like the name `_AdminClientImpl` in this context.

PR #388, opened 2018-05-29
Initial POC for generic serde support

rnpridgeon:
Rough draft for Generic SERDE support. In retrospect its probably a good idea to allow simply passing a dict to the serializer as serializer_opts from the application. This adds a lot more flexibility and could facilitate per produce request overrides such as the schemas passed to the AvroProducer.produce() function. 

i.e. `serialize(**kwargs)` 

Before going to deep into this I wanted get some initial comments on the implementation though.

### DIFF
--- a/confluent_kafka/consumer.py
+++ b/confluent_kafka/consumer.py
@@ -1,0 +1,35 @@
+from .cimpl import ConsumerImpl
+from confluent_kafka.serializer import Deserializer
+
+
+class Consumer(ConsumerImpl):
+    def __init__(self, *args, **kwargs):
+        self.conf = {
+            'key_deserializer': Deserializer.verify(kwargs.pop("key_deserializer", None), True),
+            'value_deserializer': Deserializer.verify(kwargs.pop("value_deserializer", None)),
+        }
+        super(Consumer, self).__init__(*args, **kwargs)
+
+    @staticmethod
+    def _decode(s, topic, payload):
+        if s is None or payload is None:
+            return payload
+        return s.deserialize(topic, payload)
+
+    def poll(self, timeout=-1):
+        msg = super(Consumer, self).poll(timeout)
+
+        # Bypass the remaining logic if there are no serializers configured
+        if msg is not None:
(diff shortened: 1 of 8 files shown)


confluent_kafka/consumer.py
```diff
@@ -0,0 +1,35 @@
...
+        if s is None or payload is None:
+            return payload
+        return s.deserialize(topic, payload)
+
+    def poll(self, timeout=-1):
+        msg = super(Consumer, self).poll(timeout)
+
+        # Bypass the remaining logic if there are no serializers configured
+        if msg is not None:
+            if self.conf['key_deserializer'] is None and self.conf['value_deserializer'] is None:
```
edenhill: Two dict lookups per message is kind of costly

examples/integration_test.py
```diff
@@ -890,6 +891,69 @@ def stats_cb(stats_json_str):
...
+    conf = {'bootstrap.servers': bootstrap_servers,
+            'api.version.request': api_version_request,
+            'max.in.flight.requests.per.connection': 1,
+            'error_cb': error_cb,
+            'acks': -1
+            }
+
+    t = "verify_serializer_{}".format(uuid.uuid4())
+
+    p = confluent_kafka.Producer(conf, key_serializer=SerializerTest())
```
edenhill: It seems like this serializer does not modify the data, does not actually serialize, so even if it fails to work the tests will pass.
Use something ubiquitous like  base64

examples/integration_test.py
```diff
@@ -890,6 +891,69 @@ def stats_cb(stats_json_str):
...
+    while True:
+        msg = c.poll()
+
+        if msg is None:
+            continue
+        if msg.error():
+            if msg.error().code() == confluent_kafka.KafkaError._PARTITION_EOF:
+                break
+            else:
+                print(msg.error())
```
edenhill: This should probably raise instead

rnpridgeon:
Moving into C

Issue #211, opened 2017-06-26
Support for non-avro message key in AvroConsumer and AvroProducer


iharkaroza:
Hi! At the moment we have a number of clients who produce/consume messages with string keys and payload encoded with avro schema. Some of them choose to use python for message consumption and unable to do so for messages with string keys, because AvroConsumer expects all the keys to be encoded with avro schema. 
Just want to know if there is a reason behind this avro message key approach? It looks restricting to me and I think it would be nice to have support for string message keys in AvroConsumer and AvroProducer.

edenhill:
That sounds reasonable, there is no overwhelming reason to force an Avro schema on the key.

michelazzo:
Is there any work being developed on that? That's a feature I need badly.

sebastianneubauer:
Am I right in my understanding that this issue prohibits the use of ksql together with the python AvroProducer as ksql cannot work with avro format keys? If yes, are there any workarounds?

rmoff:
@sebastianneubauer the workaround is to ensure that the key is also available in the value fields, and use KSQL's `PARTITION BY` functionality to repartition your streams with `STRING` keys (which KSQL does support).

sebastianneubauer:
The issue is, that the avro producer can only generate streams with avro keys. So for the workaround one would create a dummy avro schema for the key in the produced stream and complete ignore it and create another stream using partition by, right?

rmoff:
>  So for the workaround one would create a dummy avro schema for the key in the produced stream and complete ignore it and create another stream using partition by, right?

Yes.

cricket007:
Link https://github.com/confluentinc/confluent-kafka-python/issues/428#issuecomment-444320103

PR #390, opened 2018-05-30
Allow non avro schema key in avro producer

sebastianneubauer:
As brought up in #211, as KSQL can only handle topics with plain string keys but avro schema values, the avro producer should allow keys with plain strings. 
This PR drops the explicit check if the key has an avro schema and only throws an KeySerializerError if the key is not a string type.

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -2,6 +2,7 @@
     Avro schema registry module: Deals with encoding and decoding of messages with avro schemas
 
 """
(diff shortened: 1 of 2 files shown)


sebastianneubauer:
I would like to believe that the failing test is unrelated to my change...seems like a race condition or something?

rnpridgeon:
Hello @sebastianneubauer, thanks for the PR. We are actually taking the project another direction however as we feel it is in our best interest

Issue #90, opened 2016-12-01
Schema Registry client and SSL


dodysw:
Would it be possible to enhance the new AvroConsumer/AvroProducer/CachedSchemaRegistryClient to optionally accept certs related parameters so we can connect to Schema Registry on SSL channel that requires client certificate authentication. Or possibly accept a requests.Session() object (I think we can put a persistent certificate parameter to session object)

ewencp:
@roopahc @criccomini Any thoughts on this? I'm a bit skeptical of exposing implementation details like the use of `requests` in the public interface. Is it possible to achieve something similar just by forwarding `*args` and `**kwargs` maybe?

criccomini:
Agree on not wanting to expose `requests`.

edenhill:
@dodysw Can you help us identify exactly what parameters are needed? Is certs= and verify= enough?

I would prefer to see this added as config dict properties rather than constructor args.

dodysw:
For python requests, those are pretty much what needed. It looks like this:
```
ssl_cafile = '/opt/logs/sandbox/cert/kafka/ca-cert.pem'
ssl_certfile = '/opt/logs/sandbox/cert/kafka/client-cert.pem'
ssl_keyfile = '/opt/logs/sandbox/cert/kafka/client-key-plain.pem'
s = requests.Session()
s.cert = (ssl_certfile, ssl_keyfile)
s.verify = ssl_cafile
```
Client private key pem file must be unencrypted so it doesn't take a password.

edenhill:
Okay, cool. What about the following new config properties:
 * `schema.registry.ssl.ca.location` = str
 * `schema.registry.ssl.certificate.location` = str | tuple
 * `schema.registry.ssl.key.location` = str

This is in line with existing SSL config properties for Kafka.
(Search for ssl.ca.location here: https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)

dodysw:
Close, I think `schema.registry.ssl.certificate.location` shouldn't be a tuple if `schema.registry.ssl.key.location` is there.

roopahc:
@ewencp  Should be possible since we are hitting rest end points after all.   We can take them as argument to AvroProducer, AvroConsumer classes

oh54:
Commenting to say that I would very much be interested in being able to turn off server certificate verification when using Schema Registry over HTTPS.

rnpridgeon:
[...]

PR #394, opened 2018-06-01
Add support for self-signed certificates and TLS mutual auth (#90)

rnpridgeon:


confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
-        elif method == 'POST':
-            response = requests.post(url, body, headers=_headers)
-        elif method == 'PUT':
-            response = requests.put(url, body, headers=_headers)
-        elif method == 'DELETE':
-            response = requests.delete(url, headers=_headers)
-        else:
+        self._session = requests.Session()
+        with self._session as s:
+            if ca_path:
```
edenhill: 'is not None'

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                s.verify = ca_path
+            if cert_location or key_location:
```
edenhill: dito

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                if not key_location and cert_location:
```
edenhill: ambigious, use parens
also, `is not None`

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                    raise ValueError("Both cert_location and key_location must be set: {} {}".format(
```
edenhill: They're exposed as schema.registry.ssl.certificate.location, etc., to the user, not cert_location

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                        cert_location,
+                        key_location))
+                s.cert = (cert_location, key_location)
+
+    def _send_request(self, url, method='GET', body=None, headers={}):
+        if method not in VALID_METHODS:
             raise ClientError("Invalid HTTP request type")
```
edenhill: Might as well print the method here

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
-            response = requests.get(url, headers=_headers)
-        elif method == 'POST':
-            response = requests.post(url, body, headers=_headers)
-        elif method == 'PUT':
-            response = requests.put(url, body, headers=_headers)
-        elif method == 'DELETE':
-            response = requests.delete(url, headers=_headers)
-        else:
+        self._session = requests.Session()
+        with self._session as s:
```
edenhill: From what I understand this closes the session when the context block closes, which is probably not what we want

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,29 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                        key_location))
+                s.cert = (cert_location, key_location)
+
+    def _send_request(self, url, method='GET', body=None, headers={}):
+        if method not in VALID_METHODS:
             raise ClientError("Invalid HTTP request type")
 
-        result = json.loads(response.text)
-        return (result, response.status_code)
+        headers['Accept'] = ACCEPT_HDR
```
edenhill: This modifies the passed headers, use a local _headers copy as the original code.
Suggest adding a test-case that covers this scenario.

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
         c.close()
 
 
+def verify_avro_https(*args):
```
edenhill: Please use named args.

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+    from confluent_kafka import avro
+    avsc_dir = os.path.join(os.path.dirname(__file__), os.pardir, 'tests', 'avro')
+
+    # Avoid potential for never ending consume/produce cycle
```
edenhill: Why would this happen?

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+    timeout = time.time() + 15
+
+    # Producer config
+    conf = {'bootstrap.servers': bootstrap_servers,
+            'error_cb': error_cb,
+            'api.version.request': api_version_request,
+            'default.topic.config': {'produce.offset.report': True}}
```
edenhill: this is true by default in the python client

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+
+    # Create producer
+    if schema_registry_url:
+        conf['schema.registry.url'] = schema_registry_url
+        conf['schema.registry.ssl.ca.path'] = args[0]
+        conf['schema.registry.ssl.certificate.location'] = args[1]
+        conf['schema.registry.key.location'] = args[2]
+    else:
+        raise ValueError("Schema Registry URL required for the avro-https test")
```
edenhill: Reverse the logic to avoid nesting, and use `is None` checks:

`if schema_registry_url is None: \n raise..`

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+
+    p = avro.AvroProducer(conf)
+
+    prim_float = avro.load(os.path.join(avsc_dir, "primitive_float.avsc"))
+    prim_string = avro.load(os.path.join(avsc_dir, "primitive_string.avsc"))
+    basic = avro.load(os.path.join(avsc_dir, "basic_schema.avsc"))
+    str_value = 'abc'
+    float_value = 32.
```
edenhill: Still not fancying this notation, it is ambigious. Did he mean int 32 and happened to add a period, or is he saving a bit of typing by stripping out the 0?
Use `32.0`

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+        dict(value=str_value, value_schema=prim_string, key={'name': 'abc'}, key_schema=basic),
+        dict(value=str_value, value_schema=prim_string, key=float_value, key_schema=prim_float),
+        # Verify identity check allows Falsy object values(e.g., 0, empty string) to be handled properly (issue #342)
+        dict(value='', value_schema=prim_string, key=0., key_schema=prim_float),
+        dict(value=0., value_schema=prim_float, key='', key_schema=prim_string),
+    ]
+
+    # Consumer config
+    cons_conf = {'bootstrap.servers': bootstrap_servers,
+                 'group.id': 'test.py',
```
edenhill: Suggest using a unique group here to avoid reading already committed offsets

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+                 'on_commit': print_commit_result,
+                 'error_cb': error_cb,
+                 'default.topic.config': {
+                     'auto.offset.reset': 'earliest'
+                 }}
+
+    for i, combo in enumerate(combinations):
+        combo['topic'] = str(uuid.uuid4())
+        combo['headers'] = [('index', str(i))]
+        p.produce(**combo)
```
edenhill: nice

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+    p.flush()
+
+    conf = copy(cons_conf)
+    conf['schema.registry.url'] = schema_registry_url
```
edenhill: You could set up this dict once as `ssl_conf` and then do `conf.update(ssl_conf)`

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+
+        tstype, timestamp = msg.timestamp()
+        print('%s[%d]@%d: key=%s, value=%s, tstype=%d, timestamp=%s' %
+              (msg.topic(), msg.partition(), msg.offset(),
+               msg.key(), msg.value(), tstype, timestamp))
+
+        # omit empty Avro fields from payload for comparison
+        record_key = msg.key()
+        record_value = msg.value()
+        index = int(msg.headers()[0][1])
```
edenhill: Making assumptions on the header layout is fragile. Convert it to a dict and get the key you need instead.

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+        if isinstance(msg.value(), dict):
+            record_value = {k: v for k, v in msg.value().items() if v is not None}
+
+        assert combinations[index].get('key') == record_key
+        assert combinations[index].get('value') == record_value
+
+        msgcount += 1
+
+        if time.time() > timeout:
+            raise RuntimeError("verify_avro_https cailed to complete within 15 seconds")
```
edenhill: failed.
Use `timeout` instead of hardcoded `15`, also suggest having a timeout of 30s to make this more robust.

examples/integration_test.py
```diff
@@ -331,6 +331,115 @@ def verify_avro():
...
+
+    # Close consumer
+    c.commit(asynchronous=False)
```
edenhill: Why are you committing here?

examples/integration_test.py
```diff
@@ -967,4 +1077,11 @@ def print_usage(exitcode, reason=None):
         print('=' * 30, 'Verifying AVRO', '=' * 30)
         verify_avro()
 
+    if 'avro-https' in modes:
```
edenhill: Positional arguments won't work well with multiple modes supported at the same time.
I suggest adding a new `--conf <json>` option that allows you to specify arbitrary configuration settings, in this case the schema.registry SSL settings, and then having an example in the README.

tests/README.md
```diff
@@ -6,3 +6,17 @@ From top-level directory run:
     $ tox
 
 **NOTE**: This requires `tox` ( please install with `pip install tox` ) and several supported versions of Python.
+
+
+HTTPS tests
+===========
+
+Https tests require access to a Schema registry configured to with at least one https listener.
```
edenhill: HTTPS

tests/README.md
```diff
@@ -6,3 +6,17 @@ From top-level directory run:
...
+
+
+HTTPS tests
+===========
+
+Https tests require access to a Schema registry configured to with at least one https listener. 
+
+For instructions on how to configure the Schema Registry please see Confleunt's documentation: 
+
+[Schema Registry documentation](https://docs.confluent.io/current/schema-registry/docs/security.html#configuring-the-rest-api-for-http-or-https)
```
edenhill: Huhm, that anchor ref looks like a doc bug?

tests/README.md
```diff
@@ -6,3 +6,17 @@ From top-level directory run:
...
+
+If client authentication has been enabled you will need to provide both the client certificate, `cert_path`, and the client's private key, `key_path`
+
+        examples/integration_test.py <bootstrap-servers> confluent-kafka-testing <schema-registry-url> [ca_path] [[cert_path] [key_path]]
```
edenhill: Isn't the --avro-https mode required?

rnpridgeon:
The failure in build #682.5 looks to be the result of 

https://github.com/confluentinc/confluent-kafka-python/commit/6d928d2347aa3c455d4eae46e96fcb691f3abef6

Line 14 I believe should also point to master but I could be mistaken

.travis.yml
```diff
@@ -6,12 +6,12 @@ matrix:
     dist: trusty
     python: "2.7"
     env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=master
-  # Source package verification with Python 2.7 and librdkafka v0.11.4
+  # Source package verification with Python 2.7 and librdkafka masster
```
ah-: typo?

rnpridgeon: thanks!

.travis.yml
```diff
@@ -6,12 +6,12 @@ matrix:
     dist: trusty
     python: "2.7"
     env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=master
-  # Source package verification with Python 2.7 and librdkafka v0.11.4
+  # Source package verification with Python 2.7 and librdkafka master
```
edenhill: If you rebase on master you can remove this diff since this section is no longer included.

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,30 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
 
-    def _send_request(self, url, method='GET', body=None, headers=None):
-        if body:
-            body = json.dumps(body)
-            body = body.encode('utf8')
-        _headers = dict()
-        _headers["Accept"] = ACCEPT_HDR
+        self._session = requests.Session()
+        with self._session as s:
+            if ca_location:
```
edenhill: check for None, not true

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,30 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                s.verify = ca_location
+            if (cert_location or key_location) is not None:
```
edenhill: This is incorrect:
 * it first checks cert_location or key_location for truth in a () group
 * that group is returning a bool, True or False.
 * .. which you then check for `not None`

what you want to do is:
`if cert_location is not None or key_location is not None:`

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,30 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                if (key_location and cert_location) is None:
```
edenhill: dito

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,30 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+                    raise ValueError(
+                        "Both schema.registry.ssl.certificate.location and schema.registry.ssl.key.location"
+                        "must be set: {} {}".format(cert_location, key_location))
```
edenhill: I don't think there is much value in printing the values

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,30 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
-        elif method == 'PUT':
-            response = requests.put(url, body, headers=_headers)
-        elif method == 'DELETE':
-            response = requests.delete(url, headers=_headers)
-        else:
-            raise ClientError("Invalid HTTP request type")
-
-        result = json.loads(response.text)
-        return (result, response.status_code)
+        with self._session as s:
```
edenhill: are you sure the session can be reused after using it in a context like this?

The [docs](http://docs.python-requests.org/en/master/user/advanced/) suggest otherwise:
> This will make sure the session is closed as soon as the with block is exited, even if unhandled exceptions occurred.

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -267,8 +266,8 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
             elif code >= 200 and code <= 299:
                 return result.get('is_compatible')
             else:
-                log.error("Unable to check the compatibility")
-                False
+                log.error("Unable to check the compatibility: " + str(code))
```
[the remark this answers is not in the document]
edenhill: is good

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -301,17 +300,24 @@ def get_compatibility(self, subject=None):
         Get the current compatibility level for a subject.  Result will be one of:
 
         @:param: subject: subject name
+        @:raises: ClientError: if the request was unsuccessful or an invalid compatibility level was returned
```
edenhill: these doc strings are fixed in adminapi, make sure the merge picks it up correctly.

examples/integration_test.py
```diff
@@ -45,14 +45,15 @@
 except ImportError as e:
     with_progress = False
 
+testconf = None
```
edenhill: add newline below

examples/integration_test.py
```diff
@@ -374,6 +375,97 @@ def verify_avro():
...
+    from confluent_kafka import avro
+    avsc_dir = os.path.join(os.path.dirname(__file__), os.pardir, 'tests', 'avro')
+
+    # Producer config
+    conf = {'bootstrap.servers': bootstrap_servers,
+            'error_cb': error_cb,
+            'api.version.request': api_version_request,
+            }
+
+    conf.update(testconf['schema_registry_https'])
```
edenhill: `testconf.get('schem..', {})` is even more robust

examples/integration_test.py
```diff
@@ -374,6 +375,97 @@ def verify_avro():
...
 
 
+def verify_avro_https():
+    from confluent_kafka import avro
+    avsc_dir = os.path.join(os.path.dirname(__file__), os.pardir, 'tests', 'avro')
+
+    # Producer config
+    conf = {'bootstrap.servers': bootstrap_servers,
+            'error_cb': error_cb,
+            'api.version.request': api_version_request,
```
edenhill: trailing comma is atypical put close-brace there instead.

examples/integration_test.py
```diff
@@ -1140,6 +1232,20 @@ def print_usage(exitcode, reason=None):
     sys.exit(exitcode)
 
 
+def _supply_group_id():
```
edenhill: generate

examples/integration_test.py
```diff
@@ -1140,6 +1232,20 @@ def print_usage(exitcode, reason=None):
...
+    return uuid.uuid1()
```
edenhill: `str(uuid.uuid1())`

examples/integration_test.py
```diff
@@ -1140,6 +1232,20 @@ def print_usage(exitcode, reason=None):
...
+
+
+# Resolve environment variables
+def _normalize(_conf):
```
edenhill: normalize is a bit generic, maybe _resolve_envs() instead?

edenhill: Use a docstring instead of pre-#-comment

examples/integration_test.py
```diff
@@ -1152,18 +1258,20 @@ def print_usage(exitcode, reason=None):
     # Parse options
     while len(sys.argv) > 1 and sys.argv[1].startswith('--'):
         opt = sys.argv.pop(1)[2:]
+
+        if opt == 'conf':
```
edenhill: this is fine, but it should default to "tests/testconf.json" if no config file was specified.

examples/integration_test.py
```diff
@@ -1217,6 +1325,10 @@ def print_usage(exitcode, reason=None):
         print('=' * 30, 'Verifying Admin API', '=' * 30)
         verify_admin()
 
+    if 'avro-https' in modes:
+        print('=' * 30, 'Verifying AVRO with https', '=' * 30)
```
edenhill: HTTPS

tests/README.md
```diff
@@ -10,30 +10,35 @@ From top-level directory run:
...
 =================
 
-**WARNING**: These tests require an active Kafka cluster and will create new topics.
+**NOTE**: Integration tests require an existing Kafka cluster and a `testconf.json` configuration file.
+At a minimum you must specify `bootstrap.servers` and `topic` within `testconf.json`. Please reference [tests/testconf-example.json](tests/testconf-example.json) for formatting.
 
+**WARNING**: These tests will create new topics and consumer groups.
 
-To run all of the integration test `modes` uncomment the following line from `tox.ini` and add the addresses to your Kafka and Confluent Schema Registry instances.
+Integration test
```
edenhill: Is this intended as a heading?

tests/testconf-example.json
```diff
@@ -0,0 +1,11 @@
+{
+    "topic": "<test-topic> or $TOPIC env",
```
edenhill: it isn't clear here that any environment variable will work, maybe add that as a comment above ("__doc": "Any value starting with $ will be treated like a environment variable and its value will be automatically retrieved"

tests/testconf-example.json
```diff
@@ -0,0 +1,11 @@
...
+    "bootstrap.servers": "<bootstrap-broker> or $BOOTSTRAP_SERVER env",
+    "schema.registry.url": "<schema-registry-url> or $SCHEMA_REGISTRY_URL env",
+    "schema_registry_https": {
+            "schema.registry.url": "<schema-regsitry-url> or $SCHEMA_REGISTRY_SSL_URL env",
```
edenhill: registry

tests/testconf-example.json
```diff
@@ -0,0 +1,11 @@
+{
+    "topic": "<test-topic> or $TOPIC env",
+    "bootstrap.servers": "<bootstrap-broker> or $BOOTSTRAP_SERVER env",
```
edenhill: bootstrap-servers

tox.ini
```diff
@@ -1,17 +1,22 @@
...
-envlist = flake8,py27,py34,py35
+envlist = flake8,py27,py36
 
 [testenv]
 setenv =
     CPPFLAGS=-I{toxinidir}/tmp-build/include
     LDFLAGS=-L{toxinidir}/tmp-build/lib
     C_INCLUDE_PATH={toxinidir}/tmp-build/include
     LD_LIBRARY_PATH={toxinidir}/tmp-build/lib
+    PYTHONIOENCODING = UTF-8
```
edenhill: why this?

rnpridgeon: To ensure the default file encoding is utf-8. This caused issues on my machine after you added non-ascii characters to the integration tests.

edenhill: Even after I added this?
https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/integration_test.py#L2

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: Fairly certain this is correct although it's admittedly not clear. 

The `or` operator, as I understand it, will return the first true value(evaluated left to right) in that expression as opposed to a boolean. As long as one of those two are set the the overall expression evaluates to true which is what we want. 

This is not abundantly clear though as you have rightly demonstrated so I agree it should probably be rewritten.

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: As with the `or` expression `and` does not evaluate to true or false here. Instead it will return the first argument to evaluated to false. If no false values can be found the last argument to evaluate true will be returned instead.

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: Perhaps but it helps when evaluating failures through a third-party. Would you be more okay with returning just the property that is missing?

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: It closes the connection pool and clears up some state but the session instance is still in play. 

https://github.com/requests/requests/blob/master/requests/sessions.py#L710-L713

The adapter instance itself remains intact although the existing pool manager is cleared as is the adapters current state. 

https://github.com/requests/requests/blob/master/requests/adapters.py#L318-L326

So far as I can tell urllib3 will take care of creating a new pool for us the next time a request is made. 

https://github.com/requests/requests/blob/master/requests/adapters.py#L291-L316

https://github.com/urllib3/urllib3/blob/master/urllib3/poolmanager.py#L266-L279

Either way we aren't leveraging the pooling aspect of sessions which is arguably it's biggest selling point. I'll revert back to using a raw request and persist the CA and Key values in the client instance instead.

rnpridgeon: The session is now left open for the duration of the SR client's existence instead of falling back to raw requests.

confluent_kafka/avro/cached_schema_registry_client.py
edenhill: Right, but you don't want to check key_location or cert_location for true/false, but for None.

confluent_kafka/avro/cached_schema_registry_client.py
edenhill: There is no indication in this format what the output is after the :, which becomes even more confusing if empty strings are passed. I don't think the values themselves add much information anyway, the error message as in it is crystal clear.

tests/testconf-example.json
rnpridgeon: How about adding it to tests/README.md instead

tests/README.md
rnpridgeon: Why not?

tests/testconf-example.json
edenhill: That should absolute be stated in the test README, but it is also good with an example in the example, .e.g, `"topic": "$MY_TOPIC_ENV"`, notice how the naming makes it more clear that any env can be used

examples/integration_test.py
rnpridgeon: I humbly disagree but I have made the change anyway

examples/integration_test.py
rnpridgeon: tests/testconf-example.json

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -59,32 +60,41 @@ def __init__(self, url, max_schemas_per_subject=1000):
...
+
+    def __exit__(self, *args):
+        self.close()
+
+    def close(self):
+        self._session.close()
+
+    def _send_request(self, url, method='GET', body=None, headers={}):
+        if method not in VALID_METHODS:
+            raise ClientError("Method {} invalid; valid methods include {}".format(method, VALID_METHODS))
```
edenhill: ".. is invalid"

confluent_kafka/src/Producer.c
```diff
@@ -508,7 +508,7 @@ static PyMethodDef Producer_methods[] = {
 	  "  :param func on_delivery(err,msg): Delivery report callback to call "
 	  "(from :py:func:`poll()` or :py:func:`flush()`) on successful or "
 	  "failed delivery\n"
-          "  :param int timestamp: Message timestamp (CreateTime) in milliseconds since epoch UTC (requires librdkafka >= v0.9.4, api.version.request=true, and broker >= 0.10.0.0). Default value is current time.\n"
+          "  :param int timestamp: Message timestamp (CreateTime) in microseconds since epoch UTC (requires librdkafka >= v0.9.4, api.version.request=true, and broker >= 0.10.0.0). Default value is current time.\n"
```
edenhill: Did you change your mind? f88c55090bb5458dc6cc4010fc3e419d085ea234

examples/integration_test.py
```diff
@@ -374,6 +376,97 @@ def verify_avro():
...
+        dict(value={'name': 'abc'}, value_schema=basic),
+        dict(value={'name': 'abc'}, value_schema=basic, key=float_value, key_schema=prim_float),
+        dict(value={'name': 'abc'}, value_schema=basic, key=str_value, key_schema=prim_string),
+        dict(value=float_value, value_schema=prim_float, key={'name': 'abc'}, key_schema=basic),
+        dict(value=float_value, value_schema=prim_float, key=str_value, key_schema=prim_string),
+        dict(value=str_value, value_schema=prim_string, key={'name': 'abc'}, key_schema=basic),
+        dict(value=str_value, value_schema=prim_string, key=float_value, key_schema=prim_float),
+        # Verify identity check allows Falsy object values(e.g., 0, empty string) to be handled properly (issue #342)
+        dict(value='', value_schema=prim_string, key=0., key_schema=prim_float),
+        dict(value=0., value_schema=prim_float, key='', key_schema=prim_string),
```
edenhill: Please use full float notation, 0.0

examples/integration_test.py
```diff
@@ -1125,7 +1218,7 @@ def verify_config(expconfig, configs):
 
 # Exclude throttle since from default list
 default_modes = ['consumer', 'producer', 'avro', 'performance', 'admin']
-all_modes = default_modes + ['throttle', 'none']
+all_modes = default_modes + ['throttle', 'avro-HTTPS', 'none']
```
edenhill: avro-https

examples/integration_test.py
```diff
@@ -1140,6 +1233,21 @@ def print_usage(exitcode, reason=None):
     sys.exit(exitcode)
 
 
+def _generate_group_id():
```
edenhill: I think you can drop the underscore prefix, this is not within a class or module, nothing is being exported anyway

examples/integration_test.py
```diff
@@ -45,14 +45,16 @@
 except ImportError as e:
     with_progress = False
 
+testconf = 'tests/testconf-example.json'
```
edenhill: testconf.json, the example file is an example / template, it shouldnt be used directly without editing.

Let it default to tests/testconf.json and if it does not exist tell the user to write their own based on testconf-example.json

examples/integration_test.py
```diff
@@ -1152,18 +1260,20 @@ def print_usage(exitcode, reason=None):
...
-        bootstrap_servers = sys.argv[1]
-        if len(sys.argv) > 2:
-            topic = sys.argv[2]
-        if len(sys.argv) > 3:
-            schema_registry_url = sys.argv[3]
-    else:
-        print_usage(1)
+    bootstrap_servers = testconf.get('bootstrap.servers', None)
+    topic = testconf.get('topic', None)
+    schema_registry_url = testconf.get('schema.registry.url', None)
```
edenhill: We should error out here if there are no bootstrap servers or topic defined

examples/integration_test.py
```diff
@@ -1217,6 +1327,10 @@ def print_usage(exitcode, reason=None):
         print('=' * 30, 'Verifying Admin API', '=' * 30)
         verify_admin()
 
+    if 'avro-HTTPS' in modes:
+        print('=' * 30, 'Verifying AVRO with https', '=' * 30)
```
edenhill: So.. I wanted https to be HTTPS in the print, since it is an acronym, but https to remain in lower-case for the test mode.

tests/README.md
```diff
@@ -43,11 +49,25 @@ In order to execute the throttle_cb test you must first set a throttle for the c
...
     kafka-configs  --zookeeper <zookeeper host>:<zookeeper port> \
         --alter --delete-config 'request_percentage' \
         --entity-name throttled_client --entity-type clients
+
+HTTPS Schema Registry test
+==========================
+
+HTTPS tests require access to a Schema Registry instance configured to with at least one HTTPS listener.
+
+For instructions on how to configure the Schema Registry please see Confleunt's documentation:
```
edenhill: Confluent

tests/avro/test_cached_client.py
```diff
@@ -125,3 +125,19 @@ def test_multi_register(self):
 
     def hash_func(self):
         return hash(str(self))
+
+    def test_cert_no_key(self):
+        with self.assertRaises(ValueError):
+            self.client = CachedSchemaRegistryClient(url='https://127.0.0.1:8081',
```
edenhill: If we're unlucky the schema registry client will connect to this host and port, which we don't want for a unit test.
Not sure there is a good alternative, but maybe we should use port 65534 instead that is much less likely to be bound.

tox.ini
```diff
@@ -1,17 +1,21 @@
...
+passenv =
+    #http://tox.readthedocs.io/en/latest/config.html#confval-passenv=SPACE-SEPARATED-GLOBNAMES
+    *
+
 commands =
     pip install -v .
     py.test -v --timeout 20 --ignore=tmp-build --import-mode append {posargs}
-    # See tests/README.md for additional notes on testing
-    #python examples/integration_test.py <bootstrap-servers> confluent-kafka-testing [<schema-registry-url>]
+    #python examples/integration_test.py --conf tests/testconf-example.json
```
edenhill: testconf.json

confluent_kafka/src/Producer.c
rnpridgeon: No it got lost in my attempt to fix the broken rebase

rnpridgeon:
I have made a giant mess retargeting master locally. In order to avoid further pain I'm going to open a new PR based off master.

Issue #231, opened 2017-08-15
Custom partitioner support


sakshamgupta1:
Is there any documentation around using custom partitioners with producer?

edenhill:
Custom partitioners are currently not exposed in the Python client due to threading issues. The partitioner may be called from an internal librdkafka thread and this does not play well with the Python interpreter. We're looking into alternatives.

edenhill:
The next version of librdkafka (the underlying Kafka client) will expose the builtin partitioners as configuration properties, allowing you to change to an alternative builtin partitioner, such as the Java compatible murmur2_random partitioner.

Search for 'partitioner' here: https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md


Custom partitioners are not yet exposed in the Python client though.

ltucker:
@edenhill does the partition argument of Producer.produce() work? (ie can I execute a routing strategy not in a callback assuming I know the partition count and have enough other info on hand? )

edenhill:
Yes it does!

rnpridgeon:
@edenhill Do you think exposing the built-in partitioners here is enough or do we intend to make upstream changes to librdkafka to make this possible. Unless there are any objections I say we mark this as closed in favor of #396 for the time being. If any compelling arguments can be made for customer partitioners beyond those which are provided we can reopen .

edenhill:
Sounds good to me, close as wontfix with a reference to the built-ins.

2018-06-29 21:20 GMT+02:00 Ryan P <notifications@github.com>:

rnpridgeon:
Marked as `won't fix` with the addition of  #396 which exposes the built-in partitioners.

Bec-k:
So there is still no support for custom partitioners and assignments?

PR #396, opened 2018-06-07
Update built-in partitioner options to include murmur2

rnpridgeon:
Referenced in #231


rnpridgeon:
Travis failure due to the lingering 0.11.4 version in `.travis.yml`

edenhill:
There's no need for this anymore since librdkafka's partitioners can be configured directly in librdkafka through string properties, search for "partitioner" here: https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

We should rather remove this code from the Python client

rnpridgeon:
Makes sense, hadn't considered the fall through to librdkafka. Update inbound

tests/test_Producer.py
```diff
@@ -168,3 +168,26 @@ def handle_dr(err, msg):
     p.produce('mytopic', "\xc2\xc2", on_delivery=handle_dr)
 
     p.flush()
+
+
+def test_set_partioner_murmur2():
```
edenhill: partitioner

tests/test_Producer.py
```diff
@@ -168,3 +168,26 @@ def handle_dr(err, msg):
...
+    Producer({'partitioner': 'murmur2_random'})
+
+
+def test_set_invalid_partioner_murmur():
+    """
+    Assert invalid partitioner raises KafkaException
+    """
+    with pytest.raises(KafkaException) as e:
+        Producer({'partitioner': 'murmur'})
+        assert e == 'unknown builtin partitioner: murmur' in e
```
edenhill: are you certain this works, referencing the exception from with the block?
I think it needs to go in the outer block

rnpridgeon: It would fail otherwise no? I have tested it without checking for the raised exception at all and it raised the appropriate exception

edenhill: the assert is never reached since Producer() raises an exception

rnpridgeon: Perhaps I'm interpreting this wrong but I read it as the person raising the exception must be the last statement. Anything else should happen outside of the context. 

https://github.com/pytest-dev/pytest/blob/master/src/_pytest/python_api.py#L528-L547

rnpridgeon: yep you are correct. 

https://github.com/pytest-dev/pytest/blob/master/src/_pytest/python_api.py#L528-L547

fixing

rnpridgeon:
@edenhill, I think I got it all this go around

tests/test_Producer.py
edenhill: still partioner

edenhill:
Dont forget retarget for master before merging

Issue #321, opened 2018-02-20
Inconsistent constructor usage in docs


ecksun:
Description
===========

The examples in the README does not correspond to the API documentation.

For example in the README it says:

```python
from confluent_kafka import Consumer, KafkaError

c = Consumer({'bootstrap.servers': 'mybroker', 'group.id': 'mygroup',
              'default.topic.config': {'auto.offset.reset': 'smallest'}})
```

However in the [API docs](https://docs.confluent.io/current/clients/confluent-kafka-python/index.html#confluent_kafka.Consumer.Consumer) the constructor looks like this:

```
Consumer(**kwargs)
```

This was rather confusing when starting out. Moreover my IDE considered the form in the README to be invalid (it did not match the signature).

[The examples](https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/consumer.py#L74) also look like what I would expect.

Thus I suggest to change the README to `Consumer(**{...})`.

Edit: The same is also true for the constructors of `Producer` and the Avro constructors as well

How to reproduce
================
N/A

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): N/A
 - [x] Apache Kafka broker version: N/A
 - [x] Client configuration: N/A
 - [x] Operating system: N/A
 - [x] Provide client logs (with `'debug': '..'` as necessary): N/A
 - [x] Provide broker log excerpts: N/A
 - [ ] Critical issue

edenhill:
We actually support both kwargs and a dict, and we're trying to move away from the `**conf` notation because it confuses people and serves no purpose.

ecksun:
> We actually support both kwargs and a dict, and we're trying to move away from the **conf notation because it confuses people and serves no purpose.

Right, that make sense.

Then perhaps the right thing to do is to change the API to be `Consumer(config, **kwargs)` ?

edenhill:
Yeah, that'd work.

What IDE are you using btw?

ecksun:
> What IDE are you using btw?

IntelliJ with the python plugin normally but I ran into some issue with it not wanting to understand the .so file of confluent-kafka so for the project that uses confluent-kafka I use PyCharm

ecksun:
Looks good to me, thank you :)

PR #397, opened 2018-06-07
Fix constructor doc (#321)

rnpridgeon:


rnpridgeon:
So silently accept kwargs then without advertising this capability in the docs? Is that the right move or perhaps a brief description linking the rdkafka configuration page

edenhill:
we should link to CONFIGURATION.md, but that is for the config dict.
The kwargs is really just an alternative to the dict, it takes the same keys (but none of the standard config properties will work since they contain dots . ).

You can do this:
`p = Producer({"bootstrap.servers":"gnägg"}, logger=mylogger)`
which is identical to:
`p = Producer({"bootstrap.servers":"gnägg", "logger": mylogger})`

so I don't think there is any need to expose kwargs.

rnpridgeon:
To encourage good behavior perhaps its a good idea to expose kwargs referencing the configs specific to the python binding in the [docs](docs/index.rst)

rnpridgeon:
There are some extra formatting changes to appease the sphinx gods

Issue #289, opened 2017-12-19
AvroConsumer — access to key & value schemas


rcoup:
Description
===========

AvroConsumer happily returns decoded Avro messages, but drops all reference to the original Avro schema/identifier. The schema is super-useful in some cases (eg. Debezium, where it contains the source database table schema).

One (relatively straightforward) solution is to:
1. update `avro.MessageSerializer.decode_message()` to return a `(schema_id, payload)` tuple
2. have `AvroConsumer.poll()` wrap the `Message` in a Python `AvroMessage` subclass which has additional `.key_schema_id` and `.value_schema_id` attributes. Could also do this in C.
3. expose the schema registry client via a `AvroConsumer.get_schema(schema_id)` method

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): 0.11.0
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

edenhill:
That makes sense.

Since 1 and 2 would be breaking changes for existing applications I guess we are stuck with 3 for the time being.

Would you care to provide a PR?

rcoup:
Was all one proposed solution, not three options... (1) is the only API change, I assumed it was internal, but could always add a different method to MessageSerializer — (2) is backwards compatible?

Which bits are considered internal vs published APIs?

edenhill:
Sorry, misinterpreted your suggestion.

AvroConsumer is the only public interface to worry about here;
changing poll() to return an AvroMessage may be a breaking change for existing applications.

What if we start out with your suggested implementation, but prior to the next release decide to break poll() or add a second method for polling AvroMessages?

Would you care to submit a PR for this?

JimMcHale:
I'm looking to find a way to get the schema (or the schema id) from doing  AvroConsumer(); Is there  a way to do this without changing the source? 
Tks

adisunw:
is there any update on this?

PR #402, opened 2018-06-15
Optionally return schemas with AvroProducer

delphyne:
When calling poll, with_schemas kwarg may be provided to return a proxy for the message.  That proxy includes both the key_schema and value_schema as properties.  The raw message is also available as a property, and all other calls are delegated to the message.

This PR is particularly helpful along with #401.  Together they allow you to push multiple message types to the same topic and to be able to understand what you are getting back when consume from that topic.

Note: #289 suggests making AvroMessage a sub-class of confluent_kafka.cimpl.Message, but I could not find any way to make this work.  This version has it delegate with access to the original if desired for performance reasons.  In order to prevent this being a breaking change, the AvroMessage version of the message is only returned if schemas are explicitly requested.

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -81,6 +81,45 @@ class AvroProducer(Producer):
         super(AvroProducer, self).produce(topic, value, key, **kwargs)
 
 
+class AvroMessage(object):
+    """
+    Provides the schemas for the key and value of a message.  All other calls are delegated to the original message.
+    """
+    __slots__ = {'_key_schema', '_value_schema', '_message'}
+
+    def __init__(self, key_schema, value_schema, message):
+        self._key_schema = key_schema
+        self._value_schema = value_schema
+        self._message = message
+
+    def __getattr__(self, item):
+        return getattr(self._message, item)
+
+    @property
+    def key_schema(self):
+        """
+        The Avro schema used to encode the `key` of this message.
+        :return:
+        """
+        return self._key_schema
+
+    @property
+    def value_schema(self):
+        """
+        The Avro schema used to encode the `value` of this message.
+        :return:
+        """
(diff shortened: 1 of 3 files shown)


confluent_kafka/avro/__init__.py
```diff
@@ -81,6 +81,45 @@ def produce(self, **kwargs):
         super(AvroProducer, self).produce(topic, value, key, **kwargs)
 
 
+class AvroMessage(object):
```
rnpridgeon: This actually becomes fairly expensive on a per message basis. Rather than pulling the message type into python I'd suggest simply setting the additional attributes at runtime. 

~~~
def decode_message(self, message, schema_attr=None):
...
if not schema_attr is None: 
    msg.setattr(schema_attr, key_schema_obj)
...
~~~

This should be fairly cheap although admittedly it takes away from the readability a bit. Proper documentation should be able to bridge the gap here though.

rnpridgeon: Actually it looks like the Message type fails to set the slot for `__setattr__`, standby

delphyne: yeah... that's why i had to go this route.  i'm open to changing it if you have a better plan though.

delphyne: @rnpridgeon do you have updates on approach for this?

madtimber:
Hey gang! Wanted to bump this PR up to get some more eyes on it. I'm currently needing this exact functionality as well, and have subclassed things locally for now to get it working. A more integrated solution is obviously preferable so if I can help out in any way, please let me know!

swenzel:
How about using the message headers for this?

rnpridgeon:
Generic Serde support will provide a means to bind your schema with your data. This is similar to java's 'GenericRecord' implementation.  #502

PR #403, opened 2018-06-18
Unblock 0.11.5 RC packaging

rnpridgeon:
Couldn't come up with a good solution in time but this should unblock us for the time being

### DIFF
--- a/tools/test-manylinux.sh
+++ b/tools/test-manylinux.sh
@@ -64,7 +64,6 @@ function run_single_in_docker {
 
     # Pytest relies on a new version of six; later versions of pip fail to remove older versions gracefully
     # https://github.com/pypa/pip/issues/5247
-    pip install pytest
     pip install pytest --ignore-installed six
 
     # Verify that OpenSSL and zlib are properly linked
--- a/tools/test-manylinux.sh
+++ b/tools/test-manylinux.sh
@@ -55,8 +55,16 @@ function run_single_in_docker {
     hash -r # let go of previous 'pip'
 
     # Install modules
+
+    # TODO: revisit to avoid hardcoding dependencies
+    pip install "futures;python_version=='2.7'"
+    pip install "enum34;python_version=='2.7'"
+
     pip install confluent_kafka --no-index -f $wheelhouse
+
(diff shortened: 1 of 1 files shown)


tools/test-manylinux.sh
```diff
@@ -55,8 +55,16 @@ function run_single_in_docker {
...
 
     # Install modules
+
+    # TODO: revisit to avoid hardcoding dependencies
+    pip install "futures;python_version=='2.7'"
+    pip install "enum34;python_version=='2.7'"
+
     pip install confluent_kafka --no-index -f $wheelhouse
+
+    # https://github.com/pypa/pip/issues/5247
```
edenhill: Can you write in the comment why the double installs are needed, providing the URL for background info.

rnpridgeon:
Thanks @edenhill , updated with a brief description

tools/test-manylinux.sh
```diff
@@ -62,6 +62,7 @@ function run_single_in_docker {
 
     pip install confluent_kafka --no-index -f $wheelhouse
 
+    # Pytest relies on a new version of 6; later versions of pip fail to remove older versions gracefully
```
edenhill: "new version of 6"?  Do you mean six?

rnpridgeon: yes...

tools/test-manylinux.sh
```diff
@@ -55,8 +55,17 @@ function run_single_in_docker {
...
+    # TODO: revisit to avoid hardcoding dependencies
+    pip install "futures;python_version=='2.7'"
+    pip install "enum34;python_version=='2.7'"
+
     pip install confluent_kafka --no-index -f $wheelhouse
+
+    # Pytest relies on a new version of six; later versions of pip fail to remove older versions gracefully
+    # https://github.com/pypa/pip/issues/5247
     pip install pytest
+    pip install pytest --ignore-installed six
```
edenhill: do we need to install pytest twice?

rnpridgeon: how embarrassing... that is likely what caused my issues with six to begin with

rnpridgeon:
@edenhill , fixed

PR #404, opened 2018-06-18
Ensure plugin path handled prior to plugin configs

rnpridgeon:


confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,6 +1568,40 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
...
+                                "expected configuration property name "
+                                "as type unicode string");
+                        rd_kafka_topic_conf_destroy(tconf);
+                        rd_kafka_conf_destroy(conf);
+                        return NULL;
+                }
+
+                v = cfl_PyUnistr_AsUTF8(vs, &vs8);
+
+                if(rd_kafka_conf_set(conf, "plugin.library.paths", v, NULL, 0) != RD_KAFKA_CONF_OK) {
```
edenhill: `if (`

keep under 80 columns wide

tests/test_Consumer.py
```diff
@@ -313,3 +314,17 @@ def test_calling_store_offsets_after_close_throws_erro():
     with pytest.raises(RuntimeError) as ex:
         c.offsets_for_times([TopicPartition("test", 0)])
     assert 'Consumer closed' == str(ex.value)
+
+
+@pytest.skipif(len([True for x in (".so", ".dylib", ".dll") if os.path.exists("monitoring-interceptor" + x)]) > 0,
+               reason="requires confluent-librdkafka-plugins to be installed")
```
edenhill: and copied to the current directory

tests/test_Consumer.py
```diff
@@ -313,3 +314,17 @@ def test_calling_store_offsets_after_close_throws_erro():
...
+def test_unordered_dict():
+    """
+    Interceptor configs can only be handled after the plugin has been loaded not before.
+    """
+    Consumer({'confluent.monitoring.interceptor.publishMs': 1000,
+              'confluent.monitoring.interceptor.sessionDurationMs': 1000,
+              'plugin.library.paths': 'monitoring-interceptor.dylib',
```
edenhill: Skip the file extension, handled by the librd of kafka

tests/test_Producer.py
```diff
@@ -191,3 +192,17 @@ def test_set_invalid_partitioner_murmur():
     with pytest.raises(KafkaException) as e:
         Producer({'partitioner': 'murmur'})
     assert 'Invalid value for configuration property "partitioner": murmur' in str(e)
+
+
+@pytest.skipif(len([True for x in (".so", ".dylib", ".dll") if os.path.exists("monitoring-interceptor" + x)]) > 0,
```
edenhill: Instead of duplicating this test, move it to a common file (e.g., test_misc) and parameterize the client types: producer, consumer, admin

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,6 +1568,40 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
         /* Enable valid offsets in delivery reports */
         rd_kafka_topic_conf_set(tconf, "produce.offset.report", "true", NULL, 0);
 
+        /* Preload Plugins since dicts are unordered */
```
edenhill: Preloading a library/plugin is something else, I think you can instead flesh out the comment a bit, something like:
Plugins must be configured prior to any configuration properties they support, since dicts are unordered
we explicitly check for, set, and delete the plugin paths here to make sure they are applied prior to
their configuration.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,7 +1568,11 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
         /* Enable valid offsets in delivery reports */
         rd_kafka_topic_conf_set(tconf, "produce.offset.report", "true", NULL, 0);
 
-        /* Preload Plugins since dicts are unordered */
+       /*
```
edenhill: Indent seems off, /* should probably be one space to the right

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,7 +1568,11 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
...
+        * Plugins must be configured prior to handling any of their configuration properties
+        * Dicts are unordered so we explicitly check for, set, and delete the plugin paths here
```
edenhill: end sentence with period.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,7 +1568,11 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
...
+        * This ensures all configuration properties are handled in the right order
```
edenhill: end sentence with period.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1584,8 +1588,10 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
 
                 v = cfl_PyUnistr_AsUTF8(vs, &vs8);
 
-                if(rd_kafka_conf_set(conf, "plugin.library.paths", v, NULL, 0) != RD_KAFKA_CONF_OK) {
-                        cfl_PyErr_Format(RD_KAFKA_RESP_ERR__INVALID_ARG, "%s", NULL);
+                if (rd_kafka_conf_set(conf, "plugin.library.paths", v, NULL, 0)
+                   != RD_KAFKA_CONF_OK) {
+                        cfl_PyErr_Format(RD_KAFKA_RESP_ERR__INVALID_ARG,
+                                "%s", NULL);
```
edenhill: "%s", NULL will raise an exception with the message "(null)".
Instead pass errstr to conf_set, and then use that here.

tests/test_misc.py
```diff
@@ -123,3 +124,19 @@ def test_throttle_event_types():
...
+
+
+@pytest.mark.skipif(len([True for x in (".so", ".dylib", ".dll")
+                         if os.path.exists("monitoring-interceptor" + x)]) == 0,
+                    reason="requires confluent-librdkafka-plugins be installed and copied to the current directory")
+def test_unordered_dict():
+    """
+    Interceptor configs can only be handled after the plugin has been loaded not before.
+    """
+    for fn in [confluent_kafka.Producer, confluent_kafka.Consumer, confluent_kafka.admin.AdminClient]:
```
edenhill: I was thinking you could use parameterization: https://docs.pytest.org/en/latest/parametrize.html

confluent_kafka/src/confluent_kafka.c
rnpridgeon: Yep sorry that was left over from when I did the initial POC

rnpridgeon:
@edenhill good call on mutating the application's dict. I have made the requested changes

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1530,7 +1530,7 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
          * precedence in case of duplicate keys.
          * All keys map to configuration properties.
          */
```
edenhill: Extend this comment to mention why we make a copy of the dict

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,6 +1569,47 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
         /* Enable valid offsets in delivery reports */
         rd_kafka_topic_conf_set(tconf, "produce.offset.report", "true", NULL, 0);
 
+        /*
+         * Plugins must be configured prior to handling any of their configuration properties.
+         * Dicts are unordered so we explicitly check for, set, and delete the plugin paths here.
+         * This ensures all configuration properties are handled in the right order.
```
edenhill: s/all/plugin/

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,6 +1569,47 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
...
+                char errstr[256];
+                PyObject *vs = NULL, *vs8 = NULL;
+
+                if (!(vs = cfl_PyObject_Unistr(vo))) {
+                        PyErr_SetString(PyExc_TypeError,
+                                "expected configuration property name "
+                                "as type unicode string");
+                        rd_kafka_topic_conf_destroy(tconf);
+                        rd_kafka_conf_destroy(conf);
+                        return NULL;
```
edenhill: confdict is leaked here, do Py_DECREF

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1568,6 +1569,47 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
...
+                        cfl_PyErr_Format(RD_KAFKA_RESP_ERR__INVALID_ARG,
+                                "%s", errstr);
+
+                        rd_kafka_topic_conf_destroy(tconf);
+                        rd_kafka_conf_destroy(conf);
+
+                        Py_XDECREF(vs8);
+                        Py_XDECREF(vs);
+
+                        return NULL;
```
edenhill: dito

edenhill: also check the other returns below

rnpridgeon:
@edenhill , slightly less leaky. There are some whitespace changes but I believe they add to readability. If its a problem I can remove them .

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1529,6 +1529,8 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
          * When both args and kwargs are present the kwargs take
          * precedence in case of duplicate keys.
          * All keys map to configuration properties.
+         *
+         * Copy configuration dict to avoid manipulating application config
```
edenhill: nit (but that's what you want..): How about a trailing period?

rnpridgeon: touché

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1530,7 +1530,7 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
          * precedence in case of duplicate keys.
          * All keys map to configuration properties.
          *
-         * Copy configuration dict to avoid manipulating application config
+         * Copy configuration dict to avoid manipulating application config.
```
edenhill: Great fix!

PR #408, opened 2018-06-21
Complete HTTPS support for schema registry (#90)

rnpridgeon:


tox.ini
```diff
@@ -7,11 +7,14 @@ setenv =
...
     LD_LIBRARY_PATH={toxinidir}/tmp-build/lib
+passenv =
+    #http://tox.readthedocs.io/en/latest/config.html#confval-passenv=SPACE-SEPARATED-GLOBNAMES
+    *
 commands =
     pip install -v .
     py.test -v --timeout 20 --ignore=tmp-build --import-mode append {posargs}
     # See tests/README.md for additional notes on testing
-    #python examples/integration_test.py <bootstrap-servers> confluent-kafka-testing [<schema-registry-url>]
+    #python examples/integration_test.py --avro-https --conf tests/testconf.json
```
edenhill: Is --avro-https meant as an example here?
It might be better to remove that so that the larger test suite is run by default if people just uncomment the line.

edenhill:
Might want to fix the build warnings before merging:
https://travis-ci.org/confluentinc/confluent-kafka-python/jobs/396459372#L2439

PR #410, opened 2018-06-25
OpenSSL version bump to 1.0.2o

rnpridgeon:


tools/RELEASE.md
```diff
@@ -43,15 +43,27 @@ Commit these changes with a commit-message containing the version:
     $ git commit -m "Version v0.11.4rc1" confluent_kafka/src/confluent_kafka.c docs/conf.py setup.py
 
 
-## 2. Create a tag
+## 2. Update OpenSSL version if necessary
```
edenhill: The version tag (in step 3) must go on the version commit (step 1), so you need to put this step before the commit in step 1, preferably before all of step 1.

PR #411, opened 2018-06-26
Bump cibuildwheel and 0.11.5rc

rnpridgeon:


confluent_kafka/src/confluent_kafka.c
```diff
@@ -2156,7 +2156,7 @@ static PyObject *libversion (PyObject *self, PyObject *args) {
 }
 
 static PyObject *version (PyObject *self, PyObject *args) {
-	return Py_BuildValue("si", "0.11.5rc5", 0x000b0500);
+	return Py_BuildValue("si", "0.11.5rc6", 0x000b0500);
```
edenhill: You should set this to 0.11.5, doesn't need to be bumped for each rc

edenhill:
It seems the osx builder isn't actually building anything, probably unrelated to this PR but it is a regression

rnpridgeon:
Added py27 and py36 builders to the build index. It's slightly hacky but unfortunately it would seem travis ci doesn't make this easy today. 

https://github.com/travis-ci/travis-ci/issues/2312

.travis.yml
```diff
@@ -12,6 +12,24 @@ matrix:
     dist: trusty
     python: "3.6"
     env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=master
+  # Source package verification with Python 2.7 and librdkafka master
+  - os: osx
+    python: "2.7"
+    env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=master
+    before_install:
+        - brew update && brew upgrade pyenv
```
edenhill: nit: no reason to use && instead of separate lines

rnpridgeon:
holding for librdkafka 0.11.5 release so we can update the version here

rnpridgeon:
This fell out of sight and has since been fixed in a few other PRs

Issue #412, opened 2018-06-28
Synchronous Functions Should Release the GIL


bpowers39:
Description
===========
Some functions that take a timeout do not release the GIL. This will cause all python threads to block until the function returns.

Functions that seem to do this include:
Consumer.committed()
Consumer.get_watermark_offsets()
Consumer.offsets_for_times()
Producer.flush()

Consumer.get_watermark_offsets() is the only function I'm completely sure of. The others are just a guess from looking at the code.

How to reproduce
================
Create a topic with one partition. Run the below example. Once the example has consumed a few messages, kill the broker hosting the partition. Notice the other thread is longer printing it's debug message.

```python
import time
import sys
import threading
import confluent_kafka
from confluent_kafka import Consumer, KafkaError
from uuid import uuid4

def debug_thread_func():
    while True:
        print("Test...")
        time.sleep(1)

if __name__ == '__main__':
    print(confluent_kafka.version(), confluent_kafka.libversion())

    debug_thread = threading.Thread(target=debug_thread_func)
    debug_thread.start()

    client = Consumer({'bootstrap.servers': 'gateway:9092', 'group.id': str(uuid4()),
              'default.topic.config': {'auto.offset.reset': 'smallest'}})

    def assigned(consumer, partitions):
        print("Assigned:", partitions)
        

    client.subscribe(['ibbot'], on_assign=assigned)

    while True:
        msg = client.poll(timeout=1)
        for partition in client.assignment():
            print(client.get_watermark_offsets(partition))

        if msg is not None:
            if msg.error():
                print("Error: ", msg.error())
            else:
                print("Data")

    client.close()
   

```

Note that this example is probably dependent on a bug in rdkafka, so it may not be reproducible 100% of the time. You can also reproduce this with a multi-partition and broker setup. In this case, the application only deadlocks for as long as it takes for a new leader to be elected.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):  kafka-python: ('0.11.4', 721920)  rdkafka: ('0.11.4', 722175)
 - [x] Apache Kafka broker version: 0.11.0.2
 - [x] Client configuration: `{'bootstrap.servers': 'gateway:9092', 'group.id': str(uuid4()),
              'default.topic.config': {'auto.offset.reset': 'smallest'}}`
 - [X] Operating system: RHEL 7
[...]

PR #415, opened 2018-07-03
Release GIL prior to blocking (#412)

rnpridgeon:


confluent_kafka/src/Consumer.c
```diff
@@ -595,8 +595,10 @@ static PyObject *Consumer_committed (Handle *self, PyObject *args,
 	if (!(c_parts = py_to_c_parts(plist)))
 		return NULL;
 
-	err = rd_kafka_committed(self->rk, c_parts,
-				tmout >= 0 ? (int)(tmout * 1000.0f) : -1);
+        Py_BEGIN_ALLOW_THREADS
```
edenhill: I know the Python docs advocate the use of this macro, but it is bad practice in C to design and use macros that end up with invalid C syntax (even though the expanded macro of course is correct).
This will mess up indentation.

Looking at the macro it should be safe to append a `;` to each of those two lines, or even:
```
Py_BEGIN_ALLOW_THREADS {
...
} PY_BEGIN_END_ALLOW_THREADS;
```

rnpridgeon: Alternatively I can use `PyEval_SaveThread` and `PyEval_RestoreThread` directly. Dealer's choice

rnpridgeon: Actually if we are going to break convention anyway we might as well keep it consistent with the others. I'll go ahead and replace the macros with `PyEval_SaveThread` and `PyEval_RestoreThread`

edenhill: I disagree, using the existing macros will make the code future-proof if additional calls are added in upcoming Python versions.

rnpridgeon: Then we can fix them all at the same time! #consistency 

I see your point, I'll fix the syntax.

rnpridgeon:
@edenhill fixed

PR #419, opened 2018-07-05
Support encoding with fastavro, fix minor doc bug

marcintustin:
This PR adds the ability to encode with fastavro, if it is available. Also, fixes a minor misprecision in the docstring for `encode_record_with_schema_id`.

### DIFF
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -108,7 +108,8 @@ class MessageSerializer(object):
             raise serialize_err(message)
 
         # cache writer
-        self.id_to_writers[schema_id] = self.make_datum_writer(schema)
+        if id not in self.id_to_writers:
+            self.id_to_writers[schema_id] = self.make_datum_writer(schema)
 
         return self.encode_record_with_schema_id(schema_id, record, is_key=is_key)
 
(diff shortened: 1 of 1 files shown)


marcintustin:
[clabot:check]

marcintustin:
The failure here appears to be with one of the workers timing out. I can't force a re-run, I'm afraid.

rnpridgeon:
Thanks @marcintustin! I'll take a look tomorrow, in the meantime I have forced a restart of the failed worker.

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -101,34 +108,35 @@ def encode_record_with_schema(self, topic, schema, record, is_key=False):
...
-        # use slow avro
         if schema_id not in self.id_to_writers:
             # get the writer + schema
-
             try:
                 schema = self.registry_client.get_by_id(schema_id)
                 if not schema:
                     raise serialize_err("Schema does not exist")
-                self.id_to_writers[schema_id] = avro.io.DatumWriter(schema)
+                self.id_to_writers[schema_id] = self.make_datum_writer(schema)
```
rnpridgeon: I would move this logic back into `_encode_record_with_schema_id` and eliminate the redundant `_slow_encode_record_with_schema_id` and `_fast_encode_record_with_schema_id` methods. If we provide the same interface for both writers there is no longer a need to maintain two code paths which is a bit more DRY.

rnpridgeon:
Thanks for the contribution @marcintustin , I went ahead and made the requested changes in #492 to try and get this in for the 1.0 release which enters code freeze at EOD.

PR #420, opened 2018-07-06
Update README to add another Install option with SASL Kerberos/GSSAPI…

danbcd:
This is an improvement of the Install section from README.md to inform user that if they want SASL Kerberos/GSSAPI support, they also have the option to Install from source from PyPi and not only to build confluent-kafka from source. I have discussed this with @edenhill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,8 @@ Install
 **NOTE:** The pre-built Linux wheels do NOT contain SASL Kerberos/GSSAPI support.
           If you need SASL Kerberos/GSSAPI support you must install librdkafka and
           its dependencies using the repositories below and then build
-          confluent-kafka from source.
+          confluent-kafka from source or use the command in the "Install from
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,7 +195,8 @@ Install
 **NOTE:** The pre-built Linux wheels do NOT contain SASL Kerberos/GSSAPI support.
           If you need SASL Kerberos/GSSAPI support you must install librdkafka and
           its dependencies using the repositories below and then build
-          confluent-kafka from source.
+          confluent-kafka from source or use the command in the "Install from
```
edenhill: > then build confluent-kafka from source using the command in the "Install .....

rnpridgeon:
Thanks for this. I went ahead and opened a new PR since the CLA has yet to be signed and this appears to have stagnated.

rnpridgeon:
#439

PR #426, opened 2018-07-19
Update packaging and bump RC

rnpridgeon:


tools/RELEASE.md
```diff
@@ -5,8 +5,8 @@ when no more blocking issues needs to be fixed a final tag is created
 and the final release is rolled.
 
 confluent-kafka-python uses semver versioning and loosely follows
-librdkafka's version, e.g. v0.11.4 for the final release and
-v0.11.4rc3 for the 3rd v0.11.4 release candidate.
+librdkafka's version, e.g. v0.11.5 for the final release and
```
edenhill: no need to update this, these are just examples.

docs/conf.py
```diff
@@ -57,7 +57,7 @@
 # The short X.Y version.
 version = '0.11.5'
 # The full version, including alpha/beta/rc tags.
-release = '0.11.5'
+release = '0.11.5rc6'
```
edenhill: this one and setup.py are on rc6 while version() returns v0.11.5. Is this intentional? (if so; fine)

tools/RELEASE.md
rnpridgeon: is there any problem with doing so?

docs/conf.py
rnpridgeon: right so we had a bit of back and forth about include or don't include. Since the comment specifically mentions the tags I decided to keep the RC designation

PR #427, opened 2018-07-20
Feature/explicit read

martintaraz:
Modifications on the python source code for working with an explicit given read schemata for decoding avro messages
for tesing purposes, the integration where ran against a docker kafka build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,13 @@ script:
  - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then cibuildwheel --output-dir wheelhouse ; ls -la wheelhouse/ ; fi
  - if [[ -n $TRAVIS_TAG && $TRAVIS_OS_NAME == linux && -n $CIBW_BEFORE_BUILD ]]; then tools/test-manylinux.sh ; fi
 
+before_install:
+ - docker pull landoop/fast-data-dev
+ - docker run -d\
+  -p 2181:2181 -p 3030:3030 -p 8081:8081 \
+  -p 8082:8082 -p 8083:8083 -p 9092:9092 \
+  -e ADV_HOST=127.0.0.1 \
+  landoop/fast-data-dev
 
 deploy:
  provider: s3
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -5,34 +5,17 @@ matrix:
     language: python
     dist: trusty
     python: "2.7"
-    env: LD_LIBRARY_PATH="$PWD/tmp-build/lib" LIBRDKAFKA_VERSION=v0.11.5
-  # Source package verification with Python 3.6 and librdkafka v0.11.5
+    services: docker
```
rnpridgeon: All changes to .travis.yml should be removed from this commit

examples/integration_test.py
```diff
@@ -21,7 +21,6 @@
 """ Test script for confluent_kafka module """
 
 import confluent_kafka
-from confluent_kafka import admin
```
rnpridgeon: Please remove any integration test changes not related to the pull request itself. This helps to keep the commit history clean. If you wish to submit a PR which cleans up formatting, white space and imports I'd be happy to review and merge however :)

fkaufer:
> I have to ask though what is the use case for setting the reader schema here. On the surface this seems like a way to handle forward-compatible changes. If that is the case allowing the Schema Registry enforce compatibility constraints seems like a more robust solution for this situation.

@rnpridgeon No, it's about actually taking advantage of Avro's back-/forward compatibility features for which you need to be able to separate writer and reader schema. Before the change, the reader schema was (implicitly) hard-coded to the writer schema (of the respective message), now the consumer can pin a reader schema with the fastavro deserializer doing back or forward schema migration to the reader schema. The Schema Registry only enforces that the sequence of published Avro schemas for a certain topic complies with back-/forward compatibility constraints.

fkaufer:
> @fkaufer you are absolutely correct, I had only considered the ability to parse the message. Sorry about that. Other than the .travis.yml changes it LGTM. If you could back those out it would be greatly appreciated.
Reverted the .travis.yml and other non-related changes in https://github.com/confluentinc/confluent-kafka-python/pull/469

rnpridgeon:
Fixed with #470

PR #434, opened 2018-08-10
Safely release resources in Handle_clear

coldeasy:
During tests we create and destroy instances of `Consumer` and `Producer` on every test case. We noticed segfaults and memory corruption (the `logger` pointer was referring to other random objects).

This seems to be due to the case where both `tp_clear` and `tp_dealloc` are called on the handle, resulting in a double release of `error_cb`, `stats_cb` and `logger`. I've tested this locally and have not encountered the segfaults/memory corruption since.

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -1324,16 +1324,25 @@ static void log_cb (const rd_kafka_t *rk, int level,
  * Clear Python object references in Handle
  */
 void Handle_clear (Handle *h) {
-	if (h->error_cb)
-		Py_DECREF(h->error_cb);
+        if (h->error_cb) {
+                Py_DECREF(h->error_cb);
+                h->error_cb = NULL;
+        }
 
-        if (h->throttle_cb)
+        if (h->throttle_cb) {
                 Py_DECREF(h->throttle_cb);
+                h->throttle_cb = NULL;
+        }
 
-	if (h->stats_cb)
-		Py_DECREF(h->stats_cb);
+        if (h->stats_cb) {
+                Py_DECREF(h->stats_cb);
+                h->stats_cb = NULL;
+        }
 
-        Py_XDECREF(h->logger);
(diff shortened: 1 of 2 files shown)


rnpridgeon:
Good catch @coldeasy, this actually came up very recently as t causes issues for python, .net and likely our golang client as well. In order to handle this more elegantly the following PR has been opened for librdkafka. Once merged the plan is to adapt python to use the new immediate closure flags https://github.com/edenhill/librdkafka/pull/1932

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1324,24 +1324,37 @@ static void log_cb (const rd_kafka_t *rk, int level,
  * Clear Python object references in Handle
  */
 void Handle_clear (Handle *h) {
-	if (h->error_cb)
-		Py_DECREF(h->error_cb);
+    if (h->error_cb)
```
edenhill: Use 8 spaces for indent.
Put open-brace on same line as `if`:
`if (h->error_cb) {`

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1324,24 +1324,37 @@ static void log_cb (const rd_kafka_t *rk, int level,
...
+    }
 
-        Py_XDECREF(h->logger);
+    if (h->logger)
+    {
+        Py_DECREF(h->logger);
+        h->logger = NULL;
+    }
 
-        if (h->initiated) {
```
edenhill: no whitespace/indent changes please

coldeasy: Fixed

confluent_kafka/src/confluent_kafka.c
coldeasy: Updated, please take another look.

coldeasy:
[clabot:check]

edenhill:
Thank you for this!

PR #440, opened 2018-08-18
Add support for HTTP Basic Auth 

rnpridgeon:


confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -42,17 +44,46 @@ class CachedSchemaRegistryClient(object):
...
+    :param: str|dict url: url(deprecated) to schema registry or dictionary containing client configuration
+    :param: str ca_location: File or directory path to CA certificate(s) for verifying the Schema Registry key
+    :param: str cert_location: Path to client's public key used for authentication.
+    :param: str key_location: Path to client's private key used for authentication.
+
     """
 
     def __init__(self, url, max_schemas_per_subject=1000, ca_location=None, cert_location=None, key_location=None):
-        """Construct a client by passing in the base URL of the schema registry server"""
+        # In order to maintain comparability the url(conf in future versions) param has been preserved for now.
```
edenhill: comparability or compatbility?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -42,17 +44,46 @@ class CachedSchemaRegistryClient(object):
...
+        conf = url
+        if isinstance(url, str):
+            conf = {
+                'url': url,
+                'ssl.ca.location': ca_location,
+                'ssl.certificate.location': cert_location,
+                'ssl.key.location': key_location
+            }
+            warnings.simplefilter('always', DeprecationWarning)  # Deprecation warnings are suppressed by default
```
edenhill: Are you really supposed to change the warnings filter from the source of warning, rather than the app?
If so; how would an app silence these?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -42,17 +44,46 @@ class CachedSchemaRegistryClient(object):
...
+                "Existing params ca_location, cert_location and key_location will be replaced with their "
+                "librdkafka equivalents as keys in the conf dict: `ssl.ca.location`, `ssl.certificate.location` and "
+                "`ssl.key.location` respectively",
+                category=DeprecationWarning, stacklevel=2)
+            warnings.simplefilter('default', DeprecationWarning)  # reset filter
+
+        """Construct a Schema Registry client"""
+
+        # Ensure URL valid scheme is included; http[s]
+        if not conf.get('url', '').startswith("http"):
```
edenhill: mix of single and double quotes

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -83,6 +111,35 @@ def __exit__(self, *args):
     def close(self):
         self._session.close()
 
+    @staticmethod
+    def _configure_basic_auth(conf):
+        url = conf['url']
+        auth_provider = conf.get('basic.auth.credentials.source', 'URL').upper()
+        if auth_provider not in VALID_AUTH_PROVIDERS:
+            raise ValueError("basic.auth.credentials.source must be one of {}"
```
edenhill: Should probably print the `schema.registry.` prefix here

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -83,6 +111,35 @@ def __exit__(self, *args):
...
+                             .format(auth_provider, VALID_AUTH_PROVIDERS))
+
+        if auth_provider == 'SASL_INHERIT':
+            if conf.get('sasl.mechanisms', '').upper() == 'GSSAPI':
+                raise ValueError("SASL_INHERIT supports SASL mechanisms PLAIN and SCRAM only")
+            auth = (conf.get('sasl.username', None), conf.get('sasl.password'))
+        elif auth_provider == 'USERINFO':
+            auth = tuple(conf.get('basic.auth.user.info', None).split(':'))
```
edenhill: This will perform None.split(':') if no basic.auth.user.info is configured, which probably raises an exception

tests/avro/test_cached_client.py
```diff
@@ -148,3 +148,53 @@ def test_context(self):
...
+        self.client = CachedSchemaRegistryClient({
+            'url': 'https://user_url:secret@127.0.0.1:65534',
+        })
+        self.assertTupleEqual(('user_url', 'secret'), self.client._session.auth)
+
+    def test_basic_auth_userinfo(self):
+        self.client = CachedSchemaRegistryClient({
+            'url': 'https://user_url:secret@127.0.0.1:65534',
+            'basic.auth.credentials.source': 'userinfo',
+            'basic.auth.user.info': 'user_userinfo:secret'
```
edenhill: use another password here to make sure it doesnt pick up the urlinfo one

tests/avro/test_cached_client.py
```diff
@@ -148,3 +148,53 @@ def test_context(self):
...
+        })
+        self.assertTupleEqual(('user_userinfo', 'secret'), self.client._session.auth)
+
+    def test_basic_auth_sasl_inherit(self):
+        self.client = CachedSchemaRegistryClient({
+            'url': 'https://user_url:secret@127.0.0.1:65534',
```
edenhill: same

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -83,6 +111,35 @@ def __exit__(self, *args):
...
+        auth_provider = conf.get('basic.auth.credentials.source', 'URL').upper()
+        if auth_provider not in VALID_AUTH_PROVIDERS:
+            raise ValueError("basic.auth.credentials.source must be one of {}"
+                             .format(auth_provider, VALID_AUTH_PROVIDERS))
+
+        if auth_provider == 'SASL_INHERIT':
+            if conf.get('sasl.mechanisms', '').upper() == 'GSSAPI':
+                raise ValueError("SASL_INHERIT supports SASL mechanisms PLAIN and SCRAM only")
+            auth = (conf.get('sasl.username', None), conf.get('sasl.password'))
+        elif auth_provider == 'USERINFO':
```
mhowlett: `SASL_INHERIT` has an underscore and `USERINFO` doesn't. in the java implementation, they both have an underscore.

confluent_kafka/avro/__init__.py
```diff
@@ -29,23 +29,17 @@ class AvroProducer(Producer):
     def __init__(self, config, default_key_schema=None,
                  default_value_schema=None, schema_registry=None):
 
-        schema_registry_url = config.pop("schema.registry.url", None)
-        schema_registry_ca_location = config.pop("schema.registry.ssl.ca.location", None)
-        schema_registry_certificate_location = config.pop("schema.registry.ssl.certificate.location", None)
-        schema_registry_key_location = config.pop("schema.registry.ssl.key.location", None)
+        sr_conf = {key.replace("schema.registry.", ""): value
+                   for key, value in config.items() if key.startswith("schema.registry.") or key.startswith("sasl")}
```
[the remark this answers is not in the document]
rnpridgeon: This will just result in a dict with a few unused keys, once it falls out of the constructor's scope it will be eligible for garbage collection and no longer a problem no?

Realistically I could just pass the entire dict to the SR client if you'd prefer.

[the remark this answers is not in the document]
edenhill: But the SR client shouldn't silently ignore configs it doesn't understand though, but error out on them.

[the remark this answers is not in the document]
rnpridgeon: Fixed, including a check for unrecognized properties

rnpridgeon:
Now with integration tests!

confluent_kafka/avro/__init__.py
```diff
@@ -29,23 +29,23 @@ class AvroProducer(Producer):
...
                  default_value_schema=None, schema_registry=None):
 
-        schema_registry_url = config.pop("schema.registry.url", None)
-        schema_registry_ca_location = config.pop("schema.registry.ssl.ca.location", None)
-        schema_registry_certificate_location = config.pop("schema.registry.ssl.certificate.location", None)
-        schema_registry_key_location = config.pop("schema.registry.ssl.key.location", None)
+        sr_conf = {key.replace("schema.registry.", ""): value
+                   for key, value in config.items() if key.startswith("schema.registry.")}
+
+        if config.get("schema.registry.basic.auth.credentials.source") == 'SASL_INHERIT':
```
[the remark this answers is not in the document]
mhowlett: oh never mind, this is a producer specialized for Avro, forgot that's how it's implemented in this client. I was mirroring here some thoughts from another client where that is not the case.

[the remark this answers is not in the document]
rnpridgeon: Yeah it doesn't get much more coupled than specialized client intended solely for use with the cached schema registry client. In the long term this will just be a convenience wrapper over a client that sets `key.serializer`, `value.serializer` for you. The Avro serializer will leverage its own configure method much like that of the java clients

confluent_kafka/avro/__init__.py
```diff
@@ -102,23 +102,23 @@ class AvroConsumer(Consumer):
...
+            sr_conf['sasl.mechanisms'] = config.get('sasl.mechanisms', '')
```
edenhill: There's a mix of `'` and `"` here. I do the same, I just wanted to point it out that we're the same, and it looks fantastic!

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -83,6 +112,35 @@ def __exit__(self, *args):
...
+    def _configure_basic_auth(conf):
+        url = conf['url']
+        auth_provider = conf.pop('basic.auth.credentials.source', 'URL').upper()
+        if auth_provider not in VALID_AUTH_PROVIDERS:
+            raise ValueError("schema.registry.basic.auth.credentials.source must be one of {}"
+                             .format(auth_provider, VALID_AUTH_PROVIDERS))
+
+        if auth_provider == 'SASL_INHERIT':
+            if conf.pop('sasl.mechanisms', '').upper() == 'GSSAPI':
+                raise ValueError("SASL_INHERIT supports SASL mechanisms PLAIN and SCRAM only")
```
edenhill: Then you should check for that, not GSSAPI.

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -83,6 +112,35 @@ def __exit__(self, *args):
...
+            auth = utils.get_auth_from_url(url)
+
+        conf['url'] = utils.urldefragauth(url)
+        return auth
+
+    @staticmethod
+    def _configure_client_tls(conf):
+        cert = conf.pop('ssl.certificate.location', None), conf.pop('ssl.key.location', None)
+        # Both values can be None or no values can be None
+        if sum(x is None for x in cert) == 1:
```
edenhill: for next time; `if bool(a) != bool(b):` is simpler.

docker/bin/certify.sh
```diff
@@ -0,0 +1,26 @@
...
+export PASS="abcdefgh"
+source ${DOCKER_BIN}/../.env
+
+if [ -f ${TLS}/ca-cert ]; then
+    echo "${TLS}/ca-cert found; skipping certificate generation.."
+    exit 0
+fi
+
+# Clean up old certs
+#for file in $(ls ${TLS});do
```
edenhill: `for file in $TLS/*` ?

docker/bin/certify.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+DOCKER_BIN="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+export PASS="abcdefgh"
+source ${DOCKER_BIN}/../.env
+
+if [ -f ${TLS}/ca-cert ]; then
```
edenhill: use new style if brackets to avoid having to escape stuff:
`if [[ -f .. ]] ; then`

docker/bin/certify.sh
```diff
@@ -0,0 +1,26 @@
...
+    echo "${TLS}/ca-cert found; skipping certificate generation.."
+    exit 0
+fi
+
+# Clean up old certs
+#for file in $(ls ${TLS});do
+#        rm ${TLS}/${file}
+#done
+
+echo "Creating ca-cert..."
```
edenhill: I'd cache the hostname in a variable

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,42 @@
...
+
+DOCKER_BIN="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+source ${DOCKER_BIN}/../.env
+
+await_http() {
+    local exit_code
+    local attempt=0
+
+    curl -s "$2" > /dev/null; exit_code=$?
+    while [ "${exit_code}" -ne 0 ] && [ "${attempt}" -lt 5 ]; do
```
edenhill: double bracketry and remove quotes

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+DOCKER_BIN="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+source ${DOCKER_BIN}/../.env
+
+await_http() {
```
edenhill: describe input parameters

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,42 @@
...
+        curl -s "$2" > /dev/null; exit_code=$?
+        let "attempt+=1"
+        sleep 5
+    done
+
+    if [ ${exit_code} -eq "0" ]; then
+        return
+    fi
+
+    echo "$1 readiness test failed aborting..."
```
edenhill: "..test failed: aborting", unless it failed to abort

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
```
edenhill: use `set -eu` at the start of all scripts to catch errors  (and then add proper handling of failing commands if they're okay to fail)

docker/bin/gen-ssl-certs.sh
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env bash
+#
```
edenhill: I recognize this script, proper attribution and licensing is typically a good idea, but in this case it is okay not to.

tests/README.md
```diff
@@ -10,65 +17,47 @@ From top-level directory run:
...
-To run individual integration test `modes` use the following syntax
-
-    python examples/integration_test.py --<test mode> --conf <testconf.json>
-
-For example:
+    ./docker/bin/cluster_down.sh
 
-    python examples/integration_test.py --producer --conf testconf.json
+### Configuration
+Tests are configured with a JSON encoded file referred to as a `testconf` to be provided as the last argument upon test execution.
```
edenhill: referred to as testconf.json

tests/README.md
```diff
@@ -10,65 +17,47 @@ From top-level directory run:
...
 
-To get a list of modes you can run the integration test manually with the `--help` flag
+Advanced users can reference the provided configuration file, [testconf.json](testconf.json), if modification is required.
```
edenhill: this should be the testconf example file, right?

tests/README.md
```diff
@@ -10,65 +17,47 @@ From top-level directory run:
...
+Most developers however should just use the defaults.
```
edenhill: -just

tests/README.md
```diff
@@ -10,65 +17,47 @@ From top-level directory run:
...
 
-    python examples/integration_tests.py --help
+### Running tests
+To run the entire test suite:
```
edenhill: will this run the unit tests as well?

tests/README.md
```diff
@@ -10,65 +17,47 @@ From top-level directory run:
...
+- Without tox (will run against current interpreter)
+  - ```./tests/run_all.sh```
 
-    kafka-configs  --zookeeper <zookeeper host>:<zookeeper port> \
-        --alter --add-config 'request_percentage=01' \
-        --entity-name throttled_client --entity-type clients
+To run a specific `mode` or set of `modes` use the following syntax
 
-Once the throttle has been set you can proceed with the following command:
+    ./tests/run.sh [test mode 1] [test mode n...]
```
edenhill: `./tests/run.sh <test-mode-1> <test-mode-2>..`

tests/avro/test_cached_client.py
```diff
@@ -148,3 +148,64 @@ def test_context(self):
...
+
+    def test_init_with_dict(self):
+        self.client = CachedSchemaRegistryClient({
+            'url': 'https://127.0.0.1:65534',
+            'ssl.certificate.location': '/path/to/cert',
+            'ssl.key.location': '/path/to/key'
+        })
+        self.assertTupleEqual(('/path/to/cert', '/path/to/key'), self.client._session.cert)
+
+    def test_emptry_url(self):
```
edenhill: emptry

tests/run.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
```
edenhill: Having tests/run.sh only run the integration tests and not the unit tests is confusing, I think we should stick all integration-test related stuff in tests/integration to avoid this.

tests/testconf.json
```diff
@@ -0,0 +1,17 @@
...
+    "bootstrap.servers": "$MY_BOOTSTRAP_SERVER_ENV",
+    "schema.registry.url": "$MY_SCHEMA_REGISTRY_URL_ENV",
+    "avro-https": {
+            "schema.registry.url": "$MY_SCHEMA_REGISTRY_SSL_URL_ENV",
+            "schema.registry.ssl.ca.location": "$MY_SCHEMA_REGISTRY_SSL_CA_LOCATION_ENV",
+            "schema.registry.ssl.certificate.location": "$MY_SCHEMA_REGISTRY_SSL_CERTIFICATE_LOCATION_ENV",
+            "schema.registry.ssl.key.location": "$MY_SCHEMA_REGISTRY_SSL_KEY_LOCATION_ENV"
+    },
+    "avro-basic-auth": {
+        "schema.registry.url": "http://localhost:8083",
```
edenhill: I don't think we should provide default hosts/ports in the configuration file (especially when no longer an example), since people could be hitting a service unrelated to this test if unlucky.

Commenting them out is fine.

confluent_kafka/avro/__init__.py
rnpridgeon: Yeah I'm trying to come up with some sort of convention since pep8 fails to make one. As you can tell I'm not sure which way to go

docker/bin/certify.sh
rnpridgeon: Primed for parameter expansion. This is inherently unsafe anyway I really should just remove it and add a comment about cleaning up old certs

docker/bin/gen-ssl-certs.sh
rnpridgeon: Yep sorry I  just ran cp $librdkafka_home/tests/gen-ssl-certs.sh. I'll copy/paste the license info from the other librdkafka source files in

tests/testconf.json
rnpridgeon: Since everything is dockerized this is largely irrelevant no?

docker/bin/certify.sh
rnpridgeon: Aren't their portability concerns with `[[ ]]`? 

Seems more prudent to stick with the tried-and-true test command

edenhill: double brackets were added to bash 2.02 back in 1998, I think we can safely assume they're supported.

It is good practice to use them since it removes a lot of corner cases with regards to quoting and escaping.

rnpridgeon:
@edenhill mind giving this another pass

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -38,39 +40,66 @@
...
+
     """
 
     def __init__(self, url, max_schemas_per_subject=1000, ca_location=None, cert_location=None, key_location=None):
-        """Construct a client by passing in the base URL of the schema registry server"""
-
-        self.url = url.rstrip('/')
+        # In order to maintain compatibility the url(conf in future versions) param has been preserved for now.
+        conf = url
+        if not isinstance(url, dict):
```
edenhill: Should also check that url is string, otherwise sub-sequent code will fail (such as startswith()).
E;g;:
```
  if not isinstance(url, dict):
     if not isisntance(url, str):
       raise TypeError...
```

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -38,39 +40,66 @@
...
-            if cert_location is None or key_location is None:
-                raise ValueError(
-                    "Both schema.registry.ssl.certificate.location and schema.registry.ssl.key.location must be set")
-            s.cert = (cert_location, key_location)
+        s = Session()
+        s.verify = conf.pop('ssl.ca.location', None)
+        s.cert = self._configure_client_tls(conf)
+        s.auth = self._configure_basic_auth(conf)
 
+        self.url = conf.pop('url')
```
edenhill: shouldn't this check for error? (notexist)

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -38,39 +40,66 @@
...
         self._session = s
 
+        if len(conf) > 0:
+            raise ValueError("Unrecognized configuration key(s): {}".format(conf.keys()))
```
edenhill: s/keys/properties/

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -94,9 +152,15 @@ def _send_request(self, url, method='GET', body=None, headers={}):
         _headers.update(headers)
 
         response = self._session.request(method, url, headers=_headers, json=body)
+
+        # Returned by Jetty not SR so the payload is not json encoded
```
edenhill: Might be better to wrap the json parsing in a try-block for all 4xx

docker/.env
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+export DOCKER_SOURCE="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
edenhill: a lot of quoting going on here, I'm guessing the inner quotes should be escaped?

docker/bin/certify.sh
```diff
@@ -0,0 +1,29 @@
...
+DOCKER_BIN="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+export PASS="abcdefgh"
+source ${DOCKER_BIN}/../.env
+
+if [[ -f ${TLS}/ca-cert ]]; then
+    echo "${TLS}/ca-cert found; skipping certificate generation.."
+    exit 0
+fi
+
+# Clean up old certs, warning this is
```
edenhill: sentence is cut off, but I think this should be removed anyway since it is commented out

docker/bin/certify.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -eu
+
+DOCKER_BIN="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+export PASS="abcdefgh"
```
edenhill: you don't need to export PASS from what I can see

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,45 @@
...
+source ${DOCKER_BIN}/../.env
+
+# Wait for http service listener to come up and start serving
+# $1 http service name
+# $2 http service address
+await_http() {
+    local exit_code
+    local attempt=0
+
+    until [[ $(curl ${2}) ]] || [[ ${attempt} -gt 5 ]]; do
```
edenhill: I dont think you should call curl in a subshell, that's to be used when collecting output.

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    if [[ ${attempt} -lt 5 ]]; then
+        return
+    fi
+
+    echo "$1 readiness test failed: aborting"
+    exit 1
+}
+
+echo "Configure Environment..."
```
edenhill: future tense

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,45 @@
...
+source ${DOCKER_SOURCE}/.env
+
+echo "Generate SSL certs..."
+${DOCKER_BIN}/certify.sh
+
+echo "Deploying cluster..."
```
edenhill: present tense

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,45 @@
...
+    fi
+
+    echo "$1 readiness test failed: aborting"
+    exit 1
+}
+
+echo "Configure Environment..."
+source ${DOCKER_SOURCE}/.env
+
+echo "Generate SSL certs..."
```
edenhill: future tense

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,45 @@
...
+${DOCKER_BIN}/certify.sh
+
+echo "Deploying cluster..."
+docker-compose -f ${DOCKER_CONTEXT} up -d
+
+echo "Setting throttle for throttle test..."
```
edenhill: present tense

docker/bin/gen-ssl-certs.sh
```diff
@@ -0,0 +1,163 @@
...
+# This scripts generates:
+#  - root CA certificate
+#  - server certificate and keystore
+#  - client keys
+#
+# https://cwiki.apache.org/confluence/display/KAFKA/Deploying+SSL+for+Kafka
+#
+
+DOCKER_BIN="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
edenhill: not used

docker/bin/gen-ssl-certs.sh
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env bash
+set -eu
```
edenhill: these go below the initial comments block.

tests/avro/test_cached_client.py
```diff
@@ -148,3 +148,65 @@ def test_context(self):
...
+            'ssl.certificate.location': '/path/to/cert',
+            'ssl.key.location': '/path/to/key'
+        })
+        self.assertTupleEqual(('/path/to/cert', '/path/to/key'), self.client._session.cert)
+
+    def test_empty_url(self):
+        with self.assertRaises(ValueError):
+            self.client = CachedSchemaRegistryClient({
+                'url': ''
+            })
```
edenhill: add test where url is neither string or dict

tests/integration/integration_test.py
```diff
@@ -952,6 +750,164 @@ def my_on_revoke(consumer, partitions):
...
 
+def verify_avro():
+    base_conf = {'bootstrap.servers': bootstrap_servers,
+                 'error_cb': error_cb,
+                 'api.version.request': api_version_request,
+                 'api.version.fallback.ms': 0,
+                 'broker.version.fallback': '0.11.0.0',
+                 'schema.registry.url': schema_registry_url,
+                 'default.topic.config': {
+                     'produce.offset.report': True
```
edenhill: dont think this is needed anymore. (set by client binding code)

tests/integration/integration_test.py
```diff
@@ -952,6 +750,164 @@ def my_on_revoke(consumer, partitions):
...
 
 
+def verify_avro():
+    base_conf = {'bootstrap.servers': bootstrap_servers,
+                 'error_cb': error_cb,
+                 'api.version.request': api_version_request,
+                 'api.version.fallback.ms': 0,
+                 'broker.version.fallback': '0.11.0.0',
+                 'schema.registry.url': schema_registry_url,
+                 'default.topic.config': {
```
edenhill: default.topic.config will be deprecated

tests/integration/integration_test.py
```diff
@@ -952,6 +750,164 @@ def my_on_revoke(consumer, partitions):
...
+
+
+def verify_avro_https(mode_conf):
+    if mode_conf is None:
+        abort_on_missing_configuration('avro-https')
+
+    base_conf = dict({'bootstrap.servers': bootstrap_servers,
+                      'error_cb': error_cb,
+                      'api.version.request': api_version_request},
+                     **mode_conf)
```
edenhill: verify_avro() did it the other way around: base_conf, **{..}. Is there a reason this is reversed here?

tests/integration/integration_test.py
```diff
@@ -952,6 +750,164 @@ def my_on_revoke(consumer, partitions):
...
+
+    consumer_conf = dict({
+        'group.id': generate_group_id(),
+        'session.timeout.ms': 6000,
+        'enable.auto.commit': False,
+        'api.version.request': api_version_request,
+        'on_commit': print_commit_result,
+        'default.topic.config': {
+            'auto.offset.reset': 'earliest'
+        }}, **base_conf)
```
edenhill: dito

tests/integration/integration_test.py
```diff
@@ -952,6 +750,164 @@ def my_on_revoke(consumer, partitions):
...
+        combo['headers'] = [('index', str(i))]
+        p.produce(**combo)
+    p.flush()
+
+    c = avro.AvroConsumer(consumer_conf)
+    c.subscribe([(t['topic']) for t in combinations])
+
+    msgcount = 0
+    while msgcount < len(combinations):
+        msg = c.poll(0)
```
edenhill: poll(0) seems aggressive, this will busy-loop

tests/integration/integration_test.py
```diff
@@ -1248,41 +1187,59 @@ def resolve_envs(_conf):
...
+    schema_registry_url = testconf.get('schema.registry.url')
 
     if len(modes) == 0:
-        modes = default_modes
+        modes = test_modes
 
     if bootstrap_servers is None or topic is None:
         print_usage(1, "Properties bootstrap.servers and topic must be set. "
-                       "Use tests/testconf-example.json as a template when creating a new conf file.")
+                       "Use {} as a template when creating a new conf file.".format(testconf_file))
```
edenhill: is testconf_file the example/template file?

tests/integration/run.sh
```diff
@@ -0,0 +1,44 @@
...
+
+if [[ ${1:-} == "help" ]]; then
+    python ${TEST_SOURCE}/integration_test.py --help
+    exit 1
+fi
+
+#start cluster
+${DOCKER_BIN}/cluster_up.sh
+
+if [[ $? -ne 0 ]]; then
```
edenhill: ` .. cluster_up.sh || exit 1` 
but this is most likely already the case since you have `set -e`

tests/README.md
```diff
@@ -10,65 +17,47 @@ From top-level directory run:
...
-
-**WARNING**: These tests will create new topics and consumer groups.
+### Requirements
+ 1. docker-compose 3.0 +
+ 2. docker-engine 1.13.0+
+ 3. **Optional:** tox
 
-To run all of the integration test `modes` uncomment the following line from `tox.ini` and provide the location to your `testconf.json`
+### Cluster setup
+**Note** Manual cluster set up is not required when using ./tests/run_all.sh
```
edenhill: can't see run_all.sh committed.

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: We already know it exists 

```
        # Ensure URL valid scheme is included; http[s]
        if not conf.get('url', '').startswith('http'):
            raise ValueError("Invalid URL provided for Schema Registry")
```

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: good catch

docker/.env
rnpridgeon: quotes allow ` dirname "${BASH_SOURCE[0]}` to expand to `.` where necessary. Without them it will resolve to `` which will fail because dirname expects a path  

```
echo $( dirname ${BASH_SOURCE[0]})
usage: dirname path
```

```
echo $( dirname "${BASH_SOURCE[0]}")
.
```

tests/integration/integration_test.py
rnpridgeon: doesn't matter as long as you unpack the second dict. I'll flip it for consistency.

tests/integration/integration_test.py
rnpridgeon: With the use of docker I think it's best to just inform the user to ensure their testconf includes bootstrap.servers

rnpridgeon: i.e it's no longer a template or an example it is a configuration file which you can edit as needed.

tests/integration/run.sh
rnpridgeon: removing outright in favor of `-eu`

rnpridgeon:
@edenhill, mind giving this another go. I have fast-forwarded to rebase the PR on master and addressed all of your comments.

docker/bin/certify.sh
rnpridgeon: gen-ssl-certs.sh uses it, without the export you will get an unbound variable error

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -21,15 +21,23 @@
...
+from requests import Session, utils
 
 from .error import ClientError
 from . import loads
 
+# Python 2 considers int an instance of str
+try:
+    string_types = basestring  # noqa
+except NameError:
+    string_types = str,
```
edenhill: I think it is better to make this tuple construct explicit so the trailing comma doesn't come off as a mistake

confluent_kafka/avro/cached_schema_registry_client.py
edenhill: Any comment on this? Still seems like a valid concern

docker/.env
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
```
edenhill: if the docker stuff is just for tests maybe it should be in tests/docker instead

docker/bin/certify.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash -eu
```
edenhill: The euro version of bash, love it! 🇪🇺

docker/bin/cluster_up.sh
```diff
@@ -0,0 +1,44 @@
...
+source ${DOCKER_BIN}/../.env
+
+# Wait for http service listener to come up and start serving
+# $1 http service name
+# $2 http service address
+await_http() {
+    local exit_code
+    local attempt=0
+
+    until [[ `curl ${2}` ]] || [[ ${attempt} -gt 5 ]]; do
```
edenhill: Are the backticks really needed here?
I think you should do: `until curl ${2} || [[ ${attempt} -gt 5 ]]; do`

docker/conf/schema-registry/login.properties
```diff
@@ -0,0 +1 @@
+ckp_tester:test_secret, Testers
```
edenhill: yeah this stuff should definitely go in tests/docker

tests/README.md
```diff
@@ -1,6 +1,13 @@
+# Conventions Used in This Document
+Unless otherwise noted all commands, file and directory references are relative to the *source root* directory.
+
+## Terminology
+ - modes: Collection of integration tests to be run; a mode(singular) represents an individual integration testing unit
```
edenhill: If a testing unit is a single test, then modes may encompass multiple testing units.

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
 =================
 
-**NOTE**: Integration tests require an existing Kafka cluster and a `testconf.json` configuration file. Any value provided
-in `testconf.json` prefixed with '$' will be treated as an environment variable and automatically resolved.
-
-At a minimum you must specify `bootstrap.servers` and `topic` within `testconf.json`. Please reference [tests/testconf-example.json](tests/testconf-example.json) for formatting.
+### Requirements
+ 1. docker-compose 3.0 +
+ 2. docker-engine 1.13.0+
+ 3. **Optional:** tox
```
edenhill: Maybe say what tox provides, if it is optional

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
 
-**WARNING**: These tests will create new topics and consumer groups.
+### Cluster setup
+**Note** Manual cluster set up is not required when using ./tests/run_all.sh
```
edenhill: run_all.sh was renamed, right?

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
 
-To run all of the integration test `modes` uncomment the following line from `tox.ini` and provide the location to your `testconf.json`
+    ./tests/cluster_up.sh
```
edenhill: these should probably reside in tests/docker/

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
-    #python examples/integration_test.py --conf <testconf.json>
+### Cluster teardown
+**Note** Manual cluster teardown is not required when using ./tests/run_all.sh
 
-You can also run the integration tests outside of `tox` by running this command from the source root directory
+    ./tests/cluster_down.sh
 
-    python examples/integration_test.py --conf <testconf.json>
+### Configuration
+Tests are configured with a JSON encoded file referred to as `testconf.json` to be provided as the last argument upon test execution.
```
edenhill:
```suggestion
Tests are configured with a JSON configuration file referred to as `testconf.json` to be provided as the last argument upon test execution.
```

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
-
-    python examples/integration_test.py --producer --conf testconf.json
+From the source root directory ...
 
-To get a list of modes you can run the integration test manually with the `--help` flag
-
-    python examples/integration_tests.py --help
+- With tox installed (will run against all supported interpreters)
+  1. Uncomment the following line from [tox.ini](../tox.ini)
+    - ```#python examples/integration_test.py```
```
edenhill: This file has moved

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
+  2. Execute the following script
+    - ```./tests/run_all_tox.sh```
```
edenhill: I don't see that file in this PR

tests/README.md
```diff
@@ -10,65 +17,54 @@ From top-level directory run:
...
-[Schema Registry documentation](https://docs.confluent.io/current/schema-registry/docs/security.html#configuring-the-rest-api-for-http-or-https)
+    ./tests/run.sh --producer --consumer
 
-If client authentication has been enabled you will need to provide both the client certificate, `schema.registry.ssl.certificate.location`,
-and the client's private key, `schema.registry.ssl.key.location`
+To get a list of integration test `modes` simply supply the `help` option
 
-    python examples/integration_test.py --avro-https --conf testconf.json
\ No newline at end of file
+    python examples/integration_tests.py --help
```
edenhill: should this be changed to tests/run.sh ?

tests/integration/integration_test.py
```diff
@@ -939,6 +746,157 @@ def my_on_revoke(consumer, partitions):
...
+                 'api.version.fallback.ms': 0,
+                 'broker.version.fallback': '0.11.0.0',
+                 'schema.registry.url': schema_registry_url}
+
+    consumer_conf = dict(base_conf, **{
+        'group.id': 'test.py',
+        'session.timeout.ms': 6000,
+        'enable.auto.commit': False,
+        'on_commit': print_commit_result,
+        'default.topic.config': {
```
edenhill: we should move away from default.topic.config, it will be deprecated in the release this is in (v1.0.0)

tests/integration/integration_test.py
```diff
@@ -939,6 +746,157 @@ def my_on_revoke(consumer, partitions):
...
+        combo['headers'] = [('index', str(i))]
+        p.produce(**combo)
+    p.flush()
+
+    c = avro.AvroConsumer(consumer_conf)
+    c.subscribe([(t['topic']) for t in combinations])
+
+    msgcount = 0
+    while msgcount < len(combinations):
+        msg = c.poll(100)
```
edenhill: timeout can be 1 to allow ctrl-c

tests/integration/integration_test.py
```diff
@@ -939,6 +746,157 @@ def my_on_revoke(consumer, partitions):
...
+
+        if msg is None or msg.error():
+            continue
```
edenhill: I believe it is valuable, for troubleshooting purposes, to print errors, even if we ignore them

tests/integration/integration_test.py
```diff
@@ -1204,20 +1162,17 @@ def verify_config(expconfig, configs):
...
     from confluent_kafka import avro
-    avsc_dir = os.path.join(os.path.dirname(__file__), os.pardir, 'tests', 'avro')
+    avsc_dir = os.path.join(os.path.dirname(__file__), os.pardir, 'avro')
 
     # Producer config
     conf = {'bootstrap.servers': bootstrap_servers,
             'error_cb': error_cb,
-            'api.version.request': api_version_request,
-            'default.topic.config': {'produce.offset.report': True}}
+            'api.version.request': api_version_request}
```
edenhill: we can probably drop all this api.version.request crap by v1.0.0

tests/integration/integration_test.py
```diff
@@ -1341,41 +1281,54 @@ def resolve_envs(_conf):
...
+    schema_registry_url = testconf.get('schema.registry.url')
 
     if len(modes) == 0:
-        modes = default_modes
+        modes = test_modes
 
     if bootstrap_servers is None or topic is None:
-        print_usage(1, "Properties bootstrap.servers and topic must be set. "
-                       "Use tests/testconf-example.json as a template when creating a new conf file.")
+        print_usage(1, "Missing property bootstrap.servers."
```
edenhill: Mentioning bootstraps twice is a bit redundant, and does not mention topic.

tests/run.sh
```diff
@@ -0,0 +1,62 @@
...
+        ${DOCKER_BIN}/cluster_down.sh
+}
+
+trap cleanup 0 2 3 6 15
+
+source ${DOCKER_BIN}/../.env
+
+if [[ ${1:-} == "help" ]]; then
+    python ${TEST_SOURCE}/integration/integration_test.py --help
+    exit 1
```
edenhill: help should exit 0, is not an error

tools/windows-build.bat
```diff
@@ -54,7 +54,7 @@ for %%W in (wheelhouse\confluent_kafka-*cp27*win32.whl) do (
   python -c "from confluent_kafka import libversion ; print libversion()" || exit /b 1
   cd confluent-kafka-python
 
-  pytest --import-mode=append tests || exit /b 1
+  pytest -v --ignore tests/integration --timeout 20 --ignore=tmp-build --import-mode=append tests || exit /b 1
```
edenhill: is there a way to tell pytest to ignore dirs in-situ? e.g., `.pytestignore` files, pytest config,  or something?

tests/README.md
rnpridgeon: I think the singular form is unnecessary to begin with so I'll just remove it.

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: Sorry missed this one. I like this idea I did something similar in the CCloud CLI as well.

confluent_kafka/avro/cached_schema_registry_client.py
rnpridgeon: That's not actually intended to be a tuple. This should be singular as well.

examples/serialization/avro/__main__.py
```diff
@@ -0,0 +1,156 @@
...
+def main():
+    # To use the cluster execute <source root>/tests/docker/bin/cluster_up.sh.
+    # Defaults assume the use of the provided test cluster.
+    parser = argparse.ArgumentParser(description="Example client for handling Avro data")
+    parser.add_argument('-b', metavar="<brokers,...>", dest="bootstrap_servers",
+                        default="localhost:29092", help="Bootstrap broker(s) (host[:port])")
+    parser.add_argument('-s', metavar="<schema registry>", dest="schema_registry",
+                        default="http://localhost:8083", help="Schema Registry (http(s)://host[:port]")
+    parser.add_argument('-t', metavar="[<topic>]", dest="topic", default="example_avro",
+                        help="topic name")
```
edenhill: Capital first letter

examples/serialization/avro/__main__.py
```diff
@@ -0,0 +1,156 @@
...
+    parser.add_argument('-u', metavar="[<userinfo>]", dest="userinfo", default="ckp_tester:test_secret",
+                        help="userinfo (username:password); requires Schema Registry with HTTP basic auth enabled")
```
edenhill: dito

examples/serialization/avro/__main__.py
```diff
@@ -0,0 +1,156 @@
...
+    parser.add_argument('mode', metavar="mode", choices=['produce', 'consume'],
+                        help="Execution mode (produce | consume)")
+    parser.add_argument('-g', metavar="[<consumer group>]", dest="group", default="example_avro",
+                        help="consumer group; required if running 'consumer' mode")
+    conf = ParserOpts()
+    parser.parse_args(namespace=conf)
+
+    getattr(sys.modules[__name__], conf.mode)(conf)
```
edenhill: This will just confuse people, use an if-statement

examples/serialization/avro/__main__.py
```diff
@@ -0,0 +1,156 @@
+import sys
```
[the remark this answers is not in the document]
edenhill: missing shebang

examples/setup.py
```diff
@@ -0,0 +1,15 @@
+from setuptools import setup
```
edenhill: why do we want to make the avro examples an installable package?

rnpridgeon:
@edenhill docs build fine although the CachedSchemaRegistryClient appears to be omitted by autodoc. This has actually always been the case though so I'd like to request that we move that into another PR

examples/serialization/avro/avro-cli.py
```diff
@@ -0,0 +1,180 @@
...
+    topic = args.topic
+    conf = args.producer_conf()
+
+    producer = AvroProducer(conf, default_value_schema=record.schema())
+
+    print("producing user records to topic {}. ^c to exit.".format(topic))
+
+    while True:
+        try:
+            record.prompt()
```
edenhill: Suggest instantiating a new User object for each loop, reusing the same object for multiple Users seems weird to me and has problems if the object is passed to delivery report.

examples/serialization/avro/avro-cli.py
```diff
@@ -0,0 +1,180 @@
...
+            producer.produce(topic=topic, partition=0, value=record, callback=on_delivery)
+            producer.poll(0.1)
+        except KeyboardInterrupt:
+            break
+        except ValueError:
+            print("Invalid input,  discarding record...")
```
edenhill: remove duplicate whitespace

examples/serialization/avro/avro-cli.py
```diff
@@ -0,0 +1,180 @@
...
+    c = AvroConsumer(conf)
+    c.subscribe([topic])
+
+    while True:
+        try:
+            msg = c.poll(1)
+
+        except SerializerError as e:
+            print("Message deserialization failed {}".format(e))
+            break
```
edenhill: Should this be a fatal error? I suggest continue here to not die on invalid input

examples/serialization/avro/avro-cli.py
```diff
@@ -0,0 +1,180 @@
...
+
+        if msg is None:
+            continue
+
+        if msg.error():
+            if msg.error().code() == KafkaError._PARTITION_EOF:
+                break
+            else:
+                print(msg.error())
```
edenhill: The recommendation is not to exit on consumer errors, they are most likely temporary

examples/serialization/avro/avro-cli.py
```diff
@@ -0,0 +1,180 @@
...
+                break
+
+        record = User(msg.value())
+        print("\nusername: {}\n\tfavorite_number:{}\n\tfavorite_color:{}".format(
+              record["name"], record["favorite_number"], record["favorite_color"]))
```
edenhill: Using a dict-type for User sort of defeats the purpose of having an explicit type, since dict can represent any complex type.
Suggest using a custom class with specific fields for name, favourite_nu.. and so on

examples/serialization/avro/avro-cli.py
```diff
@@ -83,75 +75,90 @@ def on_delivery(err, msg):
...
-    topic = args.topic
-    conf = args.consumer_conf()
-    print("consuming user records from topic {}".format(topic))
+    print("consuming user records from topic {} with group {}".format(topic, conf["group.id"]))
 
-    c = AvroConsumer(conf)
-    c.subscribe([topic])
+    # Configure avro schema when the reader schema associated with our UserRecord
+    c = AvroConsumer(conf, reader_value_schema=UserRecord.schema())
+    c.assign([TopicPartition(topic, 0, 0)])
```
edenhill: I think subscribe() is a better example, most people would not use assign()

examples/serialization/avro/avro-cli.py
```diff
@@ -83,75 +75,90 @@ def on_delivery(err, msg):
...
                 break
             else:
                 print(msg.error())
-                break
+                continue
 
-        record = User(msg.value())
-        print("\nusername: {}\n\tfavorite_number:{}\n\tfavorite_color:{}".format(
-              record["name"], record["favorite_number"], record["favorite_color"]))
+        print(UserRecord(**msg.value()))
```
edenhill: For clarity I suggest instantiating the UserRecord first, and then printing it.
Also, it would look better if we could get rid of the `**`

edenhill: And I would still look to see individual UserRecord fields being printed to show case that this is actually a usable object after deserialization, relying on __str__ doesn't really say much.

examples/serialization/avro/avro-cli.py
```diff
@@ -19,61 +19,53 @@
 
 from six.moves import input
 
-from confluent_kafka import avro as avro
+from confluent_kafka import avro, TopicPartition
 
-record_schema_str = """
+
+# UserRecord represnts a user type which can be encoded and decoded as Avro.
+class UserRecord(dict):
```
edenhill: Suggest not using a dict since it is too forgiving (a dict matches any complex avro type).

examples/serialization/avro/avro-cli.py
```diff
@@ -19,61 +19,53 @@
 
 from six.moves import input
 
-from confluent_kafka import avro as avro
+from confluent_kafka import avro, TopicPartition
 
-record_schema_str = """
+
+# UserRecord represnts a user type which can be encoded and decoded as Avro.
```
edenhill: represents

examples/serialization/avro/avro-cli.py
rnpridgeon: Right I guess I should stop fighting the fact that the python avro implementation doesn't support `specific` serialization/deserialization

examples/avro-cli.py
```diff
@@ -0,0 +1,167 @@
...
+    print("producing user records to topic {}. ^c to exit.".format(topic))
+    while True:
+        # Instantiate new User, populate fields, produce record, execute callbacks.
+        record = User()
+        try:
+            record.name = input("Enter name: ")
+            record.favorite_number = int(input("Enter favorite number: "))
+            record.favorite_color = input("Enter favorite color: ")
+
+            producer.produce(topic=topic, partition=0, value=record.to_dict(), callback=on_delivery)
```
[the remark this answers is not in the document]
rnpridgeon: ack, leftover from usage of assign

examples/avro-cli.py
```diff
@@ -0,0 +1,167 @@
...
+                continue
+
+            record = User(msg.value())
+            print("name: {}\n\tfavorite_number: {}\n\tfavorite_color: {}\n".
+                  format(record.name,
+                         record.favorite_number,
+                         record.favorite_color))
+        except SerializerError as e:
+            # Report malformed record, discard results, continue polling
+            print("Message deserialization failed {}".format(e))
```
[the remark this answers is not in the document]
rnpridgeon: The message object isn't returned to the user of the AvroConsumer so I don't believe we will have access to the Message members

examples/avro-cli.py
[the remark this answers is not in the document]
rnpridgeon: we could use this opportunity to demonstrate how you would do externalized custom partitioning.

examples/avro-cli.py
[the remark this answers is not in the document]
edenhill: Okay, then the SerializerError exception needs to contain that information.
Otherwise it will be very hard to troubleshoot bad messages on a topic.

examples/avro-cli.py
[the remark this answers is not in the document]
edenhill: But the python client does not support pluggable partitioners, and it doesn't seem relevant for show-casing avro.

examples/avro-cli.py
```diff
@@ -0,0 +1,167 @@
...
+    while True:
+        # Instantiate new User, populate fields, produce record, execute callbacks.
+        record = User()
+        try:
+            record.name = input("Enter name: ")
+            record.favorite_number = int(input("Enter favorite number: "))
+            record.favorite_color = input("Enter favorite color: ")
+
+            producer.produce(topic=topic, partition=0, value=record.to_dict(), callback=on_delivery)
+            producer.poll(0)
```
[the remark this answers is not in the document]
rnpridgeon: ack, although this seems a bit redundant given our documentation covers this for the produce method.

examples/avro-cli.py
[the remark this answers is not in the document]
rnpridgeon: ack

confluent_kafka/avro/__init__.py
```diff
@@ -138,13 +139,19 @@ def poll(self, timeout=None):
...
-                message.set_key(decoded_key)
+            try:
+                if message.value() is not None:
+                    decoded_value = self._serializer.decode_message(message.value(), is_key=False)
+                    message.set_value(decoded_value)
+                if message.key() is not None:
+                    decoded_key = self._serializer.decode_message(message.key(), is_key=True)
+                    message.set_key(decoded_key)
+            except SerializerError as e:
+                raise SerializerError("Message deserialization failed for message at {} [{}] offset {}: {}".format(
```
edenhill: 👍

examples/avro-cli.py
```diff
@@ -36,19 +37,37 @@
 """)
 
 
-# User stores the deserialized user Avro record.
+"""
+    User stores the deserialized user Avro record.
```
edenhill: docstrings for classes go inside the class, no?

examples/avro-cli.py
```diff
@@ -36,19 +37,37 @@
...
+"""
+
+
 class User(object):
 
-    # Explicitly declare data members to prevent creation of dynamic properties
-    __slots__ = ["name", "favorite_number", "favorite_color"]
+    """
```
edenhill: I believe this will be interpreted as the docstring for the class, not for `__slots__`.
A comment is probably sufficient here.

examples/avro-cli.py
```diff
@@ -36,19 +37,37 @@
...
+        Slots are both more performant and memory efficient than the standard class __dict__.
+        This has the added benefit of adding stronger type safety by disabling dynamic attribute creation.
+        If the deserialized Avro object does not match the User class definition exactly an attribute error
```
edenhill: Is that really correct? The deserialized Avro object (dict) only needs to be a perfect subset of the User object, right? I.e., there could be additional members to User that are not reflected in the schema?

examples/avro-cli.py
```diff
@@ -36,19 +37,37 @@
...
+        will be raised.
+    """
+    __slots__ = ["name", "favorite_number", "favorite_color", "correlation_id"]
 
     def __init__(self, name=None, favorite_number=None, favorite_color=None):
         self.name = name
         self.favorite_number = favorite_number
         self.favorite_color = favorite_color
+        """
```
edenhill: I don't think you can a doc string for an arbitrary code line, you'd use comments for that.

examples/avro-cli.py
```diff
@@ -36,19 +37,37 @@
...
+            The coorelation_id is used to track requests, not users.
+            This comes in handy when investigating failures reported in the delivery report.
+            Do *not* include in the serialized object.
+        """
+        self.correlation_id = uuid4()
+
+    """
+        The Avro Python library does not support code generation.
```
edenhill: again, docstrings go inside what it is documenting, in this case the method.

examples/avro-cli.py
```diff
@@ -57,21 +76,34 @@ def to_dict(self):
         }
 
 
-# Handle delivery reports served from producer.poll.
-def on_delivery(err, msg):
+"""
+    Handle delivery reports served from producer.poll.
```
edenhill: dito

examples/avro-cli.py
```diff
@@ -57,21 +76,34 @@ def to_dict(self):
...
+"""
+
+
+def on_delivery(err, msg, request):
     if err is not None:
-        print('Message delivery failed {}'.format(err))
+        print('Message {} delivery failed for user {} with error {}'.format(
+                                                                            request.correlation_id,
```
edenhill: I would use just `id`, since `correlation_id` might be confused with the Kafka protocol correlation id (and there isn't much correlation needed here since the object is passed.

examples/avro-cli.py
```diff
@@ -57,21 +76,34 @@ def to_dict(self):
...
 
 
-# Handle delivery reports served from producer.poll.
-def on_delivery(err, msg):
+"""
+    Handle delivery reports served from producer.poll.
+"""
+
+
+def on_delivery(err, msg, request):
```
edenhill: suggest renaming `request` to `user` or `obj`, since that is what it is, it is not a request.

examples/avro-cli.py
```diff
@@ -92,11 +128,16 @@ def produce(topic, conf):
     producer.flush()
 
 
+"""
+    produce User records to the configured topic
```
edenhill: this docstring looks malplaced

examples/avro-cli.py
rnpridgeon: In my brain I was drawing a correlation between the response to my produce request with that object. Although I do pass in the object itself the response type be it error or success is tied to the request. So to me the id represents the relationship between those two. Flawed logic yes but I figured I'd explain myself anyway.

examples/avro-cli.py
rnpridgeon: can't or won't, flake8 seemed to be okay with it and a few places mention using ''' for block comments but I agree. fixed

examples/avro-cli.py
```diff
@@ -0,0 +1,194 @@
...
+        Handle delivery reports served from producer.poll.
+        This callback takes an extra argument, obj.
+        This allows the original contents to be included for debugging purposes.
+    """
+    if err is not None:
+        print('Message {} delivery failed for user {} with error {}'.format(
+                                                                            obj.id,
+                                                                            obj.name, err))
+    else:
+        print('Message {} successfully produced to {} [{}] at offset [{}]'.format(
```
edenhill: No brackets around offset.
librdkafka logs are standardized to emit topic and partition as `topic [part]` since they're effectively a single unit, but not offsets within brackets.

examples/avro-cli.py
```diff
@@ -0,0 +1,194 @@
...
+
+def on_delivery(err, msg, obj):
+    """
+        Handle delivery reports served from producer.poll.
+        This callback takes an extra argument, obj.
+        This allows the original contents to be included for debugging purposes.
+    """
+    if err is not None:
+        print('Message {} delivery failed for user {} with error {}'.format(
+                                                                            obj.id,
```
edenhill: micro nit:
this indent-style is kind of weird, continuations are typically current_indent + 1 tab, otherwise
there isn't much horizontal space saved by the continuation.

I guess this is a vim thing?

edenhill:
LGTM!

edenhill:
Might want to consider fixing the history of recent commit-review-ping-pong commits.

ADDale:
I am still getting the below error. please confirm if this has been fixed and can you share working exmaple for the same.
Running MacOS, Python3.6.5
Installed packages with
pip install confluent-kafka
pip install confluent-kafka[avro]

run the example avro producer script from https://github.com/confluentinc/confluent-kafka-python/blob/master/README.md having edited for my broker, schema registry and topic name.

error stack is :
File "/Users/07HO38/Documents/Python/OEHCS/kafka_produce_avro.py", line 43, in 
avroProducer.produce(topic='idcs-5bea815f69d54c6daa5701a57b5914fa-JDBCUSERS', value=value, key=key)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/init.py", line 80, in produce
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/serializer/message_serializer.py", line 98, in encode_record_with_schema
schema_id = self.registry_client.register(subject, schema)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/cached_schema_registry_client.py", line 141, in register
result, code = self._send_request(url, method='POST', body=body)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/cached_schema_registry_client.py", line 97, in _send_request
return response.json(), response.status_code
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/requests/models.py", line 892, in json
return complexjson.loads(self.text, **kwargs)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/init.py", line 354, in loads
return _default_decoder.decode(s)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 339, in decode
obj, end = self.raw_decode(s, idx=_w(s, 0).end())
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 357, in raw_decode
raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

ADDale:
I am still getting the error. Can you please share the working code for the same?
run the example avro producer script from https://github.com/confluentinc/confluent-kafka-python/blob/master/README.md having edited for my broker, schema registry and topic name.

error stack is :
File "/Users/07HO38/Documents/Python/OEHCS/kafka_produce_avro.py", line 43, in 
avroProducer.produce(topic='idcs-5bea815f69d54c6daa5701a57b5914fa-JDBCUSERS', value=value, key=key)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/init.py", line 80, in produce
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/serializer/message_serializer.py", line 98, in encode_record_with_schema
schema_id = self.registry_client.register(subject, schema)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/cached_schema_registry_client.py", line 141, in register
result, code = self._send_request(url, method='POST', body=body)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/confluent_kafka/avro/cached_schema_registry_client.py", line 97, in _send_request
return response.json(), response.status_code
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/requests/models.py", line 892, in json
return complexjson.loads(self.text, **kwargs)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/init.py", line 354, in loads
return _default_decoder.decode(s)
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 339, in decode
obj, end = self.raw_decode(s, idx=_w(s, 0).end())
File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/json/decoder.py", line 357, in raw_decode
raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

Issue #443, opened 2018-08-31
get_compatibility does not work for me


charlie-white:
Description
===========
When I call the method get_compatibility of the CachedSchemaRegistryClient class I get an exception.

```
Traceback (most recent call last):
  File "utils/show_compatibilty.py", line 42, in <module>
    compatibility = schema_registry.get_compatibility()
  File "/Users/brgl/code/schema/env/lib/python2.7/site-packages/confluent_kafka/avro/cached_schema_registry_client.py", line 335, in get_compatibility
    raise ClientError('Invalid compatibility level received: %s' % error_msg_suffix)
confluent_kafka.avro.error.ClientError: Invalid compatibility level received: No compatibility was returned

```
I looked at the code, and on line 326 of cached_schema_registry_client.py I see the following line.
`compatibility` = result.get('compatibility', None)`

If I change the line to the following it works for me.
`compatibility = result.get('compatibilityLevel', None)`

While I am at it, in the same file on line 306:
            `raise ClientError("Unable to update level: %s. Error code: %d" % (str(level)), code)`
should read:
            `raise ClientError("Unable to update level: %s. Error code: %d" % (str(level), code))`
Paren in the wrong spot.

How to reproduce
================

```
schema_registry = CachedSchemaRegistryClient(SCHEMA_REGISTRY_URL)
compatibility = schema_registry.get_compatibility()
```

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version:
confluent_kafka.version()
('0.11.5', 722176)
confluent_kafka.libversion()
('0.11.5', 722431)

 - [x] Apache Kafka broker version:
$ confluent version
Confluent Open Source: 4.1.1
$ confluent version kafka
1.1.1-cp1

 - [x] Operating system: macOS - Sierra 10.12.6

rnpridgeon:
Added integration tests and fix to #443. Thanks for you contribution

PR #444, opened 2018-09-04
Fix for issue #443.

charlie-white:
Fixing two small coding issues in update_compatibility and get_compatibility methods.
No acccompanying tests as the mock object is incomplete and I didn't have time to dive in deep enough to fix that.

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -303,7 +303,7 @@ class CachedSchemaRegistryClient(object):
         if code >= 200 and code <= 299:
             return result['compatibility']
         else:
-            raise ClientError("Unable to update level: %s. Error code: %d" % (str(level)), code)
+            raise ClientError("Unable to update level: %s. Error code: %d" % (str(level), code))
 
     def get_compatibility(self, subject=None):
         """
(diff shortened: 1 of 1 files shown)


edenhill:
[clabot:check]

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -323,7 +323,7 @@ def get_compatibility(self, subject=None):
         if not is_successful_request:
             raise ClientError('Unable to fetch compatibility level. Error code: %d' % code)
 
-        compatibility = result.get('compatibility', None)
+        compatibility = result.get('compatibilityLevel', None)
```
edenhill: this seems to contradict the API docs:
https://docs.confluent.io/current/schema-registry/docs/api.html#put--config

code:
https://github.com/confluentinc/schema-registry/blob/8831bbd73ea720fcd279032a5fffabbdb1a1f2b1/client/src/main/java/io/confluent/kafka/schemaregistry/client/rest/entities/Config.java#L25

rnpridgeon: I have opened a docs jira for this. This absolutely the case I ran into the same issue implementing the SR client in golang. 

I had to handle it as follows: 

```
/* NOTE: GET uses compatibilityLevel, POST uses compatibility */
type compatibilityLevel struct {
	CompatibilityUpdate Compatibility `json:"compatibility,omitempty"`
	Compatibility       Compatibility `json:"compatibilityLevel,omitempty"`
}
```

PR #451, opened 2018-09-19
Support for building Windows wheels

edenhill:


edenhill:
I think we'll want to publish Windows wheels for the next maintenance release, but perhaps mark them as experimental for the first release.

confluent_kafka/src/Admin.c
```diff
@@ -881,7 +891,7 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
          * Parse the list of ConfigResources and convert to
          * corresponding C types.
          */
-        c_objs = alloca(sizeof(*c_objs) * cnt);
+        c_objs = malloc(sizeof(*c_objs) * cnt);
```
rnpridgeon: What if any implications are there to allocating on the heap as opposed to the stack for these Admin operations

rnpridgeon: *aside from needing to add free

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1444,7 +1444,7 @@ static int producer_conf_set_special (Handle *self, rd_kafka_conf_t *conf,
 				      rd_kafka_topic_conf_t *tconf,
 				      const char *name, PyObject *valobj) {
 
-	if (!strcasecmp(name, "on_delivery")) {
+	if (!strcmp(name, "on_delivery")) {
```
rnpridgeon: Is there an `strcasecmp` equivalent for windows ? Otherwise `tolower()` or some equivalent  seem appropriate prior to comparison. While I agree function arguments should be case-sensitive they haven't been in the past so doesn't this technically break the api?

edenhill: There are Windows alternatives, but since python is case-sensitive, and we haven't made any claims that on_delivery, et.al, may be specified in non lower-case, I think it is safe to remove this "feature".

rnpridgeon: If you're good with it I'm good with it

edenhill: Then I'm good with it

PR #458, opened 2018-09-28
Fix memory leak in msg.headers() 

edenhill:
**Moved interceptor inclusion work to own branch**

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -1001,6 +1001,7 @@ static int py_header_to_c (rd_kafka_headers_t *rd_headers,
                         if (cfl_PyBin(_AsStringAndSize(value, (char **)&v,
                                                        &vsize)) == -1) {
                                 Py_DECREF(ks);
+                                Py_XDECREF(ks8);
                                 return 0;
                         }
                 } else if (cfl_PyUnistr(_Check(value))) {
@@ -1008,6 +1009,7 @@ static int py_header_to_c (rd_kafka_headers_t *rd_headers,
                         v = cfl_PyUnistr_AsUTF8(value, &vo8);
                         if (!v) {
                                 Py_DECREF(ks);
+                                Py_XDECREF(ks8);
                                 return 0;
                         }
                         vsize = (Py_ssize_t)strlen(v);
@@ -1017,6 +1019,8 @@ static int py_header_to_c (rd_kafka_headers_t *rd_headers,
                                      "None, binary, or unicode string, not %s",
                                      ((PyTypeObject *)PyObject_Type(value))->
                                      tp_name);
+                        Py_DECREF(ks);
+                        Py_XDECREF(ks8);
                         return 0;
                 }
         }
@@ -1027,11 +1031,13 @@ static int py_header_to_c (rd_kafka_headers_t *rd_headers,
                                  "%s",
                                  k, rd_kafka_err2str(err));
                 Py_DECREF(ks);
+                Py_XDECREF(ks8);
                 Py_XDECREF(vo8);
                 return 0;
         }
 
         Py_DECREF(ks);
+        Py_XDECREF(ks8);
         Py_XDECREF(vo8);
 
         return 1;


tools/install-interceptors.sh
```diff
@@ -0,0 +1,111 @@
...
+# Usage: $0 tools/install-interceptors.sh
+#
+
+set -e
+
+# pkgtype (rpm, deb, osx), will be empty on outer invocation.
+pkgtype=$1
+
+# Confluent Platform release version
+CPVER=5.0
```
rnpridgeon: Add this file to tools/RELEASE.md

edenhill:
Will merge as soon as these packages are verified.

rnpridgeon:
Added an additional step to validate interceptor packaging in the wheels. I did not add this as an `after_*` lifecycle as that would not fail the overall build job. Locally python was unable to locate the interceptor in its desired location using the `v0.11.6-test6`.

.travis.yml
```diff
@@ -66,7 +66,8 @@ script:
  - if [[ -z $CIBW_BEFORE_BUILD ]]; then py.test -v --timeout 20 --ignore=tmp-build --import-mode append ; fi
  - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then cibuildwheel --output-dir wheelhouse1 && tools/fixup-wheels.sh wheelhouse1 wheelhouse ; fi
  - if [[ -n $TRAVIS_TAG && $TRAVIS_OS_NAME == linux && -n $CIBW_BEFORE_BUILD ]]; then tools/test-manylinux.sh ; fi
-
+ - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then pip install --no-index -f wheelhouse/ confluent-kafka;\
+		 py.test -v --timeout 20 --ignore=tmp-build --import-mode append; tests/test_misc.py::test_unordered_dict; fi
```
edenhill: `... append; tests/test_misc...`

Should there really be a ; in there?

Also, the tab indent is very red in the diff

edenhill: You should probably use `&&` instead of `;` between pip install and py.test.
Another thing to note is that it is error prone to run the tests from within the confluent-kafka-python source tree since `import confluent_kafka` may pick up the local directory, better to cd out to the parent directory.

.travis.yml
```diff
@@ -66,8 +66,7 @@ script:
  - if [[ -z $CIBW_BEFORE_BUILD ]]; then py.test -v --timeout 20 --ignore=tmp-build --import-mode append ; fi
  - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then cibuildwheel --output-dir wheelhouse1 && tools/fixup-wheels.sh wheelhouse1 wheelhouse ; fi
  - if [[ -n $TRAVIS_TAG && $TRAVIS_OS_NAME == linux && -n $CIBW_BEFORE_BUILD ]]; then tools/test-manylinux.sh ; fi
- - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then pip install --no-index -f wheelhouse/ confluent-kafka;\
-		 py.test -v --timeout 20 --ignore=tmp-build --import-mode append; tests/test_misc.py::test_unordered_dict; fi 
+ - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then pip install --no-index -f wheelhouse/ confluent-kafka && cd .. && pytest -v --timeout 20 --ignore=tmp-build --import-mode append confluent-kafka-python/tests/test_misc.py::test_unordered_dict; fi
```
edenhill: Put the cd and the command in a subshell, or cd .. back out, otherwise any sub-sequent commands will fail because the directory changed:

` - if [[ -n $TRAVIS_TAG && -n $CIBW_BEFORE_BUILD ]]; then pip install --no-index -f wheelhouse/ confluent-kafka && (cd .. && pytest -v --timeout 20 --ignore=tmp-build --import-mode append confluent-kafka-python/tests/test_misc.py::test_unordered_dict); fi`

rnpridgeon: True, I guess there are no guarantees this will always be the last script executed

edenhill: Right, it will help the next poor chap that adds a command here.

PR #470, opened 2018-10-12
avro consumer with explicitly specified read schema

fkaufer:
Allow to configure reader schema explicitly in Avro Consumer to take advantage of Avro's back-/forward compatibility features by separation of writer and reader schema. Before the change, the reader schema was (implicitly) hard-coded to the writer schema (of the respective message), now the consumer can pin a reader schema with the fastavro deserializer doing back or forward schema migration to the reader schema.


tests/avro/read_test_schema.avsc
```diff
@@ -0,0 +1,7 @@
+{
+ "type": "record",
+ "name": "UserKey",
```
rnpridgeon: I believe the two schema must have the same record name in order to satisfy the schema resolution rules laid out in the Avro documentation. 

http://avro.apache.org/docs/current/spec.html#Schema+Resolution

```
It is an error if the two schemas do not match.
...
both schemas are records with the same name
```

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -169,6 +170,7 @@ def _get_decoder_func(self, schema_id, payload):
             # try to use fast avro
             try:
                 schema_dict = schema.to_json()
+                reader_schema_dict = schema.to_json()
```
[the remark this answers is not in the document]
fkaufer: fixed 10637a9

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -177,7 +179,8 @@ def _get_decoder_func(self, schema_id, payload):
                 # normal path.
                 payload.seek(curr_pos)
 
-                self.id_to_decoder_func[schema_id] = lambda p: schemaless_reader(p, schema_dict)
+                self.id_to_decoder_func[schema_id] = lambda p: schemaless_reader(p, schema_dict,
```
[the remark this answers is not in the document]
fkaufer: `reader_schema` is the only keyword argument, the others are positional (see https://fastavro.readthedocs.io/en/latest/reader.html?highlight=schemaless_reader#fastavro._read_py.schemaless_reader), but fair enough - changed and by that added improved bit more naming consistency: https://github.com/confluentinc/confluent-kafka-python/pull/470/commits/1f170fb2229b078568de5337c6d16fbbdc995459

confluent_kafka/avro/__init__.py
```diff
@@ -98,9 +98,10 @@ class AvroConsumer(Consumer):
     Constructor takes below parameters
 
     :param dict config: Config parameters containing url for schema registry (``schema.registry.url``)
-                        and the standard Kafka client configuration (``bootstrap.servers`` et.al).
+                        and the standard Kafka client configuration (``bootstrap.servers`` et.al)
+    :param optional a read schema for the messages
     """
-    def __init__(self, config, schema_registry=None):
+    def __init__(self, config, schema_registry=None, read_schema=None):
```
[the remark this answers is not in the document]
fkaufer: indeed, added separate key/value read schema arguments: https://github.com/confluentinc/confluent-kafka-python/pull/470/commits/2acbf4d6a4d8ac83439e8306f9dd07d351d45614

tests/avro/read_test_schema.avsc
fkaufer: Thanks, makes sense. Naming it `UserKey` was anyway not optimal. However, I fixed it by means of another Avro schema evolution feature: aliases (e902f2f)

rnpridgeon:
Looks like they dropped the `s` in the  py3 variant of the avro library. Which I actually like more but that's sort of a cruel thing to do. 

https://github.com/apache/avro/blob/master/lang/py3/avro/io.py#L459

fkaufer:
@rnpridgeon seems we are good to go. Can you approve and merge?

flxfbrx:
@rnpridgeon anything missing?
And as a reminder, you can close #427 which is obsolete with this PR.

fkaufer:
@rnpridgeon anything missing?
And as a reminder, you can close #427 which is obsolete with this PR.

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -165,19 +167,24 @@ def _get_decoder_func(self, schema_id, payload):
...
         curr_pos = payload.tell()
+
+        reader_schema_obj = self.reader_key_schema if is_key else self.reader_value_schema
+
         if HAS_FAST:
             # try to use fast avro
             try:
-                schema_dict = schema.to_json()
-                schemaless_reader(payload, schema_dict)
+                writer_schema = schema.to_json()
```
rnpridgeon: To keep naming consistent we should replace schema(declared a few lines up on 162) with `writer_schema_obj`

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -186,7 +193,13 @@ def _get_decoder_func(self, schema_id, payload):
         # here means we should just delegate to slow avro
         # rewind
         payload.seek(curr_pos)
-        avro_reader = avro.io.DatumReader(schema)
+        # Avro DatumReader py2/py3 inconsistency, hence no param keywords
```
rnpridgeon: 👍

confluent_kafka/avro/__init__.py
```diff
@@ -98,9 +98,11 @@ class AvroConsumer(Consumer):
     Constructor takes below parameters
 
     :param dict config: Config parameters containing url for schema registry (``schema.registry.url``)
-                        and the standard Kafka client configuration (``bootstrap.servers`` et.al).
+                        and the standard Kafka client configuration (``bootstrap.servers`` et.al)
+    :param optional a reader schema for the message key
+    :param optional a reader schema for the message value
     """
-    def __init__(self, config, schema_registry=None):
+    def __init__(self, config, schema_registry=None, reader_key_schema=None, reader_value_schema=None):
```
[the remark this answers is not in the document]
flxfbrx: I had it first that way (or `key_schema_reader` considering `reader` as var index), but this would be inconsistent with `default_key_schema` and `default_value_schema` in the constructor: 

https://github.com/confluentinc/confluent-kafka-python/blob/3b69697a7500e3dcf888c801dfa7d4edf8ae6c0a/confluent_kafka/avro/__init__.py#L29

Furthermore I consider `key_schema`/`value_schema` compounds naming an attribute/member of the object/namespace `reader`

This is also more consistent with `fastavro` (`writer_schema`, `reader_schema`)
https://fastavro.readthedocs.io/en/latest/reader.html#fastavro._read_py.schemaless_reader

and the REST Proxy (`key_schema` and `value_schema`)
https://docs.confluent.io/current/kafka-rest/docs/api.html#post--topics-(string-topic_name)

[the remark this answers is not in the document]
fkaufer: I had it first that way (or key_schema_reader considering reader as var index), but this would be inconsistent with default_key_schema and default_value_schema in the constructor:

See https://github.com/confluentinc/confluent-kafka-python/blob/443177e1c83d9b66ce30f5eb8775e062453a738b/confluent_kafka/avro/__init__.py#L29-L30

Furthermore I consider `key_schema`/`value_schema` compounds naming an attribute/member of the object/namespace reader

This is also more consistent with fastavro (`writer_schema`, `reader_schema`)
https://fastavro.readthedocs.io/en/latest/reader.html#fastavro._read_py.schemaless_reader

and the REST Proxy (`key_schema` and `value_schema`)
https://docs.confluent.io/current/kafka-rest/docs/api.html#post--topics-(string-topic_name)

confluent_kafka/avro/__init__.py
```diff
@@ -119,7 +121,7 @@ def __init__(self, config, schema_registry=None):
             raise ValueError("Cannot pass schema_registry along with schema.registry.url config")
 
         super(AvroConsumer, self).__init__(config)
-        self._serializer = MessageSerializer(schema_registry)
+        self._serializer = MessageSerializer(schema_registry, reader_key_schema, reader_value_schema)
```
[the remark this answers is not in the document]
fkaufer: see above

confluent_kafka/avro/serializer/message_serializer.py
fkaufer: Actually kept it intentionally as `schema` to be consistent with method param `schema_id`, but you are right, changed it: 70453e8
(kept `schema_id` as is though since there is no reader `schema_id` option; could be added later as alternative)

fkaufer:
@rnpridgeon I have closed and reopened the PR to retrigger the CI now that the dependencies seem to be fixed. With success. So, can you merge?

examples/integration_test.py
```diff
@@ -1202,9 +1202,117 @@ def verify_config(expconfig, configs):
...
+                 'default.topic.config': {
+                     'auto.offset.reset': 'earliest'
+                 }}
+
+    for i, combo in enumerate(combinations):
+        reader_key_schema = combo.pop("reader_key_schema")
+        reader_value_schema = combo.pop("reader_value_schema")
+        combo['topic'] = str(uuid.uuid4())
+        p.produce(**combo)
+        p.poll(0)
```
edenhill: no need for poll here since flush is called

examples/integration_test.py
```diff
@@ -1323,6 +1431,10 @@ def resolve_envs(_conf):
         print('=' * 30, 'Verifying Admin API', '=' * 30)
         verify_admin()
 
+    if 'explicit-read' in modes:
```
edenhill: This does not need to be its own mode, call it from verify_avro()

examples/integration_test.py
```diff
@@ -1202,9 +1202,117 @@ def verify_config(expconfig, configs):
     print("Topic {} marked for deletion".format(our_topic))
 
 
-# Exclude throttle since from default list
-default_modes = ['consumer', 'producer', 'avro', 'performance', 'admin']
+def verify_explicit_read():
```
edenhill: verify_avro_explicit_read_schema

call it from verify_avro()

rnpridgeon:
Thanks @fkaufer, I think I'm happy with it the way it is now so I have asked @edenhill to give it a quick secondary review.

examples/integration_test.py
fkaufer: Changed - fcf903d

examples/integration_test.py
fkaufer: ok - c4ff376

examples/integration_test.py
fkaufer: makes sense - c4ff376

fkaufer:
@edenhill @rnpridgeon gents, many thanks for your reviews, I changed as requested, all checks have passed.

rnpridgeon:
Ack, thank you for the secondary review @edenhill

rnpridgeon:
Thanks @fkaufer for your contribution!

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -151,33 +153,38 @@ def encode_record_with_schema_id(self, schema_id, record, is_key=False):
...
+
+        reader_schema_obj = self.reader_key_schema if is_key else self.reader_value_schema
+
         if HAS_FAST:
             # try to use fast avro
             try:
-                schema_dict = schema.to_json()
-                schemaless_reader(payload, schema_dict)
+                writer_schema = writer_schema_obj.to_json()
+                reader_schema = reader_schema_obj.to_json()
```
arthurandres: Sorry I'm a bit late to the party here.

@fkaufer , correct me if I'm wrong, but my understanding is that `self.reader_value_schema` can be null, and so `reader_schema_obj` could be null as well. Wouldn't that line cause an NPE?

Also I wonder if the next line should not be `schemaless_reader(payload, writer_schema, reader_schema)` to be consistent with the code executed by the lambda this function returns.

rnpridgeon: The next line is a sanity check as alluded to in the comment below it. This is in place to catch edges which fastavro did not covet at the time the serializer was written. It's unclear if this is still actually an issue or not but it only happens upon instantiating the writer. 

You won't get an NPE as the `None` type is an actual null but you will get an attribute error which would cause the serializer to fall back to `slow avro` as we call it. If you have already done so please feel free to open an issue which I will address tomorrow.

arthurandres: @rnpridgeon , thanks this is helpful. I had already opened an issue for a related issue: https://github.com/confluentinc/confluent-kafka-python/issues/518 I think both can be addressed at the same time. 

I actually tried to fix it my self, but couldn't get the test to run. I would need to spend more time trying to set up my development environment.

PR #475, opened 2018-10-26
Issue #449: AvroConsumer: provide method to get schemas from last message decode

soxofaan:
This is an alternative take on issue #449 that tries to address some concerns raised in PR #453 :
- overhead is minimal (no mixins, no dynamic wrapper type creation and other magic), so not being able to turn it off is less of a concern
- no non-optional `_schema` attribute on the wrapper that might cause collisions or raise code style complaints for accessing a "private" attribute

On the other hand, this solution is conceptually bit uglier than #453 as the schema (id) is not associated with the returned message directly and has to be requested with an explicit calls.

Very rough usage example:
```python
consumer = AvroConsumer()

# Get a message
msg = consumer.poll()

# Get key/value schema ids and schema objects
key_sid, value_sid = consumer.last_decode_schema_ids
key_schema, value_schema = consumer.last_decode_schemas()
```

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -124,6 +124,7 @@ class AvroConsumer(Consumer):
 
         super(AvroConsumer, self).__init__(ap_conf)
         self._serializer = MessageSerializer(schema_registry, reader_key_schema, reader_value_schema)
+        self.last_decode_schema_ids = (None, None)
 
     def poll(self, timeout=None):
         """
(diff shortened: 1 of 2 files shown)


soxofaan:
Please note: this is proof of concept to collect initial feedback (e.g. from @saabeilin, @rnpridgeon, @edenhill, involved in #453)
I'm planning to finetune documentation a bit and add tests

soxofaan:
[clabot:check]

soxofaan:
rebased branch to resolve merge conflicts

rnpridgeon:
Generic Serde support will provide a means to bind your schema with your data. This is similar to java's 'GenericRecord' implementation. #502

PR #478, opened 2018-11-02
Fixes for confluent-librdkafka-plugins CP5.0.1

edenhill:


rnpridgeon:
LGTM as long as we are good holding off on Windows builds until the next rc or real release

PR #484, opened 2018-11-13
Idempotent Producer support (fatal errors)

edenhill:
also includes a soak test client.

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -62,8 +62,11 @@ typedef struct {
 	char *str;   /* Human readable representation of error, if one
 		      * was provided by librdkafka.
 		      * Else falls back on err2str(). */
+        int   fatal; /**< Set to true if a fatal error. */
 } KafkaError;
 
+
+static void cfl_PyErr_Fatal (rd_kafka_resp_err_t err, const char *reason);
 
 static PyObject *KafkaError_code (KafkaError *self, PyObject *ignore) {
 	return cfl_PyInt_FromInt(self->code);
@@ -79,6 +82,20 @@ static PyObject *KafkaError_str (KafkaError *self, PyObject *ignore) {
 static PyObject *KafkaError_name (KafkaError *self, PyObject *ignore) {
 	/* FIXME: Pre-create name objects */
 	return cfl_PyUnistr(_FromString(rd_kafka_err2name(self->code)));
+}
+
+static PyObject *KafkaError_fatal (KafkaError *self, PyObject *ignore) {
+        PyObject *ret = self->fatal ? Py_True : Py_False;
+        Py_INCREF(ret);
+        return ret;
+}
+
+
+static PyObject *KafkaError_test_raise_fatal (KafkaError *null,
+                                              PyObject *ignore) {
+        cfl_PyErr_Fatal(RD_KAFKA_RESP_ERR__INVALID_ARG,
+                        "This is a fatal exception for testing purposes");
+        return NULL;
 }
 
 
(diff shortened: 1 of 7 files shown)


confluent_kafka/src/confluent_kafka.c
```diff
@@ -105,6 +122,14 @@ static PyMethodDef KafkaError_methods[] = {
 	  "  :rtype: str\n"
 	  "\n"
 	},
+        { "fatal", (PyCFunction)KafkaError_fatal, METH_NOARGS,
+          "  :returns: True if this a fatal error, else False.\n"
+          "  :rtype: bool\n"
+          "\n"
+        },
+        { "_test_raise_fatal", (PyCFunction)KafkaError_test_raise_fatal,
```
mhowlett: does this really provide enough value in testability to warrant messying up the `KafkaError` class?

tests/test_KafkaError.py
```diff
@@ -31,6 +33,22 @@ def test_error_cb():
...
+def test_fatal():
+    """ Test fatal exceptions """
+
+    # Configure an invalid broker and make sure the ALL_BROKERS_DOWN
+    # error is seen in the error callback.
+    p = Producer({'error_cb': error_cb})
+
+    with pytest.raises(KafkaException) as exc:
+        KafkaError._test_raise_fatal()
+    err = exc.value.args[0]
```
mhowlett: is `exc` in scope here? doesn't seem right

confluent_kafka/src/confluent_kafka.c
edenhill: ain't no other way to trigger a fatal error, short of introducing broker-side errors, so this is a must.
underscore prefixed methods are considered internal anyway.

tests/test_KafkaError.py
edenhill: Yeah, I had the same reaction initially, but this is according to docs:
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1177,6 +1216,16 @@ static void error_cb (rd_kafka_t *rk, int err, const char *reason, void *opaque)
 	CallState *cs;
 
 	cs = CallState_get(h);
+
+        /* If the client raised a fatal error we'll raise an exception
+         * rather than calling the error callback. */
+        if (err == RD_KAFKA_RESP_ERR__FATAL) {
```
mhowlett: Interestingly, in C#, I think the behavior is going to be the opposite. produce methods can (and mostly will) throw for non-fatal reasons and I think it will be common for people to handle those in a generic catch-all way and typical to check for fatal errors in the error callback (though testing the exception would also be possible). i think what you have here makes sense for the python api though.

confluent_kafka/src/confluent_kafka.c
edenhill: The fatal() functionality needs a test, which to me is more important than not messying up the class with a non-public and clearly internal method.
This is particularly important in this case where real fatal errors are really hard to trigger.

mhowlett: what we really want is a mock broker, or proxy that can inject problems.

mhowlett: a single underscore means protected by convention which does make it part of the public interface. better to make it __ -> private.

edenhill: I don't believe that's correct:
https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

> _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore.
> ..
> __double_leading_underscore: when naming a class attribute, invokes name mangling (inside class FooBar, __boo becomes _FooBar__boo; see below).

> what we really want is a mock broker, or proxy that can inject problems.
surely, but I feel that might be outside the scope of this PR ;)

PR #446, opened 2018-09-06
deprecate default.topic.configuration

rnpridgeon:

PR #493, opened 2018-11-26
Remove two default.topic.config uses

RasmusWL:
as outlined in 34baea6 (#446)

Having to add the `default.topic.config` dictionary really caught me off guard, so I'm really happy to see that this is going to be removed in 1.0 :heart:


RasmusWL:
[clabot:check]

rnpridgeon:
Thanks for catching this @RasmusWL, this is actually already taken care of in #440  which will hopefully be merged soon.

PR #494, opened 2018-11-26
Record name subject raw keys

bbqbaron:
Two high-level changes, to match what we're doing from Scala now:
* we want to be able to send Kafka messages with string keys, not mediated by Avro at all
* we want schema subjects to be derived from the names of records, not topic name/schema role

(For background: the non-python Kafka libs allow configurable key serialization and schema subject selection, which is why we haven't run into this outside Admiral. The python lib is just a bit more narrow.)

I took the approach of just hardcoding _our_ assumptions, in place of the originals. This has some corollaries:

* primitive values cannot be sent. If it's not a record, we can't get a subject name, so we can't serialize it
* all keys are just strings. If you send a non-string key, no dice; we don't serialize it.

I can certainly see a version of this fork that basically ports over the various JVM lib features that allow us to have it all, but that seemed like a deep well to go down.

The tests pass. I'm puzzling over whether to go through the setup to run the full integration tests, but they're reasonably involved. Thought I'd open this now to get eyes on it!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ dl-*
 *.whl
 .pytest_cache
 staging
+env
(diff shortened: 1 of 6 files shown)


bbqbaron:
Whoops; fork selection went 100% wrong.

PR #498, opened 2018-12-03
V1.0 deprecations

rnpridgeon:
Note two remaining `default.topic.config` are already handled in #440

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,8 @@ while True:
     if msg is None:
         continue
     if msg.error():
-        if msg.error().code() == KafkaError._PARTITION_EOF:
-            continue
-        else:
-            print(msg.error())
-            break
+        print("Consumer error: {}".format(msg.error()))
+        continue
 
     print('Received message: {}'.format(msg.value().decode('utf-8')))
 
@@ -172,11 +169,8 @@ while True:
         continue
 
     if msg.error():
-        if msg.error().code() == KafkaError._PARTITION_EOF:
-            continue
-        else:
-            print(msg.error())
-            break
+        print("AvroConsumer error: {}".format(msg.error()))
+        continue
 
     print(msg.value())
 
--- a/examples/confluent_cloud.py
+++ b/examples/confluent_cloud.py
@@ -65,9 +65,9 @@ p = Producer({
 def acked(err, msg):
     """Delivery report callback called (from flush()) on successful or failed delivery of the message."""
     if err is not None:
-        print("failed to deliver message: {0}".format(err.str()))
+        print("failed to deliver message: {}".format(err.str()))
     else:
-        print("produced to: {0} [{1}] @ {2}".format(msg.topic(), msg.partition(), msg.offset()))
+        print("produced to: {} [{}] @ {}".format(msg.topic(), msg.partition(), msg.offset()))
 
 
 p.produce('python-test-topic', value='python test value', callback=acked)
@@ -101,11 +101,11 @@ try:
             #   the group to rebalance and start consuming.
             continue
         if msg.error():
-            # Most error message are typically temporary log error and continue.
+            # Errors are typically temporary, print error and continue.
             print("Consumer error: {}".format(msg.error()))
             continue
 
-        print('consumed: {0}'.format(msg.value()))
+        print('consumed: {}'.format(msg.value()))
 
 except KeyboardInterrupt:
     pass
--- a/examples/confluent_cloud.py
+++ b/examples/confluent_cloud.py
@@ -48,7 +48,7 @@
 
 import uuid
 
-from confluent_kafka import Producer, Consumer, KafkaError
+from confluent_kafka import Producer, Consumer
 
 
 p = Producer({
@@ -98,14 +98,14 @@ try:
         if msg is None:
             # No message available within timeout.
             # Initial message consumption may take up to `session.timeout.ms` for
-            #   the group to rebalance and start consuming
+            #   the group to rebalance and start consuming.
             continue
-        elif not msg.error():
-            print('consumed: {0}'.format(msg.value()))
-        elif msg.error().code() == KafkaError._PARTITION_EOF:
-            print('end of partition: {0} [{1}] @ {2}'.format(msg.topic(), msg.partition(), msg.offset()))
-        else:
-            print('error: {0}'.format(msg.error().str()))
+        if msg.error():
+            # Most error message are typically temporary log error and continue.
+            print("Consumer error: {}".format(msg.error()))
+            continue
+
+        print('consumed: {0}'.format(msg.value()))
 
 except KeyboardInterrupt:
     pass
--- a/examples/consumer.py
+++ b/examples/consumer.py
@@ -18,7 +18,7 @@
 #
 # Example high-level Kafka 0.9 balanced Consumer
 #
-from confluent_kafka import Consumer, KafkaException, KafkaError
+from confluent_kafka import Consumer, KafkaException
 import sys
 import getopt
 import json
@@ -95,14 +95,7 @@ if __name__ == '__main__':
             if msg is None:
                 continue
             if msg.error():
-                # Error or event
-                if msg.error().code() == KafkaError._PARTITION_EOF:
-                    # End of partition event
-                    sys.stderr.write('%% %s [%d] reached end at offset %d\n' %
-                                     (msg.topic(), msg.partition(), msg.offset()))
-                else:
-                    # Error
-                    raise KafkaException(msg.error())
+                raise KafkaException(msg.error())
             else:
                 # Proper message
                 sys.stderr.write('%% %s [%d] at offset %d with key %s:\n' %
--- a/tests/integration/integration_test.py
+++ b/tests/integration/integration_test.py
@@ -29,7 +29,6 @@ import sys
 import json
 import gc
 import struct
-from copy import copy
 
 try:
     # Memory tracker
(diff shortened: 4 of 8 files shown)


README.md
```diff
@@ -85,11 +85,8 @@ while True:
     if msg is None:
         continue
     if msg.error():
-        if msg.error().code() == KafkaError._PARTITION_EOF:
-            continue
-        else:
-            print(msg.error())
-            break
+        print(msg.error())
+        break
```
edenhill: Let's change this to a continue while we are at it, errors are typically temporary.
Change to:
```
print("Consumer error: {}".format(msg.error()))
continue
```

README.md
```diff
@@ -172,11 +169,8 @@ while True:
         continue
 
     if msg.error():
-        if msg.error().code() == KafkaError._PARTITION_EOF:
-            continue
-        else:
-            print(msg.error())
-            break
+        print(msg.error())
```
edenhill: dito

examples/confluent_cloud.py
```diff
@@ -87,7 +87,8 @@ def acked(err, msg):
     'sasl.username': '<ccloud key>',
     'sasl.password': '<ccloud secret>',
     'group.id': str(uuid.uuid1()),  # this will create a new consumer group on each invocation.
-    'auto.offset.reset': 'earliest'
+    'auto.offset.reset': 'earliest',
+    'enable.partition.eof': True
```
edenhill: do we really want eof-checking in this example, which is likely to be widely used?

examples/consumer.py
```diff
@@ -52,7 +52,7 @@ def print_usage_and_exit(program_name):
     # Consumer configuration
     # See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
     conf = {'bootstrap.servers': broker, 'group.id': group, 'session.timeout.ms': 6000,
-            'auto.offset.reset': 'earliest'}
+            'auto.offset.reset': 'earliest', 'enable.partition.eof': True}
```
edenhill: we don't want eof-checking in the standard examples.

examples/integration_test.py
```diff
@@ -1282,10 +1268,7 @@ def verify_avro_explicit_read_schema():
                 continue
 
             if msg.error():
-                if msg.error().code() == confluent_kafka.KafkaError._PARTITION_EOF:
-                    break
```
edenhill: this is a change in behaviour, it would previously break out on EOF. Now there is no exit condition in this loop.
Did this test really pass?

tests/test_misc.py
```diff
@@ -169,9 +169,11 @@ def test_unordered_dict(init_func):
 
 
 def test_topic_config_update():
+    # *NOTE* default.topic.config has been deprecated.
+    # This example remains to ensure backward-compatibility until it's removal
     confs = [{"message.timeout.ms": 600000, "default.topic.config": {"message.timeout.ms": 1000}},
              {"message.timeout.ms": 1000},
-             {"default.topic.config": {"message.timeout.ms": 1000}}]
```
edenhill: This should not change, since it tests that default.topic.config works.
Also, the change just creates a duplicate test case.

rnpridgeon: doesn't the first case test that default.topic.config works?

edenhill: Why were there three tests to begin with then?
First test is: global + default
Second is: global
Third is: default

README.md
rnpridgeon: ack

examples/confluent_cloud.py
rnpridgeon: I left any example that does something with the error. We can remove it though.

edenhill: Let's create a new example (later) that does EOF reading, to keep these examples clean.

tests/test_misc.py
rnpridgeon: I agree, so two tests it is

edenhill: No, we haven't removed any functionality yet, just deprecated it.
We still need to test all three combinations.

examples/integration_test.py
rnpridgeon: Only ran the unit tests the first go to be honest. Fixed, running the full suite before pushing

tests/test_misc.py
rnpridgeon: ack

examples/confluent_cloud.py
```diff
@@ -100,12 +100,12 @@ def acked(err, msg):
...
             #   the group to rebalance and start consuming
             continue
-        elif not msg.error():
-            print('consumed: {0}'.format(msg.value()))
-        elif msg.error().code() == KafkaError._PARTITION_EOF:
-            print('end of partition: {0} [{1}] @ {2}'.format(msg.topic(), msg.partition(), msg.offset()))
-        else:
-            print('error: {0}'.format(msg.error().str()))
+        if msg.error():
+            # Most error message are self-recting; log error and continue
```
edenhill: recting?

examples/integration_test.py
```diff
@@ -1263,6 +1253,7 @@ def verify_avro_explicit_read_schema():
         conf = copy(cons_conf)
         if schema_registry_url:
             conf['schema.registry.url'] = schema_registry_url
+            conf['enable.partition.eof'] = True
```
edenhill: this should be applied for both clients, not just for when a real SR is available.

examples/integration_test.py
```diff
@@ -1304,7 +1295,7 @@ def verify_avro_explicit_read_schema():
             assert combo.get('key') == record_key
             assert combo.get('value')['name'] == record_value['name']
             c.commit(msg, asynchronous=False)
-        # Close consumer
+        # Close consume
```
edenhill: incorrect

tests/test_misc.py
```diff
@@ -169,6 +169,8 @@ def test_unordered_dict(init_func):
 
 
 def test_topic_config_update():
+    # *NOTE* default.topic.config has been deprecated.
+    # This example remains to ensure backward-compatibility until it's removal
```
edenhill: its

examples/integration_test.py
rnpridgeon: see commit message about this being removed in #440. I will be rebasing shortly

examples/confluent_cloud.py
rnpridgeon: rectifying, sorry. I'm not sure that is the right word either. Just going back to 'typically temporary'

examples/integration_test.py
rnpridgeon: post fast-forward:

```bash
git grep "default.topic.config"
confluent_kafka/src/confluent_kafka.c:        if ((vo = PyDict_GetItemString(confdict, "default.topic.config"))) {
confluent_kafka/src/confluent_kafka.c:                             "default.topic.config has being deprecated, "
confluent_kafka/src/confluent_kafka.c:                             "set default topic configuration values in the global dict");
confluent_kafka/src/confluent_kafka.c:                PyDict_DelItemString(confdict, "default.topic.config");
docs/index.rst:* ``default.topic.config``: value is a dict of client topic-level configuration
tests/test_misc.py:    # *NOTE* default.topic.config has been deprecated.
tests/test_misc.py:    confs = [{"message.timeout.ms": 600000, "default.topic.config": {"message.timeout.ms": 1000}},
tests/test_misc.py:             {"default.topic.config": {"message.timeout.ms": 1000}}]
```

tests/integration/integration_test.py
```diff
@@ -1227,6 +1216,7 @@ def verify_avro_explicit_read_schema():
         conf = copy(cons_conf)
         if schema_registry_url:
             conf['schema.registry.url'] = schema_registry_url
+            conf['enable.partition.eof'] = True
```
edenhill: the check for schema_registry_url should be much earlier, and there's also one for the producer.
If schema_registry_url is always set (otherwise an error) then the ifs are not needed

rnpridgeon: Right this test was added with the explicit read PR. It would appear I didn't not review the integration tests well enough. In order to keep the commit history cleaner can I refactor the integration test in a subsequent PR.

edenhill: Suggest you fix it in this PR, an additional commit for a cleanup is fine.

rnpridgeon: ack

tests/integration/integration_test.py
```diff
@@ -1173,100 +1162,69 @@ def verify_config(expconfig, configs):
...
     from confluent_kafka import avro
-    avsc_dir = os.path.join(os.path.dirname(__file__), os.pardir, 'avro')
 
-    # Producer config
-    conf = {'bootstrap.servers': bootstrap_servers,
-            'error_cb': error_cb}
+    """ verify that reading Avro with explicit reader schema works"""
+    base_conf = {'bootstrap.servers': bootstrap_servers,
+                 'error_cb': error_cb,
+                 'api.version.fallback.ms': 0,
```
edenhill: these two are not needed

tests/integration/integration_test.py
```diff
@@ -1173,100 +1162,69 @@ def verify_config(expconfig, configs):
...
+    reader_schema = avro.load(os.path.join(avsc_dir, "user_v2.avsc"))
+
+    user_value1 = {
+        "name": " Rogers Nelson"
     }
-    val1 = {
-        "name": "abc"
+
+    user_value2 = {
+        "name": "Kenny Loggins"
```
edenhill: :notes:

tests/integration/integration_test.py
```diff
@@ -1173,100 +1162,69 @@ def verify_config(expconfig, configs):
...
 
     combinations = [
-        dict(value=val, value_schema=schema2, key=float_value, key_schema=key_schema,
-             reader_value_schema=schema1, reader_key_schema=key_schema),
-        dict(value=val1, value_schema=schema1, key=float_value, key_schema=key_schema,
-             reader_value_schema=schema2, reader_key_schema=key_schema),
+        dict(key=user_value1, key_schema=writer_schema, value=user_value2, value_schema=writer_schema),
+        dict(key=user_value2, key_schema=writer_schema, value=user_value1, value_schema=writer_schema)
     ]
+    topic = str(uuid.uuid4())
```
edenhill: use the testconf topic prefix to be able to identify topics created by these tests

tests/integration/integration_test.py
```diff
@@ -1173,100 +1162,69 @@ def verify_config(expconfig, configs):
...
-        c.close()
-    pass
+        msgcount += 1
+        # Avro schema projection should return the two fields not present in the writer schema
+        try:
+            assert(msg.key().get('favorite_number') == 42)
+            assert(msg.key().get('favorite_color') == "purple")
+            assert(msg.value().get('favorite_number') == 42)
+            assert(msg.value().get('favorite_color') == "purple")
+            print("success: schema project worked for explicit reader schema")
```
edenhill: schema projection

rnpridgeon: 🥉

examples/confluent_cloud.py
```diff
@@ -98,14 +98,14 @@ def acked(err, msg):
...
+            #   the group to rebalance and start consuming.
             continue
-        elif not msg.error():
-            print('consumed: {0}'.format(msg.value()))
-        elif msg.error().code() == KafkaError._PARTITION_EOF:
-            print('end of partition: {0} [{1}] @ {2}'.format(msg.topic(), msg.partition(), msg.offset()))
-        else:
-            print('error: {0}'.format(msg.error().str()))
+        if msg.error():
+            # Most error message are typically temporary log error and continue.
```
edenhill: Punctuation, and stuff: "Most error are typically temporary, log error and continue."

examples/confluent_cloud.py
```diff
@@ -98,14 +98,14 @@ def acked(err, msg):
...
+            print("Consumer error: {}".format(msg.error()))
+            continue
+
+        print('consumed: {0}'.format(msg.value()))
```
edenhill: The positional {0} is not really needed, {} is fine and more consistent with other code.

tests/integration/integration_test.py
```diff
@@ -1173,100 +1161,67 @@ def verify_config(expconfig, configs):
...
+    user_value2 = {
+        "name": "Kenny Loggins"
     }
 
     combinations = [
-        dict(value=val, value_schema=schema2, key=float_value, key_schema=key_schema,
-             reader_value_schema=schema1, reader_key_schema=key_schema),
-        dict(value=val1, value_schema=schema1, key=float_value, key_schema=key_schema,
-             reader_value_schema=schema2, reader_key_schema=key_schema),
+        dict(key=user_value1, key_schema=writer_schema, value=user_value2, value_schema=writer_schema),
```
edenhill: Suggest to still use different schemas for key and value to validate that they don't get mixed up internally.

edenhill: This is actually changing (reducing), not extending, existing tests.

rnpridgeon: This test is intended to verify that schema projection works, the other avro tests ensure differing key/value schemas are handled appropriately no?

edenhill: Okay, sounds good!

PR #502, opened 2018-12-05
Add generic SerDe support

rnpridgeon:
I'm bad at naming things, open to discussion on `PARTITION_UA` and anything else you might find awkward. Unless it breaks compatibility or requires a lot of work to make compatible <3.

### DIFF
--- a/confluent_kafka/avro/schema.py
+++ b/confluent_kafka/avro/schema.py
@@ -52,7 +52,7 @@ def get_schema(datum):
         return datum.schema
     elif (datum is None):
         return PrimitiveSchema("null")
-    elif isinstance(datum, basestring):
+    elif isinstance(datum, string_type):
         return PrimitiveSchema('string')
     elif isinstance(datum, bool):
         return PrimitiveSchema('boolean')
--- a/confluent_kafka/avro/schema.py
+++ b/confluent_kafka/avro/schema.py
@@ -26,18 +26,18 @@ except NameError:
 
 class GenericAvroRecord(dict):
     """
-    Pairs a AvroRecord with it's schema
+    Pairs an AvroRecord with it's schema
 
     :param schema schema: A parsed Avro schema.
-    :param dict record: Wraps existing dict in GenericAvroRecord
-    :raises ValueError: If schema is None
-    :returns: Avro record with its schema 
+    :param dict record: Wraps existing dict in GenericAvroRecord.
+    :raises ValueError: If schema is None.
+    :returns: Avro record with its schema.
     :rtype: GenericAvroRecord
     """
     __slots__ = ['schema']
 
     def __init__(self, schema, record=None):
-        if schema is None: 
+        if schema is None:
             raise ValueError("schema must not be None")
         self.schema = schema
         if record is not None:
--- a/confluent_kafka/avro/schema.py
+++ b/confluent_kafka/avro/schema.py
@@ -1,0 +1,36 @@
+from avro.schema import PrimitiveSchema
+
+# Python 2 considers int an instance of str
+try:
+    string_type = basestring  # noqa
+except NameError:
+    string_type = str
+
+
+class GenericAvroRecord(dict):
+    __slots__ = ['schema']
+
+    def __init__(self, schema, datum=None):
+        self.schema = schema
+        if datum is not None:
+            self.update(datum)
+
+    def put(self, key, value):
+        self[key] = value
+
+
+def get_schema(datum):
+    if isinstance(datum, GenericAvroRecord):
+        return datum.schema
+    elif (datum is None):
+        return PrimitiveSchema("null")
+    elif isinstance(datum, basestring):
+        return PrimitiveSchema('string')
+    elif isinstance(datum, bool):
+        return PrimitiveSchema('boolean')
+    elif isinstance(datum, int):
+        return PrimitiveSchema('int')
+    elif isinstance(datum, float):
+        return PrimitiveSchema('float')
+    else:
+        raise ValueError("Unsupported Avro type {}.".format(type(datum)))
--- a/confluent_kafka/avro/schema.py
+++ b/confluent_kafka/avro/schema.py
@@ -1,3 +1,20 @@
+#!/usr/bin/env python
+#
+# Copyright 2018 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 from avro.schema import PrimitiveSchema
 
 # Python 2 considers int an instance of str
@@ -8,12 +25,23 @@ except NameError:
 
 
 class GenericAvroRecord(dict):
+    """
+    Pairs a AvroRecord with it's schema
+
+    :param schema schema: A parsed Avro schema.
+    :param dict record: Wraps existing dict in GenericAvroRecord
+    :raises ValueError: If schema is None
+    :returns: Avro record with its schema 
+    :rtype: GenericAvroRecord
+    """
     __slots__ = ['schema']
 
-    def __init__(self, schema, datum=None):
+    def __init__(self, schema, record=None):
+        if schema is None: 
+            raise ValueError("schema must not be None")
         self.schema = schema
-        if datum is not None:
-            self.update(datum)
+        if record is not None:
+            self.update(record)
 
     def put(self, key, value):
         self[key] = value
--- a/confluent_kafka/avro/serializer/__init__.py
+++ b/confluent_kafka/avro/serializer/__init__.py
@@ -61,7 +61,8 @@ class SerializerError(Exception):
 
 
 class KeySerializerError(SerializerError):
-    pass
+    def __new__(cls, message):
+        return super(KeySerializerError, cls).__new__(SerializerError, message, True)
 
 
 class ValueSerializerError(SerializerError):
--- a/confluent_kafka/avro/serializer/__init__.py
+++ b/confluent_kafka/avro/serializer/__init__.py
@@ -15,9 +15,37 @@
 # limitations under the License.
 #
 
+import avro
+import avro.io
+import io
+import logging
+import struct
+import sys
+import traceback
+
+from confluent_kafka.avro.schema import GenericAvroRecord, get_schema
+from confluent_kafka.avro import ClientError
+
+log = logging.getLogger(__name__)
+
+MAGIC_BYTE = 0
+
+HAS_FAST = False
+try:
+    from fastavro import schemaless_reader, schemaless_writer
+
+    HAS_FAST = True
+except ImportError:
+    pass
+
 
 class SerializerError(Exception):
     """Generic error from serializer package"""
+
+    def __new__(cls, message, is_key=False):
+        if is_key:
+            return super(SerializerError, cls).__new__(KeySerializerError, message)
+        return super(SerializerError, cls).__new__(ValueSerializerError, message)
 
     def __init__(self, message):
         self.message = message
@@ -38,3 +66,190 @@ class KeySerializerError(SerializerError):
 
 class ValueSerializerError(SerializerError):
     pass
+
+
+class ContextStringIO(io.BytesIO):
+    """
+    Wrapper to allow use of StringIO via 'with' constructs.
+    """
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+        return False
+
+
+class AvroSerializer(object):
+
+    __slots__ = ["registry_client", "codec_cache", "is_key"]
+
+    def __init__(self, registry_client, is_key=False):
+        self.registry_client = registry_client
+        self.codec_cache = {}
+        self.is_key = is_key
+
+    def __call__(self, topic, record):
+        """
+        Given a parsed avro schema, encode a record for the given topic.
+
+        The schema is registered with the subject of 'topic-value'
+        :param str topic: Topic name
+        :param GenericAvroRecord record: An object to serialize
+        :returns: Encoded record with schema ID as bytes
+        :rtype: bytes
+        """
+
+        if record is None:
+            return None
+
+        subject_suffix = '-key' if self.is_key else '-value'
+        subject = topic + subject_suffix
+
+        schema_id = self.registry_client.register(subject, get_schema(record))
+        if not schema_id:
+            message = "Failed to register schema with subject {}.".format(subject)
+            raise SerializerError(message, is_key=self.is_key)
+
+        if schema_id not in self.codec_cache:
+            self.codec_cache[schema_id] = self._get_encoder_func(get_schema(record))
+
+        return self._encode_record_with_schema_id(schema_id, record)
+
+    def _get_encoder_func(self, writer_schema):
+        if HAS_FAST:
+            return lambda record, fp: schemaless_writer(fp, writer_schema.to_json(), record)
+        writer = avro.io.DatumWriter(writer_schema)
+        return lambda record, fp: writer.write(record, avro.io.BinaryEncoder(fp))
+
+    def _encode_record_with_schema_id(self, schema_id, record):
+        """
+        Encode a record with a given schema id.  The record must
+        be a python dictionary.
+        :param int schema_id: integer ID
+        :param dict record: An object to serialize
+        :param bool is_key: If the record is a key
+        :param SerializerErr err_type: Error type to raise on serialization exception
+        :returns: decoder function
+        :rtype: func
+        """
+
+        if schema_id not in self.codec_cache:
+            try:
+                schema = self.registry_client.get_by_id(schema_id)
+                if not schema:
+                    raise SerializerError("Schema does not exist", self.is_key)
+                self.codec_cache[schema_id] = self._get_encoder_func(schema)
+            except ClientError:
+                exc_type, exc_value, exc_traceback = sys.exc_info()
+                raise SerializerError(
+                    repr(traceback.format_exception(exc_type, exc_value, exc_traceback)),
+                    self.is_key)
+
+        # get the writer
+        writer = self.codec_cache[schema_id]
+        with ContextStringIO() as outf:
+            # Write the magic byte and schema ID in network byte order (big endian)
+            outf.write(struct.pack('>bI', MAGIC_BYTE, schema_id))
+
+            # write the record to the rest of the buffer
+            writer(record, outf)
+            return outf.getvalue()
+
+
+class AvroDeserializer(object):
+
+    __slots__ = ["registry_client", "codec_cache", "is_key", "reader_schema"]
+
+    def __init__(self, registry_client, is_key=False, reader_schema=None):
+        self.registry_client = registry_client
+        self.codec_cache = {}
+        self.is_key = is_key
+        self.reader_schema = reader_schema
+
+    def __call__(self, topic, message):
+        """
+        Decode a message from kafka that has been encoded for use with
+        the schema registry.
+        :param str|bytes or None message: message key or value to be decoded
+        :returns: Decoded message contents.
+        :rtype GenericAvroRecord:
+        """
+
+        if message is None:
+            return None
+
+        if len(message) <= 5:
+            raise SerializerError("message is too small to decode")
+
+        with ContextStringIO(message) as payload:
+            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            if magic != MAGIC_BYTE:
+                raise SerializerError("message does not start with magic byte", self.is_key)
+
+            decoder_func = self._get_decoder_func(schema_id, payload)
+            return decoder_func(payload)
+
+    def _get_decoder_func(self, schema_id, payload):
+        if schema_id in self.codec_cache:
+            return self.codec_cache[schema_id]
+
+        try:
+            writer_schema = self.registry_client.get_by_id(schema_id)
+        except ClientError as e:
+            raise SerializerError("unable to fetch schema with id %d: %s" % (schema_id, str(e)))
+
+        if writer_schema is None:
+            raise SerializerError("unable to fetch schema with id %d" % (schema_id))
+
+        reader_schema = writer_schema if self.reader_schema is None else self.reader_schema
+
+        curr_pos = payload.tell()
+
+        if HAS_FAST:
+            # try to use fast avro
+            try:
+                schemaless_reader(payload, writer_schema.to_json())
+
+                # If we reach this point, this means we have fastavro and it can
+                # do this deserialization. Rewind since this method just determines
+                # the reader function and we need to deserialize again along the
+                # normal path.
+                payload.seek(curr_pos)
+
+                def fast_decoder(p):
+                    return schemaless_reader(p, writer_schema.to_json(), reader_schema.to_json())
+
+                def fast_record_decoder(p):
+                    return GenericAvroRecord(reader_schema,
+                                             schemaless_reader(p, writer_schema.to_json(), reader_schema.to_json()))
+
+                if writer_schema.type is 'record':
+                    self.codec_cache[schema_id] = fast_record_decoder
+                else:
+                    self.codec_cache[schema_id] = fast_decoder
+
+                return self.codec_cache[schema_id]
+            except Exception:
+                # Fast avro failed, fall thru to standard avro below.
+                pass
+
+        payload.seek(curr_pos)
+
+        avro_reader = avro.io.DatumReader(writer_schema, reader_schema)
+
+        def record_decoder(p):
+            bin_decoder = avro.io.BinaryDecoder(p)
+            return GenericAvroRecord(reader_schema, avro_reader.read(bin_decoder))
+
+        def decoder(p):
+            bin_decoder = avro.io.BinaryDecoder(p)
+            return avro_reader.read(bin_decoder)
+
+        if writer_schema.get_prop('type') is 'record':
+            self.codec_cache[schema_id] = record_decoder
+        else:
+            self.codec_cache[schema_id] = decoder
+
+        return self.codec_cache[schema_id]
--- a/confluent_kafka/avro/serializer/__init__.py
+++ b/confluent_kafka/avro/serializer/__init__.py
@@ -24,7 +24,7 @@ import sys
 import traceback
 
 from confluent_kafka.avro.schema import GenericAvroRecord, get_schema
-from confluent_kafka.avro import ClientError
+from confluent_kafka.avro.error import ClientError
 
 log = logging.getLogger(__name__)
 
@@ -47,7 +47,7 @@ class SerializerError(Exception):
             return super(SerializerError, cls).__new__(KeySerializerError, message)
         return super(SerializerError, cls).__new__(ValueSerializerError, message)
 
-    def __init__(self, message):
+    def __init__(self, message, is_key=False):
         self.message = message
 
         def __repr__(self):
@@ -81,31 +81,101 @@ class ContextStringIO(io.BytesIO):
         return False
 
 
+def TopicNameStrategy(topic=None, is_key=False, schema=None):
+    """
+    Constructs the subject name under which a schema is registered with the Confluent Schema Registry.
+
+    TopicNameStrategy returns the schema's subject in the form of <topic>-key or <topic>-value.
+
+    :param str topic: Topic name.
+    :param is_key: True if subject is being registered for a message key.
+    :param schema schema: Parsed Avro schema. *Note* Not used by TopicNameStrategy
+    :raises ValueError: If topic is unset.
+    :returns: The subject name with which to register the schema.
+    :rtype: str
+    """
+    if topic is None:
+        raise ValueError("Topic must be set when using TopicNameStrategy")
+
+    return "-".join([topic, '-key' if is_key else '-value'])
+
+
+def RecordNameStrategy(topic=None, is_key=False, schema=None):
+    """
+    Constructs the subject name under which a schema is registered with the Confluent Schema Registry.
+
+    RecordNameStrategy returns the fully-qualified record name regardless of the topic.
+
+    Compatibility checks of the same record name across all topics.
+    This strategy allows a topic to contain a mixture of different record types.
+
+    :param str topic: Topic name.  *Note* Not used by RecordNameStrategy
+    :param is_key: True if subject is being registered for a message key. *Note* Not used by RecordNameStrategy.
+    :param schema schema: Parsed Avro schema.
+    :raises ValueError: If schema is not set.
+    :returns: The subject name with which to register the schema.
+    :rtype: str
+    """
+    if schema is None:
+        raise ValueError("Schema must be set when using RecordNameStategy")
+
+    return schema.fullname
+
+
+def TopicRecordNameStrategy(topic=None, is_key=False, schema=None):
+    """
+    Constructs the subject name under which a schema is registered with the Confluent Schema Registry.
+
+    TopicRecordNameStrategy returns the topic name appended by the fully-qualified record name.
+
+    Compatibility checks are performed against all records of the same name within the same topic.
+    Like the RecordNameStrategy mixed record types are allowed within a topic.
+    This strategy is more flexible in that records needn't be complaint across the cluster.
+
+    :param str topic: Topic name.
+    :param schema schema: Parsed Avro schema.
+    :param is_key: True if used by a key_serializer.
+    :raises ValueError: If topic and schema are not set.
+    :returns: The subject name with which to register the schema.
+    :rtype: str
+    """
+    if not any([topic, schema]):
+        raise ValueError("Both Topic and Schema must be set when using TopicRecordNameStrategy")
+    return "-".join([topic, schema.fullname])
+
+
 class AvroSerializer(object):
+    """
+    Encodes kafka messages as Avro; registering the schema with the Confluent Schema Registry.
 
-    __slots__ = ["registry_client", "codec_cache", "is_key"]
+    :param registry_client CachedSchemaRegistryClient: Instance of CachedSchemaRegistryClient.
+    :param bool is_key: True if configured as a key_serializer.
+    :param func(str, bool, schema): Returns the subject name used when registering schemas.
+    """
 
-    def __init__(self, registry_client, is_key=False):
+    __slots__ = ["registry_client", "codec_cache", "is_key", "subject_strategy"]
+
+    def __init__(self, registry_client, is_key=False, subject_strategy=TopicNameStrategy):
         self.registry_client = registry_client
         self.codec_cache = {}
         self.is_key = is_key
+        self.subject_strategy = subject_strategy
 
     def __call__(self, topic, record):
         """
         Given a parsed avro schema, encode a record for the given topic.
 
-        The schema is registered with the subject of 'topic-value'
-        :param str topic: Topic name
-        :param GenericAvroRecord record: An object to serialize
-        :returns: Encoded record with schema ID as bytes
+        The schema is registered with the subject of 'topic-value'.
+        :param str topic: Topic name.
+        :param GenericAvroRecord record: An object to serialize.
+        :returns: Encoded record with schema ID as bytes.
         :rtype: bytes
         """
 
         if record is None:
             return None
 
-        subject_suffix = '-key' if self.is_key else '-value'
-        subject = topic + subject_suffix
+        subject = self.subject_strategy(topic, self.is_key, get_schema(record))
 
         schema_id = self.registry_client.register(subject, get_schema(record))
         if not schema_id:
@@ -115,7 +185,7 @@ class AvroSerializer(object):
         if schema_id not in self.codec_cache:
             self.codec_cache[schema_id] = self._get_encoder_func(get_schema(record))
 
-        return self._encode_record_with_schema_id(schema_id, record)
+        return self._encode(schema_id, record)
 
     def _get_encoder_func(self, writer_schema):
         if HAS_FAST:
@@ -123,12 +193,11 @@ class AvroSerializer(object):
         writer = avro.io.DatumWriter(writer_schema)
         return lambda record, fp: writer.write(record, avro.io.BinaryEncoder(fp))
 
-    def _encode_record_with_schema_id(self, schema_id, record):
+    def _encode(self, schema_id, datum):
         """
-        Encode a record with a given schema id.  The record must
-        be a python dictionary.
+        Encode a datum with a given schema id.
         :param int schema_id: integer ID
-        :param dict record: An object to serialize
+        :param object datum: An object to serialize
         :param bool is_key: If the record is a key
         :param SerializerErr err_type: Error type to raise on serialization exception
         :returns: decoder function
@@ -154,12 +223,18 @@ class AvroSerializer(object):
             outf.write(struct.pack('>bI', MAGIC_BYTE, schema_id))
 
             # write the record to the rest of the buffer
-            writer(record, outf)
+            writer(datum, outf)
             return outf.getvalue()
 
 
 class AvroDeserializer(object):
+    """
+    Decodes Kafka messages encoded by Confluent Schema Registry compliant Avro Serializers.
 
+    :param registry_client CachedSchemaRegistryClient: Instance of CachedSchemaRegistryClient.
+    :param bool is_key: True if configured as a key_serializer.
+    :param schema reader_schema: Optional reader schema to be used during deserialization.
+    """
     __slots__ = ["registry_client", "codec_cache", "is_key", "reader_schema"]
 
     def __init__(self, registry_client, is_key=False, reader_schema=None):
@@ -168,22 +243,21 @@ class AvroDeserializer(object):
         self.is_key = is_key
         self.reader_schema = reader_schema
 
-    def __call__(self, topic, message):
+    def __call__(self, topic, datum):
         """
-        Decode a message from kafka that has been encoded for use with
-        the schema registry.
-        :param str|bytes or None message: message key or value to be decoded
-        :returns: Decoded message contents.
+        Decode a datum from kafka that has been encoded for use with the Confluent Schema Registry.
+        :param str|bytes or None datum: message key or value to be decoded.
+        :returns: Decoded message key or value contents.
         :rtype GenericAvroRecord:
         """
 
-        if message is None:
+        if datum is None:
             return None
 
-        if len(message) <= 5:
+        if len(datum) <= 5:
             raise SerializerError("message is too small to decode")
 
-        with ContextStringIO(message) as payload:
+        with ContextStringIO(datum) as payload:
             magic, schema_id = struct.unpack('>bI', payload.read(5))
             if magic != MAGIC_BYTE:
                 raise SerializerError("message does not start with magic byte", self.is_key)
@@ -247,7 +321,7 @@ class AvroDeserializer(object):
             bin_decoder = avro.io.BinaryDecoder(p)
             return avro_reader.read(bin_decoder)
 
-        if writer_schema.get_prop('type') is 'record':
+        if writer_schema.type is 'record':
             self.codec_cache[schema_id] = record_decoder
         else:
             self.codec_cache[schema_id] = decoder
--- a/confluent_kafka/consumer.py
+++ b/confluent_kafka/consumer.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#
+# Copyright 2018 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from .cimpl import Consumer as _impl
+from warnings import warn
+
+
+class Consumer(_impl):
+    """
+        Create a new Kafka Consumer instance.
+
+        To avoid spontaneous calls from non-Python threads all callbacks will only be served upon
+        calling ```client.poll()``` or ```client.flush()```.
+
+        :param dict conf: Configuration properties.
+            See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md for more information.
+        :param func key_deserializer(topic, key): Converts message key bytes to object.
+            **note** deserializers are responsible for handling NULL keys
+        :param func value_deserializer(topic, value): Converts message value bytes to object.
+            **note** deserializers are responsible for handling NULL values
+        :param func on_commit(err, [partitions]): Callback used to indicate success or failure
+            of an offset commit.
+        :param func stats_cb(json_str): Callback for statistics emitted every ``statistics.interval.ms``.
+            See https://github.com/edenhill/librdkafka/wiki/Statistics” for more information.
+        :param func throttle_cb(confluent_kafka.ThrottleEvent): Callback for throttled request reporting.
+        :param logging.handlers logger: Forwards logs from the Kafka client to the provided handler instance.
+            Log messages will only be forwarded when ``client.poll()`` or ``producer.flush()`` are called.
+        :raises TypeError: If conf is not a dict.
+    """
+
+    __slots__ = ["_key_deserializer", "_value_deserializer"]
+
+    # conf must remain optional as long as kwargs are supported
+    def __init__(self, conf={}, key_deserializer=None, value_deserializer=None,
+                 on_commit=None, stats_cb=None, throttle_cb=None, logger=None, **kwargs):
+
+        if not isinstance(conf, dict):
+            raise TypeError("expected configuration dict")
+
+        if kwargs:
+            # Handle kwargs for backwards compatibility
+            conf.update(kwargs)
+            warn("The use of kwargs is being deprecated. "
+                 "In future releases `conf` will be mandatory and "
+                 "all keyword arguments must match the constructor signature explicitly.",
+                 category=DeprecationWarning, stacklevel=2)
+
+        self._key_deserializer = key_deserializer
+        self._value_deserializer = value_deserializer
+
+        # Callbacks can be set in the conf dict or *ideally* as parameters.
+        # Handle both cases prior to passing along to _impl
+        # If callbacks are configured in both places parameter values take precedence.
+        if not on_commit:
+            on_commit = conf.get('on_commit', None)
+
+        if not stats_cb:
+            stats_cb = conf.get('stats_cb', None)
+
+        if not throttle_cb:
+            throttle_cb = conf.get('throttle_cb', None)
+
+        if not logger:
+            logger = conf.get('logger', None)
+        super(Consumer, self).__init__(conf, on_commit=on_commit, stats_cb=stats_cb,
+                                       throttle_cb=throttle_cb, logger=logger)
+
+    def poll(self, timeout=-1.0, key_deserializer=None, value_deserializer=None):
+        """
+            Consumes a message, triggers callbacks, returns an event.
+
+            The application must check the returned Message object’s Message.error() method to distinguish
+            between proper messages, an error(see error().code() for specifics), or an event.
+
+            :param float timeout:  Maximum time in seconds to block waiting for message, event or callback.
+            :param func key_deserializer(topic, key): Converts message key bytes to object.
+                **note** deserializers are responsible for handling NULL keys
+            :param func value_deserializer(topic, value): Converts message value bytes to object.
+                **note** deserializers are responsible for handling NULL values
+            :returns: A confluent_kafka.Message or None on timeout.
+            :raises RuntimeError: If called on a closed consumer.
+        """
+
+        msg = super(Consumer, self).poll(timeout)
+
+        if not msg or msg.error():
+            return msg
+
+        topic = msg.topic()
+
+        # parameter overrides take precedence over instance functions
+        if not key_deserializer:
+            key_deserializer = self._key_deserializer
+
+        if key_deserializer:
+            msg.set_key(key_deserializer(topic, msg.key()))
+
+        if not value_deserializer:
+            value_deserializer = self._value_deserializer
+
+        if value_deserializer:
+            msg.set_value(value_deserializer(topic, msg.value()))
+
+        return msg
+
+    def consume(self, num_messages=1, timeout=-1):
+        """
+            Consume messages, calls callbacks and returns a list of messages. (possibly empty on timeout)
+
+            The application must check Message.error() to distinguish between
+                proper messages, an error(see error().code() for specifics), or an event. for each
+                Message in the list.
+
+            :param int num_messages: Maximum number of messages to return (default: 1)
+            :param float timeout: Maximum time in seconds to block waiting for message, event or callback.
+                (default: infinite (-1))
+            :returns: A list of Message objects (possibly empty on timeout)
+            :rtype: list(Message)
+            :raises NotImplementedError: If used with key/value serializers.
+            :raises RuntimeError: If called on a closed consumer.
+            :raises KafkaError: In case of internal error.
+            :raises ValueError: If num_messages > 1M.
+        """
+
+        # Disable consume() method when serializers are in use.
+        if self._key_deserializer or self._value_deserializer:
+            raise(NotImplementedError, "Batch consumption does not support the use of deserializers")
+
+        return super(Consumer, self).consume(num_messages, timeout)
--- a/confluent_kafka/consumer.py
+++ b/confluent_kafka/consumer.py
@@ -21,9 +21,17 @@ from .cimpl import Consumer as _impl
 from warnings import warn
 
 
+def byteDeserializer(topic, payload):
+    """
+        byteDeserializer returns an unaltered payload to the caller
+    """
+
+    return payload
+
+
 class Consumer(_impl):
     """
-        Create a new Kafka Consumer instance with or without serializer support.
+        Create a new Kafka Consumer instance.
 
         To avoid spontaneous calls from non-Python threads all callbacks will only be served upon
         calling ```client.poll()``` or ```client.flush()```.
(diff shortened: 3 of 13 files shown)


tests/integration/integration_test.py
```diff
@@ -1269,6 +1269,108 @@ def verify_avro_explicit_read_schema():
...
+             key_decoder=b16_decoder, value_decoder=b64_decoder),
+        # Test key_serializer and value_serializer overrides
+        dict(key="b16encode".encode('utf-8'), value="b16encode".encode('utf-8'),
+             key_encoder=lambda topic, key: b16encode(key),
+             value_encoder=lambda topic, value: b16encode(value),
+             key_decoder=b16_decoder, value_decoder=b16_decoder),
+    ]
+
+    # Test Producer serializer
+    def assert_cb(err, msg, conf):
```
edenhill: this name is a bit vague

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+        Create a new Kafka Consumer instance.
+
+        To avoid spontaneous calls from non-Python threads all callbacks will only be served upon
+        calling ```client.poll()``` or ```client.flush()```.
+
+        :param dict conf: Configuration properties.
+            See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md for more information.
+        :param func key_deserializer(topic, key): Converts message key bytes to object.
+            **note** deserializers are responsible for handling NULL keys
```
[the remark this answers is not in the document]
rnpridgeon: Doesn't the fact it's a kwarg with a default implicitly state that it's optional? I only explicitly mention conf because the plan is to make it required in the future.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+        :param func value_deserializer(topic, value): Converts message value bytes to object.
+            **note** deserializers are responsible for handling NULL values
+        :param func on_commit(err, [partitions]): Callback used to indicate success or failure
```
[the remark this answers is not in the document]
rnpridgeon: hm, I though `[stuff]` implied list but I could see how this is a problem.

tests/integration/integration_test.py
```diff
@@ -1269,6 +1269,108 @@ def verify_avro_explicit_read_schema():
...
+    # base16 deserializer for converting message contents back to a string
+    def b16_decoder(topic, data):
+        if data is None:
+            return None
+        return b16decode(data)
+
+    # Test matrix fo use when evaluating alternative kv serialization strategies.
+    combinations = [
+        # Test default serializers, use encode for python 2 and 3 compatibility.
+        dict(key="b64encode".encode('utf-8'), value="b64encode".encode('utf-8'),
```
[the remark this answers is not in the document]
rnpridgeon: covered in subsequent combinations where key/value serializers differ. 

i.e.   `dict(key="b64encode".encode('utf-8'), value="b16encode".encode('utf-8'),`

tests/integration/integration_test.py
rnpridgeon: The name or the comment or both?

tests/integration/integration_test.py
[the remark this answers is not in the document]
edenhill: that is not the same thing though, here we're using the same serializer for both key and value, so use different values-to-serialize to check for mixups.

tests/integration/integration_test.py
edenhill: the method name should say what the method does.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+            throttle_cb = conf.get('throttle_cb', None)
+
+        if not logger:
+            logger = conf.get('logger', None)
+
+        super(Producer, self).__init__(conf, error_cb=error_cb, stats_cb=stats_cb,
+                                       throttle_cb=throttle_cb, logger=logger)
+
+    def produce(self, topic, value=None, key=None, partition=PARTITION_UA,
+                on_delivery=None, callback=None, timestamp=0, headers=None,
```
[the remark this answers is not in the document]
rnpridgeon: This default is set by the extension code upon initialization any way. 

````c
        long long timestamp = 0;
````

This is how I came up with the default.

File: confluent_kafka/producer.py [hunk 6]
```diff
@@ -0,0 +1,130 @@
...
+                key_serializer=None, value_serializer=None):
+        """
+            Produces message to Kafka
+
+            :param str topic: Topic to produce message to.
+            :param str|bytes|obj value: Message payload; value_serializer required for non-character values.
+            :param str|bytes|obj key: Message key payload; key_serializer required for non-character keys.
+            :param int partition: Partition to produce to, else uses the configured built-in partitioner.
+                Default value is PARTITION_UA(round-robin over all partitions).
```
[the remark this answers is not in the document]
rnpridgeon: 🤦‍♂️ , yep thank you

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+            :param func on_delivery(confluent_kafka.KafkaError, confluent_kafka.Message):
+                Delivery report callback to call on successful or failed delivery.
+            :param func callback(confluent_kafka.KafkaError, confluent_kafka.Message):
+                See on_delivery.
+            :param int timestamp: Message timestamp (CreateTime) in milliseconds since epoch UTC
+                (requires librdkafka >= v0.9.4, api.version.request=true, and broker >= 0.10.0.0).
```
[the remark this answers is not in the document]
rnpridgeon: Unless you chose to build and link yourself.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1692,6 +1695,12 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
                         goto inner_err;
 		}
 
+                /* Treat key=Py_None as if it were never set */
+                if (vo == Py_None) {
```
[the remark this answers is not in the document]
rnpridgeon: We are only silently ignoring configs which have been set to `None`. Invalid properties with actual values will still error out. This takes away from the complexity of handling `x=None` in each application config property while providing the same benefits of erroring out on invalid configs. 

e.g. 
stats_cbb=None  *ignored
stats_cbb=MyStatsCb *error
stats_cb=None *handled

All without checking for none and removing the key individually.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -2352,7 +2356,7 @@ static PyObject *_init_cimpl (void) {
         PyModule_AddIntConstant(m, "OFFSET_END", RD_KAFKA_OFFSET_END);
         PyModule_AddIntConstant(m, "OFFSET_STORED", RD_KAFKA_OFFSET_STORED);
         PyModule_AddIntConstant(m, "OFFSET_INVALID", RD_KAFKA_OFFSET_INVALID);
-
+        PyModule_AddIntConstant(m, "PARTITION_UA", RD_KAFKA_PARTITION_UA);
```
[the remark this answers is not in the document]
rnpridgeon: ack

confluent_kafka/consumer.py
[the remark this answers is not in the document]
mhowlett: we should add pip 484 compatible type annotations across the library like so: https://mypy.readthedocs.io/en/latest/python2.html

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+                 "In future releases `conf` will be mandatory and "
+                 "all keyword arguments must match the constructor signature explicitly.",
+                 category=DeprecationWarning, stacklevel=2)
+
+        self._key_deserializer = key_deserializer
+        self._value_deserializer = value_deserializer
+
+        # Callbacks can be set in the conf dict or *ideally* as parameters.
+        # Handle both cases prior to passing along to _impl
+        # If callbacks are configured in both places parameter values take precedence.
```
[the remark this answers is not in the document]
mhowlett: Is there a benefit of allowing cb's to be specified in more than one place?
Also, I think we might end up regretting it if we allow non-string values in the config.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+
+        msg = super(Consumer, self).poll(timeout)
+
+        if not msg or msg.error():
+            return msg
+
+        topic = msg.topic()
+
+        # parameter overrides take precedence over instance functions
```
mhowlett: the below code silently omits to set anything for message key and value if the deserializers aren't specified - that's not a good idea. we could either throw an exception of pass the value straight through. the latter seems better on first instinct, but that requires further thought.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if not value_deserializer:
+            value_deserializer = self._value_deserializer
+
+        if value_deserializer:
+            msg.set_value(value_deserializer(topic, msg.value()))
+
+        return msg
+
+    def consume(self, num_messages=1, timeout=-1):
```
mhowlett: 1. I'd question the need for both consume and poll at the python api level. also, the justification for the naming is going to be non-obvious to many people and cause friction. batch consume is old skool librdkafka right? can we just get rid of / depreciate `consume`?
2. if it's not depreciated, I think we probably want to enable deserializers. that you don't take this method seriously (barf out if the serializers are not there) is an indication it should be depreciated...
3. batch consuming could actually be valuable (even in .NET I think it would represent a significant perf improvement), lets discuss.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+            Consume messages, calls callbacks and returns a list of messages. (possibly empty on timeout)
+
+            The application must check Message.error() to distinguish between
+                proper messages, an error(see error().code() for specifics), or an event. for each
+                Message in the list.
+
+            :param int num_messages: Maximum number of messages to return (default: 1)
```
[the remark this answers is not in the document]
mhowlett: complete agreement 👍

confluent_kafka/producer.py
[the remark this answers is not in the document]
mhowlett: regardless, `None` is semantically what we want here.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+            warn("The use of kwargs is being deprecated. "
+                 "In future releases `conf` will be mandatory and "
+                 "all keyword arguments must match the constructor signature explicitly.",
+                 category=DeprecationWarning, stacklevel=2)
+
+        self._key_serializer = key_serializer
+        self._value_serializer = value_serializer
+
+        # Callbacks can be set in the conf dict or *ideally* as parameters.
+        # Handle both cases prior to passing along to _impl
```
mhowlett: same comment as before (why allow them in config?

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+                Delivery report callback to call on successful or failed delivery.
+            :param func callback(confluent_kafka.KafkaError, confluent_kafka.Message):
+                See on_delivery.
+            :param int timestamp: Message timestamp (CreateTime) in milliseconds since epoch UTC
+                (requires librdkafka >= v0.9.4, api.version.request=true, and broker >= 0.10.0.0).
+                Default value is current time.
+            :param dict|list headers: Message headers to set on the message. The header key must be a string while
+                the value must be binary, unicode or None. Accepts a list of (key,value) or a dict.
+                (Requires librdkafka >= v0.11.4 and broker version >= 0.11.0.0)
+            :param func key_serializer(topic, key): Producer key_serializer override;
```
mhowlett: maybe call these `default_key_serializer` and `default_value_serializer`.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+                Converts message value to bytes. **note** serializers are responsible for handling NULL values
+
+        """
+
+        # on_delivery is an alias for callback and take precedence
+        if callback and not on_delivery:
+            on_delivery = callback
+
+        # parameter overrides take precedence over instance functions
+        if not key_serializer:
```
mhowlett: as with the consumer, we want something else to happen if a serializer is not specified - exception, or value passthrough (probably the latter).

unlike the consumer, there is value in allowing serializers to be different on each produce. in python, as you are doing seems ok to me, though I reserve the right to change my mind :-). however, the nomenclature on serializers passed into constructor should be changed to imply they are defaults.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+        if key_serializer:
+            key = key_serializer(topic, key)
+
+        if not value_serializer:
+            value_serializer = self._value_serializer
+
+        if value_serializer:
+            value = value_serializer(topic, value)
+
+        super(Producer, self).produce(topic, value, key, partition, on_delivery=on_delivery,
```
mhowlett: there is value in having the un-serialized values in the delivery reports, so i think this should be a feature of the library (though it has a cost so it should be togglable). however you're throwing this information away here.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,130 @@
...
+                (Requires librdkafka >= v0.11.4 and broker version >= 0.11.0.0)
+            :param func key_serializer(topic, key): Producer key_serializer override;
+                Converts message key to bytes. **note** serializers are responsible for handling NULL keys
+            :param func value_serializer(topic, value): Producer value_serializer override;
+                Converts message value to bytes. **note** serializers are responsible for handling NULL values
+
+        """
+
+        # on_delivery is an alias for callback and take precedence
+        if callback and not on_delivery:
```
mhowlett: do we have a timeline for depreciating this?

tests/integration/integration_test.py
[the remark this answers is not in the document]
rnpridgeon: ack

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if not throttle_cb:
+            throttle_cb = conf.get('throttle_cb', None)
+
+        if not logger:
+            logger = conf.get('logger', None)
+        super(Consumer, self).__init__(conf, on_commit=on_commit, stats_cb=stats_cb,
+                                       throttle_cb=throttle_cb, logger=logger)
+
+    def poll(self, timeout=-1.0, key_deserializer=None, value_deserializer=None):
```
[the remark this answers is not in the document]
rnpridgeon: We can't maintain compatibility with the existing AvroProducer/Consumer without it. Since we have already provided this in the API we can't just rip it out without notice.

confluent_kafka/consumer.py
rnpridgeon: If no serializers the message contents remain in place unaltered.

rnpridgeon: Note the original consumer/producer are no longer exposed directly. If you run the existing tests you will see everything works as if these wrappers were never written.

confluent_kafka/consumer.py
rnpridgeon: 1. I would like to deprecate batch consuming with the python client entirely. 
2. see 1 
3. In order to do batch consumption correctly we would want to error out if `auto.offset.reset=true`. This would need to be accompanied with an example or two of how you would appropriately manage offset commits when using the batch consumer.

confluent_kafka/producer.py
rnpridgeon: backward compatibility

confluent_kafka/producer.py
rnpridgeon: I would like to encourage per client serializer configs despite having to support per request serializer configurations because of the existing Avro client API. With that said I'd like to keep the name `default` out of the name if possible with hope that we will find in practice noone actually wants to override their serializers per request.

confluent_kafka/producer.py
rnpridgeon: We don't have access to the message state from python which is one of the shortcomings of pulling this into python. What we can do however is provide several examples of delivery report callbacks which preserve the message contents pre-serialization which was the plan. Alternatively we can push this down into the extension, set the message contents with the msg key/value an avoid the extra work. Either one is doable although I think starting to pull things int python is nice. We could also wrap user provided callbacks in a lambda and change the docstring.

confluent_kafka/producer.py
rnpridgeon: My understanding of proper feature deprecation is one major release. i.e. Deprecation warning in 1.0 removal in 2.0.

confluent_kafka/consumer.py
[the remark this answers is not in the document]
rnpridgeon: replaced with `**must** handle NULL values.`

confluent_kafka/consumer.py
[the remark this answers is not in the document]
rnpridgeon: As long as it handles the str|byte compatibility issues between py2 and py3 I'm game

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: I don't know what a long long is (double?) but you tell me and I'll parse it as a object check for Py_None and fall back to a 0.

confluent_kafka/consumer.py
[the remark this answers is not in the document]
mhowlett: this is a new class - i don't think ensuring users can swap out the old avro producer and consumer for this one has any priority - you're already breaking compatibility by requiring they change classes.

confluent_kafka/consumer.py
mhowlett: oh right, i miss-read this, and also assumed you did the same thing later...

confluent_kafka/producer.py
mhowlett: semantically they are treated as defaults, so it's less muddled to call them that. if you don't think the feature will be used, then it'd be better to exclude it than providing a sub-optimal api. but it's not un-common to want to use a single librdkafka instance to produce messages of different types. three ways to achieve this are 1. as you have it, 2. one duo of serializers per producer + ability to create other producer instances based on the same underlying librdkafka instance. 3. registering / lookup on type (or even topic).

i'd have to try and use the api to get a feel for whether specifying on each call is too cumbersome (it is in C#, but might not be in python).

confluent_kafka/producer.py
mhowlett: this seems like a required feature to me. i didn't understand your plan with the extension - would you get a pointer to the original python value, pass that via the opaque parameter to the produce method, and get the value again using this pointer in the delivery report?

confluent_kafka/consumer.py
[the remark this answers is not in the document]
rnpridgeon: The idea actually is to get away from having a specific class for Avro. Instead this should be configurable using the serializer interface proposed in this PR. You are right though per request overrides don't make sense in the Consumer. You can't actually pass your own schema per `poll()` in the AvroConsumer either. I just got ahead of myself.

confluent_kafka/producer.py
rnpridgeon: 1. I think is the most flexible and how I personally would like to proceed. 

2. Sharing an underlying librdkafka insance brings about its own problems. For instance what do you do when one of the other producer instances calls a blocking operation such as `flush`?

3. I think you should register alternative strategies with your serializer instead of with the producer. This keeps the state of stateful serializer where it belongs, in the serializer. 

Realistically I don't foresee anyone overriding the serializer per call. If they did there was no reason to implement generic serdes in the first place. This could have been done with the existing interface. Instead I think the serializer should have logic built into it which handles the various states you need it to. For Avro this would mean pulling the schema from the generic record, creating a datum writer if you don't have one for this schema already and then writing the record. This is the job of the serializer though not  the producer. The producer's job should be to handle over the necessary information and take the resulting output and that's it.

confluent_kafka/producer.py
rnpridgeon: What is the purpose on erroring out when serializers are not configured? The producers and consumers currently produce without serializer without issue. I do not see a reason to break this.

confluent_kafka/producer.py
rnpridgeon: We can call the serializer and hand off the message data in two places. 

1. In the extension, not the approach we took 
2. In python as we have done here. This means we do not have access to the message opaque pointer which is passed to produce however. 

If we went with step 1 we could simply provide the message state object with the data before calling serializer. Alternatively we could make a copy of the data and pass that to the message state object(which would be our opaque pointer). Either way we have access to the existing message state and the opaque pointer in the extension but not from python. 

For option 2 we can use a closure to bring the original message contents withit as we do in the avro-cli example. 

https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/avro-cli.py#L103

rnpridgeon: @mhowlett I've thought about this more and I'm not sure I agree that providing the unserialized payload is the correct move here. If we do this then the delivery report is reporting what we *wanted* to produce as opposed to what we actually did produce. That to me breaks the contract that the user has with the delivery report. 

Using a lambda to carry the payload with your callback seems like the right move here. This ensures that the delivery report upholds its intended purpose while giving users the flexibility to cater to more specific use cases.

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: ack, I'll add documentation that explains None, 0 == current time today but in the future 0 will mean 0

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+                **note** deserializers are responsible for handling NULL keys
+            :param func value_deserializer(topic, value): Converts message value bytes to object.
+                **note** deserializers are responsible for handling NULL values
+            :returns: A confluent_kafka.Message or None on timeout.
+            :raises RuntimeError: If called on a closed consumer.
+        """
+
+        msg = super(Consumer, self).poll(timeout)
+
+        if not msg or msg.error():
```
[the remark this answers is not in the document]
edenhill: https://amir.rachum.com/blog/2012/08/25/you-cant-handle-the-truth/

confluent_kafka/consumer.py
edenhill: auto.offset.reset=true -> enable.auto.offset.store=true

confluent_kafka/producer.py
edenhill: There's also the possibly undesired or unexpected reference to the original user object that would be kept alive and out of gc until the delivery report was triggered, which may have unwanted bad consequences (resource usage, etc)

confluent_kafka/consumer.py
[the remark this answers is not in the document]
edenhill: I think it is fine to enforce the callback being set in only one place, any other use is ambigious.
Let's follow Matt's rule of least surprise: enforce the callback is only specified in one place.
I don't consider that a breaking API change, it will only break bad apps.
It is literally just a couple of lines of code, something like:
```
  for var, name in [(logger, 'logger', on_commit, 'on_commit', ...)]:
     if var is not None and conf.get(name, None) is not None:
       raise ...
```

confluent_kafka/consumer.py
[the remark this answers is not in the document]
edenhill: Why are AvroProducer/AvroConsumer a bad idea?

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+            logger = conf.get('logger', None)
+        super(Consumer, self).__init__(conf, on_commit=on_commit, stats_cb=stats_cb,
+                                       throttle_cb=throttle_cb, logger=logger)
+
+    def poll(self, timeout=-1.0, key_deserializer=None, value_deserializer=None):
+        """
+            Consumes a message, triggers callbacks, returns an event.
+
+            The application must check the returned Message object’s Message.error() method to distinguish
+            between proper messages, an error(see error().code() for specifics), or an event.
```
edenhill: "an error (see msg.error().code() for specifics)"

confluent_kafka/consumer.py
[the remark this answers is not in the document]
edenhill: no

confluent_kafka/consumer.py
edenhill: I agree with Matt on 2, consume() should support serdes even with num_messages > 1.
The offset-store problem does not have anything to do with serdes being used or not.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if key_deserializer:
+            msg.set_key(key_deserializer(topic, msg.key()))
+
+        if not value_deserializer:
+            value_deserializer = self._value_deserializer
+
+        if value_deserializer:
+            msg.set_value(value_deserializer(topic, msg.value()))
+
```
edenhill: what are the performance implications of this wrapping and extra checks? (for both consumer and producer)
I want numbers!

confluent_kafka/producer.py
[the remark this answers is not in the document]
edenhill: We should probably bump MIN_RD_KAFKA_VERSION to v1.0.0, can you do that? And remove doc references to older versions.

The smaller support matrix the better

confluent_kafka/consumer.py
rnpridgeon: The `offset-store problem` was not specific to serialziation but instead to the use of Consume, and thus batch consuming, in general. 

I actually decided to forego implementing deserialization in the batch consumer based on a conversation you and I had previously. I'll go ahead and add it though as it seems there has been a change of heart. 

The question remains however. Do we or do we not want to deprecate the use of batch consumption.

confluent_kafka/consumer.py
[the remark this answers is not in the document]
rnpridgeon: Bad apps based off the fact we provided an API which allowed for it; sins of our fathers and what not. The amount of code is immaterial; the fact is we provided an API which allowed this. Aren't we under some obligation to deprecate this behavior responsibly? If you two think it's best then sure but this seems like the exact sort of thing that make API design so hard; you have to live with the decisions you made in the past for a long time.

confluent_kafka/consumer.py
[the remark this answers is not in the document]
rnpridgeon: >this is a new class 

Thank you for this @mhowlett , you are right and this got me thinking. The avro producer isn't actually allowing for serializer overrides so much as it's allowing you to optionally introduce a new datum writer by providing a new schema. Either way this is an example of the serializer's responsibility leaking into the producer itself which is not what we want. Instead the serializer should handle this. What makes this weird for Avro is that the python implementation does not actually have this notion of `Specific` or even `Generic` record types. If you want a record you pass a dict with no schema tied to it. That said this is a new class, and a new `GenericRecord` serializer(and supporting GenericRecord type (binding schema to dict)  can also be introduced 
 to handle this. In the end this will make for a much better Avro experience for everyone while also removing all these checks from the produce path.

confluent_kafka/consumer.py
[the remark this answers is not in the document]
edenhill: The previous behaviour was undefined and no priority was documented, fixing that is not an API breaking change since there were no promises made and an application that relies on both being specified is clearly wrong.

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: ack

confluent_kafka/avro/schema.py
```diff
@@ -0,0 +1,64 @@
...
+
+# Python 2 considers int an instance of str
+try:
+    string_type = basestring  # noqa
+except NameError:
+    string_type = str
+
+
+class GenericAvroRecord(dict):
+    """
```
mhowlett: Consider composition instead of inheritance here because 1. it's a looser form of coupling (allows for more future flexibility), so is generally a better default 2. composition is used in other languages, e.g. https://github.com/confluentinc/avro/blob/master/lang/csharp/src/apache/main/Generic/GenericRecord.cs

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+        raise ValueError("Both Topic and Schema must be set when using TopicRecordNameStrategy")
+    return "-".join([topic, schema.fullname])
+
+
+class AvroSerializer(object):
+    """
+    Encodes kafka messages as Avro; registering the schema with the Confluent Schema Registry.
+
+    :param registry_client CachedSchemaRegistryClient: Instance of CachedSchemaRegistryClient.
+    :param bool is_key: True if configured as a key_serializer.
```
mhowlett: It a much better abstraction imo if `is_key` is specified at serialization time, not as a constructor argument. that would also give consistency with how you've implemented `subject_strategy`.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+        return False
+
+
+def TopicNameStrategy(topic=None, is_key=False, schema=None):
```
mhowlett: `DefaultSubjectNameStrategy` is more descriptive.

confluent_kafka/avro/schema.py
```diff
@@ -0,0 +1,64 @@
...
+        self.schema = schema
+        if record is not None:
+            self.update(record)
+
+    def put(self, key, value):
+        self[key] = value
+
+
+def get_schema(datum):
+    if isinstance(datum, GenericAvroRecord):
```
mhowlett: useful function, but doesn't belong in this file - belongs with `AvroSerializer`. could be a private `@staticmethod` there.

confluent_kafka/avro/serializer/__init__.py
mhowlett: oh, i see you have others. `TopicNameStrategy` makes sense.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+
+    __slots__ = ["registry_client", "codec_cache", "is_key", "subject_strategy"]
+
+    def __init__(self, registry_client, is_key=False, subject_strategy=TopicNameStrategy):
+        self.registry_client = registry_client
+        self.codec_cache = {}
+        self.is_key = is_key
+        self.subject_strategy = subject_strategy
+
+    def __call__(self, topic, record):
```
mhowlett: In C#, the serialize method is provided with everything - suggest doing the same (this was well thought through!)
https://github.com/confluentinc/confluent-kafka-dotnet/blob/1.0.x/src/Confluent.Kafka/ISerializer.cs#L47

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+    """
+    Decodes Kafka messages encoded by Confluent Schema Registry compliant Avro Serializers.
+
+    :param registry_client CachedSchemaRegistryClient: Instance of CachedSchemaRegistryClient.
+    :param bool is_key: True if configured as a key_serializer.
+    :param schema reader_schema: Optional reader schema to be used during deserialization.
+    """
+    __slots__ = ["registry_client", "codec_cache", "is_key", "reader_schema"]
+
+    def __init__(self, registry_client, is_key=False, reader_schema=None):
```
mhowlett: same comments re: `is_key` and `__call__` method signature as serializer.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,162 @@
...
+        :param func on_commit(err, [partitions]): Callback used to indicate success or failure
+            of an offset commit.
+        :param func stats_cb(json_str): Callback for statistics emitted every ``statistics.interval.ms``.
+            See https://github.com/edenhill/librdkafka/wiki/Statistics” for more information.
+        :param func throttle_cb(confluent_kafka.ThrottleEvent): Callback for throttled request reporting.
+        :param logging.handlers logger: Forwards logs from the Kafka client to the provided handler instance.
+            Log messages will only be forwarded when ``client.poll()`` or ``producer.flush()`` are called.
+        :raises TypeError: If conf is not a dict.
+    """
+    def __new__(cls, *args, **kwargs):
```
mhowlett: fancy. this seems like a good solution to the requirements to me, though the scenario is unique to python and i'm not 100% confident i'm not missing any downsides.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,162 @@
...
+        :raises TypeError: If instantiated directly.
+    """
+
+    __slots__ = ["_key_deserializer", "_value_deserializer"]
+
+    def __new__(cls, *args, **kwargs):
+        raise TypeError("SerializingConsumer is a non user-instantiable class")
+
+    @staticmethod
+    def byteSerializer(topic, data):
```
mhowlett: best thought of an implementation detail, mark as private.

confluent_kafka/consumer.py
mhowlett: Not sure about the details, but it seems like we might ideally want a super fast path for (at least) `UTF8`, `int` and `long` where the deserialization happens in C. Have you investigated a three-pronged approach (no/fast/custom deserialization) along these lines? I think we might get string for free with python 2, but not python 3... not an expert though and haven't investigated.

tests/avro/test_avro_serializer.py
```diff
@@ -0,0 +1,83 @@
...
+
+from tests.avro import data_gen
+from confluent_kafka.avro.serializer import AvroSerializer, AvroDeserializer
+from confluent_kafka.avro.schema import GenericAvroRecord
+from tests.avro.mock_schema_registry_client import MockSchemaRegistryClient
+from confluent_kafka import avro
+
+
+class TestAvroSerializer(unittest.TestCase):
+    def setUp(self):
```
mhowlett: it would be good also to supply off-the-shelf implementations for other common data types in this PR. amongst other things, it will validate the generality of the approach taken.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,139 @@
...
+            :param int timestamp: Message timestamp (CreateTime) in milliseconds since epoch UTC
+                (requires librdkafka >= v0.9.4, api.version.request=true, and broker >= 0.10.0.0).
+                Default value is current time.
+            :param dict|list headers: Message headers to set on the message. The header key must be a string while
+                the value must be binary, unicode or None. Accepts a list of (key,value) or a dict.
+                (Requires librdkafka >= v0.11.4 and broker version >= 0.11.0.0)
+        """
+
+        super(SerializingProducer, self).produce(topic, self._value_serializer(topic, value),
+                                                 self._key_serializer(topic, key), partition, on_delivery=on_delivery,
```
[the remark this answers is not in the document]
edenhill: I think it is a bad idea to hold on to the original key/value object (pre-serialization) since it will keep them from being garbage collected.
A user can bind the on_delivery callback to their original object to have it available in the delivery report callback.

tests/avro/test_avro_serializer.py
edenhill: nit: setUp casing seems atypical to me, setup.

confluent_kafka/avro/schema.py
rnpridgeon: This is actually quite intentional as it allows me bind the schema with the record without needing to dereference the record contents in the serializer. Since all records are represented as dicts and the datum writer will only accept a dict I think this is the right move for python.

confluent_kafka/avro/schema.py
rnpridgeon: Fair point

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: All naming conventions come directly from the Java implementation to keep things simple.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: I disagree. Since I am explicitly setting a key or value serializer I see no need to restate that with each record I intend to serialize. Subject naming strategies are different as they are a schema registry client construct and not directly related to serialization.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: I considered this but instead decided to convert my serializer into a callable as you see here. I believe this adds maximum flexibility and ease of use.

confluent_kafka/consumer.py
rnpridgeon: It's not terribly easy to read but it does allow me to offer both APIs at almost no cost to those who wish to forego using the serializing producer/consumer

confluent_kafka/consumer.py
rnpridgeon: Originally it was set as a parameter default which is why I didn't mark it private. Then I realized bad things will happen if you pass None. Thanks for reminding me to come back to this!

confluent_kafka/consumer.py
rnpridgeon: All Python objects are of type PyObject and will need to be converted to their c counterparts so I'm not sure I understand what you mean here. Even bytes are of type PyBytes in Py3 so I'm not sure you get anything free in either version. 

https://docs.python.org/2/c-api/string.html

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: GC aside I personally feel that violates the implicit contract that comes with the delivery report. The way I interpret it the DR is meant to tell me what was produced not what I meant to produce. So the counter argument is that this gives me quick validation that my payload was serialized as I had intended. If I need verification it was serialized correctly I can use a lambda to carry state with it. Otherwise I have no way of telling if it was actually serialized unless I consume the data back down.

tests/avro/test_avro_serializer.py
rnpridgeon: ack
ack

confluent_kafka/avro/serializer/__init__.py
mhowlett: why is needing to restate this with each record serialized a disadvantage? it's a trivial internal detail. by contrast, having serializer instances tied keys or values results in an externally exposed abstraction that is more complex. strongly disagree with your disagree.

confluent_kafka/avro/schema.py
mhowlett: i don't quite understand what you're trying to say, but there would be only minor difference in code between the two approaches (the composition approach would be a bit more but not much). Anyway, I'm ok with the inheritance relationship here.

rnpridgeon: I guess I'm saying why call serialize(record.data) vs serialize(data). I'm not convinced I gain anything from composition here. On the contrary however Users can't treat this GenericAvroRecord exactly as if it were a dictionary. Just  like they could before I introduced a new type. i.e. record['stuff']='a' or record.get('stuff',None).

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: But I already have serializer instances tied to key or values do I not. That's how I instantiate the producer/consumer.  The type is not tied to a key or value implementation only the instance which you configure the client with. I guess in theory I could share the same instance between my key and value serializers but I'm not sure I gain much there.

confluent_kafka/avro/serializer/__init__.py
mhowlett: you only pass the topic and data to be serialized to the serializer. I'm suggesting you provide everything, to be maximally flexible.

confluent_kafka/producer.py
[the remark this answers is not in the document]
mhowlett: 'quick validation that my payload was serialized as I had intended' - it seems like you don't trust python or your serialization format of choice? Can I suggest you get a new language or serialization format of choice? No one should want/need to do this. Also, if they did for some reason, it is possible for them to drive the API such that they can (but that should not be the easy path).

Re: additional memory use: there should be a config switch to allow this to be turned off for people who don't need it / don't want to pay for the memory overhead. In C# I have it on by default so people don't get confused with null Key / Value fields (i.e. cater for beginner by default), but in python you could have it off by default since you would just not provide the fields.

Re: binding this to the on_delivery callback: you're right, but users couldn't use a static callback method. If the user does want state maintained, it will very often be unserialized message data, so it's is a nice feature to provide this out of the box, allowing for simpler usage patterns in the common case.

If you do want to expose the serialized bytes and not the original data (i think this is not a good idea), how about using a different name, to indicate this is the serialized data.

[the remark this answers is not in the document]
edenhill: The problem is that it will hold on to any resources referenced by the object, say a database response, a file, or something else that is costly.
If you produce 1 gig of data with this approach, your application will, worst case, peak out at 2 gigs of memory usage just for the original objects and the copied message payloads.


As for validating the serialized data, I agree with Matt that's not valuable.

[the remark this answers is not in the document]
mhowlett: holding onto additional resources would certainly be an important (dominating) consideration if that were likely, but I don't think it is. I'd be convinced if i saw a concrete example of where you think someone might accidentally do this.

re memory usage, yes, which is why you need a flag to turn it off (or a flag to turn it on for those who want the feature - simply not providing the property in python is an option - in statically typed languages it's not, hence my preference for it to be on by default there to avoid confusion). that said, i'd also be inclined to err on the side of non-power-user in python as well.

confluent_kafka/avro/serializer/__init__.py
mhowlett: > But I already have serializer instances tied to key or values do I not.

my first point is you are making the user specify this. they don't have to. 

>  I guess in theory I could share the same instance between my key and value serializers but I'm not sure I gain much there.

you gain simplicity in structure. the serializer is a simpler thing. it might seem small, but 100 like things across the library adds up to a lot.

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: Right I'm not saying serialization verification is the primary motivator here just that you can still do that even without returning the message in its original form. Instead I was referring to the fact that if needed you can absolutely see both forms of the message with minimal effort. A `delivery report` in my eyes should be reporting on the actual message produced. Which in this case is the post-serialized form. Either way providing a lambda as a callback requires a marginal amount of additional effort over providing a configuration flag.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: Realistically it's only 2 like things right. You can share the serializer the key serializer with as many producer instances as you see fit. 

I do understand what you are saying though with regard to constructing the serializer itself. We don't actually remove any inflexibility here though we just move it to another place. I would argue that most serializers will not care if they are working on a key or a value. The only reason the Avro Serializer cares is because it needs to know this information in order to register the schema using the default subject naming strategy. If it weren't for the original design of the subject registration strategy Avro too would not care if it were serializing a key or a value.

rnpridgeon: Also note specifying a key is specific to the Avro serializer not serializers in general. 

I should actually clarify that in the doc string as the param doc is a bit misleading. I use the words `key` and `value` in the interface for a serializer but really I just mean `payload`. The fact it is the message's key or value is actually irrelevant.

confluent_kafka/producer.py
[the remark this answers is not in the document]
edenhill: I would argue that with a serializer the key and value of a message isn't of much interest in the delivery report, the original object is.

Agree that a lambda or object-bound method is sufficient to pass state between produce and delivery report.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: I don't think serializers care nor should they care if the contents they are serializing are a key or not. This is more a schema registry client constraint which leaked into the java clients which we are still paying for.

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: When your writing code in a scripting fashion, as you are in python, quick indicators of success are handy. If I send `hello world` as  as a string that I wanted to be base64 encoded it's nice to have some sort of visual indication that it worked. I'm not saying this is the right way to approach things but it's how a lot of people make incremental progress with more complex bash/python scripts. You write some code, you run it, you sanity check the output, you write some more code.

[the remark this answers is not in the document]
edenhill: That doesn't belong here, you do that verification on the serializer.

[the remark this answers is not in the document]
rnpridgeon: That' fine, as I said it may not be right but it happens. I also agree with the GC implications either way so I think for the time being this particular issue is resolved.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: Honestly the same should probably be true for `topic` but once we get to that point we have pushed serialization down to the per message level which can already be done without changes to the client.

confluent_kafka/consumer.py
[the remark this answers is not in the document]
mhowlett: re: @edenhill's question. At one point with the .NET client, I was seriously considering separate producer/consumer and avro producer/consumer classes. The motivation for this was the observation that the serde interfaces are very simple, except for the requirements of the SR integration. It seemed reasonable to have P/Cs for simple serdes (which furthermore allowed for beautiful terse inline specification of them), which was applicable to most scenarios and defer avro functionality to elsewhere. Initially, i was thinking this could even be done with C# extension methods, not requiring a the classes for avro (but that was not the case). But it turned out to work much better to have completely general serde delegate types (two of them, one async, one not) and have the structure @rnpridgeon is suggesting above. @rnpridgeon's comment about avro being too specific also re-enforces this choice. 

@rnpridgeon - don't quite follow what you're saying, but broadly sounds good and look forward to reviewing it.

confluent_kafka/producer.py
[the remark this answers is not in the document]
mhowlett: to be concrete, to get the original k/v in the delivery report, I think the simplest you could do is:

```
if __name__ == '__main__':
    ....
    for line in sys.stdin:
        def line_cb_creator(l):
            def line_cb(err, msg):
                sys.stderr.write('delivered: %s' % l)
            return line_cb
        p.produce(topic, line.rstrip(), callback=line_cb_creator(line))
        p.poll(0)
```

the lexical scope of the loop parameter in python here is the main function, which is why the `line_cb_creator` function is necessary. I think that's too hard for an 'easy to use' python api, and think that at least ultimately we would want to plan to be adding the original values to the delivery report, and so need to make sure the behavior introduced now doesn't conflict with the foreseen API. we include serialized data in `key()` and `value()` in the existing api. I think that means that to avoid making breaking changes, we need to overload those methods now to return the unserialized data.

confluent_kafka/avro/serializer/__init__.py
mhowlett: > We don't actually remove any inflexibility here though we just move it to another place.

that other place is internal, so it's hidden from the user - big difference. the way you have it results in a more complex API for no benefit.

confluent_kafka/avro/serializer/__init__.py
mhowlett: > I don't think serializers care nor should they care if the contents they are serializing are a key or not.

yours do!

mhowlett: of course if you push all this to the message level, the user of the API would need to specify stuff twice (topic, key/value), so it's not an option, also the client can't encapsulate DRs with the original objects.

I iterated on many different strategies for this with the .net client, and the only one i couldn't fault was passing everything that could ever possibly wanted into the serdes: https://github.com/confluentinc/confluent-kafka-dotnet/blob/1.0.x/src/Confluent.Kafka/ISerializer.cs#L47

confluent_kafka/avro/schema.py
rnpridgeon: Sure thing the other thing that is nice is your record types work exactly as they did in the past with no material changes. You can still use `record.put` and `record[key]` as you did previously since all records are dicts

confluent_kafka/producer.py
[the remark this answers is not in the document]
rnpridgeon: For now we are going to move forward with leaving the delivery report key/value in their serialized form.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
-    pass
+    def __new__(cls, message):
+        return super(KeySerializerError, cls).__new__(SerializerError, message, True)
 
 
 class ValueSerializerError(SerializerError):
     pass
+
+
+class ContextStringIO(io.BytesIO):
```
swenzel: I always found this thing funny... you do know that there is `contextlib.closing`, right?

confluent_kafka/avro/schema.py
```diff
@@ -0,0 +1,64 @@
...
+    elif (datum is None):
+        return PrimitiveSchema("null")
+    elif isinstance(datum, string_type):
+        return PrimitiveSchema('string')
+    elif isinstance(datum, bool):
+        return PrimitiveSchema('boolean')
+    elif isinstance(datum, int):
+        return PrimitiveSchema('int')
+    elif isinstance(datum, float):
+        return PrimitiveSchema('float')
```
mhowlett: python floats are double precision and if you don't use a `'double'` primitive schema here, serialization will be lossy in most cases which seems bad.

same with int - python seems to have unified int's to be mean long in more recent versions of the language, and I think just using a `'long'` primitive schema type for all int values is best.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -15,11 +15,39 @@
...
+    HAS_FAST = True
+except ImportError:
+    pass
+
 
 class SerializerError(Exception):
     """Generic error from serializer package"""
 
-    def __init__(self, message):
+    def __new__(cls, message, is_key=False):
```
mhowlett: I don't think you'd ever want to have different handlers for key vs value serialization errors, so I don't think you need two types - just `SerializerError` would be better I think, with a property to indicate whether it's a key or not.

Also, you've used `__new__` again here. This seems like a bad pattern to me - the benefit of doing it your way is it makes usage terser, but in general, this is at the expense of giving up on the idea of knowing WTF you're going to get on instantiation without looking at the implementation. I'm not familiar with working in languages that have this capability, but can't see how this tradeoff is a good one.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+    """
+    __slots__ = ["registry_client", "codec_cache", "is_key", "reader_schema"]
+
+    def __init__(self, registry_client, is_key=False, reader_schema=None):
+        self.registry_client = registry_client
+        self.codec_cache = {}
+        self.is_key = is_key
+        self.reader_schema = reader_schema
+
+    def __call__(self, topic, datum):
```
mhowlett: `datum` is not a good name for the serialized data, as it suggests the actual object itself (especially in avro terminology). maybe `payload`?

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+        :param str|bytes or None datum: message key or value to be decoded.
+        :returns: Decoded message key or value contents.
+        :rtype GenericAvroRecord:
+        """
+
+        if datum is None:
+            return None
+
+        if len(datum) <= 5:
+            raise SerializerError("message is too small to decode")
```
mhowlett: more of a `DeserializerError`

mhowlett: also text suggestion: 'deserialization error: message too small'.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+        with ContextStringIO(datum) as payload:
+            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            if magic != MAGIC_BYTE:
+                raise SerializerError("message does not start with magic byte", self.is_key)
+
+            decoder_func = self._get_decoder_func(schema_id, payload)
+            return decoder_func(payload)
+
+    def _get_decoder_func(self, schema_id, payload):
+        if schema_id in self.codec_cache:
```
mhowlett: you call the things you put in here decoders. If you use a different word (`codec`) for the name of the collection, it's less obvious what the collection is for.

confluent_kafka/avro/serializer/__init__.py
```diff
@@ -33,8 +61,270 @@ def __str__(self):
...
+            return self.codec_cache[schema_id]
+
+        try:
+            writer_schema = self.registry_client.get_by_id(schema_id)
+        except ClientError as e:
+            raise SerializerError("unable to fetch schema with id %d: %s" % (schema_id, str(e)))
```
mhowlett: I think this is throwing away useful information - users may want to handle connectivity issues differently than data serialization errors, so it would be good to be able to catch on the different types. I think it's ok to just let the `ClientError` bubble up. i.e. probably going over-board to invent a special type of serializer error for connectivity issues.

confluent_kafka/consumer.py
mhowlett: Here's what the manual says on `__new__`:

>__new__() is intended mainly to allow subclasses of immutable types (like int, str, or tuple) to customize instance creation. It is also commonly overridden in custom metaclasses in order to customize class creation.

this is neither of those use cases, which makes me wary. I still **think** it's justified as a performance optimization (but again, wary that there might be something important i don't know about). i'd be interested in the perf difference between using a byte deserializer and bypassing the `DeserializingConsumer` as you're doing now as the former is more conservative.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,162 @@
...
+            Log messages will only be forwarded when ``client.poll()`` or ``producer.flush()`` are called.
+        :raises TypeError: If conf is not a dict.
+    """
+    def __new__(cls, *args, **kwargs):
+        if 'key_deserializer' in kwargs or 'value_deserializer' in kwargs:
+            return super(Consumer, cls).__new__(SerializingConsumer, *args, **kwargs)
+        return super(Consumer, cls).__new__(cls, *args, **kwargs)
+
+
+class SerializingConsumer(Consumer):
```
mhowlett: doesn't do any serializing, so perhaps `DeserializingConsumer` is a better name?

confluent_kafka/producer.py
```diff
@@ -0,0 +1,139 @@
...
+    @staticmethod
+    def byteSerializer(topic, data):
+        """ Pass-through serializer """
+        return data
+
+    # conf must remain optional as long as kwargs are supported
+    def __init__(self, conf={}, key_serializer=None, value_serializer=None,
+                 error_cb=None, stats_cb=None, throttle_cb=None, logger=None, **kwargs):
+
+        if not isinstance(conf, dict):
```
mhowlett: same comments here on type checking arguments and prefer not to have the fancy `__new__` (depending on perf benefit).

confluent_kafka/avro/schema.py
rnpridgeon: I did forget to add the double, I'll add that in. I do not think we should be silently promoting floats. The avro implementation actually ends up calling write_long within write_int anyway so there is nothing to be gained there. It's much more straightforward to say an int is an int. 

https://github.com/apache/avro/blob/master/lang/py/src/avro/io.py#L300-L304

confluent_kafka/consumer.py
rnpridgeon: >It is also commonly overridden in custom metaclasses in order to customize class creation.

Technically the extension's implementation is not a metaclass no but is this not an example of customizing class creation?

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: >yours do! 

That's not technically true, the registry cares about this not the serializer. The serializer itself is agnostic to the this idea of key or value. I guess technically the serializer cares in that it throws a different type of SerializerException but as you later note this may not really be needed or desired in the first place.

confluent_kafka/producer.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Callbacks can be set in the conf dict or *ideally* as parameters.
+        # Raise a SyntaxError if a callback is set in both places.
+        for var, name in [(logger, 'logger'), (error_cb, 'error_cb'),
+                          (stats_cb, 'stats_cb'), (throttle_cb, 'throttle_cb')]:
+            if all([var, conf.get(name, None)]):
+                raise SyntaxError("{} parameter repeated".format(name))
+            if var is None:
+                var = conf.get(name, None)
+
```
redsk: Unless here I add
```
conf.pop("key_serializer")
conf.pop("value_serializer")
```

instance creation fails with 
```
cimpl.KafkaException: KafkaError{code=_INVALID_ARG,val=-186,str="No such configuration property: "key_serializer""}
```

I think `librdkafka` does not accept these properties, I tried with both `1.0.0rc5` and `master`.

confluent_kafka/avro/schema.py
mhowlett: python doesn't have any data type that corresponds to avro's float or avro's int.

python float is equivalent to avro double.

pre-python3, I *think* python int's are platform dependent. On my machine they'e 64 bit (i.e. equivalent to avro long).

in python3, all integer values are arbitrary precision (i.e. there is no corresponding avro type, but the best choice is long, presumably the avro library has some error checking).

confluent_kafka/avro/serializer/__init__.py
mhowlett: it is true - a serializer instance is specifically for either serializing keys or values.
it's both bad UX to need to specify this at construction time, and is more prone to error than what i'm suggesting.

tests/integration/integration_test.py
```diff
@@ -1224,6 +1227,83 @@ def verify_avro_explicit_read_schema():
...
+    from base64 import standard_b64decode as b64decode
+
+    from confluent_kafka.avro.serializer import (AvroSerializer,
+                                                 AvroDeserializer,
+                                                 RecordNameStrategy,
+                                                 TopicRecordNameStrategy)
+    from confluent_kafka.avro.schema import GenericAvroRecord
+    from confluent_kafka import avro
+
+    def b64_decoder(topic, data):
```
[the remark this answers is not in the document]
rnpridgeon: I didn't plan to originally no as its fairly trivial to derive your own serde implementations. Presumably these are serdes that ship with .NET now?

confluent_kafka/producer.py
rnpridgeon: ack, I'm partial to checking the type over try except personally but I don't have terribly strong feelings either way.

confluent_kafka/consumer.py
```diff
@@ -0,0 +1,162 @@
...
+    @staticmethod
+    def byteSerializer(topic, data):
+        """ Pass-through serializer """
+        return data
+
+    # conf must remain optional as long as kwargs are supported
+    def __init__(self, conf={}, key_deserializer=None, value_deserializer=None,
+                 on_commit=None, stats_cb=None, throttle_cb=None, logger=None, **kwargs):
+
+        if not isinstance(conf, dict):
```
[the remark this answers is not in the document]
rnpridgeon: ack

confluent_kafka/consumer.py
rnpridgeon: Yeah, I thought I snagged this name from .NET but I guess not. Will fix.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: ack

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: Bear with me here as I know this goes against all my other arguments where I say I don't want to let Avro dictate python's serde API... Where it makes sense I'm trying to make the new serde api match the behavior of the existing Avro Producer and Consumer. For this reason we differentiate between a key and value serializer. I personally do not have strong feelings for either approach. If you on the other hand feel strongly that this should just be one exception type we can back it out. 

As for `__new__` I honestly don't feel this is too terrible when the resulting object is still the same class. There are no real surprises here, I wanted a Serialization exception I got one. This one is just a tad more specific than the other but they will both be caught by except SerializerError. If we decide to just return one SerializationError type however there is no need for this at all. Anyway let me know what you think, as I said I don't have particularly strong feelings either way. Now that you know my intent I'll defer to your better judgement here.

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: Nope, although to be honest I just reused the existing code. Patches welcome though :)

confluent_kafka/avro/serializer/__init__.py
rnpridgeon: Honestly if I could get away with it I wouldn't even pass the topic name in as this is an example of the Registry Client API leaking into our serde API. I would prefer to just pass in a function pointer for some callable which is agnostic of both topic and key to make things incredibly simple. That's the case though and since `is_key` matches the java serializer API I'll go ahead and add it in despite my feelings about it.

mhowlett: the need to pass in a `topic` 'cause of SR kind of ruins the api from the perspective of being able to beautifully specify callables inline. Furthermore, I think you'll also find that this isn't something you can do in most cases anyway. For example, how easy is it to write a bigendian/littleendian architecture aware integer serializer? I haven't looked into this in python, but i suspect it's something you wouldn't define inline (and probably something we'll provide with the library). So this idea - although very attractive (i was also attracted to it initially) - is the wrong thing to be prioritizing. what you have is kind of a middle ground that isn't optimal in any way.

given the above, I think the most important thing to be optimizing on is the point of view of someone using an existing serializer, and the callable interface should be extended to be completely general - include everything that could ever possibly be wanted (more than topic), including `is_key`. this simplifies the surface area of the parts of the library users will see/use in most cases. it's also allows for maximum flexibility. and it's also not a half-arsed middle ground - it's taking a point of view that is well justified.

ferozed:
hi! When will this PR be merged ?

rnpridgeon:
duplicate of #787

PR #506, opened 2018-12-08
Make group.id required for consumer instantiation(#472)

rnpridgeon:
I accidentally trashed the original PR.

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -1386,7 +1386,11 @@ static int Consumer_init (PyObject *selfobj, PyObject *args, PyObject *kwargs) {
 
         rd_kafka_poll_set_consumer(self->rk);
 
-        self->u.Consumer.rkqu = rd_kafka_queue_get_consumer(self->rk);
+        if (!(self->u.Consumer.rkqu = rd_kafka_queue_get_consumer(self->rk))) {
+                PyErr_SetString(PyExc_ValueError,
+                                "Failed to initialize consumer: group.id must be set.");
+                return -1;
+        }
 
         return 0;
 }
@@ -1423,8 +1427,8 @@ PyTypeObject ConsumerType = {
         "\n"
         ".. py:function:: Consumer(config)\n"
         "\n"
-        "  :param dict config: Configuration properties. At a minimum ``bootstrap.servers`` and "
-        "``group.id`` **should** be set"
+        "  :param dict config: Configuration properties. At a minimum ``group.id`` **must** be set,"
+                " ``bootstrap.servers`` **should** be set"
         "\n"
         "Create new Consumer instance using provided configuration dict.\n"
         "\n"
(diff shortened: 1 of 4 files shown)


confluent_kafka/src/Consumer.c
```diff
@@ -1386,7 +1386,11 @@ static int Consumer_init (PyObject *selfobj, PyObject *args, PyObject *kwargs) {
 
         rd_kafka_poll_set_consumer(self->rk);
 
-        self->u.Consumer.rkqu = rd_kafka_queue_get_consumer(self->rk);
+        if (!(self->u.Consumer.rkqu = rd_kafka_queue_get_consumer(self->rk))) {
```
edenhill: This is checking a side-effect of not setting group.id, that might change in librdkafka.
Explicitly check that group.id was configured.

You can leave this check in, but make sure that memory resources are freed up in the error case, this is leaking memory.

tests/test_Consumer.py
```diff
@@ -313,3 +313,11 @@ def test_calling_store_offsets_after_close_throws_erro():
     with pytest.raises(RuntimeError) as ex:
         c.offsets_for_times([TopicPartition("test", 0)])
     assert 'Consumer closed' == str(ex.value)
+
+
+def test_consumer_withot_groupid():
+    """ consumer should raise exception if group.id is not set """
```
edenhill: Capital C

tests/test_misc.py
```diff
@@ -155,7 +155,8 @@ def test_unordered_dict(init_func):
     """
     Interceptor configs can only be handled after the plugin has been loaded not before.
     """
-    client = init_func({'confluent.monitoring.interceptor.publishMs': 1000,
+    client = init_func({'group.id': "test-group",
```
edenhill: single quotes

rnpridgeon: Since there doesn't seem to be a strict convention around when to use one or the other I've been trying to come up with what I think feels right. I was thinking that identifiers such as dict keys should be in single quotes while there values should be in double quotes. Thoughts?

edenhill: Right, but there is existing code in this dict that uses single quotes for values, so that is what should be used for consistency.

confluent_kafka/src/Consumer.c
rnpridgeon: Good call, I didn't think about how closely this couples the validation to the librdkafka implementation. I'll remove it in favor of a check for 'group.id`'s presence. And clean up after myself of course.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -1629,6 +1629,18 @@ rd_kafka_conf_t *common_conf_setup (rd_kafka_type_t ktype,
                 PyDict_Update(confdict, kwargs);
         }
 
+        if (ktype == RD_KAFKA_CONSUMER) {
+                ko = cfl_PyUnistr(_FromString("group.id"));
+                if (PyDict_Contains(confdict, ko) != 1) {
```
edenhill: Using `PyDict_GetItemString()` avoids the extra object ko object creation.

rnpridgeon: good point, I see now it returns a borrowed reference; stand-by I'll fix that I just though `_Contains` felt more propper.

PR #510, opened 2018-12-14
Added the capability to retrieve a specific version for a subject

jpardobl:
- Added method get_schema_by_version to confluent_kakfa.avro.CachedSchemaRegistryClient. This method retrieves a specific version for a subject.

 - Modified method get_latest_schema from confluent_kakfa.avro.CachedSchemaRegistryClient. Now this method uses method get_schema_by_version passing latest as the version to retrieve

  **** All the above changes keep total backward compatibility ***

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -275,8 +275,6 @@ class CachedSchemaRegistryClient(object):
             except ClientError as e:
                 # bad schema - should not happen
                 raise ClientError("Received bad schema (id %s) from registry: %s" % (schema_id, e))
-
-
     
     
     def get_schema_by_version(self, subject, version):
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -276,7 +276,10 @@ class CachedSchemaRegistryClient(object):
                 # bad schema - should not happen
                 raise ClientError("Received bad schema (id %s) from registry: %s" % (schema_id, e))
 
-    def get_latest_schema(self, subject):
+
+    
+    
+    def get_schema_by_version(self, subject, version):
         """
         GET /subjects/(string: subject)/versions/(versionId: version)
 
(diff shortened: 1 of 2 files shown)


rnpridgeon:
Seems reasonable to me, please sign the CLA and extend the existing schema registry client integration tests to cover this new feature. 

https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/integration/integration_test.py#L737

jpardobl:
Already done both steps: added the tests and signed off the CLA. Thanks!

cricket007:
Handled this in a bit safer way, that more closely matches the Java API. Ref #515

rnpridgeon:
Thanks @cricket007 and @jpardobl. 

@cricket007 as discussed in pr #515 I don't think it's necessary to handle integer overflow issues on the client. Please feel free to follow up in the PR if you have further questions

@jpardobl in addition to the unit test you provided can you please add an integration test as well. You can simply call `get_compatibility` again with the `latest_id` returned earlier in the test and ensure you get the same value. 

https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/integration/integration_test.py#L737-L756

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -315,6 +316,25 @@ def get_latest_schema(self, subject):
 
         self._cache_schema(schema, schema_id, subject, version)
         return (schema_id, schema, version)
+    
+
+    def get_latest_schema(self, subject):
+        """
+        GET /subjects/(string: subject)/versions/(versionId: version)
```
cricket007: `(versionId: version)` should be 'latest' here

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -291,7 +292,7 @@ def get_latest_schema(self, subject):
         :returns: (schema_id, schema, version)
         :rtype: (string, schema, int)
         """
-        url = '/'.join([self.url, 'subjects', subject, 'versions', 'latest'])
+        url = '/'.join([self.url, 'subjects', subject, 'versions', version])
```
cricket007: The join will fail if the version isn't a string, so best to put `str(version)`

confluent_kafka/avro/cached_schema_registry_client.py
rooom13: why not adding a parameter "version" so we could retrieve the schema given a subject & the verssion?

rooom13:
Tha issue would really help a lot

rnpridgeon:
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/schema_registry/schema_registry_client.py#L492

PR #515, opened 2018-12-21
Add in 'versionId' rather than only 'latest' to GET /subjects/(string: subject)/versions/

cricket007:
From what I can tell, the schema URL call always looks like this

```python
url = '/'.join([self.url, 'subjects', subject, 'versions', 'latest'])`
```

Which doesn't adhere to the parameterized API

```
GET /subjects/(string: subject)/versions/(versionId: version)
```

This PR is to add in the versionId as a method parameter. 

---

Mocks and tests still need updated, probably

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -276,7 +276,7 @@ class CachedSchemaRegistryClient(object):
                 # bad schema - should not happen
                 raise ClientError("Received bad schema (id %s) from registry: %s" % (schema_id, e))
 
-    def get_latest_schema(self, subject):
+    def get_schema(self, subject, version_id='latest'):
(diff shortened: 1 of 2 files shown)


rnpridgeon:
Although normally I agree that it's best to fail early I think this particular instance is a case of `optimizing for failure` which we do not want to do. 

If a version id in excess of `Integer.MAX_VALUE` is passed the Schema Registry it will actually hit a `NumberFormatException` upon parsing the value. This exception will be caught  and then the schema registry will then throw an `InvalidVersionException` which should ultimately be returned to the client. 

`https://github.com/confluentinc/schema-registry/blob/93c0297b11ccc755e232ba77de87f88819ee3958/core/src/main/java/io/confluent/kafka/schemaregistry/rest/VersionId.java#L34

PR #516, opened 2019-01-05
update, fix cached registry client and ClientError

Rosswell:
- updated registry client docstrings
- removed ability specify subject when acting on /config endpoint, as it is invalid
- normalized string formatting for log.error and ClientError strings
- changed ClientError calls to use http_code param. error message changed to reflect usage
- updated involved integration test assertions to reflect new signature
- minor formatting changes

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -295,7 +295,7 @@ class CachedSchemaRegistryClient(object):
         :param str subject: subject name
         :raises ClientError: if the schema is invalid
         :returns: (schema_id, schema, version)
-        :rtype: (string, schema, int) | (None, None, None)
+        :rtype: (str, schema, int) | (None, None, None)
         """
         url = '/'.join([self.url, 'subjects', subject, 'versions', 'latest'])
 
@@ -362,7 +362,7 @@ class CachedSchemaRegistryClient(object):
         By default the latest version is checked against.
         :param str subject: subject name
         :param schema avro_schema: Avro schema
-        :param string | int version: version of the schema to test
+        :param str | int version: version of the schema to test
         :return: True if compatible, False if not compatible
         :rtype: bool
         """
@@ -394,14 +394,18 @@ class CachedSchemaRegistryClient(object):
         compatibility level.
 
         :param str level: ex: 'NONE','FULL','FORWARD', or 'BACKWARD'
+        :param str subject: subject of the schema whose compatibility is to be updated (optional)
         :raises: ClientError: if request was unsuccessful or an invalid compatibility level was provided
         :returns: new compatibility level, one of 'NONE','FULL','FORWARD', or 'BACKWARD'
         :rtype: str
         """
         if level not in VALID_LEVELS:
-            raise ClientError("Invalid level specified: {}".format(level))
+            raise ClientError("Invalid compatibility level specified: {}".format(level))
 
-        url = '/'.join([self.url, 'config'])
+        if subject:
+            url = '/'.join([self.url, 'config', subject])
+        else:
+            url = '/'.join([self.url, 'config'])
 
         body = {"compatibility": level}
         result, code = self._send_request(url, method='PUT', body=body)
@@ -416,11 +420,15 @@ class CachedSchemaRegistryClient(object):
         Get the current subject-level compatibility level if subject is specified. Otherwise, get the current global
         compatibility level for the schema registry.
 
+        :param str subject: subject of the schema whose compatibility is to be retrieved (optional)
         :raises: ClientError: if the request was unsuccessful or an invalid compatibility level was returned
         :returns: one of 'NONE','FULL','FORWARD', or 'BACKWARD'
         :rtype: str
         """
-        url = '/'.join([self.url, 'config'])
+        if subject:
+            url = '/'.join([self.url, 'config', subject])
+        else:
+            url = '/'.join([self.url, 'config'])
 
         result, code = self._send_request(url)
 
@@ -429,7 +437,10 @@ class CachedSchemaRegistryClient(object):
 
         compatibility = result.get('compatibilityLevel', None)
         if compatibility not in VALID_LEVELS:
-            raise ClientError('Invalid compatibility level received: {}'.format(
-                compatibility if compatibility else 'No compatibility was returned'), code)
+            if compatibility is None:
+                error_msg_suffix = 'No compatibility was returned'
+            else:
+                error_msg_suffix = str(compatibility)
+            raise ClientError('Invalid compatibility level received: {}'.format(error_msg_suffix), code)
 
         return compatibility
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -343,10 +343,10 @@ class CachedSchemaRegistryClient(object):
 
         result, code = self._send_request(url, method='POST', body=body)
         if code == 404:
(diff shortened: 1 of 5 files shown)


Rosswell:
[clabot:check]

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -366,67 +372,62 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
...
                 return result.get('is_compatible')
             else:
-                log.error("Unable to check the compatibility: " + str(code))
+                log.error("Unable to check the compatibility. Error code: {}".format(code))
                 return False
         except Exception as e:
             log.error("_send_request() failed: %s", e)
             return False
 
-    def update_compatibility(self, level, subject=None):
```
rnpridgeon: Setting per subject compatibility is part of the Schema Registry API and should be maintained. 

https://docs.confluent.io/current/schema-registry/docs/api.html#put--config-(string-%20subject)

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -366,67 +372,62 @@ def test_compatibility(self, subject, avro_schema, version='latest'):
...
         body = {"compatibility": level}
         result, code = self._send_request(url, method='PUT', body=body)
-        if code >= 200 and code <= 299:
+        if 200 <= code <= 299:
             return result['compatibility']
         else:
-            raise ClientError("Unable to update level: %s. Error code: %d" % (str(level)), code)
+            raise ClientError("Unable to update level: {}".format(level), code)
 
-    def get_compatibility(self, subject=None):
```
rnpridgeon: This too is a valid Schema Registry operation at the subject level.

https://docs.confluent.io/current/schema-registry/docs/api.html#get--config-(string-%20subject)

confluent_kafka/avro/error.py
```diff
@@ -20,7 +20,10 @@ class ClientError(Exception):
     """ Error thrown by Schema Registry clients """
 
     def __init__(self, message, http_code=None):
-        self.message = message
+        if http_code:
+            self.message = '{}. Error code: {}'.format(message, str(http_code))
+        else:
+            self.message = message
```
cricket007: Not sure which is more preferred, but this could also be written as

```python
self.message = message
if http_code:
    self.message += ". Error code: {}".format(code)
```

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -259,10 +264,10 @@ def get_by_id(self, schema_id):
 
         result, code = self._send_request(url)
         if code == 404:
-            log.error("Schema not found:" + str(code))
+            log.error("Schema not found. Error code: {}".format(code))
             return None
-        elif not (code >= 200 and code <= 299):
-            log.error("Unable to get schema for the specific ID:" + str(code))
+        elif not (200 <= code <= 299):
+            log.error("Unable to get schema with an ID of {}. Error code: {}".format(schema_id, code))
```
cricket007: Is there a reason why the wording changed here?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -338,10 +343,10 @@ def get_version(self, subject, avro_schema):
 
         result, code = self._send_request(url, method='POST', body=body)
         if code == 404:
-            log.error("Not found:" + str(code))
+            log.error("Schema for {} Not found. Error code: {}".format(subject, code))
```
cricket007: Should `N` in not be lower cased? 

How about `Schema for subject '{}' not found`?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -338,10 +343,10 @@ def get_version(self, subject, avro_schema):
...
             return None
-        elif not (code >= 200 and code <= 299):
-            log.error("Unable to get version of a schema:" + str(code))
+        elif not (200 <= code <= 299):
+            log.error("Unable to get version of the {} schema. Error code: ".format(subject, code))
```
cricket007: How about `for the '{}' schema `?

tests/avro/mock_schema_registry_client.py
```diff
@@ -142,8 +142,8 @@ def get_id_for_schema(self, subject, avro_schema):
     def test_compatibility(self, subject, avro_schema, version='latest'):
         raise ClientError("not implemented")
 
-    def update_compatibility(self, level, subject=None):
+    def update_compatibility(self, level):
         raise ClientError("not implemented")
 
-    def get_compatibility(self, subject=None):
+    def get_compatibility(self):
```
cricket007: Along with the other comments, both these parameters should be maintained

confluent_kafka/avro/cached_schema_registry_client.py
Rosswell: my mistake, I must have missed that. Do you think it should revert to the global compatibility level if there's a `404` response?

confluent_kafka/avro/cached_schema_registry_client.py
Rosswell: SGTM 👍

tests/avro/mock_schema_registry_client.py
Rosswell: Agreed - reverting these

confluent_kafka/avro/cached_schema_registry_client.py
cricket007: I think "with specific ID {}" is grammatically fine. I was just curious. No reason to change

confluent_kafka/avro/cached_schema_registry_client.py
cricket007: I don't think it would be a good idea edit the compatibility for all other subjects if just one subject has a 404

Rosswell: @cricket007 Sorry you're right I meant to write this on `get_compatibility` - thoughts on that?

cricket007: I think the spec says that the 404 corresponds to the subject not being found. Personally, I think it is a good idea to return the global compatibility as a fallback, but I don't think the Java clients do that (yet)?

Rosswell: Ok. I'll stay safe and leave as-is for now, unless there's a specific need to make it behave otherwise

Rosswell:
@rnpridgeon any more changes requested?

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -338,10 +343,10 @@ def get_version(self, subject, avro_schema):
 
         result, code = self._send_request(url, method='POST', body=body)
         if code == 404:
-            log.error("Not found:" + str(code))
+            log.error("Schema for subject '{}' not found. Error code: {}".format(subject, code))
```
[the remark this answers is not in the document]
Rosswell: Do you think we should begin addressing that level of granularity, or stick with the generic `Not found`?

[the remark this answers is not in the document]
rnpridgeon: I'm fine with stating either the schema or the subject were not found. If memory serves this is what we do further down in the code. 

My only issue with the current wording is it could leave the unsuspecting user to believe that the schema was not registered at all. This as you and I know is not entirely correct; most users however do not know, nor should they care, how the schemas are stored.

Rosswell:
@rnpridgeon look good?

PR #529, opened 2019-01-25
Add support for getting message persistence status

rnpridgeon:


confluent_kafka/__init__.py
```diff
@@ -100,3 +105,25 @@ def _resolve_plugins(plugins):
             resolved.append(plugin)
 
     return paths_sep.join(resolved)
+
+
+class MessageStatus(Enum):
```
mhowlett: In .net magnus didn't try to block block me on `PersistenceStatus`, which i think is more informative (though doesn't roll off the tongue as well).

I see you've called the property `status()`, where as i used `PersistenceStatus` for that as well. I'm thinking I like just status better and might change the .net api.

we probably should be consistent here. I like the meeting half way as proposed above :-).

rnpridgeon: I think I like `PersistenceStatus` better than `MessageStatus` as well.  I think we are in agreement then; let it be so!

PR #548, opened 2019-02-27
Add producer purge method with optional blocking argument

peteryin21:
## Background
Currently, if a producer times out (i.e when a broker goes down), messages are added to an internal queue to be published. The number of remaining messages is returned via the `flush()` command. 
In use cases where duplicate messages might be highly undesirable (e.g sending task requests to a factory), it would be useful to purge failed messages from the internal queue so they are not unintentionally sent the next time `flush()` is called with the brokers back online.

## Description
Luckily, `librdkafka` exposes a [`purge` api](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L3455-L3485
) that purges messages in internal queues. This PR adds a binding for this functionality to this library so we can access it via the python `Producer` class. There are three strategies of purging that are defined by the `PurgeStrategy` Enum that can be passed in as first positional argument. 
- `PURGE_INTERNAL_QUEUES`: Purge messages from internal queues.
- `PURGE_INFLIGHT`: Purge messages in flight to or from the broker.
- `PURGE_ALL`: Purge messages from internal queues and those in flight.
By default, the method will block, but if the optional argument `blocking` is set to False, we will not wait on the background thread queue.

In addition, the user will need to call `poll()` or `flush()` after `purge()` to serve the delivery report callbacks of the purged messages. I conservatively decided to leave it out, but am open to discussion for how to best make this intuitive to the user.

### DIFF
--- a/confluent_kafka/admin/__init__.py
+++ b/confluent_kafka/admin/__init__.py
@@ -11,6 +11,9 @@ from ..cimpl import (KafkaException, # noqa
                      CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG,
                      CONFIG_SOURCE_STATIC_BROKER_CONFIG,
                      CONFIG_SOURCE_DEFAULT_CONFIG,
+                     PURGE_F_QUEUE,
+                     PURGE_F_INFLIGHT,
+                     PURGE_ALL,
                      RESOURCE_UNKNOWN,
                      RESOURCE_ANY,
                      RESOURCE_TOPIC,
@@ -33,6 +36,15 @@ class ConfigSource(Enum):
     DYNAMIC_DEFAULT_BROKER_CONFIG = CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG  #:
     STATIC_BROKER_CONFIG = CONFIG_SOURCE_STATIC_BROKER_CONFIG  #:
     DEFAULT_CONFIG = CONFIG_SOURCE_DEFAULT_CONFIG  #:
+
+
+class PurgeStrategy(Enum):
+    """
+    Purge strategies for the Kafka producer.
+    """
+    PURGE_INTERNAL_QUEUES = PURGE_F_QUEUE  #:
+    PURGE_INFLIGHT = PURGE_F_INFLIGHT  #:
+    PURGE_ALL = PURGE_ALL  #:
 
 
 class ConfigEntry(object):
--- a/confluent_kafka/admin/__init__.py
+++ b/confluent_kafka/admin/__init__.py
@@ -11,9 +11,6 @@ from ..cimpl import (KafkaException, # noqa
                      CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG,
                      CONFIG_SOURCE_STATIC_BROKER_CONFIG,
                      CONFIG_SOURCE_DEFAULT_CONFIG,
-                     PURGE_F_QUEUE,
-                     PURGE_F_INFLIGHT,
-                     PURGE_ALL,
                      RESOURCE_UNKNOWN,
                      RESOURCE_ANY,
                      RESOURCE_TOPIC,
@@ -36,15 +33,6 @@ class ConfigSource(Enum):
     DYNAMIC_DEFAULT_BROKER_CONFIG = CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG  #:
     STATIC_BROKER_CONFIG = CONFIG_SOURCE_STATIC_BROKER_CONFIG  #:
     DEFAULT_CONFIG = CONFIG_SOURCE_DEFAULT_CONFIG  #:
-
-
-class PurgeStrategy(Enum):
-    """
-    Purge strategies for the Kafka producer.
-    """
-    PURGE_INTERNAL_QUEUES = PURGE_F_QUEUE  #:
-    PURGE_INFLIGHT = PURGE_F_INFLIGHT  #:
-    PURGE_ALL = PURGE_ALL  #:
 
 
 class ConfigEntry(object):
--- a/confluent_kafka/src/Producer.c
+++ b/confluent_kafka/src/Producer.c
@@ -547,18 +547,18 @@ static void *Producer_purge (Handle *self, PyObject *args,
         if (in_queue)
                 purge_strategy = RD_KAFKA_PURGE_F_QUEUE;
         if (in_flight) 
-                purge_strategy = purge_strategy | RD_KAFKA_PURGE_F_INFLIGHT;
+                purge_strategy |= RD_KAFKA_PURGE_F_INFLIGHT;
         if (blocking)
-                purge_strategy = purge_strategy | RD_KAFKA_PURGE_F_NON_BLOCKING;
+                purge_strategy |= RD_KAFKA_PURGE_F_NON_BLOCKING;
 
         err = rd_kafka_purge(self->rk, purge_strategy);
 
-        if (err == RD_KAFKA_RESP_ERR_NO_ERROR) {
-                Py_RETURN_NONE;
-        }
         if (err) {
-                cfl_PyErr_Format(err, "%s", rd_kafka_err2str(err));
+                cfl_PyErr_Format(err, "Purge failed: %s", rd_kafka_err2str(err));
+                return NULL;
         }
+
+        Py_RETURN_NONE;
 }
 
 
@@ -628,7 +628,7 @@ static PyMethodDef Producer_methods[] = {
 	  "\n"
 	},
 	{ "purge", (PyCFunction)Producer_purge, METH_VARARGS|METH_KEYWORDS,
-          ".. py:function:: purge(purge_strategy, [blocking])\n"
+          ".. py:function:: purge([in_queue=True], [in_flight=True], [blocking=True])\n"
           "\n"
 	  "   Purge messages currently handled by the producer instance.\n"
 	  "   The application will need to call poll() or flush() "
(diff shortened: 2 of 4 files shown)


confluent_kafka/src/Producer.c
```diff
@@ -463,6 +479,17 @@ static PyMethodDef Producer_methods[] = {
...
 	  "\n"
 	},
+	{ "purge", (PyCFunction)Producer_purge, METH_VARARGS|METH_KEYWORDS,
+          ".. py:function:: purge([blocking])\n"
+          "\n"
+	  "   Purge messages in internal queues currently handled by producer instance.\n"
+	  "   The application will need to call poll() or flush() "
+	  "afterwards to serve the delivery report callbacks of the purged messages."
+	  "\n"
+          "  :param: bool blocking: If set to False, will not wait on background thread queue)\n"
```
rnpridgeon: I think we should expose both  RD_KAFKA_PURGE_F_QUEUE  and RD_KAFKA_PURGE_F_INFLIGHT as an enum. In addition to the two options provided by librdkafka lest add RD_KAFKA_PURGE_F_ALL for both in-flight and queued messages. blocking should remain a bool though that's nicer than making users execute bitwise operators. 

See `AdminTypes.c` and `admin/__init__.py` for an example for exposing enums.

confluent_kafka/src/Producer.c
```diff
@@ -398,6 +398,22 @@ static PyObject *Producer_flush (Handle *self, PyObject *args,
...
+
+static PyObject *Producer_purge (Handle *self, PyObject *args,
+                                 PyObject *kwargs) {
+	int blocking = 0;
+        static char *kws[] = { "blocking", NULL };
+
+	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|p", kws, &blocking))
+			return NULL;
+	if (blocking==1)
+		rd_kafka_purge(rk, RD_KAFKA_PURGE_F_QUEUE)
```
rnpridgeon: Make sure to handle any errors which may be raised from the execution of rd_kafka_purge 

https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L3478-L3482

peteryin21:
[clabot:check]

peteryin21:
Thanks for the review @rnpridgeon! I've made the requested changes and fixed the build errors. Let me know how it looks.

Also, I think it would be worth adding some tests, but wanted to get your thoughts. I was thinking of adding a `def test_purge` to [test_Producer.py](https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/test_Producer.py) and adding `purge()` usage to [this test](https://github.com/confluentinc/confluent-kafka-python/blob/0ed0f09eb2d918b2e581bc71038f39e7841837d6/tests/integration/integration_test.py#L154) in `integration_test.py`.

rnpridgeon:
Sorry I didn't get back to this in time. At any rate I believe the Transactional Producer API is better suited for this. I'll close this for now but if we deem the Transactional API to not be sufficient we can revisit.

edenhill:
purge() is useful for other purposes, for instance when terminating and the application wants to know the exact status of each message delivery without having to wait the full message.timeout.ms on flush.

E.g., something like this:
```python
# terminate

p.flush(5)  # wait for messages to be delivered, but much shorter than message.timeout.ms

# purge any remaining messages from queues/inflight
p.purge(in_queue=True, in_flight=True)

p.flush(0)  # get delivery reports for the purged messages

sys.exit(0)
```

I think we should reconsider adding this functionality since Peter has put the effort in to implement it.

peteryin21:
@edenhill @rnpridgeon It's been a while! I will look more into the transactional producer api. In the meantime, if `purge()` would still be a useful method to expose, then I'm happy to try and get this across the finish line. Let me know what the remaining steps are before we can merge this.

confluent_kafka/admin/__init__.py
```diff
@@ -35,6 +38,15 @@ class ConfigSource(Enum):
     DEFAULT_CONFIG = CONFIG_SOURCE_DEFAULT_CONFIG  #:
 
 
+class PurgeStrategy(Enum):
```
edenhill, line 41: While this technically maps nicely to librdkafka, I think from a user-experience it might be better to simply go with optional boolean kwargs: `purge(in_queue=True, in_flight=True, blocking=True)` (those are the defaults)

confluent_kafka/src/Producer.c
```diff
@@ -398,6 +398,34 @@ static PyObject *Producer_flush (Handle *self, PyObject *args,
         return cfl_PyInt_FromInt(qlen);
 }
 
+
+static void *Producer_purge (Handle *self, PyObject *args,
```
edenhill, line 402: I know the source is messed up tab/space-wise, but please try to use 8-space indents in new code, no tabs.
Thank you!

confluent_kafka/src/Producer.c
```diff
@@ -398,6 +398,34 @@ static PyObject *Producer_flush (Handle *self, PyObject *args,
...
+	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i|p", kws, &blocking))
+			return NULL;
+	if (blocking==1)
+		err = rd_kafka_purge(self->rk, purge_strategy);
+	else
+		err = rd_kafka_purge(self->rk, purge_strategy|RD_KAFKA_PURGE_F_NON_BLOCKING);
+	if (err == RD_KAFKA_RESP_ERR_NO_ERROR) {
+		Py_RETURN_NONE;
+	}
+	if (err == RD_KAFKA_RESP_ERR__INVALID_ARG) {
```
edenhill, line 418: I don't think the specific error code handling is needed here, it is very unlikely to happen, so to keep it simple simply do `cfl_PyErr_format(err, "%s", rd_kafka_err2str(err))`

edenhill:
It also needs a unit-test (tests/test_Producer.py)

peteryin21:
@edenhill Thanks for the comments. I've addressed them and also added a new test for purge in `test_Producer.py`. It seems to work! The things I don't think I can test functionally are the blocking and in_flight flags, as well as the error, since these will probably require integration testing. Let me know what you think

peteryin21:
@edenhill Ping in case this fell off your radar :)

confluent_kafka/src/Producer.c
```diff
@@ -532,6 +532,36 @@ static PyObject *Producer_abort_transaction(Handle *self, PyObject *args) {
...
+
+        rd_kafka_resp_err_t err;
+        static char *kws[] = { "in_queue", "in_flight", "blocking", NULL};
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|bbb", kws, &in_queue, &in_flight, &blocking))
+                return NULL;
+        if (in_queue)
+                purge_strategy = RD_KAFKA_PURGE_F_QUEUE;
+        if (in_flight) 
+                purge_strategy = purge_strategy | RD_KAFKA_PURGE_F_INFLIGHT;
```
edenhill, line 550: nit: `purge_strategy |= RD_KAFKA_PURGE_F_INFLIGHT;`

confluent_kafka/src/Producer.c
```diff
@@ -532,6 +532,36 @@ static PyObject *Producer_abort_transaction(Handle *self, PyObject *args) {
...
+        if (blocking)
+                purge_strategy = purge_strategy | RD_KAFKA_PURGE_F_NON_BLOCKING;
+
+        err = rd_kafka_purge(self->rk, purge_strategy);
+
+        if (err == RD_KAFKA_RESP_ERR_NO_ERROR) {
```
edenhill, line 556: To be consistent with other functions, flip this logic around:
```c
if (err) {
    cfl_PyErr_Format(err, "Purge failed: %s", rd_kafka_err2str(err));
    return NULL;
}

Py_RETURN_NONE;
```

confluent_kafka/src/Producer.c
```diff
@@ -597,6 +627,18 @@ static PyMethodDef Producer_methods[] = {
 	  "callbacks may be triggered.\n"
 	  "\n"
 	},
+	{ "purge", (PyCFunction)Producer_purge, METH_VARARGS|METH_KEYWORDS,
+          ".. py:function:: purge(purge_strategy, [blocking])\n"
```
edenhill, line 631: Should read `purge([in_queue=True], [in_flight=True], [blocking=True]`)

confluent_kafka/src/Producer.c
```diff
@@ -597,6 +627,18 @@ static PyMethodDef Producer_methods[] = {
...
+          "\n"
+	  "   Purge messages currently handled by the producer instance.\n"
+	  "   The application will need to call poll() or flush() "
+	  "afterwards to serve the delivery report callbacks of the purged messages."
+	  "\n"
+	  "  :param: bool in_queue: Purge messages from internal queues. By default, true.\n"
+	  "  :param: bool in_flight: Purge messages in flight to or from the broker. By default, true.\n"
+	  "  :param: bool blocking: If set to False, will not wait on background thread queue)\n"
```
edenhill, line 639: remove paren

peteryin21:
Looks like there are some issues with prepare-osx.sh that are also being experienced in recent PRs https://github.com/confluentinc/confluent-kafka-python/pull/909. I can rebase this branch on top of the fix once that's figured out.

edenhill:
Let's get #909 merged first, then we'll restart your CI jobs to make sure they pass before merging.

peteryin21:
@edenhill It looks like #909 is merged -- would you be able to restart the CI jobs and then merge? 

I'm also curious what the release schedule looks like, and what the timeline for releasing this change might look like?

peteryin21:
Hey @edenhill, would it be possible to restart the CI jobs and then merge if it looks good?

peteryin21:
@edenhill Can we get this PR merged?

edenhill:
Thanks for this, @peteryin21 !

PR #557, opened 2019-03-13
Use Environment markers instead of version_info

matt2000:
PEP-508 recommends Environment markers for better compatibility with modern build tools and dependency managers such as Poetry.

https://www.python.org/dev/peps/pep-0508/#environment-markers

E.g., See also: 
 - https://github.com/sdispater/poetry/issues/844
 - https://github.com/sdispater/poetry/issues/881

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,8 +5,8 @@ from setuptools import setup, find_packages
 from distutils.core import Extension
 import platform
 
-INSTALL_REQUIRES = ["futures; python_version<'3.3'",
-                    "enum34; python_version<'3.5'",
+INSTALL_REQUIRES = ['futures; python_version<"3.3"',
+                    'enum34; python_version<"3.5"',
                     'requests']
 
 # On Un*x the library is linked as -lrdkafka,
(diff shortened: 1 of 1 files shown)


matt2000:
[clabot:check]

setup.py
```diff
@@ -3,16 +3,11 @@
...
-INSTALL_REQUIRES = list()
-
-if sys.version_info[0] < 3:
-    avro = 'avro'
-    INSTALL_REQUIRES.extend(['futures', 'enum34', 'requests'])
-else:
-    avro = 'avro-python3'
+INSTALL_REQUIRES = ["futures; python_version<'3.3'",
+                    "enum34; python_version<'3.5'",
+                    'requests']
```
edenhill: suggesting using double-quotes for requests here to be consistent with the above quoting.

setup.py
```diff
@@ -51,6 +46,6 @@ def get_install_requirements(path):
       data_files=[('', ['LICENSE.txt'])],
       install_requires=INSTALL_REQUIRES,
       extras_require={
-          'avro': ['fastavro', 'requests', avro],
+          'avro': ['fastavro', 'requests', "avro; python_version<'3'", "avro-python3; python_version>='3'"],
```
edenhill: dito;
suggesting using double-quotes here to be consistent in the same list.

matt2000:
Looks liek an equivalent change has since been merged. Thanks!

PR #558, opened 2019-03-14
Fix docs for Consumer.subscribe

QinMing:


QinMing:
[clabot:check]

rnpridgeon:
Looks like this was already handled.

https://github.com/confluentinc/confluent-kafka-python/commit/d9102788c218f4aa66d6ee7ec224898600841662

A bit odd the diff doesn't reflect this but the commit from above consistent with master.
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/src/Consumer.c#L1012

PR #567, opened 2019-04-01
Add environment markers to python2-only dependencies

mellon85:
[PEP 496](https://www.python.org/dev/peps/pep-0496/) describes environment markers and [PEP 508](https://www.python.org/dev/peps/pep-0508/) describes the semantics of the python_version environment marker, tools which allow us to mark certain dependencies as necessary only when the user's python interpreter is within a certain version range.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,16 +3,8 @@
 import os
 from setuptools import setup, find_packages
 from distutils.core import Extension
-import sys
 import platform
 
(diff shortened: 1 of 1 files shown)


matt2000:
This change got merged elsewhere. I suggest this PR can be closed.

mellon85:
cool, sorry but didn't get the approval yet -.-

PR #569, opened 2019-04-03
ST-1322: Trogdor Produce Runner

yangxi:
The Trogdor Runner executes Trogdor ExternalCommandSpec tasks. It can be invoked directly with `./tests/trogdor_runner --spec ExternalCommandSpecFile`, or started by Trogdor agents. When directly invoking, the Trogdor Runner parse the spec file and execute the workload. When started by Trogdor agents, it wants on stdin for command, then execute the workload.

* Add Trogdor Runner (./tests/trogdor/trogdor_runner.py) to execute Trogdor tasks.
* Add Trogdor Produce Runner (./tests/trogdor/produce_spec_runner.py) to execute Trogdor Produce task.
* Add Trogdor Produce Test to ./tests/run.sh.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -27,6 +27,12 @@ run_tox() {
     tox -r "$@"
 }
 
+run_trogdor() {
+    start_cluster
+    echo "Executing Trogdor"
+    python ${TEST_SOURCE}/trogdor/trogdor_runner.py --spec ${TEST_SOURCE}/trogdor/example-produce-spec.json
+}
+
 run_native() {
     pip install -v .[avro]
     start_cluster
@@ -55,6 +61,10 @@ case ${1:-} in
         shift
         run_unit $@
         run_native $@
+        ;;
+    "trogdor")
+        shift
+        run_trogdor $@
         ;;
     *)
         run_native $@
--- a/tests/trogdor/example-produce-spec.json
+++ b/tests/trogdor/example-produce-spec.json
@@ -1,0 +1,29 @@
+{
+    "class": "org.apache.kafka.trogdor.workload.ExternalCommandSpec",
+    "command": [
+	"python",
+	"./trogdor-runner.py"
+    ],
+    "durationMs": 10000000,
+    "producerNode": "node0",
+    "workload": {
+	"class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
+	"durationMs": 10000000,
+	"producerNode": "node0",
+	"bootstrapServers": "localhost:9092",
+	"targetMessagesPerSec": 10000,
+	"maxMessages": 50000,
+	"activeTopics": {
+	    "foo[1-3]": {
+		"numPartitions": 10,
+		"replicationFactor": 1
+	    }
+	},
+	"inactiveTopics": {
+	    "foo[4-5]": {
+		"numPartitions": 10,
+		"replicationFactor": 1
+	    }
+	}
+    }
+}
--- a/tests/trogdor/example-produce-spec.json
+++ b/tests/trogdor/example-produce-spec.json
@@ -2,28 +2,26 @@
     "class": "org.apache.kafka.trogdor.workload.ExternalCommandSpec",
     "command": [
 	"python",
-	"./trogdor-runner.py"
+	"./tests/trogdor/trogdor-runner.py"
     ],
     "durationMs": 10000000,
     "producerNode": "node0",
     "workload": {
-	"class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
-	"durationMs": 10000000,
-	"producerNode": "node0",
-	"bootstrapServers": "localhost:9092",
-	"targetMessagesPerSec": 10000,
-	"maxMessages": 50000,
-	"activeTopics": {
-	    "foo[1-3]": {
-		"numPartitions": 10,
-		"replicationFactor": 1
-	    }
-	},
-	"inactiveTopics": {
-	    "foo[4-5]": {
-		"numPartitions": 10,
-		"replicationFactor": 1
-	    }
+		"class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
+		"bootstrapServers": "localhost:29092",
+		"targetMessagesPerSec": 10000,
+		"maxMessages": 50000,
+		"activeTopics": {
+			"foo[1-3]": {
+				"numPartitions": 10,
+				"replicationFactor": 1
+			}
+		},
+		"inactiveTopics": {
+			"foo[4-5]": {
+				"numPartitions": 10,
+				"replicationFactor": 1
+			}
+		}
 	}
-    }
 }
--- a/tests/trogdor/example-produce-spec.json
+++ b/tests/trogdor/example-produce-spec.json
@@ -1,11 +1,11 @@
 {
     "class": "org.apache.kafka.trogdor.workload.ExternalCommandSpec",
     "command": [
-	"python",
-	"./tests/trogdor/trogdor-runner.py"
+	    "python",
+	    "./tests/trogdor/trogdor-runner.py"
     ],
     "durationMs": 10000000,
-    "producerNode": "node0",
+    "commandNode": "node0",
     "workload": {
 		"class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
 		"bootstrapServers": "localhost:29092",
--- a/tests/trogdor/produce_spec_runner.py
+++ b/tests/trogdor/produce_spec_runner.py
@@ -48,14 +48,30 @@
 # "bootstrapServers": #string#
 # "targetMessagePerSec": #Java int#
 # "maxMessages": #Java long#
-# "keyGenerator": #Option, sequential generator by default# { "type": "constant" | "sequential" | "uniformRandom" | "null" }
-# "valueGenerator"#Same as the keyGenerator#"
-# "transactionGenerator":#Option, by default empty# {"type":"uniform"}
+# "keyGenerator": #payloadGenerator, optional, sequential generator (default)# { "type": "constant" | "sequential" | "uniformRandom" | "null" }
+# "valueGenerator"#payloadGenerator, optional, constant generator (default)#" {"type": "constant" | "sequential" | "uniformRandom" | "null" }
+# "transactionGenerator":#Object, optional, empty (default)# {"type":"uniform"}
 # "producerConf": #a JSON object#
 # "commonClientConf": "a JSON object"
 # "adminClientConf": "a JSON object"
 # "activeTopics" : "a JSON object"
 # "inactiveTopics": "a JSON object"
+#
+# PayloadGenerator:
+# ConstantPayLoadGenerator:
+# "type": #string, "constant"
+# "size": #Java Int, "size of the value"
+# "value": #Java byte array, optional, "constant value"
+# SequentialPayloadGenerator:
+# "type": "sequential"
+# "size": #Java Int, "size of the value"
+# "startOffset": #Java Long, "start value of the increasing value"
+# uniformRandomPayloadGenerator:
+# "type": "uniformRandom"
+# "size": #Java Int", "size of the value"
+# "seed": #Java Long", "seed of the value"
+# "padding" #Java Long", #padding size of the value, size - padding is the size of random bytes"
+
 import threading
 import time
 
@@ -63,7 +79,7 @@ from hdrh.histogram import HdrHistogram
 
 from trogdor_utils import expand_topics, trogdor_log, update_trogdor_error, merge_topics, create_topics, \
     update_trogdor_status, create_admin_conn, partition_set, PayloadGenerator, SeqGenerator, ConstGenerator, \
-    create_kafka_conf, create_producer_conn
+    create_kafka_conf, create_producer_conn, get_payload_generator
 
 
 def execute_produce_spec(spec):
@@ -177,8 +193,12 @@ class ProduceSpecRunner:
         self.report_status_interval = 10
         self.start_timestamp = time.time()
         self.create_spec_topics(self.producer_spec)
-        self.key_generator = PayloadGenerator(SeqGenerator(4, 0))
-        self.val_generator = PayloadGenerator(ConstGenerator(4,0xabcd))
+        self.key_generator_spec = self.producer_spec.get("keyGenerator", {"type":"sequential", "size":4, "startOffset":0})
+        self.value_generator_spec = self.producer_spec.get("valueGenerator", {"type": "constant", "size": 512})
+        key_payload_generator = get_payload_generator(self.key_generator_spec)
+        value_payload_generator = get_payload_generator(self.value_generator_spec)
+        self.key_generator = PayloadGenerator(key_payload_generator)
+        self.val_generator = PayloadGenerator(value_payload_generator)
         self.qps = self.producer_spec.get("targetMessagePerSec", 10000)
         self.max_messages = self.producer_spec.get("maxMessages", 100000)
         self.nr_finished_messages = 0
--- a/tests/trogdor/produce_spec_runner.py
+++ b/tests/trogdor/produce_spec_runner.py
@@ -1,0 +1,194 @@
+# Copyright 2016 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# One example of a produce Trogdor task, the workload is a JSON object following the ProduceBenchSpec.
+#
+# {
+#     "id": "$TASK_ID",
+#
+#     "spec": {
+#         "class": "org.apache.kafka.trogdor.workload.ExternalCommandSpec",
+#         "command": ["python", "./tests/bin/ExternalCommandExample.py"],
+#         "durationMs": 10000000,
+#         "commandNode": "node0",
+#         "workload":{
+#             "class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
+#             "bootstrapServers": "localhost:9092",
+#             "targetMessagesPerSec": 10000,
+#             "maxMessages": 50000,
+#             "activeTopics": {
+#                 "foo[1-3]": {
+#                     "numPartitions": 10,
+#                     "replicationFactor": 1
+#                 }
+#             },
+#             "inactiveTopics": {
+#                 "foo[4-5]": {
+#                     "numPartitions": 10,
+#                     "replicationFactor": 1
+#                 }
+#             }
+#         }
+#     }
+# }
+
+# ProduceBenchSpec:
+# "startMs": #start
+# "bootstrapServers": #string#
+# "targetMessagePerSec": #Java int#
+# "maxMessages": #Java long#
+# "keyGenerator": #Option, sequential generator by default# { "type": "constant" | "sequential" | "uniformRandom" | "null" }
+# "valueGenerator"#Same as the keyGenerator#"
+# "transactionGenerator":#Option, by default empty# {"type":"uniform"}
+# "producerConf": #a JSON object#
+# "commonClientConf": "a JSON object"
+# "adminClientConf": "a JSON object"
+# "activeTopics" : "a JSON object"
+# "inactiveTopics": "a JSON object"
+import threading
+import time
+
+from hdrh.histogram import HdrHistogram
+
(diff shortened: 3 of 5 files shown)


tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+        else:
+            update_trogdor_error(spec_class + " is not supported.")
+    except Exception as exp:
+        update_trogdor_error("Exception: " + exp)
+
+
+def get_task_from_specfile(specFile):
+    """
+    :param specFile: One Trogdor ExternalCommandSpec JSON file
+     One example of the ExternalCommandSpec spec:
```
edenhill: Is this supposed to show or link to an example?

tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+
+    :return specFile["spec"]["workload"] or None
+    """
+    specString = ""
+    workload = None
+    with open(specFile) as f:
+        for line in f:
+            if line.startswith('#'):
```
edenhill: suggest whitespace trimming the line

tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+                continue
+            specString += line
+        f.close()
```
edenhill: I dont think f.close is needed when using `with open` .. contexts.

tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+    task = json.loads(specString)
+    if "class" in task and "workload" in task and \
+            task["class"] == "org.apache.kafka.trogdor.workload.ExternalCommandSpec":
+        workload = task["workload"]
+    return workload
+
+
+def get_task_from_input():
+    """
+    Waiting for a new task on stdin, formatted in {"id":<task ID string>, "workload":<configured workload JSON object>}
```
edenhill: formatted as

tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+        for line in f:
+            if line.startswith('#'):
+                continue
+            specString += line
+        f.close()
+    if (specString == ""):
+        return None
+    task = json.loads(specString)
+    if "class" in task and "workload" in task and \
+            task["class"] == "org.apache.kafka.trogdor.workload.ExternalCommandSpec":
```
edenhill: `if task.get("class", "") == "org.." and "workload" in task"` is slightly more effecient

tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+
+def get_task_from_input():
+    """
+    Waiting for a new task on stdin, formatted in {"id":<task ID string>, "workload":<configured workload JSON object>}
+    :return: spec["workload"]
+    """
+    for line in sys.stdin:
+        try:
+            comm = json.loads(line)
+            if "workload" in comm and "id" in comm:
```
edenhill: Why is "id" needed, it is not being used?

tests/trogdor/trogdor_runner.py
```diff
@@ -0,0 +1,89 @@
...
+                trogdor_log("Received one workload:{}".format(comm))
+                return comm["workload"]
+        except ValueError:
+            trogdor_log("Input line " + line + " is not a valid external runner command.")
+
+
+if __name__ == '__main__':
+    spec = None
```
edenhill: spec can't be undefined after the if statement, so no need to initialize here.

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
+# Copyright 2016 Confluent Inc.
```
edenhill: 2019

edenhill: Missing shebang

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
...
+            topic_partitions.append((t_name, i))
+    return topic_partitions
+
+
+def expand_topics(topics):
+    """
+    Return extended topics.
+    For example, for topics { "foo[1-3]" : {} }, return { "foo1" : {}, "foo2" : {}, "foo3" : {}}
+    """
+    topic_expand_matter = re.compile(r'(.*?)\[([0-9]*)\-([0-9]*)\](.*)')
```
edenhill: the lo and hi integers should be required, and no need to escape -, thus:
`re.compile(r'(.*?)\[([0-9]+)\-([0-9]+)\](.*)')`

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
...
+    expanded = {}
+    for topicName in topics:
+        match = topic_expand_matter.match(topicName)
+        if match:
+            pre = match.group(1)
+            start = int(match.group(2))
+            end = int(match.group(3))
+            last = match.group(4)
```
edenhill: suffix. last could be confused with end.

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
...
+        k_conf.update(conf)
+    return k_conf
+
+
+def create_admin_conf(bootstrap_servers, common_client_config, admin_client_config):
+    # Refer Java Trogdor tools/src/main/java/org/apache/kafka/trogdor/common/WorkerUtils.java#L305
+    admin_request_timeout_ms = 25000
+    create_kafka_conf(bootstrap_servers, common_client_config, admin_client_config)
+    admin_conf = create_kafka_conf(bootstrap_servers, common_client_config, admin_client_config)
+    admin_conf["socket.timeout.ms"] = admin_request_timeout_ms
```
edenhill: Why is this needed? The defaults should be good.

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
...
+    return admin_conf
+
+
+def create_producer_conn(bootstrap_servers, common_client_config, producer_config):
+    producer_conf = create_kafka_conf(bootstrap_servers, common_client_config, producer_config)
+    return Producer(**producer_conf)
```
edenhill: no need for double **, the constructor takes a dict as is.

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def create_admin_conn(bootstrap_servers, common_client_config, admin_client_config):
```
edenhill: change conn to client.

tests/trogdor/trogdor_utils.py
```diff
@@ -0,0 +1,202 @@
...
+    return admin_conn
+
+
+def create_topics(admin_conn, topics):
+    for topic_name in topics:
+        topic = topics[topic_name]
+        create_topic(admin_conn, topic_name, topic)
+
+
+def create_topic(admin_conn, topic_name, topic):
```
edenhill: admin_client

tests/trogdor/example-produce-spec.json
```diff
@@ -0,0 +1,27 @@
...
+	    "python",
+	    "./tests/trogdor/trogdor-runner.py"
+    ],
+    "durationMs": 10000000,
+    "commandNode": "node0",
+    "workload": {
+		"class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
+		"bootstrapServers": "localhost:29092",
+		"targetMessagesPerSec": 10000,
+		"maxMessages": 50000,
```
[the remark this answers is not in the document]
yangxi: Yeah, this is just an example spec file for demonstrating how to config the workload. For real product perf tests, this is too short.

tests/trogdor/example-produce-spec.json
```diff
@@ -0,0 +1,27 @@
...
+		"activeTopics": {
+			"foo[1-3]": {
```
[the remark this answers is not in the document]
yangxi: This is a demo spec file. The idea is to use the similar one as the example of Java Trogdor Produce runner. https://github.com/apache/kafka/blob/trunk/tests/spec/simple_produce_bench.json

tests/trogdor/produce_spec_runner.py
```diff
@@ -0,0 +1,222 @@
...
+        if realQPS:
+            exp_status["realMPS"] = realQPS
+        update_trogdor_status(exp_status)
+
+    def message_on_delivery(self, err, msg, sent_time):
+        if err is not None:
+            trogdor_log("ProduceSpecRunner: delivery failed: {} [{}]: {}".format(msg.topic(), msg.partition(), err))
+            self.nr_failed_messages += 1
+        now = time.time()
+        latency = now - sent_time
```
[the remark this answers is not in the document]
yangxi: I am working on adding an `iteration` field to the produce spec. With `iteration`, you can control how many times Trogdor should run the same workload. That's a better way to compare `warmup` Vs. `stable` cases.

tests/trogdor/produce_spec_runner.py
```diff
@@ -0,0 +1,222 @@
...
+        self.latency_histogram.record_value(math.ceil(latency * 100000))
+        self.totalLatency += latency
+        self.nr_finished_messages += 1
+
+    def get_msg_callback(self):
+        product_time = time.time()
```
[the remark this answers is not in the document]
yangxi: sure.

tests/trogdor/trogdor_utils.py
yangxi: Try to be same as the Java Trogdor produce worker.

tests/trogdor/example-produce-spec.json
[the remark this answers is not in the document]
yangxi: How about this? 
Let's keep it as the 5s case. I am working on creating a regression Jenkins job that compares performance between Python clients and Java clients. We will have complex spec files for that Jenkins job. After I finish that, I will add those spec files to this `tests/trogdor/` directory.

tests/trogdor/produce_spec_runner.py
```diff
@@ -0,0 +1,222 @@
...
+                      "totalRecorded": self.latency_histogram.get_total_count(),
+                      "totalError": self.nr_failed_messages,
+                      "planMPS": self.qps,
+                      "averageLatencyMs": self.latency_histogram.get_mean_value()/100.0,
+                      "p50LatencyMs": self.latency_histogram.get_value_at_percentile(50)/100.0,
+                      "p95LatencyMs": self.latency_histogram.get_value_at_percentile(95)/100.0,
+                      "p99LatencyMs": self.latency_histogram.get_value_at_percentile(99)/100.0,
+                      "maxLatencyMs": self.latency_histogram.get_max_value()/100.0
+                      }
+        if realQPS:
```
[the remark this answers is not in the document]
yangxi: Fixed

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
yangxi: Fixed

tests/trogdor/trogdor_utils.py
yangxi: Fixed

tests/trogdor/trogdor_utils.py
yangxi: Fixed

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
edenhill: That is probably true for the cluster itself, but each client instance will need some warmup to amortize for connection setup times, etc, and this should not be included in the final throughput or latency results.

tests/trogdor/trogdor_utils.py
edenhill: Okay, let's avoid setting specific configs unless we know it is needed.

tests/trogdor/trogdor_utils.py
yangxi: Good catch, will update the Java version too.

tests/trogdor/trogdor_runner.py
yangxi: Yeah, removed.

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
yangxi: The warmup will be the first iteration, say first 5000 messages. Also, the warmup latency is important, I suggest we should show that too.

tests/trogdor/produce_spec_runner.py
```diff
@@ -0,0 +1,222 @@
...
+            self.status = "exception"
+
+    def execute_spec(self):
+        """ Executed in the context of the producer thread"""
+        start_produce_time = time.time()
+        nr_topics = len(self.topic_partitions)
+        nr_message = 0
+        pause = 1.0 / self.qps
+        next_fire_time = start_produce_time + pause
+        while nr_message < self.max_messages:
```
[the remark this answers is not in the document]
yangxi: One potential problem here is that whether there will be strong contentions on the Global Interpreter Lock between the produce thread the and polling thread. I will run some tests, see whether it is problem.

yangxi:
@edenhill Thanks for the review. I addressed most of your comments.

tests/trogdor/trogdor_utils.py
yangxi: fixed

tests/trogdor/trogdor_utils.py
yangxi: fixed

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
edenhill: Are the iterations using the same producer instance?
If so, then this is fine, but if new producer instances are used for each iteration we still need a warmup period per iteration to avoid including startup costs in the throughput measurements.

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
edenhill: Python will only execute one thread at a time, there is no real concurrency. But for this case where the main  loop just calls poll() which is a non-GIL-holding blocking call it should be fine as long as the producer thread does not wait for the main thread, or vice versa.

tests/trogdor/produce_spec_runner.py
```diff
@@ -212,11 +212,11 @@ def __init__(self, workload):
...
         self.key_generator = PayloadGenerator(key_payload_generator)
         self.val_generator = PayloadGenerator(value_payload_generator)
-        self.qps = self.produce_workload.get("targetMessagePerSec", 10000)
+        self.mps = self.produce_workload.get("targetMessagePerSec", 10000)
         self.max_messages = self.produce_workload.get("maxMessages", 100000)
         self.nr_finished_messages = 0
         self.nr_failed_messages = 0
         self.producer = create_producer_conn(self.bootstrap_servers, self.common_client_conf, self.producer_conf)
-        trogdor_log("Produce {} at message-per-sec {}".format(self.max_messages, self.qps))
+        trogdor_log("Produce {} at message-per-sec {}".format(self.max_messages, self.mps))
```
edenhill: This could be clearer, "Produce {} *messages* at .."

tests/trogdor/example-produce-spec.json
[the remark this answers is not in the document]
edenhill: Sounds good

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
edenhill: Any comment on this?

tests/trogdor/trogdor_utils.py
edenhill: Please remove the unneeded configuration.

tests/trogdor/example-produce-spec.json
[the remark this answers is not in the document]
yangxi: Changed the names to py_trogdor.

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
yangxi: I added the `iterations` field to the spec file.  When the field > 1, the trogdor_runner.py reports an JSON array of performance results, e.g. {"status": [{"totalSent": 50000, "totalRecorded": 50000, "totalError": 0, "planMPS": 10000, "averageLatencyMs": 13814.083504000002, "p50LatencyMs": 13864.95, "p95LatencyMs": 29470.71, "p99LatencyMs": 30719.99, "maxLatencyMs": 31006.71}, {"totalSent": 50000, "totalRecorded": 50000, "totalError": 0, "planMPS": 10000, "averageLatencyMs": 19482.478866999998, "p50LatencyMs": 19394.55, "p95LatencyMs": 35573.75, "p99LatencyMs": 36597.75, "maxLatencyMs": 36823.03}]}

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
yangxi: Yeah, I know. Now, the performance of the produce benchmark can be very bad in some cases, for example, compared with sending requests to a local Kafka that is not in the docker instance, sending requests to a local Kafka in Docker is much slower.

I suspect that this performance issue has some connections with the python GIL. I will investigate the issue when I have time.

tests/trogdor/trogdor_utils.py
yangxi: @edenhill Done.

tests/trogdor/produce_spec_runner.py
yangxi: Sure.

tests/run.sh
```diff
@@ -29,6 +29,12 @@ run_tox() {
     tox -r "$@"
 }
 
+run_trogdor() {
+    start_cluster
+    echo "Executing Trogdor"
+    python ${TEST_SOURCE}/trogdor/trogdor_runner.py --spec ${TEST_SOURCE}/trogdor/example-produce-spec.json
```
edenhill: this should not point to the example, but a proper file. The user/dev should copy and modify the example file into place. Alternatively we call this file `produce-spec.json`, without the example part, to allow it to run out of the box.

tests/trogdor/example-produce-spec.json
```diff
@@ -0,0 +1,27 @@
...
+    "class": "org.apache.kafka.trogdor.workload.ExternalCommandSpec",
+    "command": [
+	    "python",
+	    "./tests/trogdor/trogdor-runner.py"
+    ],
+    "durationMs": 10000000,
+    "commandNode": "node0",
+    "workload": {
+		"class": "org.apache.kafka.trogdor.workload.ProduceBenchSpec",
+		"bootstrapServers": "localhost:29092",
```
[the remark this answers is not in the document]
edenhill: That is probably the docker mapped port.
@yangxi ?

tests/trogdor/example-produce-spec.json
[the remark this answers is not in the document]
edenhill: I don't think there is much value in a perf test that runs for this short, the results will be too noisy. So let's provide a file with proper defaults. At least 30s, preferably 60s of runtime.

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
edenhill: Are the iterations reusing the same producer instance?
If not then there is no warm-up benefit on the client side, so I really think they should be reused.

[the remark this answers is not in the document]
yangxi: Yeah, they use the same producer.

tests/trogdor/example-produce-spec.json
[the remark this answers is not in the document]
yangxi: @stanislavkozlovski It is the exposed Docker port, declared here https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/docker/docker-compose.yaml#L14

tests/trogdor/produce_spec_runner.py
```diff
@@ -0,0 +1,222 @@
+# Copyright 2016 Confluent Inc.
```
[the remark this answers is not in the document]
mhowlett: nit: 2020 :P

mhowlett:
I'm going to progress the work in this PR over here: https://github.com/mhowlett/confluent-kafka-python/tree/trogdor

I'll add to the comments in this PR, where relevant, but make updates to my own branch. we can sort out anything already raised in comments here, and continue the discussion in my branch when done.

tests/trogdor/trogdor_runner.py
mhowlett: I guess you had a specific reason to want to add support for comments, but i don't see this as critical, and since JSON doesn't support comments I think we should just leave this out completely.

tests/trogdor/trogdor_runner.py
mhowlett: afaict it's the nicest way to do it though? `spec` needs to existing in this scope, and this seems the most straightforward way to this.

tests/trogdor/produce_spec_runner.py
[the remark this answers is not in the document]
yangxi: I am still working on this PR, will fix it :)

PR #575, opened 2019-04-08
Have consistent string formating

hrchu:
Use the %d operator in both consumer and producer for the offset.


edenhill:
[clabot:check]

edenhill:
Thank you!

PR #578, opened 2019-04-11
SR client: Don't disable cert verification if no ssl.ca.location set

edenhill:
This is a regression in v1.0.0


ybyzek:
Just for searchability, the issue that this PR addresses presents itself with a warning message that resembles:

```bash
InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
```

PR #583, opened 2019-04-16
implement environment markers to fix poetry

fishman:
pep 508 recommends the usage of environment markers rather than version_info
https://www.python.org/dev/peps/pep-0508/#environment-markers

currently installing using poetry resolves the wrong libraries

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,16 +3,13 @@
 import os
 from setuptools import setup, find_packages
 from distutils.core import Extension
-import sys
 import platform
 
-INSTALL_REQUIRES = list()
-
-if sys.version_info[0] < 3:
-    avro = 'avro'
-    INSTALL_REQUIRES.extend(['futures', 'enum34', 'requests'])
-else:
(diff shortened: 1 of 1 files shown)


fishman:
[clabot:check]

fishman:
@edenhill can we get this merged? It breaks pretty much all python 3.6/3.7 installs by installing enum34 on an unsupported platform. It also goes against https://www.python.org/dev/peps/pep-0508/#environment-markers

It's a non intrusive change that fixes a lot of environments. https://github.com/sdispater/poetry/issues/844#issuecomment-472252717

edenhill:
Awesome, thank you!

fishman:
@edenhill is there any chance you can do a 1.0.1 service release ? right now i still have to manually remove enum34 after installing anywhere i use poetry

edenhill:
We're aiming to do a 1.0.1 maintenance release (with this fix and other ones) within a couple of weeks.

PR #588, opened 2019-04-25
DEVX-812: add link to confluentinc/examples

ybyzek:


examples/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+* [adminapi.py](adminapi.py): library of Kafka AdminClient operations
+* [avro-cli.py](avro-cli.py): produces Avro messages with Confluent Schema Registry and then reads them back again 
+* [consumer.py](consumer.py): reads messages from a Kafka topic
+* [producer.py](producer.py): reads lines from stdin and sends them to Kafka
+
+Additional examples for [Confluent Cloud](https://www.confluent.io/confluent-cloud/):
+
+* [confluent_cloud.py](confluent_cloud.py): produces messages to Confluent Cloud and then reads them back again
+* [confluentinc/examples](https://github.com/confluentinc/examples/tree/5.2.1-post/clients/cloud/python): integrates Confluent Cloud and Confluent Cloud Schema Registry
```
edenhill: Do we really want to link to 5.2...? That'll get outdated real quickly and we'll want to avoid bumping this version dependency in the future  (because we'll forget).
Can we use master instead?

examples/README.md
```diff
@@ -0,0 +1,11 @@
+The scripts in this directory provide code examples using Confluent's Python client:
+
+* [adminapi.py](adminapi.py): library of Kafka AdminClient operations
```
edenhill: "library" implies that this is a library/module, which it is not. Maybe "collection" instead?

edenhill: "Collection of Kafka Admin API operations" is probably better, since it is more descriptive of what it is.

examples/README.md
ybyzek: Agree

examples/README.md
ybyzek: Fixed

Issue #518, opened 2019-01-07
Use a parsed schema in the cached decoder function


arthurandres:
Description
===========

MessageSerializer#_get_decoder_func should call  fastavro#parse_schema to improve performance.

There is a minor overhead to this function call (parse_schema), so it would be beneficial if it was only done once, and the parsed schema would be binded to schemaless_reader when creating the lambda that constitutes the cached decoder function

Later, the call to parse_schema within schemaless_reader is smart enough to figure out that the schema is already parsed (https://github.com/fastavro/fastavro/blob/master/fastavro/_schema_py.py#L83)

How to reproduce
================

I'm not sure if it's applicable as it is not a bug. This is covered by TestMessageSerializer either.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): 0.11.4
 - [x] Apache Kafka broker version: 0.11.0.3 
 - [  ] Client configuration: `{...}`
 - [x] Operating system: Linux
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

arthurandres:
This looks like it's done: 
https://github.com/confluentinc/confluent-kafka-python/pull/627

PR #601, opened 2019-05-17
Fixed fastavro not being used if schemas aren't manually specified…

mparry:
… Don't force fastavro to re-parse from JSON every time; cache the result. Added a flag to disable message key decoding.

As the commit message summarises, this improves a few things. I'll elaborate:
1) Addresses #616, where `fastavro` would no longer be used by default.
2) I also addressed #518 while I was working on the above, so that `fastavro` will no longer be forced to reparse the schema dict for every message.
3) Finally, I made the key decoding in `AvroConsumer` optional (but on by default). The main reason for this is that it is not, as far as I can tell, a requirement to encode both for a given topic. One very good reason for only Avro-encoding message values and not keys is that KSQL actually requires this! See confluentinc/ksql#824, for example.

Potentially point 3 at least would be superseded by #502, if that's still planned. However, we needed a way of disabling this more immediately. Let me know if you would therefore like me to minimise this PR at all.

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -5,7 +5,7 @@
 
 from confluent_kafka import Producer, Consumer
 from confluent_kafka.avro.error import ClientError
-from confluent_kafka.avro.load import load, loads  # noqa
+from confluent_kafka.avro.load import load, loads, loads_fast  # noqa
 from confluent_kafka.avro.cached_schema_registry_client import CachedSchemaRegistryClient
 from confluent_kafka.avro.serializer import (SerializerError,  # noqa
                                              KeySerializerError,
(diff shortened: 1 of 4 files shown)


mparry:
We have attempted to sign the CLA as an organisation previously, but the bot still seems unsatisfied. Is there someone that I can contact about this, to try and figure out why? Thanks.

swenzel:
I ran into that issue with fastavro not being used since after the last update our tests failed. The bad thing about it is that it is not only slower but also behaves differently on logical types which it keeps as bytes instead of turning e.g. decimals into floats.
This makes me think whether it's maybe better to explicitly choose whether to use a FastAvroConsumer or a "normal" AvroConsumer so you actually notice when it fails. But that's another story.
I look forward to seeing your pull request being merged!

lowercase24:
Is this stalled?

mparry:
I think that all of this is addressed in other, smaller PRs - respectively to my original bullet points: #773, #627, #787. (One is yet to be merged, but is still a better bet than mine now.)

PR #603, opened 2019-05-21
Add trove classifiers.

gandalf013:
caniusepython3 project uses trove classifiers to determine if a package
can be used in Python 3. This commits adds some classifier data to
allow caniusepython3 to detect that this package supports Python 3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,16 @@ def get_install_requirements(path):
     ]
 
 
+trove_classifiers = [
+    'Development Status :: 5 - Production/Stable',
+    'Intended Audience :: Developers',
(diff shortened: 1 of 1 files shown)


gandalf013:
[clabot:check]

faph:
PyPI uses the licence information in the classifiers to display the OSS licence. This is then picked up by other systems to filter packages by OSS licence. Some systems may also simply block packages being loaded if there is no standard OSS licence in the meta data. 

Is there any chance this could be merged and released soon?

edenhill:
Thank you @gandalf013 !

gandalf013:
Thank you!

PR #611, opened 2019-05-29
v1.0.1 release 

rnpridgeon:


tools/build-manylinux.sh
```diff
@@ -44,7 +44,7 @@ fi
 #
 
 echo "# Installing basic system dependencies"
-yum install -y zlib-devel gcc-c++
+yum install -y zlib-devel gcc-c++ƒ
```
edenhill: this looks weird

rnpridgeon: Interesting I thought I pushed the removal of this prior to opening a pr

edenhill: Good thing there is a web page showing what has actually been pushed.
This one
;)

rnpridgeon:
OSX build failure indicated curl exited with error code 56 which indicates we hit a failure receiving data. This seems like it should be intermittent so I reran the task.

Feandil:
Is there any plan to release this in the close future?

PR #614, opened 2019-06-05
Encode avro schemas using MappingProxyEncoder

hermesdt:
When encoding nested types like this:

```
{
    "name": "field_name",
    "type": [
        "null",
        {
            "type": "long",
            "connect.name": "org.apache.kafka.connect.data.Timestamp"
        }
    ]
}
```

Apache avro uses a MappingProxyType for it, resulting into this schema:

```
{'type': ['null',
    mappingproxy({'type': 'long',
                  'connect.name': 'org.apache.kafka.connect.data.Timestamp'})]
```

mappingproxy objects aren't json serializable thus a custom class is required.

Custom class comes from apache avro library:

https://github.com/apache/avro/blob/d3011c98d746e211c06a0308f759ef4d8312566c/lang/py3/avro/schema.py#L154

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -29,6 +29,11 @@ from requests import Session, utils
 from .error import ClientError
 from . import loads
 
+try:
+    from types import MappingProxyType
+except ImportError:
+    MappingProxyType = None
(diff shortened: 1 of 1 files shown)


hermesdt:
> [clabot:check]

[clabot:check]

PR #634, opened 2019-07-09
V1.1.0rc

rnpridgeon:


test-requirements.txt
```diff
@@ -1,2 +1,2 @@
-pytest
-flake8
\ No newline at end of file
+pytest==4.6.4
```
edenhill: What's this for? The commit message should say why.

tools/fixup-wheels.sh
```diff
@@ -82,10 +82,12 @@ fixup_wheel_macosx () {
 
         # Change the name to be local
         install_name_tool -id "$lib" $lib
-        if otool -L $lib | grep -q /usr/local/lib/librdkafka.1.dylib ; then
+        # Extract existing(old) reference
+        old=$(otool -L $lib | grep -o .*librdkafka.1.dylib)
+        if [[ ! -z "old" ]]; then
```
edenhill: forgot something?

tools/fixup-wheels.sh
```diff
@@ -82,10 +82,12 @@ fixup_wheel_macosx () {
 
         # Change the name to be local
         install_name_tool -id "$lib" $lib
-        if otool -L $lib | grep -q /usr/local/lib/librdkafka.1.dylib ; then
+        # Extract existing(old) reference
+        old=$(otool -L $lib | grep -o .*librdkafka.1.dylib)
```
edenhill: You need to use single quotes around the regex to avoid shell expansion of `*`.

also, technically, the returned string will have a bunch of leading spaces that should be trimmed off.

tools/fixup-wheels.sh
```diff
@@ -82,10 +82,12 @@ fixup_wheel_macosx () {
...
+        if [[ ! -z "old" ]]; then
             # Change the librdkafka reference to load from the same
             # directory as the plugin
-            install_name_tool -change /usr/local/lib/librdkafka.1.dylib '@loader_path/librdkafka.1.dylib' $lib
+            install_name_tool -change $old '@loader_path/librdkafka.1.dylib' $lib
```
edenhill: Paths ($old) should be quoted to support spaces-in-paths.

tools/fixup-wheels.sh
```diff
@@ -82,10 +82,12 @@ fixup_wheel_macosx () {
 
         # Change the name to be local
         install_name_tool -id "$lib" $lib
-        if otool -L $lib | grep -q /usr/local/lib/librdkafka.1.dylib ; then
+        # Extract existing(old) reference
+        old=$(otool -L $lib | grep -o '.*librdkafka.1.dylib' | xargs)
```
edenhill: nit: Nice trick, but will not handle multiple spaces in the path itself (however unlikely).
```
$ echo '  y/    strange path/x' | xargs
y/ strange path/x
```

That's ok for now though.,

PR #636, opened 2019-07-11
Change version to 3.0.1

sp1rs:
Change lib version to `3.0.1`.


sp1rs:
[clabot:check]

edenhill:
This package is already correctly versioned.

PR #646, opened 2019-07-24
DOCS: Remove timeout param from Consumer_position function(@hrcu)

hrchu:
Consumer_position does not have the timeout parameter.


justinfx:
What a coincidence. I just hit this problem 1 hour ago and was about to submit a ticket, and you just fixed it :-)

edenhill:
[clabot:check]

PR #647, opened 2019-07-24
TESTS: Update Consumer and Producer tests with proper exception testing(@hrcu)

hrchu:
Let tests fail if the expected exception not thrown.

### DIFF
--- a/tests/test_Consumer.py
+++ b/tests/test_Consumer.py
@@ -10,10 +10,9 @@ def test_basic_api():
     """ Basic API tests, these wont really do anything since there is no
         broker configured. """
 
-    try:
+    with pytest.raises(TypeError) as ex:
         kc = Consumer()
-    except TypeError as e:
-        assert str(e) == "expected configuration dict"
+    assert ex.match('expected configuration dict')
 
     def dummy_commit_cb(err, partitions):
         pass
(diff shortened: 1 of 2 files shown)


hrchu:
CI seems broken..

```
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: Failed to fetch https://packagecloud.io/github/git-lfs/ubuntu/dists/trusty/InRelease  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
E: Failed to fetch http://dl.hhvm.com/ubuntu/dists/trusty/main/binary-amd64/Packages  Hash Sum mismatch
W: Some index files failed to download. They have been ignored, or old ones used instead.
The command "tools/install-interceptors.sh" failed and exited with 100 during .
‵

File: tests/test_Consumer.py [hunk 1]
```diff
@@ -14,6 +14,8 @@ def test_basic_api():
         kc = Consumer()
     except TypeError as e:
         assert str(e) == "expected configuration dict"
+    else:
```
edenhill: should use http://doc.pytest.org/en/latest/reference.html#pytest-raises

hrchu: Yeap that will be better.

PR #652, opened 2019-08-06
Fix doc avro 495

lowercase24:
This is a fix for 495.  This includes the cached schema registry in the doc generation.


edenhill:
Thank you!

PR #653, opened 2019-08-06
Fix for 644 signature mismatch

lowercase24:
An extra argument was specified in the doc signature.  This fixes the discrepancy.  This resolves bug 644


edenhill:
Thank you!

PR #661, opened 2019-08-23
Add wheel metadata for additional libs

beezz:
This PR adds correct metadata to the wheel RECORD file for additional libraries.

Since some libraries are added to the wheels without also adding their correct metadata to the wheel RECORD file wheel verification process is failing.

That means that for example creating a [pex](https://github.com/pantsbuild/pex) distribution will fail with:

```
$ pex confluent_kafka -o test.pex
...
pex.vendor._vendored.wheel.wheel.install.BadWheelFile: No expected hash for file 'confluent_kafka/.libs/monitoring-interceptor.so'
```

### DIFF
--- a/tools/fixup-wheels.sh
+++ b/tools/fixup-wheels.sh
@@ -12,6 +12,7 @@ set -e
 wheelhouse=$1
 fixed_wheelhouse=$2
 platform=$(uname -s)
+gen_record_row=$(dirname $(readlink -f $0))/generate-record-row.py
 
 stagedir=$PWD/staging
 
@@ -26,7 +27,13 @@ fi
 
 mkdir -p $fixed_wheelhouse
 
+find_record_file () {
+  find . -name RECORD -type f | head -1
+}
+
 fixup_wheel_linux () {
+
+    local record_path=$(find_record_file)
 
     pushd confluent_kafka/.libs
 
@@ -62,6 +69,8 @@ fixup_wheel_linux () {
         fi
         echo "$lib dependencies:"
         ldd $lib
+
+        $gen_record_row $lib confluent_kafka/.libs/ >> ../../$record_path
     done
 
     popd # confluent_kafka/.libs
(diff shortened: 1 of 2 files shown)


tools/fixup-wheels.sh
```diff
@@ -26,8 +27,14 @@ fi
 
 mkdir -p $fixed_wheelhouse
 
+find_record_file () {
+  find . -name RECORD
```
edenhill: `find . -name RECORD -type f | head -1`

tools/generate-record-row.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+"""
+Generate wheel metadata RECORD file row for given file.
+
+Usefull in cases where you want to postprocess the wheel archive e.g. include
+additional libraries after the wheel has been build.
+"""
```
edenhill: Provide a Usage: line with what parameters are expected

tools/fixup-wheels.sh
beezz: Thanks for suggestion. Applied.

tools/generate-record-row.py
beezz: I added argparse which will provide `--help` and usage info.

beezz:
[clabot:check]

edenhill:
CI flake8 fails on Python3: 
https://travis-ci.org/confluentinc/confluent-kafka-python/jobs/575757575#L3345

beezz:
I added noqa for the line as it is for python2 compatibility so for python3 it will always complain.

edenhill:
@rnpridgeon The ball is yours on this.
I suggest you try this out for the upcoming RC before merging to see if there is anything else needed.

rnpridgeon:
Ack

beezz:
@rnpridgeon I see there was a release few days ago but this PR was not included and the published wheel is broken for use with pex.

Is there a problem with this PR? How can I help you?

rnpridgeon:
Nope sorry slipped through the cracks. If you want to reopen it we can create a toss away tag to build the wheels and confirm the behavior.

PR #663, opened 2019-08-26
V.1.2.0 rc1

rnpridgeon:
Openssl version remained the same
Tested against 2.7.16, 3.6.9, 3.7.4
Interceptors will be bumped in v1.2.0rc2 after manual validation of the latest build


setup.py
```diff
@@ -48,11 +62,6 @@ def get_install_requirements(path):
       data_files=[('', ['LICENSE.txt'])],
       install_requires=INSTALL_REQUIRES,
       extras_require={
-          'avro': [
-              'fastavro',
-              'requests',
-              'avro;python_version<"3.0"',
-              'avro-python3;python_version>"3.0"'
-          ],
-          'dev': get_install_requirements("test-requirements.txt")
```
edenhill: Is this file still relevant now?

PR #673, opened 2019-09-30
Show the error details when failing to register a schema

jacopofar:
I found out that if there are issues communicating with the schema registry the library raises an exception including the HTTP status but no further description of the problem.

This PR adds the content of the response to the exception to facilitate troubleshooting.


jacopofar:
[clabot:check]

mhowlett (approved):
thanks @jacopofar . ideally this information would be exposed programmatically on ClientError as well, but this is certainly a step forward, so we'll take it :-).

maybe worth noting this being depreciated eventually in favor of the new serde approach.

mhowlett:
merged via #1056

PR #676, opened 2019-10-01
Calling consumer.close() multiple times should not raise an Exception

mkmoisen:
Currently, calling `consumer.close` more than once will raise a `RuntimeError: Consumer already closed`.

The proposed changes will just return NULL if the user calls `close` on a consumer that has already been closed.

This will make it rerunnable and allow the developer to not have to trap this exception.

### DIFF
--- a/confluent_kafka/src/Consumer.c
+++ b/confluent_kafka/src/Consumer.c
@@ -981,11 +981,8 @@ static PyObject *Consumer_consume (Handle *self, PyObject *args,
 static PyObject *Consumer_close (Handle *self, PyObject *ignore) {
         CallState cs;
 
-        if (!self->rk) {
-                PyErr_SetString(PyExc_RuntimeError,
-                                "Consumer already closed");
-                return NULL;
-        }
+        if (!self->rk)
+                Py_RETURN_NONE;
 
         CallState_begin(self, &cs);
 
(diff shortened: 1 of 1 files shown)


mkmoisen:
[clabot:check]

edenhill:
What's the use-case for calling close multiple times?

mkmoisen:
Hi @edenhill please check the new PR [here](https://github.com/confluentinc/confluent-kafka-python/pull/678) which is passing. I've copied and pasted the following over there.

The use case is to make `close` rerunnable, for example if you wanted to make a context manager, or if you had a large function wrapped in a try/except/finally. Another programmer who calls your context manager can close inside if he wants; otherwise the context manager will close for him.


    import contextlib
    from confluent_kafka import Consumer
    
    @contextlib.contextmanager
    def kafka():
        consumer = Consumer({...})
        try:
            yield consumer
        finally:
            # currently, would need to wrap this in a try/except
            consumer.close()

    # ...
    # Some programmers may want to call close explicitly
    with kafka() as consumer:
            # do something
            consumer.close()

    # Other programmers would prefer the context manager to take care
    with kafka() as consumer:
        # do something


I'm also under the impression that the majority of the other python libraries that close resources also make close rerunnable. I think it would be a good idea to have confluent_kafka follow this pattern.

    from requests import Session
    s = Session()
    s.close()
    s.close()  # does not raise an exception

    import sqlite3
    conn = sqlite3.connect(':memory:')
    conn.close()
    conn.close()  # does not raise an exception

edenhill:
@rnpridgeon What are your thoughts on this?

PR #683, opened 2019-10-03
Fix for doc issues

joel-hamill:
I believe this is generated content, but here is a start on what will fix the doc issues.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,3 +1,4 @@
+
 Welcome to Confluent's Python client for Apache Kafka documentation
 ===================================================================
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1,3 +1,4 @@
+
```
joel-hamill: a space required at the top of the file, or and RST anchor (e.g. `.. _confluent-kafka-index:`

joel-hamill:
@ewencp is this index.rst page auto-generated - or is it hand-coded?

edenhill:
@rnpridgeon Can you verify that the removal of `:members:` does not actually remove the class member listings from the generated docs?

ewencp:
@joel-hamill afaik it is manually generated. the autogenerated stuff is just via the `automodule` directive.

joel-hamill:
@ewencp well then this is ridiculously easy to fix. 🤷‍♂

Issue #610, opened 2019-05-29
confluent-kafka 1.0.0 is incompatible with avro-python3 1.9.0


hoesler:
Description
===========

Since version 1.9.0 of avro-python3 [Schema.props](https://github.com/apache/avro/blob/3c76495e9524ef322726d03d7ee406be89e8fde0/lang/py3/avro/schema.py#L211) returns a `MappingProxyType` object which not json-serializable by default and leads to following exceptions:

```
...
  File "<venv>/lib/python3.7/site-packages/confluent_kafka/avro/__init__.py", line 80, in produce
    value = self._serializer.encode_record_with_schema(topic, value_schema, value)
  File "<venv>/lib/python3.7/site-packages/confluent_kafka/avro/serializer/message_serializer.py", line 105, in encode_record_with_schema
    schema_id = self.registry_client.register(subject, schema)
  File "<venv>/lib/python3.7/site-packages/confluent_kafka/avro/cached_schema_registry_client.py", line 214, in register
    body = {'schema': json.dumps(avro_schema.to_json())}
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type mappingproxy is not JSON serializable
```

To fix this, the registry client (and everything else) should use [`Schema.__str__`](https://github.com/apache/avro/blob/3c76495e9524ef322726d03d7ee406be89e8fde0/lang/py3/avro/schema.py#L218) to get the json representation instead of calling `json.dumps(avro_schema.to_json())`

mpyatishev:
I have the same problem

rnpridgeon:
[...]

PR #694, opened 2019-10-17
Fix using fastavro if reader schema is None

jamesjrg:
1. Don't attempt to call to_json on reader schema if it's None
Reader schema is optional, yet currently an exception in calling reader_schema_obj.to_json() will cause the `if HAS_FAST:` block to raise an exception and hence a fallback to the Python avro implementation

2. Use json.loads(str(schema)) rather than schema.to_json(), due to #610

### DIFF
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -172,7 +172,11 @@ class MessageSerializer(object):
             # try to use fast avro
             try:
                 fast_avro_writer_schema = parse_schema(writer_schema_obj.to_json())
-                fast_avro_reader_schema = parse_schema(reader_schema_obj.to_json())
+                fast_avro_reader_schema = (
+                    parse_schema(reader_schema_obj.to_json())
+                    if reader_schema_obj
+                    else None
+                )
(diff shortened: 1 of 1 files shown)


jamesjrg:
[clabot:check]

rnpridgeon:
Thanks @jamesjrg , I am actually refactoring the avro module this week to eliminate the fallback behavior all together as its a bit messy and I don't believe its needed anymore. It may have been when this client was originally drafted but fastavro appears to support most if not all the things the apache distribution does without the semver headaches. 

Moving forward we will provide two avro serializers, Apache and FastAvro which will operate independently of one another. The default however will be Fast Avro. The idea being that we can eventually phase out the apache module all together. 

That said have I'd like to ask if you have ever run into a situation where the fallback behavior was actually needed (barring bugs in the fallback logic itself which is a bit messy)? 

Thanks, 
Ryan

jamesjrg:
Never needed the fallback behaviour nope! Though did at one point need to pass an extra argument to fastavro to get back union names: https://github.com/fastavro/fastavro/commit/eb530765e91a252922ea5ebb57dd2632a7a346d5

PR #695, opened 2019-10-17
Iterating on the docs

mhowlett:


README.md
```diff
@@ -4,27 +4,19 @@ Confluent's Python Client for Apache Kafka<sup>TM</sup>
 **confluent-kafka-python** is Confluent's Python client for [Apache Kafka](http://kafka.apache.org/) and the
 [Confluent Platform](https://www.confluent.io/product/compare/).
 
-Features:
+We provide high-level Producer, Consumer and AdminClient classes analogous to the Java client. These are backwards and forwards compatible with all Kafka brokers from version 0.8.
 
-- **High performance** - confluent-kafka-python is a lightweight wrapper around
-[librdkafka](https://github.com/edenhill/librdkafka), a finely tuned C
-client.
+Characteristics:
```
edenhill: should the characteristics be 'Reliable, Performant, Future Proof'? The current ones don't seem to be in the same tense, sense, fence, hence

README.md
```diff
@@ -4,27 +4,19 @@ Confluent's Python Client for Apache Kafka<sup>TM</sup>
 **confluent-kafka-python** is Confluent's Python client for [Apache Kafka](http://kafka.apache.org/) and the
 [Confluent Platform](https://www.confluent.io/product/compare/).
 
-Features:
+We provide high-level Producer, Consumer and AdminClient classes analogous to the Java client. These are backwards and forwards compatible with all Kafka brokers from version 0.8.
```
edenhill: we is vague, 'confluent-kafka-python provides a high-level Pro..` ? Or Confluent .. ?

README.md
```diff
@@ -4,27 +4,19 @@ Confluent's Python Client for Apache Kafka<sup>TM</sup>
...
-- **High performance** - confluent-kafka-python is a lightweight wrapper around
-[librdkafka](https://github.com/edenhill/librdkafka), a finely tuned C
-client.
+Characteristics:
 
-- **Reliability** - There are a lot of details to get right when writing an Apache Kafka
-client. We get them right in one place (librdkafka) and leverage this work
-across all of our clients (also [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go)
-and [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet)).
+- **Reliability** - The Python client is a lightweight wrapper around [librdkafka](https://github.com/edenhill/librdkafka) which is widely deployed in a diverse set of demanding production scenarios. It's tested using [the same set of system tests](https://github.com/confluentinc/confluent-kafka-python/tree/master/confluent_kafka/kafkatest) as the Java client [and more](https://github.com/confluentinc/confluent-kafka-python/tree/master/tests). It's supported by [Confluent](https://confluent.io).
```
edenhill: I'm not sure lightweight wrapper is what people are after, how about 'The Python client provides a robust wrapper?/python client on top of librdkafka .. '

README.md
```diff
@@ -4,27 +4,19 @@ Confluent's Python Client for Apache Kafka<sup>TM</sup>
...
 
-- **Reliability** - There are a lot of details to get right when writing an Apache Kafka
-client. We get them right in one place (librdkafka) and leverage this work
-across all of our clients (also [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go)
-and [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet)).
+- **Reliability** - The Python client is a lightweight wrapper around [librdkafka](https://github.com/edenhill/librdkafka) which is widely deployed in a diverse set of demanding production scenarios. It's tested using [the same set of system tests](https://github.com/confluentinc/confluent-kafka-python/tree/master/confluent_kafka/kafkatest) as the Java client [and more](https://github.com/confluentinc/confluent-kafka-python/tree/master/tests). It's supported by [Confluent](https://confluent.io). 
 
-- **Supported** - Commercial support is offered by
-[Confluent](https://confluent.io/).
+- **Performance** - Performance is a key design consideration. Maximum throughput is comparable to the Java client for larger message sizes (where the overhead of the Python interpreter has less impact).
```
edenhill: Minimal latency and maximum throughput .. ?

README.md
```diff
@@ -177,8 +172,31 @@ while True:
...
+        try:
+            f.result()  # The result itself is None
+            print("Topic {} created".format(topic))
+        except Exception as e:
+            print("Failed to create topic {}: {}".format(topic, e))
+```
+
+
+
+Thread Safety
```
edenhill: I'm not sure this is the place, better off having this in the API docs for each client type.

README.md
mhowlett: I flip flopped between the two a number of times corresponding to different wording. Yes, i think this variation is better here.

README.md
mhowlett: I iterated on the wording - see what you think. found this remarkably difficult to write...

README.md
mhowlett: I can remove the word lightweight. i my mind this reinforced that the client would be more robust, but i take your point. I don't like the word robust here (too many salesy words already, fine line).

README.md
mhowlett: ok, i added latency into the mix (with slightly more precise wording)

README.md
mhowlett: I think this should definitely be on the front page - it seems an important thing people want to check off, and it adds to the vibe of credibility, which we want to be getting across up front.

mhowlett: its also on the front page of other clients, and people incorrectly interpret it as a differentiator for them.

README.md
mhowlett: i also removed the word demanding, as i started to think it takes more away in terms of coming across too salesy for the amount of value it adds.

README.md
```diff
@@ -177,18 +177,22 @@ c.close()
...
-        try:
-            f.result()  # The result itself is None
-            print("Topic {} created".format(topic))
-        except Exception as e:
-            print("Failed to create topic {}: {}".format(topic, e))
+from confluent_kafka.admin import AdminClient, NewTopic
+
+a = AdminClient({'bootstrap.servers': 'mybroker'})
+
+new_topics = [NewTopic(topic, num_partitions=3, replication_factor=1) for topic in ["topic1", "topic2"]]
```
edenhill: We should use the default replication factor of 3 to not trick people into using replication_factor=1

README.md
```diff
@@ -2,29 +2,21 @@ Confluent's Python Client for Apache Kafka<sup>TM</sup>
 =======================================================
 
 **confluent-kafka-python** is Confluent's Python client for [Apache Kafka](http://kafka.apache.org/) and the
-[Confluent Platform](https://www.confluent.io/product/compare/).
+[Confluent Platform](https://www.confluent.io/product/compare/). It provides a high-level Producer, Consumer
+and AdminClient analogous to the Java client which are backwards and forwards compatible with all
```
edenhill: I'm not sure I like the `analogous to the Java client` bit, do Python users care?
And analogous is a foreign word for most non-native english speakers

README.md
```diff
@@ -2,29 +2,21 @@ Confluent's Python Client for Apache Kafka<sup>TM</sup>
...
+- **Reliable** - The Python client is a wrapper around [librdkafka](https://github.com/edenhill/librdkafka) (provided automatically via binary wheels) which is widely deployed in a diverse set of production scenarios. It's tested using [the same set of system tests](https://github.com/confluentinc/confluent-kafka-python/tree/master/confluent_kafka/kafkatest) as the Java client [and more](https://github.com/confluentinc/confluent-kafka-python/tree/master/tests). It's supported by [Confluent](https://confluent.io). 
 
-- **Reliability** - There are a lot of details to get right when writing an Apache Kafka
-client. We get them right in one place (librdkafka) and leverage this work
-across all of our clients (also [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go)
-and [confluent-kafka-dotnet](https://github.com/confluentinc/confluent-kafka-dotnet)).
-
-- **Supported** - Commercial support is offered by
-[Confluent](https://confluent.io/).
+- **Performant** - Performance is a key design consideration. Latency is on par with the Java client, as is maximum throughput for larger message sizes (where the overhead of the Python interpreter has less impact).
```
edenhill: "as is maximum throughput" does not read well.

README.md
```diff
@@ -177,8 +172,35 @@ while True:
 c.close()
 ```
 
-See the [examples](examples) directory for more examples, including [how to configure](examples/confluent_cloud.py) the python client for use with
-[Confluent Cloud](https://www.confluent.io/confluent-cloud/).
+**AdminClient**
```
edenhill: Missing ":" at end of "AdminClient", or remove from "Producer:" and "Consumer:"
while you're editing the file, the Avro example headers are missing a ":"

README.md
mhowlett: i'd agree with that.

README.md
mhowlett: i think it's ok, but ok. Two sentences will do.

README.md
mhowlett: these are all effectively headings, so all better off without the ":" imo, removed everywhere.

README.md
mhowlett: there's a tradeoff here between making sure there is zero friction for people running examples on a dev machine and best practice for running in production. I would argue the former is more important because by the time people get to the latter they've already adopted kafka and they're not going to stop due to this misunderstanding (but every point of friction early on is a risk of giving up). I'll add a comment instead, which I think is good enough to address your concern. The other way to do this would be push people towards a docker setup script for local testing that has 3 brokers (this seems optimal to me, but we don't have it yet).

README.md
```diff
@@ -1,30 +1,19 @@
-Confluent's Python Client for Apache Kafka<sup>TM</sup>
```
edenhill: Why this change? I believe the former wording was more legally correct

README.md
```diff
@@ -1,30 +1,19 @@
...
-=======================================================
+Confluent's Kafka Python Client
+===============================
 
-**confluent-kafka-python** is Confluent's Python client for [Apache Kafka](http://kafka.apache.org/) and the
-[Confluent Platform](https://www.confluent.io/product/compare/).
+**confluent-kafka-python** provides high-level Producer, Consumer and AdminClient classes compatible with all
```
edenhill: "classes" does not really mean anything, they could be empty.
Use "clients", "implementations", or something similar.

README.md
```diff
@@ -34,7 +23,10 @@ See the [API documentation](http://docs.confluent.io/current/clients/confluent-k
 Usage
 =====
 
-**Producer:**
+Below are some examples of typical usage. For more examples, including [how to configure](examples/confluent_cloud.py) the python client for use with
```
edenhill: @ybyzek will definitely want a link to https://github.com/confluentinc/examples/tree/5.3.1-post/clients/cloud/python

README.md
```diff
@@ -190,11 +211,11 @@ Install
           confluent-kafka  using the command in the "Install from
           source from PyPi" section below.
 
-**Install self-contained binary wheels for OSX and Linux from PyPi:**
+**Install self-contained binary wheels for OSX and Linux from PyPi**
```
edenhill: And Windows, so not really sure we need to spell out the platforms since we provide wheels for all relevant ones.

README.md
```diff
@@ -261,7 +282,7 @@ Tests
 =====
 
 
-**Run unit-tests:**
+**Run unit-tests**
```
edenhill: Not sure this is relevant in the top level README, maybe we should just link to tests/README.md, or let people find it themselves.

README.md
```diff
@@ -270,7 +291,7 @@ In order to run full test suite, simply execute:
 **NOTE**: Requires `tox` (please install with `pip install tox`), several supported versions of Python on your path, and `librdkafka` [installed](tools/bootstrap-librdkafka.sh) into `tmp-build`.
 
 
-**Integration tests:**
+**Integration tests**
```
edenhill: dito

README.md
mhowlett: It's an attempt to better match how people will search for this in google. I thought the legal implications were probably negligible - we're not trying to mislead, and the TM is present immediately below.

It may or may not meaningfully help us (exact phrase matches are less important than they once were). I'm pretty sure it won't be worse.

Anyway, I'll change it back.

README.md
mhowlett: yeah. that was a last minute change as it flowed better. i'll change it back to having no qualifier.

README.md
mhowlett: i'll add a link

README.md
mhowlett: completely agree, was thinking the same. same in .net. this was just supposed to be a quick iterative change, small PR, merge quickly, that sort of thing. but since you bring it up, i'll reorg it.

README.md
mhowlett: I want to also add links to blog posts, and generally organize information/links out logically as per: https://www.divio.com/blog/documentation/

README.md
mhowlett: sounds good, will make the change in this PR

README.md
mhowlett: the README.md should be completely focused on users, so i moved all this stuff to a new DEVELOPER.md doc.

PR #713, opened 2019-10-31
Don't install the LICENSE as a datafile

dirkmueller:
Data files are installed in the wrong directory when packaging this,
and it is very uncommon to do so. data files should be files needed
for the runtime only.


dirkmueller:
I can not easily sign such a CLA.

rnpridgeon:
Thanks for bringing this to our attention. We will look into this a bit more to figure out what exactly the best practice is here. I suspect omitting the license entirely is less than ideal.

Atheuz:
Add in license_files key instead?
See here: https://setuptools.pypa.io/en/latest/references/keywords.html

MSeal:
This issue appears to have been resolved by a later move to pyproject.toml -- I noticed the manifest for licenses was still odd and used the newer license fields in https://github.com/confluentinc/confluent-kafka-python/pull/2001 to further normalize the project file registration.

PR #724, opened 2019-11-25
Significant performance boost: only write to the writer schema cache once.

fimmtiu:
We noticed a performance problem with confluent-kafka-python while profiling our app: `_get_decoder_func` was being called a small number of times and took up a tiny amount of time, but `_get_encoder_func` seemed as if it was being called once per message and was taking up about 10% of our total runtime. This one-line patch gave us a 9% speed boost.

What's happening is that we have a cache for the encoder functions, but we're not actually using it in the `encode_record_with_schema` method: https://github.com/confluentinc/confluent-kafka-python/blob/50e09cf35445a7c2709c3e97683c17f171d4a9eb/confluent_kafka/avro/serializer/message_serializer.py#L113 Every time that method is called, it will set a new entry in the `id_to_writers` cache, even if an identical one already exists. The fix is trivial: we should actually use the cache and not set new entries if there's already one in there.

### DIFF
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -110,7 +110,8 @@ class MessageSerializer(object):
             raise serialize_err(message)
 
         # cache writer
-        self.id_to_writers[schema_id] = self._get_encoder_func(schema)
+        if schema_id not in self.id_to_writers:
+            self.id_to_writers[schema_id] = self._get_encoder_func(schema)
 
         return self.encode_record_with_schema_id(schema_id, record, is_key=is_key)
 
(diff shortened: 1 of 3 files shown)


fimmtiu:
[clabot:check]

fimmtiu:
@edenhill / @mhowlett : Any thoughts on this? It's a pretty trivial change with a nice benefit, and we've been using it in production for a few weeks now. (The CI failures seem unrelated to the actual code in the PR.)

fimmtiu:
@rnpridgeon Done! Thanks for the suggestion.

AndreiLieb:
It seems to me that this can be fixed simply by removing line 113. Line 113 is not needed as `encode_record_with_schema` delegates the work to `encode_record_with_schema_id` where this logic is already present.

mhowlett:
well pointed out @AndreiLieb . since the code does the job as is, and it'll be depreciated soon, just going with what's there to avoid mucking about. merged via #1057

Issue #722, opened 2019-11-21
SSL handshake error gets swallowed


naartjie:
Description
===========

This issue might be with the way librdkafka does the error propagation, and the issue might need to be logged there. I am logging it here, because I encountered it while using this library.

It seems when there is an error in the SSL handshake connecting to the broker, the client code doesn't get an error, and looks like it's silently swallowed. You can see it in the logs by setting debug in config:

```python
 'debug' : 'security,broker'
```
The logs show something like this:

```
sasl_ssl://xyz.gcp.confluent.cloud:9092/bootstrap: failed: err: Local: SSL error: (errno: Undefined error: 0)
```

(Here is another issue which appears to have the same problem: #697)

How to reproduce
================

Set the SSL CA manually, to one which does not match the CA of the broker. I used [this one](https://www.geotrust.com/resources/root_certificates/certificates/GeoTrust_Global_CA2.pem) to test, since it doesn't match Confluent Cloud's one, it gets us a repro:

```python
  'ssl.ca.location': '/dir/invalid-ca.pem',
```

The error won't be reported in client code - the client just sits there waiting, but it will be visible in librdkafka debug logs.

Checklist
=========

 - confluent-kafka-python: 1.2.0
 - librdkafka version: 1.2.0
 - Apache Kafka broker version: Confluent Cloud
 - Client configuration:

```python
{
    'bootstrap.servers': 'xyz.gcp.confluent.cloud:9092',
    'sasl.mechanisms': 'PLAIN',
    'security.protocol': 'SASL_SSL',
    'sasl.username': '${USERNAME}',
    'sasl.password': '${PASSWORD}'
}
```
 - Operating system: OSX 10.14.6

colinhicks:
Hi @naartjie,

Supplying `error_cb` to the producer or consumer instance should provide a way to handle connection errors like you describe.

For example, if I provide a similar producer configuration as above, plus a sample `error_cb`:
```python
def error_cb(err):
    print(err.str())

 p = Producer({
    'bootstrap.servers': 'pkc-______.us-east1.gcp.confluent.cloud:9092',
    'sasl.mechanisms': 'PLAIN',    
    'security.protocol': 'SASL_SSL',
    'sasl.username': '<>',
    'sasl.password': '<>',
    'ssl.ca.location': 'invalid.pem',
}, error_cb=error_cb)
```

Then, when I flush the producer instance, the callback prints respective errors:
```
[...]

PR #726, opened 2019-11-27
add error_cb to confluent example

naartjie:
Hi. I thought it might be handy to add the `error_cb` to the confluent example, in case someone else gets tripped up by errors getting swallowed up, like in #722.

### DIFF
--- a/examples/confluent_cloud.py
+++ b/examples/confluent_cloud.py
@@ -50,6 +50,8 @@ import uuid
 
 from confluent_kafka import Producer, Consumer
 
+def error_cb(err):
+    print(err.str())
 
 p = Producer({
     'bootstrap.servers': '<ccloud bootstrap servers>',
@@ -57,7 +59,7 @@ p = Producer({
     'security.protocol': 'SASL_SSL',
     'sasl.username': '<ccloud key>',
     'sasl.password': '<ccloud secret>'
-})
+}, error_cb=error_cb)
 
 
 def acked(err, msg):
@@ -84,7 +86,7 @@ c = Consumer({
     'sasl.password': '<ccloud secret>',
     'group.id': str(uuid.uuid1()),  # this will create a new consumer group on each invocation.
     'auto.offset.reset': 'earliest'
-})
+}, error_cb=error_cb)
 
 c.subscribe(['python-test-topic'])
 


naartjie:
[clabot:check]

mhowlett:
good idea! even better, it would be good to add the following note to the callback method so that people understand better what the error events are for (this creates a lot of confusion).

```
# called to expose error events e.g. connection failures or all brokers down that don't correspond
# to any specific API function call. Note that the client will try to automatically recover from errors that
# are not marked as fatal. Non-fatal errors events should be interpreted as informational rather 
# than catastrophic.
```

mhowlett:
(but stick to 80 char wide limit - i don't think the above does)

edenhill:
There is a bit of librdkafka nuance here:
if an error_cb is not configured any broker/connection errors will be logged by default,
but if an error_cb is configured the connection errors will be enqueued on the event queue that is served by poll() and that will eventually trigger the callback.
The problem, as shown in the original issue, is that while there is no error_cb provided by the python script, the confluent-kafka-python bindings still set up an internal default error_cb that simply ignores the errors if no py error_cb was set.
So while I think that adding error_cbs is a good addition to the examples, we should perhaps not register the C error_cb if no Py error_cb is set, allowing the default logger to emit the connection failures.
But then there is the case with fatal errors that needs to be handled by the C error_cb regardless if a Py error_cb was set... see https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/src/confluent_kafka.c#L1224


Another problem is that if you register an error_cb and call something blocking that is not poll(), such as metadata(), you will not get any error logs (because error_cb is set) and you wont get any error_cbs because you are not calling poll().

mhowlett:
this is such a complicated arrangement to have exposed to the end user - no one is going to use this correctly - we need to simplify it.

examples/confluent_cloud.py
```diff
@@ -50,14 +50,16 @@
...
+def error_cb(err):
+    print(err.str())
 
 p = Producer({
     'bootstrap.servers': '<ccloud bootstrap servers>',
     'sasl.mechanisms': 'PLAIN',
     'security.protocol': 'SASL_SSL',
     'sasl.username': '<ccloud key>',
     'sasl.password': '<ccloud secret>'
-})
```
rnpridgeon: It may look a read a little easier if we set the error cb in the conf dict itself. 

i.e. https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/integration/integration_test.py#L369

edenhill:
I think we should close this PR and fix so that the Producer does not set the C error_cb unless a Py error_cb is specified. This way librdkafka will log broker failures by default, and it wont require the application to provide an error_cb, nor poll() to have it triggered.

rnpridgeon:
https://github.com/confluentinc/confluent-kafka-python/issues/745

PR #727, opened 2019-11-28
v1.3.0 release preparations and doc/build fixes

edenhill:


docs/index.rst
```diff
@@ -19,62 +19,56 @@ Consumer
 ********
 
 .. autoclass:: confluent_kafka.Consumer
-
+   :members:
```
mhowlett: i'm doing this over in #730, but i guess leave this here in case it doesn't get in for 1.3

mhowlett: actually - i don't know the full story, but sphinx is spitting out warnings due to comments in the .c file(s) that you haven't updated i think and i think the docs repo has warnings == error set. so these change might upset the docs people. I'm addressing that in #730 (note: I could be wrong about this).

mhowlett: oh, you have updated the c file, ignore me.

PR #730, opened 2019-12-02
iterating on api docs

mhowlett:
some quick wins for the API docs. most importantly, the top of the index page currently looks like a mess. this tidies that up. also adds back the auto generated member docs.

lots of other improvements possible. let's keep the updates time-bounded for now. incremental quick wins...


confluent_kafka/src/Consumer.c
```diff
@@ -1421,26 +1421,17 @@ PyTypeObject ConsumerType = {
...
-        "\n"
-        "\n"
-        ".. py:function:: on_commit(err, partitions)\n"
-        "\n"
-        "  :param KafkaError err: Commit error object, or None on success.\n"
-        "  :param list(TopicPartition) partitions: List of partitions with "
-        "their committed offsets or per-partition errors.\n"
+        "Create a new Consumer instance using the provided configuration *dict* ("
+        "including properties and callback functions). Additional information: "
+        "https://docs.confluent.io/current/clients/confluent-kafka-python/#configuration"
```
mhowlett: this way of linking is not ideal, but I don't know how to do better.

confluent_kafka/src/Admin.c
```diff
@@ -1540,7 +1540,7 @@ PyTypeObject AdminType = {
         "\n"
         ".. py:function:: Admin(**kwargs)\n"
         "\n"
-        "  Create new AdminClient instance using provided configuration dict.\n"
+        "  Create a new AdminClient instance using the provided configuration dict.\n"
```
edenhill: I once reviewed a book, and the editor removed most uses of "a/an" and "the".
The book wasn't very successful.

edenhill:
incorporated in #727

PR #733, opened 2019-12-04
Correct the SchemaRegistry authentication for SASL_INHERIT

abij:
We ran into an issue connecting the SchemaRegistry inside the Producer.

I looked into the problem, it turns out the Server(broker) can use multiple `sasl.mechanisms` (plural) and the clients (producer / consumer) `sasl.mechanism` (singular). There was an issue passing through this config, when the setting `schema.registry.basic.auth.credentials.source` is set to `SASL_INHERIT`.

I think there is no point in inheriting the credentials, since the SchemaRegistry has separate key/secret, but maybe there is a way to use the same credentials for both (SR + Cluster).

This PR will fix the SASL_INHERIT in SchemaRegistry.

I did add an example for `confluent_cloud.py`, but looks like the `avro-cli.py` has a working solution already, so I reverted this commit.

### DIFF
--- a/confluent_kafka/avro/__init__.py
+++ b/confluent_kafka/avro/__init__.py
@@ -33,7 +33,7 @@ class AvroProducer(Producer):
                    for key, value in config.items() if key.startswith("schema.registry")}
 
         if sr_conf.get("basic.auth.credentials.source") == 'SASL_INHERIT':
-            sr_conf['sasl.mechanisms'] = config.get('sasl.mechanisms', '')
+            sr_conf['sasl.mechanism'] = config.get('sasl.mechanism', '')
             sr_conf['sasl.username'] = config.get('sasl.username', '')
             sr_conf['sasl.password'] = config.get('sasl.password', '')
 
@@ -110,7 +110,7 @@ class AvroConsumer(Consumer):
                    for key, value in config.items() if key.startswith("schema.registry")}
 
         if sr_conf.get("basic.auth.credentials.source") == 'SASL_INHERIT':
-            sr_conf['sasl.mechanisms'] = config.get('sasl.mechanisms', '')
+            sr_conf['sasl.mechanism'] = config.get('sasl.mechanism', '')
             sr_conf['sasl.username'] = config.get('sasl.username', '')
             sr_conf['sasl.password'] = config.get('sasl.password', '')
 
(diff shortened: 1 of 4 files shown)


abij:
[clabot:check]

confluent_kafka/avro/__init__.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, config, default_key_schema=None,
                    for key, value in config.items() if key.startswith("schema.registry")}
 
         if sr_conf.get("basic.auth.credentials.source") == 'SASL_INHERIT':
-            sr_conf['sasl.mechanisms'] = config.get('sasl.mechanisms', '')
+            sr_conf['sasl.mechanism'] = config.get('sasl.mechanism', '')
```
edenhill: sasl.mechanism is just an alias for sasl.mechanisms, so this change will break existing applications that use sasl.mechanisms.

What we'll need to do here is something like:
`sr_conf['sasl.mechanism'] = config.get('sasl.mechanism', config.get('sasl.mechanisms', ''))`

confluent_kafka/avro/__init__.py
```diff
@@ -110,7 +110,7 @@ def __init__(self, config, schema_registry=None, reader_key_schema=None, reader_
...
```
edenhill: dito

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -135,8 +135,8 @@ def _configure_basic_auth(conf):
             raise ValueError("schema.registry.basic.auth.credentials.source must be one of {}"
                              .format(VALID_AUTH_PROVIDERS))
         if auth_provider == 'SASL_INHERIT':
-            if conf.pop('sasl.mechanism', '').upper() is ['GSSAPI']:
-                raise ValueError("SASL_INHERIT does not support SASL mechanisms GSSAPI")
+            if conf.pop('sasl.mechanism', '').upper() == 'GSSAPI':
```
edenhill: Good catch!

abij:
I did add the plural fallback for mechanism(s), if you have a suggestion for the deprecation warning I could include this in the PR as-well.

rnpridgeon:
In retrospect deprecation since they are one and the same for the underlying producer/consumer clients. 

Can you add a quick test to `tests/avro/test_cached_client.py` though. Use the singular variant with `GSSAPI` so we have slightly better coverage.

andharris:
@abij are you planning on still pursuing getting this PR merged? I recently hit this issue and would like to help however possible.  thanks!

abij:
Added a testcase to cover these lines as suggested by @rnpridgeon.

edenhill:
Thank you for your contribution!

PR #735, opened 2019-12-09
download-s3: only match confluent-kafka-python builds

edenhill:
And update S3 collector to handle more than 1000 objects in listing.


edenhill:
@rnpridgeon Merge when you see fit, based on your release work.

rnpridgeon:
@edenhill looks like you have some flake errors. Can you push a change to `tox.ini` to ignore tools or address the flake errors.

Issue #741, opened 2019-12-12
AvroProducer fails silently when not setting `"security.protocol": "ssl"`


ruurtjan:
Description
===========

Today, we forgot to set the security protocol to ssl. This resulted in flush taking a couple of minutes, then proceeding without storing anything to Kafka. The schemas got persisted to the schema registry though.

Expected behaviour is some error message, preferably indicating that you forgot to specify the security protocol.

Output with debug on:
```
%7|1576156120.005|BROKER|rdkafka#producer-1| [thrd:app]: [...broker1...]:9094/bootstrap: Added new broker with NodeId -1
%7|1576156120.005|BRKMAIN|rdkafka#producer-1| [thrd::0/internal]: :0/internal: Enter main broker thread
%7|1576156120.005|BRKMAIN|rdkafka#producer-1| [thrd:[...broker1...].]: [...broker1...]:9094/bootstrap: Enter main broker thread
%7|1576156120.005|BROKER|rdkafka#producer-1| [thrd:app]: [...broker2...]:9094/bootstrap: Added new broker with NodeId -1
%7|1576156120.005|BROKER|rdkafka#producer-1| [thrd:app]: [...broker3...]:9094/bootstrap: Added new broker with NodeId -1
%7|1576156120.005|BRKMAIN|rdkafka#producer-1| [thrd:[...broker2...].]: [...broker2...]:9094/bootstrap: Enter main broker thread
%7|1576156120.005|CONNECT|rdkafka#producer-1| [thrd:app]: [...broker3...]:9094/bootstrap: Selected for cluster connection: bootstrap servers added (broker has 0 connection attempt(s))
%7|1576156120.005|INIT|rdkafka#producer-1| [thrd:app]: librdkafka v1.2.0 (0x10200ff) rdkafka#producer-1 initialized (builtin.features gzip,snappy,ssl,sasl,regex,lz4,sasl_plain,sasl_scram,plugins,zstd,sasl_oauthbearer, GCC GXX PKGCONFIG INSTALL GNULD LDS LIBDL PLUGINS ZLIB SSL ZSTD HDRHISTOGRAM SNAPPY SOCKEM SASL_SCRAM SASL_OAUTHBEARER CRC32C_HW, debug 0x46)
%7|1576156120.005|BRKMAIN|rdkafka#producer-1| [thrd:[...broker3...].]: [...broker3...]:9094/bootstrap: Enter main broker thread
[...]

PR #742, opened 2019-12-12
README: Add delivery report to Avro example

edenhill:
The AvroProducer.produce() call should also take an on_delivery|callback argument.

See discussion in https://github.com/confluentinc/confluent-kafka-python/issues/741#issuecomment-565042361


edenhill:
That seems like too much work for a small README example, and the message contents should not be used to identify the message, but rather a bound variable, meaning the payload isn't that meaningful here.

Issue #705, opened 2019-10-25
how to check connection before producer or consumer start.


hackrole:
Description
===========
the errors was that, we have not install or run kafka in localhost, while I write a example to test
the producer and consumer, it still will block forever. 
there maybe sometime as we config the url error or forget to start the kafka server.
but the behavor above make it hard to discover or debug.

so I wonder if there any way to check the connnection for the producer and consumer before loop to poll or produce.

mhowlett:
set a callback via `error_cb` in the producer or consumer constructor and check for `code=_ALL_BROKERS_DOWN` (which will happen quickly if no brokers are available). you generally won't want to exit out of your application if it's been up for some time, since the clients will automatically recover, so you'll want to use sort of timer to check how long the application has been running before exiting.

https://docs.confluent.io/current/clients/confluent-kafka-python/#configuration

mhowlett:
this is a common question, and I think very valid usage. we could consider providing some blocking/awaitable constructor mechanism in the future in addition to the existing one.

edenhill:
I disagree, why differentiate between the case where brokers are unavailable at startup vs during runtime? It is a distributed system so the downtime can happen at any time and singleing out specific cases does not make the system more robust.

Instead I recommend focusing on the data needs
 * what time is allowed to produce this piece of data (`message.timeout.ms`)?
 * what is the longest period without new incoming data before the pipeline is deemed malfunctional (consumer)? e.g., if we haven't received a new message within 10 minutes something is fishy.

mhowlett:
They're very different scenarios:

If all brokers are down on startup, the probabilities are very much in favor of this being a config / firewall / routing / forgot to start my local test server etc. problem that's not going to resolve itself, and you don't want to handle this by pretending everything is fine. This happens very frequently - we shouldn't assume it never happens. Also, if the server is genuinely down and will eventually come back up, and an app is started, then I think good chance whoever/whatever started the app might not know that, and would like to!

More color / examples:

If I have an app i'm using on my local machine, and I start it up and all brokers are down, I don't want it to hang there without knowing what's going on - the app should say kafka seems to be down as soon as it knows it, then quit or take some action as soon as it comes up. How do you implement this now? You need to add an 'error' handler, react to the all brokers down message, also set a log handler, work out what debug level you need to set and what message you need to react to there to know when the cluster is up. It's hard to implement this very common requirement. Much better UX to have blocking/awaitable startup code.

[...]

PR #744, opened 2019-12-13
Add broker connectivity verification example to FAQ

rnpridgeon:
in response to #705

### DIFF
--- a/examples/faq/are_you_there.py
+++ b/examples/faq/are_you_there.py
@@ -1,0 +1,66 @@
+from confluent_kafka import Producer, KafkaError
+
+
+class ConfluentException(Exception):
+    """
+        Converts KafkaError into an exception for handling by an application.
+
+        :param KafkaError kafka_error: The KafkaError to convert.
+        :param str msg: Optional Error message override.
+    """
+    def __init__(self, kafka_error, msg=None):
+        self.code = kafka_error.code()
+
+        if msg is None:
+            msg = kafka_error.str()
+
+        self.message = msg
+
+
+class DeliveryReporter(object):
+    __slots__ = ["success"]
+
+    def __init__(self):
+        self.success = False
+
+    def __call__(self, err, msg):
+        """
+        Handles per message delivery reports indicating success or failed delivery attempts.
+
+        :param err: KafkaError indicating message delivery failed.
+        :param msg: The message delivered to the broker.
+        """
+        if err:
+            self.error(err)
+
+        self.success = True
+        print("Message {} delivered!".format(msg.offset()))
+
+    def error(self, err):
+        """
+        Handler for global/generic and message delivery errors.
+
+        Raises ConfluentException for `KafkaError.__ALL_BROKERS_DOWN` if seen before a successful delivery.
+        Often times this is the result of a misconfiguration.
+        """
+        if not self.success and err.code() is KafkaError._ALL_BROKERS_DOWN:
+            raise ConfluentException(err)
+
+        print('Error: %s' % err)
+
+
+if __name__ == "__main__":
+    """
+    This example is designed to demonstrate how a user might report broker connection failures.
+
+    https://github.com/confluentinc/confluent-kafka-python/issues/705
+    """
+
+    dr = DeliveryReporter()
+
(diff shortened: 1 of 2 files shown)


examples/faq/are_you_there.py
```diff
@@ -0,0 +1,65 @@
+from confluent_kafka import Producer, KafkaError
```
edenhill: copyright header

examples/faq/are_you_there.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class ConfluentException(Exception):
```
edenhill: There is a KafkaException already.

examples/faq/are_you_there.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        if err:
+            self.error(err)
+
+        self.success = True
+        print("Message {} delivered!".format(msg.offset()))
+
+    def error(self, err):
+        """
+        Handler for global/generic and message delivery errors.
```
edenhill: This seems to be mixing per-message status propagation with global error notifications, I'm not sure the latter fits.

Also, since this call is not really used for any delivery reports in this example I'm not sure it is warranted.

Keep things minimal.

examples/faq/are_you_there.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Raises ConfluentException for `KafkaError.__ALL_BROKERS_DOWN` if seen before a successful delivery.
+        Often times this is the result of a misconfiguration.
+        """
+        if not self.success and err.code() is KafkaError._ALL_BROKERS_DOWN:
```
edenhill: This is problematic, all brokers could go down at any time for whatever reason and be back up the next second, librdkafk has been carefully crafted not to make this a fatal error, but instead abstract the required retries, restoration, etc. The approach to error handling layed out here voids all that, at no benefit.

This is indeed a big and complex subject, but I think we need to all agree on the how,when and what of error propagation before sending our users down any specific error handling path.

E.g., a common ask is to have a way for an application to quickly know if its broker list, security settings, are erroneously configured, ALL_BROKERS_DOWN is suitable for this, but only at startup.
IF an application has been running successfully for 3 weeks and suddenly gets an ALL_BROKERS_DOWN it is probably not due to configuration error.
How do we differentiate? That's the big question,

examples/faq/are_you_there.py
rnpridgeon: The delivery report callback is actually used to set the success or failure of a delivered message. That flag is then used to decide if we have ever connected before or not. In an attempt to keep things minimal I did not include the actual producing of a message. I can do so to more fully convey how you would use such a tactic.

examples/faq/are_you_there.py
rnpridgeon: Understood, that's why I'm offering an example of something one might do as opposed to trying to put it in the client. This example is not intending to solve distributed computing rather it  provides an example of how one might notify the application that a particular type of error occured. In this case that error is ALL_BROKERS_DOWN in the absence of any proof a connection (successful send). Raising exceptions under certain error conditions so the application can decide what to do doesn't seem terribly misguided to me but maybe I'm wrong. 

What if we include a healthy disclaimer in the example documentation and a fuller example where we wrap our main producer logic in a try.

examples/faq/are_you_there.py
rnpridgeon: I purposefully avoided using it since this is not raised by the underlying client. I wanted to avoid any confusion about this being something handled by confluent-kafka-python because this is an imperfect solution to a complicated problem.

rnpridgeon:
I understand and sympathise with your concerns. I'm not pretending this is a perfect solution for this as there are none. Rather I want to provide an example, perfect or not, of something a developer could do if they want to watch for this (they do, that's where the idea came from). 

It's not uncommon for systems to raise an exception when they are having trouble contacting a remote peer. This is usually controlled by some configurable connect timeout and possibly some number of retries. Rather than raising a FATAL exception you just raise a connection exception and then leave it up to the application developer to decide what they want to do.

rnpridgeon:
Closing until we come up with a proper answer for how we want to address this common issue.

PR #747, opened 2019-12-17
Schema Registry wire format uses int not uint for ids

### LINKED ISSUE #679: out of range schema id in avroProducer for IBM Event Streams
Description
===========

When working with IBM Event Streams and the `confluent_kafka.avro` library I get the following error:

```
  File "ContainerAvroProducer.py", line 75, in <module>
    kp.publishEvent(TOPIC_NAME,container_event,"containerID")
  File "/home/kafka/KcAvroProducer.py", line 43, in publishEvent
    self.producer.produce(topic=topicName,value=json.loads(value),key=json.loads(value)[key], callback=self.delivery_report)
  File "/root/.local/lib/python3.7/site-packages/confluent_kafka/avro/__init__.py", line 80, in produce
  File "/root/.local/lib/python3.7/site-packages/confluent_kafka/avro/serializer/message_serializer.py", line 113, in encode_record_with_schema
    return self.encode_record_with_schema_id(schema_id, record, is_key=is_key)
  File "/root/.local/lib/python3.7/site-packages/confluent_kafka/avro/serializer/message_serializer.py", line 144, in encode_record_with_schema_id
    outf.write(struct.pack('>bI', MAGIC_BYTE, schema_id))
struct.error: argument out of range
```

the schema id that IBM Event Streams produces is `avrotest-value-6cqbo6`

Apparently, the Java client does not have such problem.

How to reproduce
================

Send a message to a topic in IBM Event Streams whose name is of some length (not much as you can see above).

Checklist
=========

 - [ ] Operating system:
 - [ ] Critical issue

rnpridgeon:
Fixes #679

### DIFF
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -143,7 +143,7 @@ class MessageSerializer(object):
         writer = self.id_to_writers[schema_id]
         with ContextStringIO() as outf:
             # Write the magic byte and schema ID in network byte order (big endian)
-            outf.write(struct.pack('>bI', MAGIC_BYTE, schema_id))
+            outf.write(struct.pack('>bi', MAGIC_BYTE, schema_id))
 
             # write the record to the rest of the buffer
             writer(record, outf)
@@ -222,7 +222,7 @@ class MessageSerializer(object):
             raise SerializerError("message is too small to decode")
 
         with ContextStringIO(message) as payload:
-            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            magic, schema_id = struct.unpack('>bi', payload.read(5))
             if magic != MAGIC_BYTE:
                 raise SerializerError("message does not start with magic byte")
             decoder_func = self._get_decoder_func(schema_id, payload, is_key)


rnpridgeon:
Technically the schema id is an int not a uint even if the SR will never assign a negative value. Instead `-1` is used internally by the id generator to indicate no schemas have been registered.

rnpridgeon:
https://github.com/confluentinc/schema-registry/blob/0e55aaa0baf26fa26803fa473d16f9df383739a0/core/src/main/java/io/confluent/kafka/schemaregistry/id/ZookeeperIdGenerator.java#L55

confluent_kafka/avro/serializer/message_serializer.py
```diff
@@ -143,7 +143,7 @@ def encode_record_with_schema_id(self, schema_id, record, is_key=False):
         writer = self.id_to_writers[schema_id]
         with ContextStringIO() as outf:
             # Write the magic byte and schema ID in network byte order (big endian)
-            outf.write(struct.pack('>bI', MAGIC_BYTE, schema_id))
+            outf.write(struct.pack('>bi', MAGIC_BYTE, schema_id))
```
edenhill: Should we even attempt to make a request with a negative id? Sounds like some previous error handling was insufficient if we get here with an invalid schema.

rnpridgeon: No we should not the negative value is internal to SR and should never be used by clients. That said in order to properly conform with the spec we should be using a signed int.

rnpridgeon:
Closing as won't do for now

PR #749, opened 2019-12-18
fix error with trailing slash at the end

IvanProdaiko94:
To whom it may concern:

if you define the url for schema registry with `/` in the end, the lib will construct the wrong one.

Example:
`schema_registy_host=http://localhost:1234/`
will lead to `http://localhost:1234//schemas/whatever_next`

The lib does strip the `/` at the end of the string [here](https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/cached_schema_registry_client.py#L95)

but [several lines below](https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/cached_schema_registry_client.py#L111) get the deleted url from config with '/'.

So, with this fix there will be no error if the path of the schema registry will be defined like so:
`http://localhost:1234/`

### DIFF
--- a/confluent_kafka/avro/cached_schema_registry_client.py
+++ b/confluent_kafka/avro/cached_schema_registry_client.py
@@ -85,7 +85,7 @@ class CachedSchemaRegistryClient(object):
             """Construct a Schema Registry client"""
 
         # Ensure URL valid scheme is included; http[s]
-        url = conf.get('url', '')
+        url = conf.pop('url', '')
         if not isinstance(url, string_type):
             raise TypeError("URL must be of type str")
 
@@ -106,9 +106,8 @@ class CachedSchemaRegistryClient(object):
         if ca_path is not None:
             s.verify = ca_path
         s.cert = self._configure_client_tls(conf)
-        s.auth = self._configure_basic_auth(conf)
-
-        self.url = conf.pop('url')
+        s.auth = self._configure_basic_auth(self.url, conf)
+        self.url = utils.urldefragauth(self.url)
 
         self._session = s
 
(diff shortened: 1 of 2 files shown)


IvanProdaiko94:
[clabot:check]

rnpridgeon:
🤦‍♂ , why not pop the url from the dict up here though 

https://github.com/confluentinc/confluent-kafka-python/pull/749/files#diff-c96bedd6523a440fc267c9e78106496bL88

Also please add a unit test that validates the trailing "/" is stripped from the url to avoid future regressions.

IvanProdaiko94:
@rnpridgeon so, I did what you asked about)

tests/avro/test_cached_client.py
```diff
@@ -180,6 +180,12 @@ def test_invalid_url(self):
                 'url': 'example.com:65534'
             })
 
+    def test_trailing_slash_removal(self):
```
rnpridgeon: 👍

confluent_kafka/avro/cached_schema_registry_client.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, url, max_schemas_per_subject=1000, ca_location=None, cert_loc
             """Construct a Schema Registry client"""
 
         # Ensure URL valid scheme is included; http[s]
-        url = conf.get('url', '')
+        url = conf.pop('url')
```
rnpridgeon: Keep the default, '', to avoid throwing KeyError.  The check validation below will throw a more appropriate ValueError when it finds the url is not prepended with 'http'.

IvanProdaiko94: Fixed)

PR #766, opened 2020-01-17
Added the capability to retrieve a specific version for a subject #510

rooom13:
Reading version as string to avoid str.join failure


rooom13:
[clabot:check]

PR #768, opened 2020-01-21
Java compatible serdes

rnpridgeon:
- Adds trivup as a test fixutre to make integration testing possible from an IDE
- Adds a simple "JavaProducer" and "JavaConsumer" for interop testing (maven or make can be used to build the underlying java client)
- Add basic serializer classes which are compatible with Java's. String serialization isn't awesome and will continue to be not awesome as long as we support python 2 in addition to python 3


rnpridgeon:
duplicated #787

Issue #616, opened 2019-06-07
fastavro is no longer used by default


mparry:
Description
===========

There seems to have been a bug introduced in #470 that meant that, unless reader schemas are manually specified in the `AvroConsumer` constructor, attempts to use `fastavro` will always fail. (That is, the `reader_schema_obj` variable will be `None`, the attempt to call `to_json()` on it will throw, and the catch-all error handling will mean that it silently falls back to plain Python avro.)

As well as being rather a lot slower, this alters deserialisation behaviour in some cases.

This seems to have first been released in 1.0.0.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

PR #773, opened 2020-01-30
Add support for fastavro when reader schema isn't provided.

0x26res:
Related issue: https://github.com/confluentinc/confluent-kafka-python/issues/616

This is causing issues for people who want to use fastavro but don't want to provide a reader schema. 

It's not just a performance issue, as records are different when using fastavro (for example logical date time are represented as python `datetime` in fastavro, instead of `int`).

I've tested it on my local environment but it'd be better if we can write unit test for this. 

Also it'd be good to get rid of this try catch block that surrounds this statement and swallows important errors. When fastavro isn't installed but doens't work, it should be explicit.

### DIFF
--- a/confluent_kafka/avro/serializer/message_serializer.py
+++ b/confluent_kafka/avro/serializer/message_serializer.py
@@ -175,7 +175,8 @@ class MessageSerializer(object):
             # try to use fast avro
             try:
                 fast_avro_writer_schema = parse_schema(writer_schema_obj.to_json())
-                fast_avro_reader_schema = parse_schema(reader_schema_obj.to_json())
+                fast_avro_reader_schema = None if reader_schema_obj is None \
+                    else parse_schema(reader_schema_obj.to_json())
                 schemaless_reader(payload, fast_avro_writer_schema)
 
                 # If we reach this point, this means we have fastavro and it can


arthurandres:
[clabot:check]

arthurandres:
@edenhill, could you have a look at this PR?

I've tried to write tests but I'm unable to set up my working environment correctly. I find the `DEVELOPER.md` a bit lacking on the subject.

I've ran `python setup.py install` successfully but I still get `No module named 'confluent_kafka.cimpl` when running `./test/run.sh unit`

rnpridgeon:
@arthurandres we are currently in the process of wrapping up a refactor of the avro module which completely removes apache avro in favor of first class FastAvro support. This eliminates the clunky fallback logic in the serializer and cuts down on the amount of cache lookups which speeds things up considerably. Functionally this will be a backwards compatible meaning the reader schema functionality will be present. The pr isn't open yet because there are some generic serialization api discussions underway so we can add protobuf and json support in an upcoming release as well. I have a few things to update from our API discussion then I'll push the PR. n the meantime please do feel free to poke around what exists today.

https://github.com/confluentinc/confluent-kafka-python/tree/FirstClassFastAvro/confluent_kafka/avro

arthurandres:
@rnpridgeon , thanks for the update.

dylrich:
Is it possible to merge this in and make a release for now while you're kicking around the new API? We haven't been able to serialize Decimal types for ages and only today after many hours of debugging did we notice that this was because `fastavro` was never actually being used post `confluent_kafka` 1.0.0. We use the registry and don't want to pass in reader schemas, and the silent fallback to Apache `avro` (because of the bug this PR fixes) fooled us.

arthurandres:
> Is it possible to merge this in and make a release for now while you're kicking around the new API? We haven't been able to serialize Decimal types for ages and only today after many hours of debugging did we notice that this was because `fastavro` was never actually being used post `confluent_kafka` 1.0.0. We use the registry and don't want to pass in reader schemas, and the silent fallback to Apache `avro` (because of the bug this PR fixes) fooled us.

If it's really an issue you can monkey patch `MessageSerializer._get_decoder_func`. We've resolved to using this in the interim and it works well for us, though it's not pretty.

dylrich:
@rnpridgeon It looks like that branch was merged but it doesn't actually fix the bug this PR addresses. Any update on when we could expect that bugfix to land in a release?

arthurandres:
Is this line causing the issue: https://github.com/confluentinc/confluent-kafka-python/blob/c8a05dbd73e06ea310de7f28267fd4714f90ef8c/confluent_kafka/schema_registry/avro.py#L273 ?

@dylrich it shouldn't be too hard in the interim for you to create your own Deserializer and provide it when creating the consuer.

dylrich:
I've monkey patched all of our code as you suggested up above, but I'd really like to delete all of that nonsense given that this is a *super* simple bugfix.

andrewegel:
Causing CI to break due to the "unknown repository" - This needs to be closed to get around that. If this change is still required, please re-do it.

PR #776, opened 2020-02-10
Transactional Producer

rnpridgeon:


.appveyor.yml
```diff
@@ -1,6 +1,6 @@
 environment:
   global:
-    LIBRDKAFKA_NUGET_VERSION: 1.3.0
+    LIBRDKAFKA_NUGET_VERSION: 1.4.0-PRE6
```
edenhill: Let me create an RC off master, this one is quite dated now.

rnpridgeon: alternatively I can just point it at master until we are ready for a release

edenhill: There is no master nuget package though.

tools/windows-build.bat
```diff
@@ -20,7 +20,7 @@ dir wheelhouse
 rem cibuildwheel installs the generated packages, but they're not ready yet,
 rem so remove them.
 rem FIXME: this only covers python27 (default)
-pip uninstall -y confluent_kafka
+pip uninstall -y confluent_kafka[dev]
```
edenhill: what's dev?

confluent_kafka/src/Producer.c
```diff
@@ -398,6 +398,131 @@ static PyObject *Producer_flush (Handle *self, PyObject *args,
...
+static PyObject *Producer_send_offsets_to_transaction(Handle *self, PyObject *args) {
+        CallState cs;
+        char errstr[256];
+        rd_kafka_resp_err_t err;
+        PyObject *offsets = NULL;
+        rd_kafka_topic_partition_list_t *c_offsets;
+        char *group_id;
+        double tmout = 1;
+
+        if(!PyArg_ParseTuple(args, "sOd", &group_id, &offsets, &tmout))
```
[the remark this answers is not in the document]
rnpridgeon: all timeouts have been made optional and default to -1.0f

confluent_kafka/src/Producer.c
```diff
@@ -466,8 +591,73 @@ static PyMethodDef Producer_methods[] = {
         { "list_topics", (PyCFunction)list_topics, METH_VARARGS|METH_KEYWORDS,
           list_topics_doc
         },
-
-	{ NULL }
+        { "init_transactions", (PyCFunction)Producer_init_transactions, METH_VARARGS|METH_KEYWORDS,
+          ".. py:function:: init_transactions([timeout])\n"
+          "\n"
+          "  :param float timeout: Maximum time to block in seconds; Must be >= 1.0.\n"
```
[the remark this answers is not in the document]
rnpridgeon: I believe this was a constraint on the original librdkafka implementation.

examples/eos-transactions.py
```diff
@@ -0,0 +1,131 @@
...
+#
+# See the following blog for additional information.
+# https://www.confluent.io/blog/transactions-apache-kafka/
+#
+
+
+"""
+The following example demonstrates how to perform a consume-transform-produce loop with exactly-once semantics.
+
+In order to achieve exactly-once semantics we use the idempotent producer with transactions enabled and a single
```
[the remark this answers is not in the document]
rnpridgeon: flake8 is 120, do we want to conform to c stlying in python files?

examples/eos-transactions.py
```diff
@@ -0,0 +1,131 @@
...
+
+    consumer.subscribe(['input_topic'])
+
+    producer = Producer({
+        'bootstrap.servers': brokers,
+        'transactional.id': 'example_transactional_id',
+    })
+
+    # Initialize producer transaction.
+    producer.init_transactions(2.0)
```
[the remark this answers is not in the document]
rnpridgeon: all timeouts have been made optional and default to -1

tools/windows-build.bat
rnpridgeon: everything, so pytest, avro, requests etc

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def drain_queue(consumer):
+    msg_cnt = 0
+    eof = {}
+    print("=== Draining {} ===".format(consumer.assignment()))
+    while (True):
+        msg = consumer.poll(timeout=1.0)
+
+        if len(consumer.assignment()) == 0:
```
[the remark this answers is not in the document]
rnpridgeon: good point,  guess its not really needed at all since the msg will always be none as long as  don't have an assignment.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+        if self._producer is None:
+            self._producer = Producer(self.client_conf({'linger.ms': 500}))
+
+        if value_source is None:
+            value_source = ['test-data{}'.format(i) for i in range(0, 100)]
+
+        if key_source is None:
+            key_source = [None]
+
+        produce_thread = threading.Thread(target=_produce,
```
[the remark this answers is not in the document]
rnpridgeon: the risk seemed low, in the event your consumer sends a fetch request before the producer sends a produce request the worst thing that happens is we need to call poll again on the consumer

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+from uuid import uuid4, uuid1
+
+from trivup.clusters.KafkaCluster import KafkaCluster
+
+from confluent_kafka import Producer, Consumer
+
+from confluent_kafka.admin import AdminClient, NewTopic
+
+
+class ClusterFixture(object):
```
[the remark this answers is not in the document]
rnpridgeon: The idea was to allow for embedded clusters or using testconf.json for external clusters.

examples/eos-transactions.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+"""
+The following example demonstrates how to perform a consume-transform-produce loop with exactly-once semantics.
+
+In order to achieve exactly-once semantics we use the idempotent producer with transactions enabled and a single
+transaction aware consumer.
+
+The following assumptions apply to the source data(input_topic below):
+    1. The input source, input_topic below, was populated using a transactional producer
```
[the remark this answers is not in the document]
rnpridgeon: I was thinking this was a needed in order to satisfy `read_committed` requirement. In retrospect that's wrong, removing.

examples/eos-transactions.py
```diff
@@ -0,0 +1,131 @@
...
+
+        processed_value, processed_key = process_input(msg)
+
+        producer.produce("output_topic", processed_value, processed_key, on_delivery=delivery_report)
+        producer.poll(1.0)
+
+    # commit processed messages to the transaction
+    producer.send_offsets_to_transaction(group_id, consumer.position(consumer.assignment()), 2.0)
+    # commit transaction
+    producer.commit_transaction(2.0)
```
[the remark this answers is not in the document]
rnpridgeon: I think we want this one since you can't ^C out if you put a bad broker address in.

examples/eos-transactions.py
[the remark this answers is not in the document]
edenhill: I thought it was 80.
```
  --max-line-length=n   Maximum allowed line length for the entirety of this
                        run. (Default: 79)
```

Anyway, it is good practice to keep things below 80, it makes it readable everywhere (terminals, github code reviews, etc), and this is very easy to do with comments that can be formatted in any way we like.
For code it is sometimes not practical to keep below 80, so that's another story.

tools/windows-build.bat
edenhill: Is that a standard package thing, or a sub-package that we specifically make?

rnpridgeon: its an "extra", you can see it in setup.py

examples/eos-transactions.py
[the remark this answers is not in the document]
rnpridgeon: https://github.com/confluentinc/confluent-kafka-python/blob/master/tox.ini#L58

Looks like 3 years ago we configured it to be 119. I'll update the tox.ini as well.

tests/integration/cluster_fixture.py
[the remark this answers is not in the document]
edenhill: I mean the `object` inheritance, I think it is implicit for new style classes so you dont need it.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+    _slots_ = ()
```
edenhill: single underscore here, double underscore in the next class.
set here, list in the next class.

Why the difference?

Also: I'm still skeptical of specifying slots, from what I understand it is mainly useful for memory allocation reasons where you have a large number of small objects. Apart from that it just limits use/extensibility.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    ClusterFixture factory.
+    Provides client configs enabling integration tests to execute against a live cluster.
+    """
+
+    def __new__(cls, options):
+        conf = {k[6:]: v for (k, v) in os.environ.items() if k.startswith('KAFKA')}
+
+        if bool(conf):
+            return ExternalClusterFixture(conf, options)
```
edenhill: Shouldn't External and Trivup ClusterFixture classes derive from the same base class, making it more obvious what methods they supply.
As it currently stands TrivupClusterFixture provides more methods than the ExternalClusterFixture (e.g., create_topic), which makes it unclear if that is a public method or not.

Not sure having a factory return incompatible types is practical.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+
+        return TrivupFixture(options)
+
+
+class ExternalClusterFixture(object):
+    __slots__ = ['_conf', '_topic']
+
+    """
+        Configures integration tests clients to communicate with external(developer provided) cluster.
+        All client configs are valid and follow the following naming convention.
```
edenhill: following?

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+            conf.update({'group.id': uuid1()})
+
+        return Consumer(conf)
+
+    @property
+    def schema_registry(self):
+        return self._conf.get('schema.registry.url')
+
+
+class TrivupFixture(object):
```
edenhill: no doc string

tests/integration/cluster_fixture.py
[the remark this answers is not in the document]
edenhill: There is either a risk of race or there is not, no such thing as a low risk when it comes to race conditions.

It is possible for the consumer to reach EOF prior to all (or any) messages being produced.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+                    producer.produce(topic,
+                                     value=value_source[i],
+                                     key=key_source[i % num_keys])
+                break
+            except BufferError:
+                producer.poll(0.1)
+            continue
+        producer.poll(0.0)
+
+    producer.flush(10.0)
```
edenhill: This timeout is too aggressive on CI.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+
+        produce_thread.daemon = True
+        produce_thread.start()
+
+    @property
+    def schema_registry(self):
+        return self._cluster.sr.get('url')
+
+
+def _produce(producer, topic, value_source, key_source, header_source=None):
```
edenhill: nit: Why the _source suffix? It is ambigious: what is a source? is it a generator?
`values, keys, headers` is clearer, simply indicating plural.

edenhill: also: doc strings

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+        :returns: a new producer instance.
+        :rtype: confluent_kafka.Producer
+        """
+        return Producer(self.client_conf(options))
+
+    def consumer(self, options=None):
+        """
+        Returns a consumer bound to this cluster.
+
+        :param dict options: additional consumer values to be used.
```
edenhill: what is a consumer value?
If this is configuration we should call it conf, 'values' and 'options' introduce new names for something that already exists and has a common name, thus bringing confusion.

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+        produce_thread = threading.Thread(target=_produce,
+                                          args=(self._producer, topic, value_source,
+                                                key_source, header_source))
+
+        produce_thread.daemon = True
+        produce_thread.start()
+
+    @property
+    def schema_registry(self):
+        return self._cluster.sr.get('url')
```
edenhill: There is code above that does the same thing but that also checks that cluster.sr exists, maybe move that here and make the other code use this method

tests/integration/cluster_fixture.py
```diff
@@ -0,0 +1,226 @@
...
+                                     key=key_source[i % num_keys])
+                break
+            except BufferError:
+                producer.poll(0.1)
+            continue
+        producer.poll(0.0)
+
+    producer.flush(10.0)
+
+    print('# finished producing {} messages to topic {}'.format(num_messages, topic))
```
edenhill: Since you are not checking the return value from flush(with a timeout) you dont know if all messages were sent.
Suggest you fail if flush() returns non-zero.

tests/integration/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pytest
+
+from cluster_fixture import ClusterFixture
+
+
```
edenhill: Is this a test for ClusterFixture (if so, what is the test pass criteria?), or a helper method for the tests (if so, the name conftest seems wrong)?

Docstrings.

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+# limit
+#
+from uuid import uuid1
+
+from confluent_kafka import Consumer, KafkaError
+
+
+def my_error_cb(err):
+    """ Reports global/generic errors to aid in troubleshooting test failures. """
+    print(err)
```
edenhill: Prefix the error message with something so we know where it comes from.

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+        return
+
+
+def test_commit_transaction(kafka_cluster):
+    output_topic = kafka_cluster.create_topic("output_topic" + str(uuid1()))
+
+    producer = kafka_cluster.producer({
+        'transactional.id': 'example_transactional_id',
+        'error_cb': my_error_cb,
+
```
edenhill: remove empty line

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+    producer = kafka_cluster.producer({
+        'client.id': 'producer1',
+        'transactional.id': 'example_transactional_id',
+        'error_cb': my_error_cb,
+    })
+
+    consumer_conf = {
+        'group.id': str(uuid1()),
+        'auto.offset.reset': 'earliest',
+        'isolation.level': 'read_committed',
```
edenhill: that's the default, isnt it?

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+
+    kafka_cluster.seed_topic(input_topic)
+    consumer.subscribe([input_topic])
+
+    read_all_msgs(consumer)
+
+    producer.init_transactions()
+    transactional_produce(producer, output_topic, 100)
+
+    print("=== Sending offsets {} to transaction ===".format(consumer.position(consumer.assignment())))
```
edenhill: nit: only call position(assignment()) once.

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+    producer.send_offsets_to_transaction(consumer_conf.get('group.id'),
+                                         consumer.position(consumer.assignment()))
+    producer.commit_transaction()
+
+    producer = kafka_cluster.producer({
```
edenhill: suggest calling this variable producer2, eases readability.

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+        'transactional.id': 'example_transactional_id',
+        'error_cb': my_error_cb
+    })
+
+    # ensure offset commits are visible prior to sending FetchOffsets request
+    producer.init_transactions()
+
+    committed = consumer.committed(consumer.assignment())
+    print("=== Committed offsets for {} ===".format(committed))
+    committed = [tp.offset for tp in committed]
```
edenhill: changing types of an existing variable is confusing to the reader, perhaps committed_offsets ?

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+
+    consumer.close()
+
+    assert committed == [100]
```
edenhill: For ease of debugging, assert as soon as possible (prior to close())

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+
+def transactional_produce(producer, topic, num_messages):
+    print("=== Producing {} transactional messages to topic {}. ===".format(
+        num_messages, topic))
+
+    producer.begin_transaction()
+
+    for value in ['test-data{}'.format(i) for i in range(0, num_messages)]:
+        producer.produce(topic, value, on_delivery=delivery)
+        producer.poll()
```
edenhill: Why are you doing synchronous producing?

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+
+    producer.flush()
+
+
+def read_all_msgs(consumer):
```
edenhill: This assumes the consumer has not already consumed the messages in the partitions, since it relies on EOFs.
May want to say that in a docstring.

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+    print("=== Draining {} ===".format(consumer.assignment()))
+    while (True):
+        msg = consumer.poll(timeout=1.0)
+
+        if msg is None:
+            continue
+
+        if msg.error():
+            if msg.error().code() == KafkaError._PARTITION_EOF:
+                eof[(msg.topic(), msg.partition())] = True
```
edenhill: Technically the partition should be deleted from the dict when a message is seen, i.e., resetting the EOF.
If this method is only called for topics that are not seeing new messages then say so in the docstring.

tests/integration/producer/test_transactions.py
```diff
@@ -0,0 +1,189 @@
...
+
+    return msg_cnt
+
+
+def consume_committed(conf, topic):
+    print("=== Consuming transactional messages from topic {}. ===".format(topic))
+
+    consumer_conf = {'group.id': str(uuid1()),
+                     'auto.offset.reset': 'earliest',
+                     'isolation.level': 'read_committed',
```
edenhill: is default

confluent_kafka/src/Producer.c
```diff
@@ -402,20 +402,20 @@ static PyObject *Producer_init_transactions (Handle *self, PyObject *args) {
...
 
-        err = rd_kafka_init_transactions(self->rk, (int)(tmout * 1000), errstr, sizeof(errstr));
+        err = rd_kafka_init_transactions(self->rk, cfl_toMS(tmout), errstr, sizeof(errstr));
 
         if (!CallState_end(self, &cs))
                 return NULL;
 
         if (err != RD_KAFKA_RESP_ERR_NO_ERROR) {
-                cfl_PyErr_Format(err, "Failed to initialize transactions within %.2f seconds", tmout);
+                cfl_PyErr_Format(err, "Failed to initialize transactions within %d seconds: %s", tmout, errstr);
```
edenhill: I dont think you need to prefix the error strings from the transacitonal API, they are rich as-is.

confluent_kafka/src/Producer.c
```diff
@@ -402,7 +402,7 @@ static PyObject *Producer_init_transactions (Handle *self, PyObject *args) {
         CallState cs;
         char errstr[256];
         rd_kafka_resp_err_t err;
-        double tmout = -1;
+        double tmout = -1.0f;
```
edenhill: remove the `f` suffix, this will generate a float that is cast to a double.

confluent_kafka/src/Producer.c
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2016 Confluent Inc.
+ * Copyright 2020 Confluent Inc.
```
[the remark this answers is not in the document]
rnpridgeon: > but don't 

Is this to say it should be range and I didn't change it to one or don't or don't do that revert to the previous copyright date.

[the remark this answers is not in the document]
edenhill: This is to say:
 - never change copyright start year.
 - copyrights should preferably be start_year-current_year
 - but until we have automatic tooling in place to verify and possibly update the copyright year, we shouldn't touch them.

tests/integration/cluster_fixture.py
rnpridgeon: It's primary purpose is actually to speed up attribute lookups. It's secondary use is memory savings. I'm not sure I understand how it limits extensibility if you want dynamic attributes, you can subclass this and not use slots.

confluent_kafka/src/Producer.c
rnpridgeon: okay, should I do the same for consumer.c or save that for another PR since its unrelated.

tests/integration/cluster_fixture.py
edenhill: Neither seems relevant here though, I don't see how it adds any value to a class like this.
But it is up to you to keep it.

confluent_kafka/src/Producer.c
edenhill: That's up to you.

tests/integration/producer/test_transactions.py
rnpridgeon: I wrongfully thought marking the parameter optional would populate the timeout with the 0 value for its type. I know now this is incorrect per your earlier comment and revisiting the cpython api docs.

tests/integration/producer/test_transactions.py
rnpridgeon: yes, although this isn't differs from java and makes following the blog on EOS a bit odd. Cleaning up to highlight the better default in librdkafka.

tests/integration/conftest.py
rnpridgeon: not sure I follow it brings ClusterFixture into scope. It looks odd because we don't have an __init__.py but that is per the pytest documentation and necessary to avoid issues importing the c extension.

edenhill: I don't understand what the purpose of the method is: is it a utility method used by the tests, or is it a test itself?
The filename suggests that it is a config test, but I don't see any verification criteria, which leads me to believe it is a utility function used by the tests, but the filename confuses me.

tests/integration/cluster_fixture.py
rnpridgeon: how about we just block for -1

edenhill: Yes, that's the recommended way to call flush(), to rely on the message.timeout.ms

tests/integration/cluster_fixture.py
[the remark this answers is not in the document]
rnpridgeon: I'm not sure that's true for python 2.7 which supports classic and new style classes: 

https://wiki.python.org/moin/NewClassVsClassicClass

tests/integration/cluster_fixture.py
rnpridgeon: Mostly I was experimenting with ways to maintain backward compatibility with different forms of encapsulation. I'll remove this entirely and use kafka_cluster within conftest.py as a factory method.

examples/eos-transactions.py
```diff
@@ -0,0 +1,163 @@
...
+
+    consumer.close()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Exactly Once Semantics(EOS) example")
+    parser.add_argument('-b', dest="brokers", required=True,
+                        help="Bootstrap broker(s) (host[:port])")
+    parser.add_argument('-g', dest="group_id", default=str(uuid4()),
```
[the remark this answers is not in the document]
rnpridgeon: the group id?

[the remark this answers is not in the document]
edenhill: Yes, the default group.id should have a prefix that gives the operator a hint where it came from/what it is used for. a uuid is untrackable.

confluent_kafka/src/Producer.c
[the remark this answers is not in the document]
rnpridgeon: ack

confluent_kafka/src/Producer.c
```diff
@@ -360,7 +360,7 @@ static PyObject *Producer_poll (Handle *self, PyObject *args,
 	int r;
 	static char *kws[] = { "timeout", NULL };
 
-	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "d", kws, &tmout))
+	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|d", kws, &tmout))
```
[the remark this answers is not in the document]
rnpridgeon: made optional, default to -1 like the others

confluent_kafka/src/Producer.c
rnpridgeon: I'll leave as is for now. I think I want to follow up with a style PR to fix all the inconsistencies

PR #784, opened 2020-02-21
Update README.md for latest macos

chicagobuss:
updating readme with ssl location that works for mojave and catalina


rnpridgeon:
We should refrain from explicitly mentioning locations for CA Bundles.

PR #786, opened 2020-02-25
Raise import error on module import failure to avoid segfault

overstre:
A dependency of confluent-kafka, `enum34`, published a broken version 1.1.8 that caused a segfault in `confluent-kafka` python library when accessing the module `confluent_kafka.admin`.  For example, the list_topics call will try to import this module in function `cfl_PyObject_lookup`.

This results in `module` pointer being nil in the return of this call:
```        PyObject *module = PyImport_ImportModule(modulename);```

and then it gets accessed here
```        obj = PyObject_GetAttrString(module, typename);```
which results in a segfault:
```

Thread 28 "python" received signal SIGSEGV, Segmentation fault.
--
[Switching to Thread 0x7fff96ffd700 (LWP 8778)]
0x00000000004efa6b in PyObject_GetAttrString ()
(gdb) backtrace
#0  0x00000000004efa6b in PyObject_GetAttrString ()
#1  0x00007ffff433dae2 in cfl_PyObject_lookup (modulename=modulename@entry=0x7ffff434a067 "confluent_kafka.admin",
typename=typename@entry=0x7ffff434a057 "ClusterMetadata") at confluent_kafka/src/confluent_kafka.c:1962
#2  0x00007ffff43418a0 in c_metadata_to_py (self=0x7ffff0127440, metadata=0x7fff84002570) at confluent_kafka/src/Metadata.c:267
#3  list_topics (self=0x7ffff0127440, args=<optimized out>, kwargs=<optimized out>) at confluent_kafka/src/Metadata.c:382

```

This pull request is to check the value of module to ensure that it's not null otherwise return a runtime exception.  This will make debugging this issue in the future.

Without patch
```
$ python ~/list.py
Segmentation fault
```

with patch:
```
$ python ~/list.py
  File "/home/brian/list.py", line 5, in <module>
    print producer.list_topics(topic=topic).topics.values()
RuntimeError: Module confluent_kafka.admin import failed.
```

### DIFF
--- a/confluent_kafka/src/confluent_kafka.c
+++ b/confluent_kafka/src/confluent_kafka.c
@@ -1952,10 +1952,10 @@ PyObject *cfl_PyObject_lookup (const char *modulename, const char *typename) {
         PyObject *module = PyImport_ImportModule(modulename);
         PyObject *obj;
 
-        if (!modulename) {
-                PyErr_Format(PyExc_TypeError,
-                             "Module %s not found when looking up %s.%s",
-                             modulename, modulename, typename);
+        if (!module) {
+                PyErr_Format(PyExc_ImportError,
+                             "Module not found when looking up %s.%s",
+                             modulename, typename);
                 return NULL;
         }
 

### CODE REVIEW

**Comment by overstre** [comment]
[clabot:check]

**Comment by rnpridgeon** [comment]
What are your thoughts on removing the enum34 dependency all together. I'm actually not a huge fan of this dependency and we can easily achieve the desired functionality with constants as we do with logical offsets.

**Comment by overstre** [comment]
I'm in favor of reducing dependencies and keeping the style consistent.

File: confluent_kafka/src/confluent_kafka.c [hunk 1]
```diff
@@ -1959,6 +1959,13 @@ PyObject *cfl_PyObject_lookup (const char *modulename, const char *typename) {
                 return NULL;
         }
 
+        if (!module) {
```
edenhill: Thanks for finding this, but I think the initial intent was for the if-statement above to handle this, could you change that one instead to check for module rather than modulename, and remove the first %s and modulename from the argument list.

Also, change it to PyExc_ImportError.

overstre: Thanks for the review.  I've updated the code.
```
brian@dev-brian-1:~/code/confluent-kafka-python$ python ~/list.py
  File "/home/brian/list.py", line 5, in <module>
    print producer.list_topics(topic=topic).topics.values()
ImportError: Module not found when looking up confluent_kafka.admin.ClusterMetadata
```

PR #791, opened 2020-02-27
Update README.md

chicagobuss:
try #2 to strike a balance between informative/useful/concise

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,16 +270,22 @@ authentication is used).
 
 The client will use CA certificates to verify the broker's certificate.
 The embedded OpenSSL library will look for CA certificates in `/usr/lib/ssl/certs/`
-or `/usr/lib/ssl/cacert.pem`. CA certificates are typically provided by the
-Linux distribution's `ca-certificates` package which needs to be installed
-through `apt`, `yum`, et.al.
+or `/usr/lib/ssl/cacert.pem`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,16 +270,22 @@ authentication is used).
...
-or `/usr/lib/ssl/cacert.pem`. CA certificates are typically provided by the
-Linux distribution's `ca-certificates` package which needs to be installed
-through `apt`, `yum`, et.al.
+or `/usr/lib/ssl/cacert.pem`.
+
+On Linux, CA certificates are typically provided by the distribution's `ca-certificates`
+package which needs to be installed through `apt`, `yum`, et.al.
+
+On MacOS, different versions can store CA certificates in different locations.
+On MacOS Mojave and later, for instance, this is usually ` '/private/etc/ssl/cert.pem'`.
```
rnpridgeon: I don't think this is entirely accurate. I think the bundled openssl version may look there but MacOS itself stores them in the keychain database. I believe this to be the reason why `openssl s_client -connect` works but the Python client complains. It'd be good to verify this statement prior to proceeding as this may be one of those "works on my machine" type situations rather than works on MacOS Mojave+.

edenhill: Inconsistent formatting, drop the `'`.

PR #811, opened 2020-03-31
JsonSerializer/JsonDeserializer for GenericSerdeApi

rnpridgeon:


tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,101 @@
+import pytest
+from confluent_kafka.cimpl import TopicPartition
+
+from confluent_kafka.schema_registry.json_schema import JsonSerializer, JsonDeserializer
```
mhowlett: If someone is using JSON and not Avro, will they be required to install any avro dependencies? 

(i think no, but just checking)

rnpridgeon: Currently yes, that said the Avro package "extra" has caught a few folks off guard though. If we think its a huge deal we can make it so you have to specify each dependency in the extras. i.e. pip install confluent-kafka[avro, protobuf, jsonschema]. I think it gets a bit weird when you have dependencies for your dependencies though. I.e. I'm not sure how you would handle pip install confluent-kafka[schema_registry, avro, protobuf, jsonschema]. It probably works but it starts to make for a sort of annoying install. Anyway open to  the idea if we think its important.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+
+    JsonSerializer configuration properties:
+    +-----------------------+----------+--------------------------------------------------+
+    | Property Name         | type     | Description                                      |
+    +=======================+==========+==================================================+
+    |                       |          | Registers schemas automatically if not           |
+    | auto.register.schemas | bool     | previously associated with a particular subject. |
+    |                       |          | Defaults to True.                                |
+    +-----------------------|----------+--------------------------------------------------+
+    |                       |          | Callable(str, SerialalizationContext) -> str     |
```
mhowlett: the arg order in this comment doesn't match reality anymore. you need to update the avro doc too, and i'm guessing protobuf.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+    Schemas are registered to namespaces known as Subjects which define how a
+    schema may evolve over time. By default the subject name is formed by
+    concatenating the topic name with the message field separated by a hyphen.
+
+    i.e. {topic name}-{message field}
+
+    Alternative naming strategies may be configured with the property
+    `subject.name.strategy`.
+
+    Supported subject name strategies:
```
mhowlett: maybe "Subject name strategies supported out of the box:" (you support more than just these three!)

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+        self._registry = schema_registry_client
+        self._schema_id = None
+        # Avoid calling registry if schema is known to be registered
+        self._known_subjects = set()
+
+        if to_dict is not None and not callable(to_dict):
+            raise ValueError("to_dict must be callable with the signature"
+                             " to_dict(Serialization Context, object)->dict")
+
+        self.to_dict = to_dict
```
mhowlett: make private

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+            bytes: Confluent Schema Registry formatted JSON bytes
+
+        """
+        if obj is None:
+            return None
+
+        subject = self._subject_name_func(ctx, self._schema_name)
+
+        # Check to ensure this schema has been registered under subject_name.
+        if self._auto_register and subject not in self._known_subjects:
```
mhowlett: i think locking of _known_subjects is a good idea because if one call to SR is underway you probably don't want another one as well (since it will almost certainly be for the same schema).

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+        from_dict (callable, optional): Callable(SerializationContext, dict) -> object.
+            Converts dict to an instance of some object.
+
+    .. _Schema definition:
+        https://json-schema.org/understanding-json-schema/reference/generic.html
+
+    """
+    __slots__ = ['_schema', '_registry', 'from_dict']
+
+    def __init__(self, schema_registry_client, schema_str, from_dict=None):
```
mhowlett: fwiw in .net i didn't even bother integrating with SR on the consume side. but good job..

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+        self._registry = schema_registry_client
+        self._schema = json.loads(schema_str)
+
+        if from_dict is not None and not callable(from_dict):
+            raise ValueError("from_dict must be callable with the signature"
+                             " from_dict(SerializationContext, dict) -> object")
+
+        self.from_dict = from_dict
```
mhowlett: make private.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,261 @@
...
+            raise SerializationError("message is too small to decode {}".format(value))
+
+        with _ContextStringIO(value) as payload:
+            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            if magic != _MAGIC_BYTE:
+                raise SerializationError(
+                    "message does not start with magic byte")
+
+            # JSON documents are self-describing; no need to query schema
+            return json.loads(payload.read(), encoding="utf8")
```
mhowlett: you're not using `from_dict`, `_schema`, or `_registry`.

confluent_kafka/schema_registry/json_schema.py
mhowlett: ahh, you don't actually use any of these.

in python you can't have more than on __init__, so i don't think you can nicely extend to add this capability later. not exactly sure what to do. i think probably document that schema_registry_client & schema_str are currently ignored and may be used in the future.

confluent_kafka/schema_registry/json_schema.py
mhowlett: i think you should apply `from_dict`, but no need to worry about `_schema` or `_registry`. open to someone else having a better idea, not sure this is the best idea here.

tests/integration/schema_registry/test_json_serializers.py
mhowlett: in .net i have 5 separate packages: https://github.com/confluentinc/confluent-kafka-dotnet/tree/1.4.x#referencing

i think it's important - dependencies can often cause headaches, and it often pays to have a strong preference to avoiding additional ones unless absolutely necessary.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: I thought about preventing multiple inflight requests but I decided to forego locking as it makes things easier to reason about. This does mean you will initially have a few more requests than intended but I'd argue its hardly a problem. If you feel strong about it we could add locking but since we will always get the same schema id (the schema is fixed) I didn't feel it was necessary.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: yeah I felt a bit weird about this one. I originally validated that the schema fetched matched the provided schema then I saw we don't do that in .NET so I decided to do away with it. Perhaps we should bring it back though.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: that was actually an accidental removal when fixing it up to match the latest version of the generic serdes/avro api. I'll add a test since that would have caught this. Thanks for pointing it out.

tests/integration/schema_registry/test_json_serializers.py
rnpridgeon: That's fair it just feels a bit clumsy to me for a scripting language because the source is there and in executable form but the dependencies are not. That said it's easy enough to make extras and have them all require the schema registry dependencies. I'll add it tomorrow.

confluent_kafka/schema_registry/json_schema.py
edenhill: I'm with Ryan, we discussed this for Avro and decided that the extra calls to  SR are okay at this time.

tests/integration/schema_registry/data/invalid.json
```diff
@@ -0,0 +1,31 @@
+{
```
edenhill: Can we indicate somehow where the invalid part is, because it is not appearant when reading this?

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
+import pytest
```
edenhill: LICENSE header

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+            "productName": product_obj.name,
+            "price": product_obj.price,
+            "tags": product_obj.tags,
+            "dimensions": product_obj.dimensions,
+            "warehouseLocation": product_obj.location}
+
+
+def _testProduct_from_dict(ctx, product_dict):
+    """
+    Returns testProduct instance from it's dict format.
```
edenhill: its

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+                        product_dict['productName'],
+                        product_dict['price'],
+                        product_dict['tags'],
+                        product_dict['dimensions'],
+                        product_dict['warehouseLocation'])
+
+
+def test_json_record_serialization(kafka_cluster, load_file):
+    """
+    Tests basic JsonSerializer and JsonDeserializer basic functionality
```
edenhill: .

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+        https://json-schema.org/learn/getting-started-step-by-step.html
+
+    Args:
+        kafka_cluster (KafkaClusterFixture): cluster fixture
+
+        load_file (callable(str)): JSON Schema file reader
+
+    """
+    topic = kafka_cluster.create_topic("serialization-json")
+    sr = kafka_cluster.schema_registry({'url': 'http://localhost:8081'})
```
edenhill: Why is the URL specified here? I thought trivup's SR allocates a port?
Having fixed port numbers should be avoided since it will collide with other instances, programs, etc.

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    schema_str = load_file("product.json")
+    value_serializer = JsonSerializer(sr, schema_str)
+    producer = kafka_cluster.producer(value_serializer=value_serializer)
+
+    record = {"contractorId": 1,
+              "contractorName": "David Davidson",
+              "contractRate": 1250,
```
edenhill: hot daymn! I need to learn masonry

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+              "trades": ["mason"]}
+
+    with pytest.raises(SerializationError,
+                       match=r"(.*) is a required property"):
+        producer.produce(topic, value=record, partition=0)
+
+
+def test_json_record_serialization_invalid(kafka_cluster, load_file):
+    """
+    Ensures ValueError raise if  JSON Schema definition lacks Title annotation.
```
edenhill: double spaces

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Args:
+        kafka_cluster (KafkaClusterFixture): cluster fixture
+
+        load_file (callable(str)): JSON Schema file reader
+
+    """
+    sr = kafka_cluster.schema_registry({'url': 'http://localhost:8081'})
```
edenhill: dito

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,266 @@
...
+class JsonDeserializer(Deserializer):
+    """
+    JsonDeserializer decodes bytes written in the Schema Registry
+    JSON format to an object.
+
+    Args:
+        schema_registry_client (SchemaRegistryClient): Confluent Schema Registry
+            client instance.
+
+        schema_str (str): JSON schema declaration.
```
edenhill: definition, they got it right.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,266 @@
...
+
+        if from_dict is not None and not callable(from_dict):
+            raise ValueError("from_dict must be callable with the signature"
+                             " from_dict(SerializationContext, dict) -> object")
+
+        self._from_dict = from_dict
+
+    def __call__(self, ctx, value):
+        """
+        Deserializes Schema Registry response to JSON object literal(dict)
```
edenhill: .

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,266 @@
...
+
+        Args:
+            ctx (SerializationContext): Metadata pertaining to the serialization
+                operation.
+
+            value (bytes): bytes
```
edenhill: Confl SR JSON formatted bytes

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,266 @@
...
+
+        Returns:
+            dict: Deserialized JSON
+
+        Raises:
+            SerializerError: If response payload and expected Message type
```
edenhill: Huhm, Not sure what "response payload" nor "Message type" is.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,266 @@
...
+            differ.
+
+        """
+        if len(value) <= 5:
+            raise SerializationError("message is too small to decode {}".format(value))
```
edenhill: Grab the error strings from the Avro counterparts, they are better.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,266 @@
...
+
+        with _ContextStringIO(value) as payload:
+            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            if magic != _MAGIC_BYTE:
+                raise SerializationError(
+                    "message does not start with magic byte")
+
+            # JSON documents are self-describing; no need to query schema
```
edenhill: But the schema dictates what fields are required, and which fields should be ignored (not included in schema).
If the Deserializer doesn't check it then the application needs to, which sort of defeats the purpose of a schema aware deserializer.

tests/integration/schema_registry/test_json_serializers.py
rnpridgeon: Accidentally left it. It's there as an override for an external SR since CP 5.5 isn't out yet.

rnpridgeon: on the bright side look how easy it is to use an external SR instance with trivup

tests/integration/schema_registry/test_json_serializers.py
rnpridgeon: David Davidson nis really good but sometimes shows up when all you wanted was an ice swan.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: you are right added with a test

edenhill: @mhowlett Just to verify; is the .NET client validating received JSON with the schema?

rnpridgeon: No, but this is consistent with the AvroDeserializer and ProtobufDeserializer which is a 1:1 relationship as was requested when the generic serdes API was proposed.

tests/integration/schema_registry/test_json_serializers.py
rnpridgeon: I think this is a bit bigger than this PR; let's tackle afterwards. The problem I have with the current setup is that all of this code is available to your application when installing confluent-kafka even without the dependencies. This means it won't  be caught until runtime which is sloppy. Instead we should move these to their own top-level packages which can be referenced as extras when installing confluent-kafka. 

So confluent-kafka-avro, confluent-kafka-json, confluent-kafka-schema-registry would all be their own projects which could be optionally installed with extras from the main confluent-kafka package. The serializer packages would list confluent-kafka-schema-registry as a dependency insuring its installed along with the serializer. 

With this setup installation would still look like this `pip install confluent-kafka[avro, json]`. Except none of these packages would be available for import until explicitly installed.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: In addition to these three we support a reference name strategy which is technically supported but only has one implementation. For now I think this is sufficient. Let's revisit this in a follow-up docs PR I'm planning to verify all the rendering.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: Resolving for now, we can tweak it later if we feel this is too burdensome.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,300 @@
...
+            # Write the magic byte and schema ID in network byte order (big endian)
+            fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))
+            # JSON dump always writes a str never bytes
+            # https://docs.python.org/3/library/json.html
+            fo.write(json.dumps(value).encode('utf8'))
+
+            return fo.getvalue()
+
+
+class JsonDeserializer(Deserializer):
```
edenhill: Should this perhaps be JSONDeserializer?
See json.JSONDecoder

confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -114,7 +114,8 @@ def __init__(self, conf):
 
         # Any leftover keys are unknown to _RestClient
         if len(conf_copy) > 0:
-            raise ValueError("Unrecognized properties: {}".format(conf_copy.keys()))
+            raise ValueError("Unrecognized properties: {}"
+                             .format(", ".join(conf_copy.keys())))
```
edenhill: 💯

rnpridgeon: <3

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: It was originally, I disliked the way it looked. I don't have strong feelings though so I'll let @mhowlett break the tie.

tests/integration/schema_registry/test_json_serializers.py
mhowlett: i'm mostly concerned about backwards compatibility. but if this is all considered "experimental" as an extension of the `SerializingProducer` being marked as such, then that's fine.

confluent_kafka/schema_registry/json_schema.py
mhowlett: in .net there is no explicit validation against any schema on deserialization for json or protobuf. you would most want to do this against the reader schema (the schema supplied to the deserializer). i should ideally be doing that. validation against the writer schema seems overly paranoid. 

although i do no explicit validation against the schema, in .net the schema is defined implicitly through attributes on the class, and the json library will do some validation due to that. but we support two types of attributes - those from the JSON library and those from the JSON schema library, and the latter aren't validated against (easy to add without breaking API change).

what i don't know the most about is the need for a schema registry instance. you might want a schema_registry client instance for two reasons. validate against writer schema. maybe some paranoid people want to turn this on. auto-register reader schema, if unknown to SR. maybe some people want that. I don't know. I don't think i'd use it. in .net, confusion about this is not an issue as we can add it later without breaking the API. in python that seems less possible. Overall i'm comfortable with the approach as it stands (document the sr param is ignored). Not ideal, but probably not a big deal.

tests/integration/schema_registry/test_json_serializers.py
rnpridgeon: fixed

examples/json_consumer.py
```diff
@@ -0,0 +1,141 @@
...
+    return User(name=obj['name'],
+                favorite_number=obj['favorite_number'],
+                favorite_color=obj['favorite_color'])
+
+
+def main(args):
+    topic = args.topic
+
+    schema_str = """
+    {
```
mhowlett: in the .net implementation, I take the stance that users will never specify the schema directly. it's always implicit via property attributes. i think this approach matches the vibe of JSON which is very much ease of use over strictness and contrasts nicely with protobuf / avro. i'm very happy with this decision.

is this possible in python? 

sorry didn't think about this before when reviewing. python does have the language constructs to allow for it, don't know if there's a library.

examples/json_consumer.py
```diff
@@ -0,0 +1,141 @@
...
+          "description": "User's favorite color",
+          "type": "string"
+        }
+      },
+      "required": [ "name", "favorite_number", "favorite_color" ]
+    }
+    """
+
+    sr_conf = {'url': args.schema_registry}
+    schema_registry_client = SchemaRegistryClient(sr_conf)
```
mhowlett: given it's not used, i don't think you should construct it. misleading example.

examples/json_consumer.py
rnpridgeon: https://json-schema.org/implementations.html#from-code

We should phase out py 2.7 but I'm not sure we can move to exclusively supporting 3.6+

examples/json_consumer.py
rnpridgeon: good point

rnpridgeon: It's now optional but this gives us an asymmetric API (which also differs from Avro's). We could also just remove it and make it optional in a future release if we deem it necessary. We'd just need to makes sure we are comfortable always having to treat it as if the SR may not be there which might be fine.

examples/json_producer.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    Reports the failure or success of a message delivery.
+
+    Args:
+        err (KafkaError): The error that occurred on None on success.
+
+        msg (Message): The message that was produced or failed.
+
+    Note:
+        In the delivery report callback the Message.key() and Message.value()
```
mhowlett: as noted previously, i think there is a good backwards compatible api here in which these are renamed `key_bytes()` and `value_bytes()` (and `key()` and `value()` retrun non-None in the bytes case). but given the experimental designation, this can wait. I just didn't want to miss an opportunity to push my opinion on this.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -243,7 +243,7 @@ class JsonDeserializer(Deserializer):
     """
     __slots__ = ['_parsed_schema', '_registry', '_from_dict']
 
-    def __init__(self, schema_registry_client, schema_str, from_dict=None):
+    def __init__(self, schema_str, from_dict=None, schema_registry_client=None):
```
mhowlett: if you're changing the order like this, just remove the `schema_registry` parameter altogether - it can be added later without breaking backwards compatibility if needed (which it probably won't be).

rnpridgeon: agreed

rnpridgeon: poof

confluent_kafka/schema_registry/json_schema.py
mhowlett: reference name strategy is arguably a different type of thing, and this text still misleading  (but minor thing)

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,294 @@
...
+    +--------------------------------------+------------------------------+
+    | topic_record_subject_name_strategy   | {topic name}-{record name}   |
+    +--------------------------------------+------------------------------+
+    | record_subject_name_strategy         | {record name}                |
+    +--------------------------------------+------------------------------+
+
+    See ``Subject name strategy`` for additional details.
+
+    Note:
+        The ``title`` annotation, referred to as a record name
```
mhowlett: this needs rewording a bit i think to be precisely correct, but that can wait.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,294 @@
...
+        schema_dict = json.loads(schema_str)
+        schema_name = schema_dict.get('title', None)
+        if schema_name is None:
+            raise ValueError("Missing required JSON schema annotation title")
+
+        self._schema_name = schema_name
+        self._parsed_schema = schema_dict
+        self._schema = Schema(schema_str, schema_type="JSON")
+
+    def __call__(self, ctx, obj):
```
mhowlett: it makes more sense to me for ctx to be last (and to match .net). if you changed this on behalf of a comment i made, i apologize, didn't mean it here. i meant ctx should be first in the subject naming case.

mhowlett: reasoning: here the extensibility you're most likely to want is going to be inside SerializationContext. In the subject naming case, the extensibility you might care about is outside it.

confluent_kafka/schema_registry/json_schema.py
mhowlett: i still disagree, weighing up the magnitude of the potential downsides, but i think in practice worst case scenario unlikely to be hit, so i don't really care.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,294 @@
...
+    def __init__(self, schema_str, from_dict=None):
+        self._parsed_schema = json.loads(schema_str)
+
+        if from_dict is not None and not callable(from_dict):
+            raise ValueError("from_dict must be callable with the signature"
+                             " from_dict(SerializationContext, dict) -> object")
+
+        self._from_dict = from_dict
+
+    def __call__(self, ctx, value):
```
mhowlett: again, flip order. sorry :-(

confluent_kafka/schema_registry/json_schema.py
mhowlett: what @magnus says. from PEP-8:

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: sigh,  yes I wanted them to be consistent across the board so I changed both. The serializers unfortunately being the more difficult one to change. Alright I'll fix it here then open a PR for generic serdes and another for avro

examples/json_producer.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    def __init__(self, name, address, favorite_number, favorite_color):
+        self.name = name
+        self.favorite_number = favorite_number
+        self.favorite_color = favorite_color
+        # address should not be serialized, see user_to_dict()
+        self._address = address
+
+
+def user_to_dict(ctx, user):
```
mhowlett: ahh, same comment about arg ordering here. flip user, ctx since user is the important thing, ctx is peripheral.

mhowlett: it's the opposite in subject naming. ctx contains the more important stuff.

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -0,0 +1,294 @@
...
+            # JSON documents are self-describing; no need to query schema
+            obj_dict = json.loads(payload.read(), encoding="utf8")
+
+            try:
+                validate(instance=obj_dict, schema=self._parsed_schema)
+            except ValidationError as ve:
+                raise SerializationError(ve.message)
+
+            if self._from_dict is not None:
+                return self._from_dict(ctx, obj_dict)
```
mhowlett: flip arg ordering - ctx last.

confluent_kafka/schema_registry/json_schema.py
edenhill: If we change the order here should we also change the order on AvroDeserializer?

rnpridgeon: it's gone entirely

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: sure thing. I actually want to do a docs specific PR to verify/correct any rendering problems, wording etc.

confluent_kafka/schema_registry/json_schema.py
rnpridgeon: np all done

PR #816, opened 2020-04-01
ProtobufSerializer/ProtobufDeserializer for Generic Serdes API

rnpridgeon:
Open question, should we extract the DESCRIPTOR from the proto class or leave it as is.


confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+from . import (_MAGIC_BYTE,
+               reference_subject_name_strategy,
+               topic_subject_name_strategy,)
+from .schema_registry_client import (Schema,
+                                     SchemaReference)
+from confluent_kafka.serialization import SerializationError
+
+if sys.version > '3':
+    def _bytes(b):
```
edenhill, line 34: Add a comment saying what type b should be in both cases

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+    msg_idx.appendleft(len(msg_idx))
+    return list(msg_idx)
+
+
+def _schema_to_str(proto_file):
+    """
+    Base64 encodes a FileDescriptor
+
+    Args:
+        proto_file (Descriptor): FileDescriptor to encode.
```
edenhill, line 108: Descriptor or FileDescriptor?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+        if not callable(self._subject_name_func):
+            raise ValueError("subject.name.strategy must be callable")
+
+        self._ref_reference_subject_func = conf_copy.pop('reference.subject.name.strategy')
+        if not callable(self._ref_reference_subject_func):
+            raise ValueError("subject.name.strategy must be callable")
+
+        if len(conf_copy) > 0:
+            raise ValueError("Unrecognized properties: {}"
+                             .format(conf_copy.keys()))
```
edenhill, line 209: think you changed this to a `", ".join..` in the other serdes?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+        proto_file = descriptor.file
```
edenhill, line 211: nit: this temporary variable seems unnecessary

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+        Args:
+            buf (BytesIO): buffer to write index to
+            index [index]: Message index
+
+        Returns:
+            [bytes]: Serialized Message index
+
+        """
+        for value in index:
```
edenhill, line 231: I might be tired, but this varint encoder does not seem correct, it writes the continuation bit for the last octet if it is >0. shouldn't it be `while (value & ~0x7f): ` ?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+        proto_file = descriptor.file
+
+        self._proto_file = proto_file
+        self.msg_index = _create_msg_index(descriptor)
+        self._schema = Schema(_schema_to_str(proto_file),
+                              schema_type='PROTOBUF')
+
+    @staticmethod
+    def _encode_index(buf, index):
```
edenhill, line 219: Add unit tests for the edge cases (0, 1, 0x7f-1, 0x7f, 0x7f+1, etc).

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+    Args:
+        descriptor (Descriptor): Protobuf Message Descriptor.
+
+    .. _Schema definition:
+        https://json-schema.org/understanding-json-schema/reference/generic.html
+
+    """
+    def __init__(self, descriptor):
+        self.msg_index = _create_msg_index(descriptor)
```
edenhill, line 318: shall these be private fields?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+            SerializerError: If response payload and expected Message type
+            differ.
+
+        """
+        if value is None:
+            return None
+
+        # SR wire protocol + msg_index length
+        if len(value) < 6:
+            raise SerializationError("message is too small to decode {}".format(value))
```
edenhill, line 423: Write the good error message here, like in the last two deserializers.

examples/Makefile
```diff
@@ -0,0 +1,14 @@
+WORK_DIR:=$(strip $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+PROTO_HOME=/usr/local/opt/include
```
edenhill, line 2: Questionable portability

examples/Makefile
```diff
@@ -0,0 +1,14 @@
...
+SRC_DIR=$(WORK_DIR)
+TARGET_DIR=$(WORK_DIR)
+
+PROTOS := user.proto
+
+compile: $(PROTOS)
```
edenhill, line 8: nit: This target will always be rebuilt, a proper target should be:
`output_file: input_file`

But I guess it is fine.

examples/Makefile
```diff
@@ -0,0 +1,14 @@
...
+	for proto in $(PROTOS); do \
+		protoc -I=$(PROTO_HOME) -I=$(SRC_DIR) --python_out=$(TARGET_DIR) $$proto ; \
+	done
+
+clean:
+	rm $(TARGET_DIR)/*_pb2.py
```
edenhill, line 14: rm -f   to avoid failure when files dont exist

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: if I use it more than once I usually just create a throw away variable. We can get rid of it though since the first time is actually just the instance assignment

rnpridgeon: actually this isn't needed at all anymore because I moved the Schema declaration here

examples/Makefile
rnpridgeon: open to suggestions

examples/Makefile
rnpridgeon: I really dislike rfing script generated recursive removals but as you wish

examples/Makefile
rnpridgeon: Resolving as "fine the way it is"

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+    | record_subject_name_strategy         | {record name}                |
+    +--------------------------------------+------------------------------+
+
+    See ``Subject name strategy`` for additional details.
+
+    Args:
+        schema_registry_client (SchemaRegistryClient): Schema Registry
+            client instance.
+
+        descriptor (Descriptor): Protobuf Message Descriptor.
```
mhowlett, line 163: Why do you need to pass the descriptor to the serializer constructor? The serializer only accepts classes generated by `protoc`, so you should be able to get it on the fly.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+            raise ValueError("subject.name.strategy must be callable")
+
+        if len(conf_copy) > 0:
+            raise ValueError("Unrecognized properties: {}"
+                             .format(conf_copy.keys()))
+
+        proto_file = descriptor.file
+
+        self._proto_file = proto_file
+        self.msg_index = _create_msg_index(descriptor)
```
mhowlett, line 214: you'd want to cache all this stuff if you were doing it on the fly, and in such a way a to support multiple types.
a bit more work, but not much.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+                 'msg_index', '_proto_file', '_schema', '_schema_id',
+                 '_ref_reference_subject_func', '_subject_name_func']
+    # default configuration
+    _default_conf = {
+        'auto.register.schemas': True,
+        'subject.name.strategy': topic_subject_name_strategy,
+        'reference.subject.name.strategy': reference_subject_name_strategy
+    }
+
+    def __init__(self, schema_registry_client, descriptor, conf=None):
```
mhowlett, line 184: i guess there's an adequate path to making the api support multiple types in the future, and that's to give `descriptor` a default value of `None` here. however it forces you to name the `conf` parameter.

i think we should just do the more general thing now as i think it's not too much more effort.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        if msg is None:
+            return None
+
+        subject = self._subject_name_func(ctx, msg.DESCRIPTOR.full_name)
+
+        if subject not in self._known_subjects:
+            self._schema.references = self.resolve_dependencies(ctx, msg.DESCRIPTOR.file)
+
+        self._schema_id = self._registry.register_schema(subject, self._schema)
```
mhowlett, line 293: what if `_auto_register` is `false?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+        with _ContextStringIO() as fo:
+            # Write the magic byte and schema ID in network byte order (big endian)
+            fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))
+            # write the record index to the buffer
+            self._encode_index(fo, self.msg_index)
```
mhowlett, line 299: this `msg_index` relates to the message descriptor passed into the constructor, however your code here will happily work if `msg` is any protobuf type. if a different type is serialized than the descriptor passed to the constructor, this will be incorrect.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Originally @magnus requested that serializers be 1:1 with the type they are serializing. This I believe was based on how Avro works though since the Schema is not bound to the instance as is the case with Protobuf. With that said this serializer knows how to  serialize one type of proto message and one only its in the constructor. If we want to be more flexible, which is substantially easier with protobuf than avro (since we use generic types) that's more than okay with me.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: see previous comment

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+            bytes: Confluent Schema Registry formatted Protobuf bytes
+
+        """
+        if msg is None:
+            return None
+
+        subject = self._subject_name_func(ctx, msg.DESCRIPTOR.full_name)
+
+        if subject not in self._known_subjects:
+            self._schema.references = self.resolve_dependencies(ctx, msg.DESCRIPTOR.file)
```
mhowlett, line 291: this isn't right if this method can work with `msg`s of different types.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: related to first comment. Originally the decision was to bind the serializer to a single type. I'm on board with supporting multiple types if we can get @edenhill to agree

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: you are right, will fix encoder/decoder and add more tests.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+        Returns:
+            int: decoded uvarint
+
+        Raises:
+            EOFError: if buffer is empty
+
+        """
+        value = 0
+        shift = 0
+        while True:
```
mhowlett, line 338: i'm pretty sure @edenhill could tell you whether this logic is correct or not with his eyes closed. since there is no comment here, i assume it's good (haven't checked it).

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+    Protobuf format to an object.
+
+    Args:
+        descriptor (Descriptor): Protobuf Message Descriptor.
+
+    .. _Schema definition:
+        https://json-schema.org/understanding-json-schema/reference/generic.html
+
+    """
+    def __init__(self, descriptor):
```
mhowlett, line 317: i think we should not pass descriptors in here.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+
+        if subject not in self._known_subjects:
+            self._schema.references = self.resolve_dependencies(ctx, msg.DESCRIPTOR.file)
+
+        self._schema_id = self._registry.register_schema(subject, self._schema)
+
+        with _ContextStringIO() as fo:
+            # Write the magic byte and schema ID in network byte order (big endian)
+            fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))
+            # write the record index to the buffer
```
mhowlett, line 298: you're not writing the schema id.

mhowlett: (you need to, it's part of the serialization format)

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+        if len(value) < 6:
+            raise SerializationError("message is too small to decode {}".format(value))
+
+        with _ContextStringIO(value) as payload:
+            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            if magic != _MAGIC_BYTE:
+                raise SerializationError(
+                    "message does not start with magic byte")
+
+            # Protobuf Messages are self-describing; no need to query schema
```
mhowlett, line 431: read schema id here

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Related to first comment, reached out in slack to spark up conversation about more flexible protobuf serde. The 1:1 instance type requirement brought this about.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: >fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        Returns int representation for a byte.
+
+        Args:
+            buf (BytesIO): buffer to read from
+
+        .. _ord:
+            https://docs.python.org/2/library/functions.html#ord
+        """
+        i = buf.read(1)
```
mhowlett, line 359: seems like there should be some better way of doing this (i don't know, just speaking what i'm thinking)

mhowlett: bit nit, it'll work...

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+        """
+        Extracts message index from Schema Registry Protobuf formatted bytes.
+
+        Args:
+            buf (BytesIO): byte buffer
+
+            expect ([int]): expected message index
+
+        Returns:
+            [int]: Message index. Not currently used.
```
mhowlett, line 375: comment that this isn't used belongs on method call, not here.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Today its bound to a specific message type.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: > magic, schema_id = struct.unpack('>bI', payload.read(5)

rnpridgeon: its more efficient to read/write the schema id with the magic byte since they are both known quantities

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,437 @@
...
+            raise SerializationError("Message index mismatch. The consumed"
+                                     " Message does not match this"
+                                     " Deserializer's Message type")
+
+        msg_index = [size]
+        for _ in range(size):
+            msg_index.append(ProtobufDeserializer._decode_uvarint(buf))
+
+        if msg_index != expect:
+            raise SerializationError("Message index mismatch. The consumed"
```
mhowlett, line 394: this is fine - they don't need to match, they just need to be compatible. `ParseFromString` will tell you that. you should remove the `expect` checking stuff.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: tests pass so I've got that going for me but they also passed with a bug in the encoder so there is that.

confluent_kafka/schema_registry/protobuf.py
mhowlett: you haven't implemented the special case where the index array is [0]. This is encoded as 0, not 1, 0. see the C# code for e.g. impl: https://github.com/confluentinc/confluent-kafka-dotnet/blob/1.4.x/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs#L147

rnpridgeon: I actually handle this while generating the message index 

```
def _create_msg_index(msg_desc):
...
    # The root element at the 0 position does not need a length prefix.
    if len(msg_idx) == 1 and msg_idx[0] == 0:
        return [0]

    msg_idx.appendleft(len(msg_idx))
    return list(msg_idx)
```

I use a double ended queue

confluent_kafka/schema_registry/protobuf.py
mhowlett: consistency is an almost insignificant priority here imo.

rnpridgeon: FWIW I agree, but I'll wait for us to all come to agreement before jumping because I don't want to do it twice.

confluent_kafka/schema_registry/protobuf.py
mhowlett: alternative comment: using the `DESCRIPTOR` from msg is misleading then and doesn't protect against incorrect use.

confluent_kafka/schema_registry/protobuf.py
mhowlett: oh, right, blind.

confluent_kafka/schema_registry/protobuf.py
mhowlett: oh right. my (minor) comment here then is if the steps were more clearly spelled out, i.e. 1. create the index array 2. modify for the optimization, i would have seen this easier. the more specific the function the better.

confluent_kafka/schema_registry/protobuf.py
mhowlett: yep, somehow i was blind to the this. good work, carry on!

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Added tests for 0, 1, 127, 128, `etc` felt unnecessary with 128 there but we can add to the test matrix pretty easily if you want more.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: I'd argue the "special case" here is really a characteristic of the index itself not the act of encoding the index which would make this the right place for it.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: There are a few annoying things to account for here. 

1. BytesIO.read(1) lets you know that you have hit EOF by returning an empty bytes object.
2. You can't actually perform bitwise operations on a bytes object (hence ord)

confluent_kafka/schema_registry/protobuf.py
mhowlett: yeah, agree with you. sorry, i wasn't giving this enough focus.

examples/Makefile
edenhill: At least add a comment what is supposed to be installed in /usr/local/opt/include, proto-something?

tests/integration/schema_registry/test_proto_serializers.py
```diff
@@ -0,0 +1,108 @@
+import pytest
```
edenhill, line 1: LICENSE....

tests/integration/schema_registry/test_proto_serializers.py
```diff
@@ -0,0 +1,108 @@
...
+                                                 one_id='oneof_str',
+                                                 is_active=False)})
+])
+def test_protobuf_message_serialization(kafka_cluster, pb2, data):
+    """
+    Validates that we get the same message back that we put in.
+
+    """
+    topic = kafka_cluster.create_topic("serialization-proto")
+    sr = kafka_cluster.schema_registry({'url': 'http://localhost:8081'})
```
edenhill, line 40: I'd rather see us use a dummy-url that does not risk being an existing SR instance, e.g., http://localhost:65534.
Otherwise we risk messing up someones unrelated dev SR.

tests/integration/schema_registry/test_proto_serializers.py
```diff
@@ -0,0 +1,108 @@
...
+    consumer = kafka_cluster.consumer(key_deserializer=deserializer)
+    consumer.assign([TopicPartition(topic, 0)])
+
+    producer.produce(topic, key=pb2_1(), partition=0)
+    producer.flush()
+
+    with pytest.raises(ConsumeError, match="Message index mismatch. The"
+                                           " consumed Message does not match"
+                                           " this Deserializer's Message type"
+                                           r" \(KafkaError code -160\)"):
```
edenhill, line 107: The KafkaError code must be written in its string (err2name) form, not the int code.

tests/schema_registry/test_proto.py
```diff
@@ -0,0 +1,77 @@
...
+    buf.close()
+
+    assert decoded_msg_idx == msg_idx
+
+
+@pytest.mark.parametrize("msg_idx, expected_hex", [
+    ([1, 0], b'00'),   # b2a_hex always returns hex pairs
+    ([1, 1], b'01'),
+    ([1, 127], b'7f'),
+    ([1, 128], b'8001')
```
edenhill, line 68: try a large value as well, occupying 5 or so encoded bytes.

tests/schema_registry/test_proto.py
```diff
@@ -0,0 +1,77 @@
...
+])
+def test_index_encoder(msg_idx, expected_hex):
+    buf = BytesIO()
+    ProtobufSerializer._encode_index(buf, msg_idx)
+    buf.flush()
+    # ignore array length prefix
+    buf.seek(1)
+
+    assert binascii.b2a_hex(buf.read()) == expected_hex
```
edenhill, line 77: add decoding to this test as well checking that the decoded value matches, ensuring symetricallity.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,440 @@
...
+        if subject not in self._known_subjects:
+            self._schema.references = self.resolve_dependencies(ctx, msg.DESCRIPTOR.file)
+
+        self._schema_id = self._registry.register_schema(subject, self._schema)
+
+        with _ContextStringIO() as fo:
+            # Write the magic byte and schema ID in network byte order (big endian)
+            fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))
+            # write the record index to the buffer
+            self._encode_index(fo, self._msg_index)
```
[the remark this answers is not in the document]
rnpridgeon: It is, I handle this during index creation. 

```
def _create_msg_index(msg_desc):
...
    # The root element at the 0 position does not need a length prefix.
    if len(msg_idx) == 1 and msg_idx[0] == 0:
        return [0]

    msg_idx.appendleft(len(msg_idx))
    return list(msg_idx)
```

examples/protobuf_producer.py
```diff
@@ -0,0 +1,106 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This is a simple example of the SerializingProducer using protobuf.
+#
+import argparse
+from uuid import uuid4
+
```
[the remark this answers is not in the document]
rnpridgeon: stdlib 

third-party

confluelnt

tests/integration/schema_registry/test_proto_serializers.py
rnpridgeon: yes sorry this is just there for running integration tests right now since there is no SR image with 5.5. I will remove it next push after running tests.

confluent_kafka/schema_registry/protobuf.py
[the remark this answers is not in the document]
edenhill: But that `[0]` list would still be encoded as length(0) + item(0), not just item(0), correct?

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Done, will follow up with Many:One Protobuf serializer

tests/integration/schema_registry/test_proto_serializers.py
edenhill: In that case maybe we need to make it configurable to use a non-trivup SR instead of hard-coding the URL in all these tests?   
This will be useful any time we need to test with a non-released version of SR, so how about adding this to trivup, maybe have the trivup SchemaRegistry honour env `TRIVUP_EXISTING_SR_URL` to expose that URL through its .url property rather than starting an SR?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,440 @@
...
+        if not callable(self._subject_name_func):
+            raise ValueError("subject.name.strategy must be callable")
+
+        self._ref_reference_subject_func = conf_copy.pop('reference.subject.name.strategy')
+        if not callable(self._ref_reference_subject_func):
+            raise ValueError("subject.name.strategy must be callable")
+
+        if len(conf_copy) > 0:
+            raise ValueError("Unrecognized properties: {}"
+                             .format(", ".format(conf_copy.keys())))
```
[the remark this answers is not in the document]
rnpridgeon: and one too few joins 🤦‍♂

examples/protobuf_consumer.py
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This is a simple example of the SerializingProducer using protobuf.
+#
+import argparse
+
+import user_pb2
```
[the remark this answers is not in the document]
rnpridgeon: Doesn't the import statement tell us that?

tests/schema_registry/test_proto.py
rnpridgeon: Ack, the thought processes there was that we have this test in the integration tests above albeit for real indexes instead of contrived ones.

tests/schema_registry/test_proto.py
rnpridgeon: ([1, 9223372036854775807], b'ffffffffffffffff7f') it is then

tests/integration/schema_registry/test_proto_serializers.py
rnpridgeon: Yes I think it would be great if trivup would accept overrides from environment variables prefixed with `TRIVUP_` (to avoid name clashing). I'd rather not bake this into the Python client tests as its applicable to anything using Trivup. I'd like  SR  to be  treated as a first class cluster as well.  I'll submit a PR after all of this is done and merged

confluent_kafka/schema_registry/protobuf.py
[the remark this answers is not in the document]
rnpridgeon: If the message we are interested is at the root level and it's the first message (index 0) we return a list with one element and that element is 0. If the message is anywhere else we return a list  prepended with the length of the list. The encoder blindly encodes each item in that list to the buffer. I didn't think it made sense to put that logic in the encoder as it should treat all indexes the same. So instead I bake it into the index itself as seen here.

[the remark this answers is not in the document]
rnpridgeon: This is covered in test_proto.test_create_index as well if you want to see it in action. 

https://github.com/confluentinc/confluent-kafka-python/pull/816/files?file-filters%5B%5D=.py#diff-a4a23da8ac61e21115dd8821f75124aaR36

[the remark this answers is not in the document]
rnpridgeon: Added `assert msg_idx[0] == len(coordinates)` to the prefixed array test case for completeness

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: resolving for now we  can address later if we decide on a cleaner suggestion. I don't want to block on cosmetics though.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Resolving based  on the decision to allow  for multiple types to be handled by the same serializer. We can discuss the implementation in the follow up PR.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Tripped up Magnus as well, renamed to `_encode_uvarints` in hopes it will more clearly convey the purpose of this method

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: We will revisit with the follow up PR which moves away from binding the  serializer to a specific protobuf type.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: We will revisit with the follow up PR which moves away from binding the  serializer to a specific protobuf type.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: This will need to be addressed in the multi-type serializer implementation to come in a follow up PR. I'll try to get it in this afternoon.

confluent_kafka/schema_registry/protobuf.py
edenhill: The optimization is a wire coding optimization, afaik, and should thus be done in the encoding function rather than this here.

examples/Makefile
rnpridgeon: Forgot this was the examples one not the tests Makefile. Fixed with proper target

rnpridgeon:
We decided offline to forego handling multiple types in with Protobuf Serializer for now while we figure out best how we want to plan this in the future. With that said the resolved comments which mention flexible protobuf serialzation may now be ignored.

confluent_kafka/deserializing_consumer.py
```diff
@@ -131,9 +131,9 @@ def poll(self, timeout=-1):
         if self._value_deserializer is not None:
             try:
                 value = self._value_deserializer(value, ctx)
-            except SerializationError as se:
+            except Exception as se:
                 raise ConsumeError(KafkaError._VALUE_DESERIALIZATION,
-                                   reason=se.message,
+                                   reason=str(se),
```
mhowlett, line 136: is there no way to embed the actual exception, not just the string representation of it? (that looses information)

confluent_kafka/error.py
```diff
@@ -15,11 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from confluent_kafka.cimpl import KafkaException, KafkaError
 
 
-class ConsumeError(Exception):
-    __slots__ = ['message', 'reason', 'error_code']
-
+class ConsumeError(KafkaException):
```
mhowlett, line 21: In an ideal world, would the naming `KafkaException` -> `KafkaError` and `KafkaError` -> `Error` make the most sense?

confluent_kafka/error.py
```diff
@@ -30,23 +29,25 @@ class ConsumeError(Exception):
     Args:
         error (KafkaError): The error that occurred.
```
mhowlett, line 30: you should support exception chaining

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+    +--------------------------------------+------------------------------+
+    | topic_record_subject_name_strategy   | {topic name}-{record name}   |
+    +--------------------------------------+------------------------------+
+    | record_subject_name_strategy         | {record name}                |
+    +--------------------------------------+------------------------------+
+
+    See ``Subject name strategy`` for additional details.
+
+    Args:
+        message (Message): Protobuf Message.
```
mhowlett, line 184: this reads like your passing an actual message to the constructor, not a type.

confluent_kafka/error.py
rnpridgeon: Yeah unfortunately we  don't have that luxury today as its technically a breaking change. I'd like to change that or come up with a better name in the future  though.

confluent_kafka/error.py
rnpridgeon: Don't think its supported until 3.0 

https://www.python.org/dev/peps/pep-3134/

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: renamed msg_type and used GeneratedProtocolMessageType as the type.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -204,7 +204,7 @@ static long KafkaError_hash (KafkaError *self) {
 	return self->code;
 }
 
-static PyTypeObject KafkaErrorType;
+PyTypeObject KafkaErrorType;
```
edenhill, line 207: Removing static makes this variable global to the program, but it is not used outside of this file so there is no point in doing so.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -271,16 +304,14 @@ static PyTypeObject KafkaErrorType = {
 	0,                         /*tp_setattro*/
 	0,                         /*tp_as_buffer*/
 	Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE |
-        Py_TPFLAGS_BASE_EXC_SUBCLASS | Py_TPFLAGS_HAVE_GC, /*tp_flags*/
+        Py_TPFLAGS_HAVE_GC, /*tp_flags*/
```
edenhill, line 307: Why are you removing this flag?

confluent_kafka/src/confluent_kafka.c
```diff
@@ -249,8 +249,41 @@ static PyObject* KafkaError_richcompare (KafkaError *self, PyObject *o2,
...
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i|z", kws, &code, &reason))
+                return -1;
+
+        KafkaError_init(self, code, reason ? reason : rd_kafka_err2str(code));
+        return 0;
+}
 
-static PyTypeObject KafkaErrorType = {
+PyTypeObject KafkaErrorType = {
```
edenhill, line 286: dito static

confluent_kafka/src/confluent_kafka.c
```diff
@@ -296,26 +327,12 @@ static PyTypeObject KafkaErrorType = {
 	0,                         /* tp_descr_get */
 	0,                         /* tp_descr_set */
 	0,                         /* tp_dictoffset */
-	0,                         /* tp_init */
-	0                          /* tp_alloc */
+	KafkaError_init0,          /* tp_init */
+	0,                         /* tp_alloc */
+        KafkaError_new             /* tp_new */
```
edenhill, line 332: Add unit tests in tests/test_Error.py (or whatever it is called) for this new functionality.

rnpridgeon: Sure we get coverage from the negatice serializer test but we can add one specific to KafkaError

confluent_kafka/error.py
```diff
@@ -30,23 +29,11 @@ class ConsumeError(Exception):
...
```
edenhill, line 30: Error seems to be an error code now, not KafkaError.
Both doc string and param name should probably be changed.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+        self.close()
+        return False
+
+
+def _create_msg_index(msg_desc):
+    """
+    Maps the location of msg_desc within a FileDescriptor.
+
+    Args:
+        msg_desc (MessageDescriptor): Message descriptor
```
edenhill, line 78: I'm thinking we might want to include the module name here, `protobuf.MessageDescriptor` to make it less ambigious with regards to this package's `Message`.

edenhill: This method modifies msg_desc, might want to point that out in the doc string here.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+
+    Returns:
+        [int]: Message
```
edenhill, line 81: This is also a bit confusing, is this a pb or kafka Message? the module would do good here.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+        self._schema = Schema(_schema_to_str(descriptor.file),
+                              schema_type='PROTOBUF')
+
+    @staticmethod
+    def _encode_uvarints(buf, uvarints):
+        """
+        Encodes each int as a uvarint onto buf
+
+        Args:
+            buf (BytesIO): buffer to write index to
```
edenhill, line 247: remove index

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+            uvarints ([int]): ints to be encoded.
```
edenhill, line 248: The parameter does not contain uvarints ,so better call this ints.

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+
+        Returns:
+            bytes: Serialized Message index
```
edenhill, line 251: Does not return anything

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+        self._msg_index = _create_msg_index(descriptor)
+        self._msg_class = MessageFactory().GetPrototype(descriptor)
+
+    @staticmethod
+    def _decode_uvarint(buf):
+        """
+        Decodes a single uvarint from a buffer.
+
+        Args:
+            - buf (BytesIO): buffer to read from
```
edenhill, line 363: remove dash?

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+        value = 0
+        shift = 0
+        try:
+            while True:
+                i = ProtobufDeserializer._read_byte(buf)
+
+                value |= (i & 0x7f) << shift
+                shift += 7
+                if not (i & 0x80):
+                    break
```
edenhill, line 381: could put the return here instead

confluent_kafka/schema_registry/protobuf.py
```diff
@@ -0,0 +1,465 @@
...
+
+        with _ContextStringIO(value) as payload:
+            magic, schema_id = struct.unpack('>bI', payload.read(5))
+            if magic != _MAGIC_BYTE:
+                raise SerializationError("Unknown magic byte. This message was"
+                                         " not produced with a Confluent"
+                                         " Schema Registry serializer")
+
+            # Protobuf Messages are self-describing; no need to query schema
+            # Move the reader cursor passed the index
```
edenhill, line 457: past

confluent_kafka/serialization/__init__.py
```diff
@@ -29,6 +29,8 @@
            'SerializationError',
            'Serializer']
 
+from confluent_kafka.cimpl import KafkaException
```
edenhill, line 32: drop .cimpl

confluent_kafka/serializing_producer.py
```diff
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from confluent_kafka.cimpl import KafkaError, KafkaException
```
edenhill, line 18: drop .cimpl

confluent_kafka/serializing_producer.py
```diff
@@ -152,16 +153,23 @@ def produce(self, topic, key=None, value=None, partition=-1,
                 the application should call :py:func:`SerializingProducer.Poll`
                 and try again.
 
-             KafkaException: for other errors, see exception code
+             KafkaError: For all other errors
```
edenhill, line 156: Add explicit mentions of Key|ValueSerializerError here, finish off with KafkaError for everything else.

confluent_kafka/serializing_producer.py
```diff
@@ -152,16 +153,23 @@ def produce(self, topic, key=None, value=None, partition=-1,
...
 
         """
         ctx = SerializationContext(topic, MessageField.KEY)
         if self._key_serializer is not None:
-            key = self._key_serializer(key, ctx)
-
+            try:
+                key = self._key_serializer(key, ctx)
+            except Exception as se:
+                raise KafkaException(
```
edenhill, line 164: This should be a KeySerializerError()

confluent_kafka/src/confluent_kafka.c
```diff
@@ -249,6 +249,39 @@ static PyObject* KafkaError_richcompare (KafkaError *self, PyObject *o2,
...
+                                 PyObject *kwargs) {
+        return type->tp_alloc(type, 0);
+}
+
+/**
+ * @brief Initialize a KafkaError object.
+ */
+static void KafkaError_init (KafkaError *self,
+			     rd_kafka_resp_err_t code, const char *str) {
+	self->code = code;
```
edenhill, line 262: Since this code is moved around we can fix the spacing, convert tabs to 8 whitespaces.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -249,6 +249,39 @@ static PyObject* KafkaError_richcompare (KafkaError *self, PyObject *o2,
...
+		self->str = strdup(str);
+	else
+		self->str = NULL;
+}
+
+static int KafkaError_init0 (PyObject *selfobj, PyObject *args, PyObject *kwargs) {
+        KafkaError *self = (KafkaError *)selfobj;
+        int code;
+        const char *reason = NULL;
+
```
edenhill, line 276: remove empty line.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -249,6 +249,39 @@ static PyObject* KafkaError_richcompare (KafkaError *self, PyObject *o2,
...
+	else
+		self->str = NULL;
+}
+
+static int KafkaError_init0 (PyObject *selfobj, PyObject *args, PyObject *kwargs) {
+        KafkaError *self = (KafkaError *)selfobj;
+        int code;
+        const char *reason = NULL;
+ 
+        static char *kws[] = { "error", "reason",  NULL };
```
edenhill, line 277: It would be good to add optional retriable, fatal, txn_requires_abort so that those errors can be mocked.
This also allows us to remove the test_create_fatal_error() or whatever it is called.

edenhill: rename error to error_code or code.

confluent_kafka/src/confluent_kafka.c
edenhill, line 332: Either have them all be whitespaces or tabs, this is a mix now.

examples/Makefile
```diff
@@ -0,0 +1,11 @@
+WORK_DIR:=$(strip $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST)))))
+SRC_DIR?=$(WORK_DIR)
```
edenhill, line 2: When would these be set by the developer?

examples/README.md
```diff
@@ -5,8 +5,14 @@ The scripts in this directory provide code examples using Confluent's Python cli
 * [consumer.py](consumer.py): reads messages from a Kafka topic
 * [producer.py](producer.py): reads lines from stdin and sends them to Kafka
 * [eos-transactions.py](eos-transactions.py): transactional producer with exactly once semantics (EOS)
+* [avro_producer.py](avro_producer.py): SerializingProducer with AvroSerializer
+* [avro_consumer.py](avro_consumer.py): DeserializingConsumer with AvroDeserializer
+* [json_producer.py](json_producer.py): SerializingProducer with JsonSerializer
+* [json_consumer.py](json_consumer.py): DeserializingConsumer with JsonDeserializer
+* [protobuf_producer.py](protobuf_producer.py): SerializingProducer with ProtobufSerializer
+* [protobuf_consumer.py](protobuf_consumer.py): DeserializingConsumer with ProtobufDeserializer
```
edenhill, line 13: 🎉

examples/README.md
```diff
@@ -5,8 +5,14 @@ The scripts in this directory provide code examples using Confluent's Python cli
...
 
 Additional examples for [Confluent Cloud](https://www.confluent.io/confluent-cloud/):
 
 * [confluent_cloud.py](confluent_cloud.py): produces messages to Confluent Cloud and then reads them back again
-* [confluentinc/examples](https://github.com/confluentinc/examples/tree/master/clients/cloud/python): integrates Confluent Cloud and Confluent Cloud Schema Registry
+* [confluentinc/examples](https://github.com/confluentinc/examples/tree/master/clients/cloud/python): integrates Confluent Cloud and Confluent Cloud Schema Registry`
```
edenhill, line 18: trailing garbage

examples/protobuf_consumer.py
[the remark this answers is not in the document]
edenhill: No, it doesnt say it is a proto-something generated class.
But now it does!

examples/protobuf_consumer.py
```diff
@@ -0,0 +1,86 @@
...
+            if user is not None:
+                print("User record {}: name: {}\n"
+                      "\tfavorite_number: {}\n"
+                      "\tfavorite_color: {}\n"
+                      .format(msg.key(), user.name,
+                              user.favorite_color,
+                              user.favorite_number))
+        except KeyboardInterrupt:
+            break
+
```
edenhill, line 73: consumer.close()

examples/protobuf_producer.py
```diff
@@ -0,0 +1,117 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This is a simple example of the SerializingProducer using protobuf.
+#
+# To regenerate Protobuf classes you must first install the protobuf
```
edenhill, line 22: Do they have to? I thought we checked in the generated classes?

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -16,11 +16,10 @@
 # limitations under the License.
 #
 import pytest
-from confluent_kafka.cimpl import TopicPartition, KafkaError
+from confluent_kafka.cimpl import TopicPartition, KafkaException
```
edenhill, line 19: drop .cimpl

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -150,7 +149,7 @@ def test_json_record_serialization_incompatible(kafka_cluster, load_file):
               "contractRate": 1250,
               "trades": ["mason"]}
 
-    with pytest.raises(SerializationError,
+    with pytest.raises(KafkaException,
```
edenhill, line 152: Serializers should raise Key|ValueSerializerError, inheritance is:
Key|ValueSerializerError -> SerializerError -> KafkaException (has KafkaError as args 0) -> Exception

tests/integration/schema_registry/test_json_serializers.py
```diff
@@ -251,6 +250,5 @@ def test_json_record_deserialization_mismatch(kafka_cluster, load_file):
 
     with pytest.raises(
             ConsumeError,
-            match=r"(.*) is a required property \(KafkaError code {}\)"
-                  .format(KafkaError._VALUE_DESERIALIZATION)):
+            match="\'productId\' is a required property"):
```
edenhill, line 253: Don't think you need to escape `'`

tests/test_KafkaError.py
```diff
@@ -56,3 +56,9 @@ def a_method(self):
     err = MyExc()
     assert err.a_method() == "yes"
     assert isinstance(err, KafkaException)
+
+
+def test_kafkaError_custom_msg():
+    err = KafkaError(KafkaError._ALL_BROKERS_DOWN, "Mayday!")
+    assert err == KafkaError._ALL_BROKERS_DOWN
+    assert err.str() == "Mayday!"
```
edenhill, line 64: Add a KafkaException-wrapped KafkaError test too, and one where you specify an unknown code, e.g., 12345.

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Where does it modify msg_desc?

edenhill: Sorry, read the code wrong, ignore that!

examples/protobuf_producer.py
rnpridgeon: That's why it says regenerate as opposed to generate. The implication being if you want to generate them again. I can remove it or say if you want to make changed to the proto definition you need to regenerate them and you can use make.

edenhill: Ah, sorry, read that as "generate". It is fine as is-is then :+1:

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: it's also wrong it returns the message index which is neither a protobuf Message nor a Kafka Message. I'll prefix Message and add index

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: So there are actually two type of message impls and depending on which one is being used the namespace will be different. Neither of which is protobuf.MessageDescriptor. Can we just clarify in the doc string instead

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: 🤦‍♂

confluent_kafka/src/confluent_kafka.c
rnpridgeon: in that order?

edenhill: Doesn't really matter, but let's do fatal, retriable, txn_requires_abort to match what is in init().

confluent_kafka/src/confluent_kafka.c
```diff
@@ -279,8 +312,6 @@ static PyTypeObject KafkaErrorType = {
 	"  - Propagation of errors\n"
 	"  - Propagation of events\n"
 	"  - Exceptions\n"
```
edenhill, line 314: Need to document the constructor and its arguments here.

examples/Makefile
rnpridgeon: When you put your Makefile in a top level directory your proto files in another directory and your generated code in yet another. Sort of like setup.py doesn't live in the source directory. If it bugs you we can remove it.

edenhill: But this Makefile is only for the user protobuf file in this directory, so why would someone need to use alternate directories?

confluent_kafka/src/confluent_kafka.c
rnpridgeon: :+1:

confluent_kafka/schema_registry/protobuf.py
rnpridgeon: Discussed offline, as long as we clarify somewhere it's good. So we will do so in the description

confluent_kafka/src/confluent_kafka.c
rnpridgeon: sorry this came back when I added the additional args. Fixing again

tests/test_KafkaError.py
```diff
@@ -56,3 +57,26 @@ def a_method(self):
...
+
+
+def test_kafkaException_unknown_KafkaError_with_subclass():
+    class MyException(KafkaException):
+        def __init__(self, error_code):
+            super(MyException, self).__init__(KafkaError(error_code))
+
+    with pytest.raises(KafkaException, match="Err-12345?") as e:
+        raise MyException(12345)
+    assert not e.value.args[0].fatal()
```
edenhill, line 82: No tests for retriable, fatal, txn_requires_abort.

confluent_kafka/src/confluent_kafka.c
```diff
@@ -249,6 +238,46 @@ static PyObject* KafkaError_richcompare (KafkaError *self, PyObject *o2,
...
+static int KafkaError_init0 (PyObject *selfobj, PyObject *args, 
+                             PyObject *kwargs) {
+        KafkaError *self = (KafkaError *)selfobj;
+        int code;
+        int fatal = 0, retriable = 0, txn_requires_abort = 0;
+        const char *reason = NULL;
+        static char *kws[] = { "error", "reason", "fatal", 
+                               "retriable", "txn_requires_abort", NULL };
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "i|ziii", kws, &code,
```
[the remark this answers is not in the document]
rnpridgeon: Is that true for py2? I'm not finding it here 

https://docs.python.org/2/c-api/arg.html

confluent_kafka/error.py
```diff
@@ -28,25 +29,111 @@ class ConsumeError(Exception):
         may be retrieved from the ``message`` attribute.
 
     Args:
-        error (KafkaError): The error that occurred.
+        error_code (KafkaError): Error code indicating the type of error.
 
-        message (Message, optional): The message returned from the broker.
+        error(Exception, optional): The original exception
```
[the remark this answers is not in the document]
rnpridgeon: +1

confluent_kafka/error.py
```diff
@@ -28,25 +29,111 @@ class ConsumeError(Exception):
...
 
-        reason (str): String description of the error.
+        message (Message, optional): The Kafka Message returned from the broker.
 
     """
+    def __init__(self, error_code, error=None, message=None):
+        if error is not None:
+            kafka_error = KafkaError(error_code, repr(error))
```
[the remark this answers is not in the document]
rnpridgeon: you typically look at the text of an exception when debugging right? It's helpful to see the original exception in the string when checking the output. 

i.e. 

```
v = ValueError('abc')
str(v)
'abc'
repr(v)
"ValueError('abc')"

File: confluent_kafka/serializing_producer.py [hunk 72]
```diff
@@ -16,9 +16,11 @@
 # limitations under the License.
 #
 
-from .cimpl import Producer as _ProducerImpl
+from confluent_kafka.cimpl import Producer as _ProducerImpl
```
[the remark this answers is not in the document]
rnpridgeon: We import SerializingProducer and DeserializingConsumer into` __init__.py` if we get confluent_kafka.cimpl from `__init__.py` we will endup with a circular dependency. We could remove SerializingProducer and DeserializingConsumer from `__init__.py` but I'm not convinced that's what we want.

PR #824, opened 2020-04-07
Producer/Consumer SASL auth examples

rnpridgeon:


examples/sasl_consumer.py
```diff
@@ -0,0 +1,103 @@
...
+                        help="Topic name")
+    parser.add_argument('-g', dest="group", default="example_sasl",
+                        help="Consumer group")
+    parser.add_argument('-m', dest="sasl_mechanism", default='PLAIN',
+                        help="SASL mechanism to use for authentication."
+                             "Defaults to PLAIN")
+    parser.add_argument('-u', dest="user_principal", required=True,
+                        help="username")
+    parser.add_argument('-s', dest="user_secret", required=True,
+                        help="password; path to keytab if using"
```
edenhill, line 100: Password for PLAIN and SCRAM, or path to..

examples/sasl_consumer.py
```diff
@@ -0,0 +1,103 @@
...
+                        help="Bootstrap broker(s) (host[:port])")
+    parser.add_argument('-t', dest="topic", default="example_sasl",
+                        help="Topic name")
+    parser.add_argument('-g', dest="group", default="example_sasl",
+                        help="Consumer group")
+    parser.add_argument('-m', dest="sasl_mechanism", default='PLAIN',
+                        help="SASL mechanism to use for authentication."
+                             "Defaults to PLAIN")
+    parser.add_argument('-u', dest="user_principal", required=True,
+                        help="username")
```
edenhill, line 98: Username

examples/sasl_consumer.py
```diff
@@ -0,0 +1,103 @@
...
+        args (argparse.Namespace): parsed args
+
+    Returns:
+         dict: SASL configuration dict
+
+    """
+    sasl_mechanism = args.sasl_mechanism
+
+    sasl_conf = {'sasl.mechanism': sasl_mechanism,
+                 'security.protocol': 'SASL_PLAINTEXT'}
```
edenhill, line 42: Might want to make a note here for SASL_SSL

examples/sasl_consumer.py
```diff
@@ -0,0 +1,103 @@
...
+
+    if sasl_mechanism != 'GSSAPI':
+        sasl_conf.update({'sasl.username': args.user_principal,
+                          'sasl.password': args.user_secret})
+    else:
+        sasl_conf.update({'sasl.kerberos.service.name', args.broker_principal,
+                          # On Windows the configured user principal is ignored.
+                          # The logged on user's credentials are used instead.
```
edenhill, line 50: and keytabs are not supported.

examples/sasl_consumer.py
```diff
@@ -0,0 +1,103 @@
...
+                          'sasl.kerberos.principal', args.user_principal,
+                          'sasl.kerberos.keytab', args.user_secrent})
```
edenhill, line 52: secrent.
Make sure to test both GSSAPI and not GSSAPI

examples/sasl_consumer.py
```diff
@@ -0,0 +1,103 @@
...
+    consumer.subscribe([topic])
+
+    while True:
+        try:
+            # SIGINT can't be handled when polling, limit timeout to 1 second.
+            msg = consumer.poll(1.0)
+            if msg is None:
+                continue
+
+            user = msg.value()
```
edenhill, line 77: user does not make sense here, just use value and key directly.

examples/sasl_producer.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python
```
edenhill, line 1: Since this is to show-case how to set up SASL and the SASL setups are identical I don't really see the need for both a consumer and producer.
If we were to keep one I suggest keeping the producer since it does not require another party to operate.

examples/sasl_consumer.py
rnpridgeon: mention or add support for

edenhill: Mentioning when to use SASL_SSL  is enough.

examples/sasl_producer.py
```diff
@@ -114,10 +116,11 @@ def main(args):
...
                              "Defaults to PLAIN")
+    parser.add_argument('--tls', dest="enab_tls", default=False)
     parser.add_argument('-u', dest="user_principal", required=True,
-                        help="username")
+                        help="Username")
     parser.add_argument('-s', dest="user_secret", required=True,
-                        help="password; path to keytab if using"
-                             "SASL Mechanism GSSAPI")
+                        help="Password for PLAIN and SCRAM, or path to keytab "
+                             "if GSSAPI.")
```
edenhill, line 124: " (ignored on Windows)"  (maybe make it non-required)

PR #827, opened 2020-04-08
Fix v1.4.0 doc rendering

rnpridgeon:


confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -250,9 +250,10 @@ def get_schema_id_by_subject(self, subject, schema):
 
 class SchemaRegistryClient(object):
     """
-    Schema Registry Client.
+    Schema Registry API Client.
```
edenhill, line 253: drop API, the API is the how, not the what.

confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -265,12 +266,12 @@ class SchemaRegistryClient(object):
     |                          |      | Path to client's private key                  |
     |                          |      | (PEM) used for authentication.                |
     | ssl.key.location         | str  |                                               |
-    |                          |      | *ssl.certificate.location must also be set.   |
+    |                          |      | ssl.certificate.location must also be set.    |
```
edenhill, line 269: These properties should probably have some form of fixed-width quoting.

docs/index.rst
```diff
@@ -3,185 +3,287 @@ The confluent_kafka API
 
 A reliable, performant and feature rich Python client for Apache Kafka v0.8 and above.
 
-Clients
+Client API 
+   - :ref:`AdminClient <pythonclient_adminclient>`
+   - :ref:`AvroProducer <avro_producer>`
```
edenhill, line 8: People will use these instead of the new Serializing producers/consumers, so I suggest adding " (legacy)" after these, and something like " (Avro, Protobuf, JSON, primitive types, etc)" or whatever catches the eye of the reader.

Also suggest pudding AvroPro and AvroCon at the bottom of this list.

docs/index.rst
```diff
@@ -3,185 +3,287 @@ The confluent_kafka API
...
 Supporting classes
     - :ref:`Message <pythonclient_message>`
     - :ref:`TopicPartition <pythonclient_topicpartition>`
+    - :ref:`SerializationContext <serde_ctx>`
+    - :ref:`Schema <schemaregistry_schema>`
+    - :ref:`RegisteredSchema <schemaregistry_registered_schema>`
+    - :ref:`SchemaRegistryError <schemaregistry_error>`
     - :ref:`KafkaError <pythonclient_kafkaerror>`
     - :ref:`KafkaException <pythonclient_kafkaexception>`
+    - :ref:`ConsumeError <serde_consumer_error>`
```
edenhill, line 40: It probably already does, but the ConsumeError should reside under confluent_kafka(.error?) and not under confluent_kafka.serializers.

docs/index.rst
```diff
@@ -304,3 +406,183 @@ neither the retriable or abortable flags set, as fatal.
...
+Kafka Client Configuration
+===========================
+
+Configuration of producer and consumer instances is performed by
+providing a dict of configuration properties to the instance constructor, e.g.
+
+.. code-block:: python
+
+  conf = {'bootstrap.servers': 'mybroker.com',
+          'group.id': 'mygroup', 'session.timeout.ms': 6000,
```
edenhill, line 421: session should be on its own line.

docs/index.rst
rnpridgeon: yes this is an artifact of when I had SerializingProducer/DeserializingConsumer under serialization in the TOC. Will fix

rnpridgeon: The ref tag that is *

confluent_kafka/deserializing_consumer.py
```diff
@@ -43,59 +43,57 @@ class DeserializingConsumer(_ConsumerImpl):
...
-        https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+    Notable DeserializingConsumer configuration properties(* indicates required field)
+
+    +----------------------+-----------------+-----------------------------------------------------+
+    | Property Name        | Type            | Description                                         |
+    +======================+=================+=====================================================+
+    | `bootstrap.servers*` | str             | Comma-separated list of brokers.                    |
+    +----------------------+-----------------+-----------------------------------------------------+
+    |                      |                 | Client group id string.                             |
+    | `group.id*`          | str             | All clients sharing the same group.id belong to the |
```
edenhill, line 54: nit: the splat should probably be outside the ticks as not to confuse it as part of the literal.
But no need to fix now.

confluent_kafka/deserializing_consumer.py
```diff
@@ -43,59 +43,57 @@ class DeserializingConsumer(_ConsumerImpl):
...
+    | `key.deserializer`   | callable        |                                                     |
+    |                      |                 | Deserializer used for message keys.                 |
+    +----------------------+-----------------+-----------------------------------------------------+
+    |                      |                 | Callable(SerializationContext, bytes) -> obj        |
+    | `value.deserializer` | callable        |                                                     |
+    |                      |                 | Deserializer used for message values.               |
+    +----------------------+-----------------+-----------------------------------------------------+
+    |                      |                 | Callable(KafkaError)                                |
+    |                      |                 |                                                     |
+    | error_cb             | callable        | Callback for generic/global error events. These     |
```
edenhill, line 67: These should be ticked as well.

The tick-no-tick rule is differentiate between normal descriptive text and programmatic literals that have meaning to the program and must be written in their exact form, such as configuration parameters, signatures, or anyhting really that relates to what ends up in the user code.

confluent_kafka/schema_registry/avro.py
```diff
@@ -95,9 +97,10 @@ class AvroSerializer(Serializer):
     i.e. {topic name}-{message field}
 
     Alternative naming strategies may be configured with the property
-    `subject.name.strategy`.
+    ``subject.name.strategy``.
```
edenhill, line 100: why is this double tick but it is single tick above?

docs/index.rst
```diff
@@ -3,185 +3,288 @@ The confluent_kafka API
...
+    - :ref:`ProtobufSerializer <schemaregistry_protobuf_serializer>`
 
+:ref:`Transactional Producer API <pythonclient_transactional>`
+
+:ref:`Kafka Client Configuration <pythonclient_configuration>`
 
 Supporting classes
     - :ref:`Message <pythonclient_message>`
     - :ref:`TopicPartition <pythonclient_topicpartition>`
+    - :ref:`SerializationContext <serde_ctx>`
```
edenhill, line 35: The schemas are not really used by the application, so dont think we need to include them in the TOC here.

confluent_kafka/deserializing_consumer.py
```diff
@@ -43,59 +43,57 @@ class DeserializingConsumer(_ConsumerImpl):
         All configured callbacks are served from the application queue upon
```
edenhill, line 43: nit: Typo in commit message, `Prepare doc rednering for v1.4.0 release`

confluent_kafka/schema_registry/json_schema.py
```diff
@@ -72,9 +73,10 @@ class JSONSerializer(Serializer):
     i.e. {topic name}-{message field}
 
     Alternative naming strategies may be configured with the property
-    `subject.name.strategy`.
+    ``subject.name.strategy``.
```
edenhill, line 76: dito

confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -252,41 +252,42 @@ class SchemaRegistryClient(object):
...
-    | basic.auth.user.info     | str  |                                               |
-    |                          |      | By default userinfo is extracted from         |
-    |                          |      | the URL if present.                           |
-    +--------------------------+------+-----------------------------------------------+
+    SchemaRegistryClient configuration properties(* indicates a required field):
+
+    +----------------------------+------+-------------------------------------------------+
+    | Property name              | type | Description                                     |
+    +============================+======+=================================================+
+    | url*                       | str  | Schema Registry URL.                            |
```
edenhill, line 260: missing ticks

confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -252,41 +252,42 @@ class SchemaRegistryClient(object):
...
+    +----------------------------+------+-------------------------------------------------+
+    |                            |      | Path to CA certificate file used                |
+    | `ssl.ca.location*`         | str  | to verify the Schema Registry's                 |
+    |                            |      | private key.                                    |
+    +----------------------------+------+-------------------------------------------------+
+    |                            |      | Path to client's private key                    |
+    |                            |      | (PEM) used for authentication.                  |
+    | `ssl.key.location`         | str  |                                                 |
+    |                            |      | ``ssl.certificate.location`` must also be set.  |
```
edenhill, line 269: double tick vs single tick

docs/index.rst
```diff
@@ -3,185 +3,288 @@ The confluent_kafka API
...
+    - :ref:`Schema <schemaregistry_schema>`
+    - :ref:`RegisteredSchema <schemaregistry_registered_schema>`
+    - :ref:`SchemaRegistryError <schemaregistry_error>`
     - :ref:`KafkaError <pythonclient_kafkaerror>`
     - :ref:`KafkaException <pythonclient_kafkaexception>`
+    - :ref:`ConsumeError <pyclient_error_consumer>`
```
edenhill, line 41: I would group these a bit so that the SR stuff comes last, not in the middle, since everything else relates to the Kafka clients.

docs/index.rst
rnpridgeon: If you extend the serializer it might which is why it's in supporting classes

docs/index.rst
rnpridgeon: ConsumeError is a KafkaClient error not a schema registry error.

confluent_kafka/schema_registry/schema_registry_client.py
rnpridgeon: double ticks for everyone

eb-emilio:
@rnpridgeon @edenhill hey, I see that `AvroConsumer` and `AvroProducer` were documented as Legacy on this PR, are they deprecated? they are still on the README as suggested way of consuming and producing using AVRO schemas, is there a better way to do it now? Thanks!

PR #828, opened 2020-04-09
v1.4.0 openssl, interceptor updates

rnpridgeon:


tools/build-openssl.sh
```diff
@@ -6,7 +6,7 @@
 
 # NOTE: Keep this updated to make sure we always build the latest
 #       version of OpenSSL in the 1.0 release train.
-OPENSSL_VERSION=1.0.2t
+OPENSSL_VERSION_SUFFIX=u
```
edenhill, line 9: Why? the suffix is not unique, only in combination with the version. revert this.

tools/build-openssl.sh
```diff
@@ -23,7 +23,7 @@ if ! grep -q "^VERSION=${OPENSSL_VERSION}$" build-openssl/Makefile ; then
     rm -rf build-openssl
     mkdir -p build-openssl
     pushd build-openssl
-    curl -s -l https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz | \
+    curl -s -l https://www.openssl.org/source/old/1.0.2/openssl-${OPENSSL_VERSION}.tar.gz | \
```
edenhill, line 26: Lessons learned from librdkafka; do `curl -fL <URL>`, otherwise it will not fail on 404.

tools/build-openssl.sh
```diff
@@ -23,7 +23,7 @@ if ! grep -q "^VERSION=${OPENSSL_VERSION}$" build-openssl/Makefile ; then
     rm -rf build-openssl
     mkdir -p build-openssl
     pushd build-openssl
-    curl -s -l https://www.openssl.org/source/old/1.0.2/openssl-${OPENSSL_VERSION}.tar.gz | \
+    curl -fl https://www.openssl.org/source/old/1.0.2/openssl-${OPENSSL_VERSION}.tar.gz | \
```
edenhill, line 26: `-fL`

PR #832, opened 2020-04-13
Fix failure to use fastavro when no Avro reader schema is given

woodlee:
When `MessageSerializer` is initialized without a `reader_schema`, the `to_json` method call on the None value at L178 results in an exception. Since this happens in a broad try/except block that captures all exceptions, it was leading to a silent failure to use the fastavro library in such cases. This is a quick fix for that.


woodlee:
[clabot:check]

woodlee:
Thanks @mparry , I had not realized this was s duplicate of nos. #694 and #773 . Closing this.

PR #837, opened 2020-04-16
Version v1.4.0.1rc

rnpridgeon:
The recent release(v1.4.0) tracks it's dependencies in requirements.txt files and references these files from the setup.py. These files do not appear to be included in the source distribution which breaks installation(from a source distribution).


confluent_kafka/src/confluent_kafka.h
```diff
@@ -41,8 +41,8 @@
  *  0xMMmmRRPP
  *  MM=major, mm=minor, RR=revision, PP=patchlevel (not used)
  */
-#define CFL_VERSION     0x01040000
-#define CFL_VERSION_STR "1.4.0"
```
edenhill, line 45: Suggest just bumping the micro version to 1.4.1, no need to add a revision.

confluent_kafka/src/confluent_kafka.h
```diff
@@ -41,8 +41,8 @@
...
+#define CFL_VERSION     0x01040100
```
edenhill, line 44: This is not consistent with the version string.

Issue #848, opened 2020-04-29
Protobuf Schema Registration fails at requests authentication


naoko:
Description
===========

when calling register_schema, it generates
```
site-packages/requests/models.py", line 549, in prepare_auth
    r = auth(self)
TypeError: 'list' object is not callable
```
because
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/schema_registry/schema_registry_client.py#L113
is list but expecting tuple

How to reproduce
================

conf = {"url": "<your-url>", "basic.auth.user.info": "<api_key>:<secret>"}
call SchemaRegistryClient(conf).register_schema(subject, schema)

after i change it the code from `self.session.auth = userinfo` to `self.session.auth = tuple(userinfo)` things worked.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): python 3.8, confluent-kafka==1.4.1
 - [x] Apache Kafka broker version: confluent cloud
 - [x] Operating system: Ubuntu
 - [x] Provide broker log excerpts

rnpridgeon:
https://github.com/confluentinc/confluent-kafka-python/pull/853

PR #853, opened 2020-05-03
requests auth property takes a tuple and not a list.

blown302:
Ran into the same issue as https://github.com/confluentinc/confluent-kafka-python/issues/848

Seems like it would be this simple to fix.

Not sure the protocol here for PRs on this repo so I'll open and let you comment.

Thanks!


blown302:
[clabot:check]

rnpridgeon:
Thanks Thomas!

PR #863, opened 2020-05-13
Don't send empty credentials to SR in Authorization Header

rnpridgeon:


tests/schema_registry/conftest.py
```diff
@@ -164,8 +169,28 @@ def __init__(self, conf):
...
 
+        adapter.add_matcher(self._auth_matcher)
         self._rest_client.session.mount('http://', adapter)
 
+    @classmethod
+    def _auth_matcher(cls, request):
+        headers = request._request.headers
+
+        # Pass request to downstream matchers
+        if 'Authorization' not in headers.keys():
```
edenhill, line 180: I'm not sure how this works under the hood, but this seems like it needs to scan keys() for 'Authorization', would suggest using `..keys().get('Authorization', None)`

tests/schema_registry/conftest.py
```diff
@@ -164,8 +169,28 @@ def __init__(self, conf):
...
+            return None
+
+        userinfo = headers.get('Authorization').split(" ")[1]
```
edenhill, line 183: And here's a second lookup. Just do one lookup.

edenhill: Inconsistent use of single/double quotes on the same line.

edenhill: Will throw an IndexError if the Authorization header does not contain a space, but maybe that's okay here.

tests/schema_registry/conftest.py
```diff
@@ -164,8 +169,28 @@ def __init__(self, conf):
...
+        if userinfo == b64encode(cls.USERINFO):
+            # Authenticated, pass downstream
+            return None
+
+        unauhtorized = {'error_code': 401,
```
edenhill, line 188: typo

tests/schema_registry/conftest.py
```diff
@@ -164,8 +169,28 @@ def __init__(self, conf):
...
+                        'message': "401 Unauthorized"}
+        return create_response(request=request,
+                               status_code=401,
+                               json=unauhtorized)
```
edenhill, line 192: typo

tools/RELEASE.md
```diff
@@ -147,12 +147,13 @@ Create a new virtualenv:
...
 Install the relevant package for your platform:
 
-    $ pip install dl-v0.11.4rc1/confluent_kafka-....whl
+    $ pip install --no-cache-dir --find-links dl-v0.11.4rc1/ confluent-kafka
 
 Verify that the package works, should print the expected Python client
 and librdkafka versions:
 
-    $ python -c 'import confluent_kafka as ck ; print "py:", ck.version(), "c:", ck.libversion()'
+    $ python -c 'import confluent_kafka as ck ; print "py: {} c: {}" \
```
edenhill, line 155: This will not work for python3

PR #868, opened 2020-05-16
configuration constants

blown302:
Liked how the java library had constants for common configuration parameters. Used the same names.


edenhill:
These need to be auto-generated like the .NET client does it.

PR #886, opened 2020-05-29
Add wheel metadata for additional libs (@beezz)

rnpridgeon:
Reviving the great work done by @beezz at #661. Manually tested with ` pex confluent_kafka -o test.pex` seemed to work as expected.

### DIFF
--- a/tools/fixup-wheels.sh
+++ b/tools/fixup-wheels.sh
@@ -12,7 +12,7 @@ set -e
 wheelhouse=$1
 fixed_wheelhouse=$2
 platform=$(uname -s)
-gen_record_row=$(pwd)/tools/generate-record-row.py
+gen_record_row=$PWD/tools/generate-record-row.py
 
 stagedir=$PWD/staging
 
--- a/tools/fixup-wheels.sh
+++ b/tools/fixup-wheels.sh
@@ -71,7 +71,7 @@ fixup_wheel_linux () {
         echo "$lib dependencies:"
         ldd $lib
 
-        ${gen_record_row} $lib confluent_kafka/.libs/ >> ../../$record_path
+        ${gen_record_row} $lib confluent_kafka/.libs/ >> ../../${record_path}
     done
 
     popd # confluent_kafka/.libs or confluent_kafka.libs
@@ -106,7 +106,7 @@ fixup_wheel_macosx () {
             otool -L $lib
         fi
 
-        ${gen_record_row} $lib confluent_kafka/.dylibs/ >> ../../$record_path
+        ${gen_record_row} $lib confluent_kafka/.dylibs/ >> ../../${record_path}
 
     done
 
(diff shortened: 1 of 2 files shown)


tools/fixup-wheels.sh
```diff
@@ -12,6 +12,7 @@ set -e
 wheelhouse=$1
 fixed_wheelhouse=$2
 platform=$(uname -s)
+gen_record_row=$(dirname $(readlink -f $0))/generate-record-row.py
```
edenhill, line 15: nit: readlink is a gnu extension and is not generally available. But since we're only doing this on linux and osx I guess we're fine?

tools/generate-record-row.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
```
edenhill, line 1: Suggest changing this to python3 since `python` is ambigious and not available on newer linux distros (ubuntu 20.04)

tools/generate-record-row.py
```diff
@@ -0,0 +1,58 @@
...
+additional libraries after the wheel has been build.
+"""
+import argparse
+import base64
+import csv
+import hashlib
+import os
+import sys
+
+if sys.version_info[0] < 3:
```
edenhill, line 15: .. and then we can skip this block too. just require python3

tools/fixup-wheels.sh
rnpridgeon: suggested alternative?

edenhill: Since we know the script is called from the top-level dir we can just do $PWD/tools/generate-record-row.py

tools/generate-record-row.py
rnpridgeon: ack that will require bumping the version used to create wheels as well which we should also do. The question being should we do so here or in a follow up PR that bumps the interpreter version for our manylinux builds.

edenhill: We can skip it in this PR, but we'll need to address it soon since none of our py scripts work out of the box on when executed through shebang as `examples/eos-transactions.py` on ubuntu 20.04 (and soon others).

rnpridgeon: agreed, cibuildwheel implies it will stop working with py2 as the execution environment in its release notes as well which will leave us with our hands tied. I'll get this in to 1.4.3 then work start attacking the migration

tools/generate-record-row.py
rnpridgeon: ack deferring to py3 build env migration

tools/fixup-wheels.sh
```diff
@@ -12,6 +12,7 @@ set -e
 wheelhouse=$1
 fixed_wheelhouse=$2
 platform=$(uname -s)
+gen_record_row=$(pwd)/tools/generate-record-row.py
```
edenhill, line 15: PWD?

lexn82:
Apologies, I just noticed this PR.

I fixed the same issue by running code from `wheel` package directly to rebuild the wheel [here](https://github.com/confluentinc/confluent-kafka-python/pull/896). Please consider it as an alternative approach.

edenhill:
Closing in favour of less hacky #896

PR #891, opened 2020-06-08
Example Alpine-based docker image

edenhill:


edenhill:
Sticking this in the examples dir voids any guarantees, but simply providing functional examples.

PR #895, opened 2020-06-10
Update README.md

smedegaard:
add paragraph about 0.8 not supporting Consumer.subscribe()

might save some other poor noob some time

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,8 @@ If you use Kafka broker 0.9 or 0.8 you must set
 `broker.version.fallback` to your broker version,
 e.g `broker.version.fallback=0.9.0.1`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -257,6 +257,8 @@ If you use Kafka broker 0.9 or 0.8 you must set
 `broker.version.fallback` to your broker version,
 e.g `broker.version.fallback=0.9.0.1`.
 
+Please note that brokers prior to version 0.9 does not support `Consumer.subscribe()`. If your server is version 0.8.x, please use [`Consumer.assign()`](https://docs.confluent.io/current/clients/confluent-kafka-python/#confluent_kafka.Consumer.assign) instead.
```
edenhill, line 260: This list of unsupported APIs in earlier broker versions will grow long. I'd prefer if this was just mentioned in each method's doc string.

The client should also raise an error (if an error_cb is set) when this happens, did it?

smedegaard: yeah, I agree

It did raise an error. I don't remember which on exactly. 

It think it said that JoinGroup was not supported and had a hint suggesting that the `api.version.request` was not set to `false` or `broker.version.fallback` was not set to the correct value

Issue #482, opened 2018-11-07
PyPI wheel for 0.11.6 (Python 3.6 x86-64, manylinux) causes BadWheelFile


vtermanis:
Description
===========
At least one of the wheels for version `0.11.6` on PyPI seems to contain an invalid hash. The previous version (`0.11.5`) for the below mentioned platform etc. does not seem to have said problem.

- `confluent_kafka==0.11.6` (via PyPI)
- Ubuntu 16.04, Python 3.6.2, x86-64
- `pip==18.1,wheel=0.31.1,setuptools=40.5.0`

How to reproduce
================
**Procedure**
```shell
wget https://files.pythonhosted.org/packages/dd/0a/c8bbf50a0b8e1d623521565e4c3211aebf7c9a7cf7eeb35b5360132a0ccc/confluent_kafka-0.11.6-cp36-cp36m-manylinux1_x86_64.whl
# Alternatively run (assuming have Linux x86-64 & CPython 3.6):
pip wheel --no-cache -w . confluent_kafka==0.11.6

md5sum confluent_kafka-0.11.6-cp36-cp36m-manylinux1_x86_64.whl 
72bf2e29ab1608e49bbc36d0a51546a8  confluent_kafka-0.11.6-cp36-cp36m-manylinux1_x86_64.whl

# Trigger wheel check
wheel unpack -d /tmp confluent_kafka-0.11.6-cp36-cp36m-manylinux1_x86_64.whl
```

**Actual outcome**
Wheel unpack/install fails with:
`wheel.install.BadWheelFile: Bad hash for file 'confluent_kafka/cimpl.cpython-36m-x86_64-linux-gnu.so`

**Expected outcome**
Wheel unpack/install succeeds.

rnpridgeon:
Thanks for reporting this @vtermanis. This seems related to the changes we made to include confluent monitoring interceptors. I'll dig into it a bit more next week to.

switchtrue:
This also appears to be occurring with the `confluent_kafka-1.0.0-cp34-cp34m-manylinux1_x86_64.whl` any progress on this?

azfaraziz:
I am getting this issue on the latest version `1.3.0` I tried earlier version up to `1.0.0` and still having the same error
`wheel.install.BadWheelFile: No expected hash for file 'confluent_kafka/.dylibs/monitoring-interceptor.dylib'`

Any updates on this?

shanwei8:
same here.

[...]

PR #896, opened 2020-06-12
Rebuild the package with wheel instead of zip for correct signatures.

lexn82:
Fixes [issue-482](https://github.com/confluentinc/confluent-kafka-python/issues/482).

Instead of re-zipping the files after fixing up binaries, we use the `WheelFile` from `bdist_wheel` package to rebuild the wheel properly.

This works as-is if the wheel package is available to the system python interpreter, otherwise we may need a little more scripting to setup venv and `pip install wheel`.


lexn82:
[clabot:check]

tools/fixup-wheels.sh
```diff
@@ -117,9 +117,10 @@ fixup_wheel () {
         fixup_wheel_macosx
     fi
 
-    zip -r $fixed_whl .
-
     popd # tmpdir
+
+    python -c "from wheel.wheelfile import WheelFile; w = WheelFile('$fixed_whl', mode='w'); w.write_files('$tmpdir'); w.close()"
```
edenhill, line 122: Can we move this little scriptlet to its own file in this directory to avoid quoting problems if $fixed_whl or $tmpdir contains special character?

lexn82: Fixed.

tools/fixup-wheels.sh
```diff
@@ -117,9 +117,10 @@ fixup_wheel () {
         fixup_wheel_macosx
     fi
 
-    zip -r $fixed_whl .
-
     popd # tmpdir
+
+    python rebuild-wheel.py --staging-dir $tmpdir --wheel-file $fixed_whl
```
lexn82, line 122: I'm not sure what's the best way to define the path to the `rebuild-wheel.py` script. Should this be `tools/rebuild-wheel.py` or do you normally anchor the path based on the shell script location?

edenhill: I believe this script is run from the top-level project directory, so you can assume that being the case and reference rebuild-wheel.py through a relative path.

tools/rebuild-wheel.py
```diff
@@ -0,0 +1,23 @@
+from wheel.wheelfile import WheelFile
```
edenhill, line 1: Add `#!/usr/bin/env python3` shebang and copyright

edenhill: Does this add a requirement on a wheel package? If so, add to tools/requirements.txt

lexn82: Added the wheel package to `requirements-manylinux.txt`

lexn82:
Looks like the `wheel` package is not intended to be used as a library so it does not provide a stable API (see docs [here](https://pypi.org/project/wheel/)). However, it turns out there is an even easier way to package a directory with wheel using CLI.

edenhill:
Superb, thank you!

PR #909, opened 2020-07-07
Add example listing committed offsets and lag

edenhill:


examples/list_offsets.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# Show committed offsets and current lag for group and topic(s).
+#
+
+
+import sys
+import confluent_kafka
+
+
+if len(sys.argv) < 4:
```
rnpridgeon, line 27: Why not just use argparse like the existing examples?

edenhill: Keeping it simple.
4 examples use sys.argv, 6 use argparse.

edenhill:
> how one might sidestep limitations in the admin client.

That's a much bigger task, including adding new APIs to librdkafka and the Python client, so until then we need to explain how to go about getting this information using what is available, since we get this question quite frequently.

PR #912, opened 2020-07-08
Correct SerializingProducer Exception doc

rnpridgeon:


rnpridgeon:
*indentation was also off

Issue #913, opened 2020-07-12
Give developers ability to pre-load and check Avro registry schema IDs.


arthurlm:
# Description

Give developers ability to pre-load and check Avro registry schema IDs.

# Details description

Problem is on function `AvroSerializer.__call__(self, obj, ctx)`.

Currently it does the serialization __and__ loading data from registry.

This results in really poor performances when inserting the first message into Kafka.
It is particularly visible in applications which restart often and does not produce messages frequently.

Second problem is when `auto.register.schemas` equals `False`.
The app crash only on first message insertion and not at init.

Moreover (I have not tested it on my own), there might also be an error when same serializer instance is used for multiple topics.
Problem may comes from having only one `self._schema_id` per `AvroSerializer`.

# Example usage

Load / init the app (non critical section):

```python
avro_serializer = AvroSerializer(json.dumps(schema), schema_registry)

# Load registry info at init (feature request)
avro_serializer.load_registry_schema_id(SerializationContext(topic_name1, MessageField.VALUE))
avro_serializer.load_registry_schema_id(SerializationContext(topic_name2, MessageField.VALUE))
avro_serializer.load_registry_schema_id(SerializationContext(topic_name3, MessageField.KEY))

producer = SerializingProducer({
    'bootstrap.servers': ...,
    'key.serializer': avro_serializer,
    'value.serializer': avro_serializer,
})
```

Critical section (app need constant performances):

```python
producer.produce(...)
```

# Example of implementation

See below a possible implementation for the new `__call__` function.
Please note that I have not run tests suite on it.

```python
class AvroSerializer(Serializer):
    ...

    def load_registry_schema_id(self, ctx):
        subject = self._subject_name_func(ctx, self._schema_name)

        # Check to ensure this schema has been registered under subject_name.
        if self._auto_register and subject not in self._known_subjects:
            # The schema name will always be the same. We can't however register
            # a schema without a subject so we set the schema_id here to handle
            # the initial registration.
            self._schema_ids[subject] = self._registry.register_schema(subject,
                                                                       self._schema)
            self._known_subjects.add(subject)
        elif not self._auto_register and subject not in self._known_subjects:
            registered_schema = self._registry.lookup_schema(subject,
                                                             self._schema)
            self._schema_ids[subject] = registered_schema.schema_id
            self._known_subjects.add(subject)

        return self._schema_ids[subject]

    def __call__(self, obj, ctx):
        if obj is None:
            return None

[...]

PR #914, opened 2020-07-14
Add AvroSerializer ability to pre load schema ID + Fix bug when the same serializer is used for multiple topics

arthurlm:
See issue #913 

# Changes to MockSchemaRegistryClient

1) ID changes

Do not use constant `SCHEMA_ID` but generate constant IDs based on subject name.

Before this change, it was impossible to check if cache is correctly used per topic.
Moreover, in real registry, IDs are different per subject.

2) Tests reset changes

Reset `MockSchemaRegistryClient.counter` before each tests.

# Changes to AvroSerializer

1) Fix bug when the same serializer is used for multiple topics
2) Add ability to force pre loading of registry schema ID

### DIFF
--- a/tests/schema_registry/test_avro_serializer.py
+++ b/tests/schema_registry/test_avro_serializer.py
@@ -15,6 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from struct import unpack
+from io import BytesIO
+
 import pytest
 
 from confluent_kafka.schema_registry import (record_subject_name_strategy,
@@ -23,6 +26,8 @@ from confluent_kafka.schema_registry import (record_subject_name_strategy,
 from confluent_kafka.schema_registry.avro import AvroSerializer
 from confluent_kafka.serialization import (MessageField,
                                            SerializationContext)
+
+from .conftest import find_schema_id
 
 # MockSchemaRegistryClient, see ./conftest.py for additional details.
 TEST_URL = 'http://SchemaRegistry:65534'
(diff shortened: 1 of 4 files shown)


arthurlm:
[clabot:check]

tests/schema_registry/test_avro_serializer.py
```diff
@@ -74,6 +79,62 @@ def test_avro_serializer_config_auto_register_schemas_false(mock_schema_registry
...
+    """
+    conf = {'url': TEST_URL}
+    test_client = mock_schema_registry(conf)
+    topic1 = "test-topic1"
+    topic2 = "test-topic2"
+
+    test_serializer = AvroSerializer("string", test_client,
+                                     conf={'auto.register.schemas': False})
+
+    test_serializer.load_registry_schema_id(SerializationContext(topic1, MessageField.KEY))
```
mhowlett, line 123: i haven't looked at this closely, but given the existing design at least, the capability to pre-cache a schema seems more like it belongs in the (caching) schema registry client, not the serializer.

Also, the `SerializationContext` is intended to be used for what it's name implies - context whilst (de)serializing, so it shouldn't be part of the interface otherwise.

arthurlm: Looking at your comment, indeed the `load_registry_schema_id` should not be part of the interface.

I have look more deeply into the `SchemaRegistryClient`.
There is indeed a cache into the `register_schema` function used when `'auto.register.schemas': True`.
But there is not in `lookup_schema` function used when `'auto.register.schemas': False`.

This is where my miss understanding comes from.

---

To go forward:

1) I can update my confluent_kafka usage to `register_schema` and force pre-loading.
2) If you are still interested by the others changes: I can revert my last commit (related to making public the usage of `load_registry_schema_id`). Otherwise you can close the PR.
3) It may still be interesting to add a cache into the `lookup_schema` and remove the one from the `Serializer` object.

Issue #945, opened 2020-09-02
Naming of asyncio.py example causes circular import


feocco:
Description
===========
The current name of the asyncio example is `asyncio.py` which matches the library `import asyncio`. This causes a circular import error.

As a result, the example doesn't run unless the file is renamed.

Could be helpful to rename the example. But, this can at least serve as a searchable reference otherwise.

How to reproduce
================
```
# venv steps
virtualenv -p python3 venv
source venv/bin/activate

# Install Libraries
pip install confluent-kafka
pip install pydantic
pip install fastapi
pip install uvicorn

# Run asyncio.py
python asyncio.py

# Receive Error
# AttributeError: partially initialized module 'asyncio' has no attribute 'iscoroutinefunction' (most likely due to a circular import)

# Change file name & run
mv asyncio.py asyc.py
python asyc.py
# INFO:     Started server process [77966]
# INFO:     Waiting for application startup.
# INFO:     Application startup complete.
# INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
```

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): 1.5.0
 - [x] Apache Kafka broker version: cp-server:5.4.1
 - [x] Client configuration: `{"bootstrap.servers": "localhost:12091"}`
 - [x] Operating system: OSX
 - [x] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

PR #946, opened 2020-09-03
Rename asyncio.py example to avoid circular import (#945)

edenhill:


mhowlett:
thanks for doing this.

mhowlett (approved):

PR #950, opened 2020-09-15
Change schema registry to work without avro

jaysonsantos:
These errors are being used just to export avro's exceptions but this means that you can never use schema-registry if you installed confluent kafka with `confluent-kafka[schema-registry]`

### DIFF
--- a/confluent_kafka/schema_registry/error.py
+++ b/confluent_kafka/schema_registry/error.py
@@ -15,7 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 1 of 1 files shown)


jaysonsantos:
[clabot:check]
it would be nice if you guys used https for the cla page

mhowlett (approved):
yeah, this is bad. solution is not ideal, but resolves the problem and we can back it out without consequence if we want, eventually doing something better.

mhowlett:
for some reason which i don't understand this is showing a conflict, and not allowing this to be resolved here, so I opened again in #1055 and merged that.

PR #959, opened 2020-09-23
Update load.py

mak600d:
fix the bug module 'avro.schema' has no attribute 'Parse'

### DIFF
--- a/confluent_kafka/avro/load.py
+++ b/confluent_kafka/avro/load.py
@@ -26,7 +26,7 @@ def loads(schema_str):
         if sys.version_info[0] < 3:
             return schema.parse(schema_str)
         else:
-            return schema.Parse(schema_str)
+            return schema.parse(schema_str)
     except schema.SchemaParseException as e:
         raise ClientError("Schema parse failed: %s" % (str(e)))
 


mak600d:
> It looks like @mak600d hasn't signed our **C**ontributor **L**icense **A**greement, yet.
> 
> > The purpose of a CLA is to ensure that the guardian of a project's outputs has the necessary ownership or grants of rights over all contributions to allow them to distribute under the chosen licence.
> > [Wikipedia](http://en.wikipedia.org/wiki/Contributor_License_Agreement)
> 
> You can read and sign our full Contributor License Agreement [here](http://clabot.confluent.io/cla).
> 
> Once you've signed reply with `[clabot:check]` to prove it.
> 
> Appreciation of efforts,
> 
> clabot

[clabot:check]

jeeftor:
I need this change!!

benjaminxscott:
Having run into this issue, I believe it's due to a recent change with the `avro` dependency which deprecates `avro-python3` in favor of [`avro`](https://pypi.org/project/avro/)

https://avro.apache.org/docs/1.10.0/gettingstartedpython.html

In this new paradigm, I believe this codepath could call `parse()` without having to switch on the python ver

Noting also that the `if` statement in this patch is a no-op

jeeftor:
I cam up with a very bad fix which was rewriting the sys.version to 2 instead of 3... PURE HACK

mmartinSophos (approved):

mhowlett:
thanks for reporting @jeeftor and sorry for the delay in actioning. this was later resolved by #1006 which I just merged.

Issue #459, opened 2018-10-04
Timing issue with create_topics


ivonindza:
Description
===========
I'm writing some integration tests for our code and doing create_topics/delete_topics in test set_up/tear_down. Code is something like this:
```
def kafka_create_topics(topics: Iterable[str]):
    new_topics = [NewTopic(t, num_partitions=1, replication_factor=1) for t in topics]
    admin = AdminClient({'bootstrap.servers': ...})

    futures = admin.create_topics(new_topics)
    for f in futures.values():
        f.result()

    time.sleep(1)
    while True:
        cluster_topics = admin.list_topics().topics
        if all(t in cluster_topics for t in topics):
            break
```

Even after this whole thing, the testing code that follows is not guaranteed to encounter created topics and occasionally fails with KafkaError.UNKNOWN_TOPIC_OR_PART. Any idea why this would happen?

How to reproduce
================

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): ('0.11.5', 722176), ('0.11.5', 722431)
 - [x] Apache Kafka broker version: 2.11-1.1.0
 - [ ] Client configuration: `{...}`
 - [x] Operating system: linux
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
by default, the broker will return a response to the create_topics call without waiting for the operation to complete (i.e. be completely synced across other brokers / zookeeper). to make the broker block until topic creation has completed, you can set the operation_timeout option parameter to a value higher than 0.

https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/admin/__init__.py#L299

ivonindza:
Hi, sorry for the slow reply. This doesn't resolve my issue. I am still unable to reliably create/delete topics through AdminClient.

[...]

PR #963, opened 2020-10-01
Upgrade fastavro to a minimum of 1.0.0.post1

mrooding:
We're using confluent-kafka-python and running into an issue in fastavro [#459](https://github.com/fastavro/fastavro/pull/459). This is fixed in 0.24.1.

I've updated fastavro to depend on at least the latest version (1.0.0.post1). Since the requirements.txt was already depending on at least version 0.23.0 it would've pulled in the latest version anyway for the next release. Do note that version 1.0.0 drops support for Python 2. If this is undesirable for confluent-kafka-python, I can change the fastavro dependency to be fixed to 0.24.2.

### DIFF
--- a/confluent_kafka/schema_registry/requirements.txt
+++ b/confluent_kafka/schema_registry/requirements.txt
@@ -1,4 +1,4 @@
-fastavro>=0.23.0
+fastavro>=1.0.0.post1
 jsonschema
 protobuf
 requests


mrooding:
[clabot:check]

mrooding:
In hindsight I noticed that you're using setuptools and that setup.py already specifies fastavro for both Python 2.7 and 3 (as of 77681916850ec5902dd0b9f270f8274d73d81b25). My change thus only affects the local requirement.txt and is probably obsolete. I'll close the PR

PR #965, opened 2020-10-01
WIP: Continue work on adding Python 3.8 support for Windows

dewert:
I need a 3.8 wheel for Windows, so I thought I'd give it a try here instead of just setting up a local build


dewert:
[clabot:check]

PR #978, opened 2020-10-29
Preps for v1.5.2 release

edenhill:
And restructured the source tree to avoid pytest/tox to load confluent-kafka from the source directory rather than what it has installed in its virtual environment.


tox.ini
```diff
@@ -11,38 +11,18 @@ passenv =
...
-[testenv:py35]
-deps =
-    {[base]deps}
-    avro-python3
-
-[testenv:py36]
-deps =
-    {[base]deps}
-    avro-python3
+    -rtests/requirements.txt
```
rnpridgeon, line 25: Missing space

rnpridgeon (approved):
Nit in tox.ini other than that LGTM

tox.ini
edenhill: That's actually required to be without space.

rnpridgeon: TIL

Issue #968, opened 2020-10-14
Confluent kafka does not support python 3.9


MikkelHJuul:
Description
===========
confluent kafka cannot be installed into docker image python:3.9-slim-buster (the problem is probably more widespread than that)
works:
```
FROM python:3.8-slim-buster

RUN pip install confluent_kafka
```
does not work
```
FROM python:3.9-slim-buster

RUN pip install confluent_kafka
```
throws errors on missing gcc
when after installing gcc, it throws errors on librdkafka, then after installing that you try installing confluent_platform, and it still doesn't work.

How to reproduce
================
Dockerfile
```Dockerfile
FROM python:3.8-slim-buster

RUN pip install confluent_kafka

FROM python:3.9-slim-buster

RUN pip install confluent_kafka
```
>/ docker build .

I have tried installing `gcc`, `librdkafka-dev` and `confluent_platform` current versions as I write (in docker image using `apt`)
Attachment: log output of stated dockerfile run with gcr.io/kaniko using a gitlab-runner

At least there should be some kind of note of this limitation in install descriptions, until it is fixed. 

[error.log](https://github.com/confluentinc/confluent-kafka-python/files/5377152/error.log)

dangreenisrael:
This happened to me. I'm putting this here in care others are googling for a similar error:

```
 #include <librdkafka/rdkafka.h>
 ^~~~~~~~~~~~~~~~~~~~~~
 compilation terminated.
 error: command '/usr/bin/gcc' failed with exit code 1
```

dangreenisrael:
If you need a quick fix to get yourself unblocked for on 3.9, you could build the binaries from source. 

The following works:

```dockerfile
# Install librdkafka from source
RUN git clone https://github.com/edenhill/librdkafka
WORKDIR /app/librdkafka/
RUN ./configure
RUN make
RUN make install
RUN ldconfig
WORKDIR /app
```

dangreenisrael:
@edenhill 

https://github.com/confluentinc/confluent-kafka-python/pull/987

olasd:
As far as I can tell and I've tested locally, confluent-kafka-python works fine with Python 3.9. You just need a recent enough version of librdkafka. However, the version of librdkafka-dev **packaged in Debian buster** (which is what you've seemingly tried to installed in the docker python3.9 image) is much too old for confluent-kafka-python 1.5.0.

In light of this I don't think #987 should be merged.

[...]

PR #981, opened 2020-11-06
Note that Python 3.9 is broken

dangreenisrael:
Python 3.9 is busted for wheel installs

See issue: https://github.com/confluentinc/confluent-kafka-python/issues/968

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ client's CA location with `'ssl.ca.location': certifi.where()`.
 Prerequisites
 =============
 
- * Python >= 2.7 or Python 3.x
+ * Python >= 2.7, >=3.9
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -284,7 +284,7 @@ client's CA location with `'ssl.ca.location': certifi.where()`.
 Prerequisites
 =============
 
- * Python >= 2.7 or Python 3.x
+ * Python >= 2.7, >=3.9
```
edenhill, line 287: If Py 3.9 is broken then this should be <3.9, but we'll want to fix it instead.

dangreenisrael: Whoops. Typo.

Agreed about fixing it instead, but I think for the time being a note that it is broken with 3.9 would be save people a lot of heartache

dangreenisrael:
Superseded by: https://github.com/confluentinc/confluent-kafka-python/pull/987

PR #992, opened 2020-11-20
Incremental rebalancing

mhowlett:
FYI: current status. testing is the "only" thing left iirc.


src/confluent_kafka/src/Consumer.c
```diff
@@ -97,8 +103,8 @@ static PyObject *Consumer_subscribe (Handle *self, PyObject *args,
 					 PyObject *kwargs) {
 
 	rd_kafka_topic_partition_list_t *topics;
-	static char *kws[] = { "topics", "on_assign", "on_revoke", NULL };
-	PyObject *tlist, *on_assign = NULL, *on_revoke = NULL;
+	static char *kws[] = { "topics", "on_assign", "on_revoke", "on_lost", NULL };
```
edenhill, line 106: Why do we want a specific callback for on_lost in the Python client, we're not doing that anywhere else, are we?

src/confluent_kafka/src/Consumer.c
```diff
@@ -206,6 +227,35 @@ static PyObject *Consumer_unsubscribe (Handle *self,
 }
 
 
+static PyObject *Consumer_incremental_assign (Handle *self, PyObject *tlist) {
+
+	rd_kafka_topic_partition_list_t *c_parts;
+	rd_kafka_error_t *error;
+
+        if (!self->rk) {
```
edenhill, line 235: mix of whitespace and tabs, stick to ws for new code.
Sorry!

src/confluent_kafka/src/Consumer.c
```diff
@@ -1039,13 +1119,20 @@ static PyMethodDef Consumer_methods[] = {
 	  "  :param callable on_revoke: callback to provide handling of "
 	  "offset commits to a customized store on the start of a "
 	  "rebalance operation.\n"
+	  "  :param callable on_lost: callback to provide handling in "
+	  "the case a partition assignment has been lost. If not "
+	  "specified, lost partition events will be delivered to "
+	  "on_revoke, if specified. Partitions that have been lost may "
+	  "already be owned by other members in the group and therefore "
+	  "commiting offsets, for example, may fail.\n"
```
edenhill, line 1127: committing.

And should the recommendation be not to commit offsets perhaps?

src/confluent_kafka/src/Consumer.c
```diff
@@ -1113,13 +1200,33 @@ static PyMethodDef Consumer_methods[] = {
...
 	  "\n"
 	},
+	{ "incremental_assign", (PyCFunction)Consumer_incremental_assign, METH_O,
+	  ".. py:function:: incremental_assign(partitions)\n"
+	  "\n"
+	  "  Incrementally add the provided list of :py:class:`TopicPartition` "
+          "to the current partition assignment.\n"
+	  "\n"
+	  "  :param list(TopicPartition) partitions: List of topic+partitions and optionally initial offsets to start consuming.\n"
+          "  :raises: RuntimeError if called on a closed consumer\n"
```
edenhill, line 1210: or KafkaException, see unassign() below, and update it for assign() too.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1113,13 +1200,33 @@ static PyMethodDef Consumer_methods[] = {
           "  :raises: RuntimeError if called on a closed consumer\n"
 	  "\n"
 	},
+	{ "incremental_assign", (PyCFunction)Consumer_incremental_assign, METH_O,
+	  ".. py:function:: incremental_assign(partitions)\n"
+	  "\n"
+	  "  Incrementally add the provided list of :py:class:`TopicPartition` "
+          "to the current partition assignment.\n"
```
edenhill, line 1207: Maybe beef this up a bit more with the constraints, e.g., no dups, etc.
Same for incr_unassign

src/confluent_kafka/src/Consumer.c
```diff
@@ -1314,15 +1421,25 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
 				   void *opaque) {
 	Handle *self = opaque;
 	CallState *cs;
+        rd_kafka_resp_err_t assign_err = RD_KAFKA_RESP_ERR_NO_ERROR;
+        rd_kafka_error_t *error = NULL;
+        char* assign_errstr = NULL;
```
edenhill, line 1426: Splat goes on the variable, not the type.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1341,10 +1458,16 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
...
 		}
 
-		result = PyObject_CallObject(
-			err == RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS ?
-			self->u.Consumer.on_assign :
-			self->u.Consumer.on_revoke, args);
+                if (err == RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS)
+                        cb = self->u.Consumer.on_assign;
+                else {
+                        if (rd_kafka_assignment_lost(rk) &&
```
edenhill, line 1464: nit: This nested if doesn't provide anything apart from intend, do else if + else

src/confluent_kafka/src/Consumer.c
```diff
@@ -1341,10 +1458,16 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
...
+                            self->u.Consumer.on_lost)
+                                cb = self->u.Consumer.on_lost;
+                        else
+                                cb = self->u.Consumer.on_revoke;
+                }
+		result = PyObject_CallObject(cb, args);
```
edenhill, line 1470: nit: reinsert the empty line for readability.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1356,17 +1479,67 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
 		}
 	}
 
+        if (!strcmp(rd_kafka_rebalance_protocol(rk), "COOPERATIVE")) {
```
edenhill, line 1482: Can return NULL (even though it shouldnt in this case, but better be careful, so assign to a temp var.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1356,17 +1479,67 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
...
+                if (self->u.Consumer.rebalance_assigned) {
+                        assign_errstr = "assign() must not be called during a "
```
edenhill, line 1484: This already enforced by librdkafka, don't need to do it here.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1356,17 +1479,67 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
...
+                if (self->u.Consumer.rebalance_incremental_assigned ||
+                    self->u.Consumer.rebalance_incremental_unassigned) {
+                        assign_errstr = "incremental_(un)assign must not be "
+                                        "called during an EAGER protocol "
+                                        "rebalance";
+                }
+        }
+
+        if (assign_errstr) {
+                cfl_PyErr_Format(RD_KAFKA_RESP_ERR__FAIL, assign_errstr);
```
edenhill, line 1505: Whytespaces, not tabbs

edenhill: Must pass format string to format functions.

edenhill: otoh this code is not needed at all, see previous comment about enforcement in the librd of kafka

src/confluent_kafka/src/Consumer.c
```diff
@@ -1356,17 +1479,67 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
...
+                                                                      c_parts);
+                } else {
+		        if (err == RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS)
+			        assign_err = rd_kafka_assign(rk, c_parts);
+		        else
+			        assign_err = rd_kafka_assign(rk, NULL);
+                }
 	}
 
+	if (assign_err) {
```
edenhill, line 1533: Keep this inside the scope where it is used, including the variable defs, don't think this will ever be true if the fallback hasn't been used.

edenhill (changes requested):

edenhill:
Missing:
 - Tests (unit and integration, the latter using Ryan's new integration test suite, not integration_tests.py)
 - CHANGELOG.md update
 - Beef up the docs to clarify incremental usage is required when using a cooperative assignor
 - Add rebalance_protocol()
 - Add example

src/confluent_kafka/src/Consumer.c
mhowlett: the only client I know of without a specific callback for lost partitions is the one you made :-).

src/confluent_kafka/src/Consumer.c
mhowlett: i think that is implicit. btw: i copy/pasted this sentence from the go impl. that you did.

mhowlett:
thanks for the early review - opened as a draft just so you could see where it was at.

src/confluent_kafka/src/Consumer.c
mhowlett: it goes further, please also be sorry about 8 space tabs

src/confluent_kafka/src/Consumer.c
mhowlett: yep. i just copy/pasted this from assign.

src/confluent_kafka/src/Consumer.c
mhowlett: yep, absolutely should.

src/confluent_kafka/src/Consumer.c
mhowlett: nit: the structure I have better expresses the logic here imo, so preferred as being easier to read even though this could be written out more tersely as you suggest.

src/confluent_kafka/src/Consumer.c
mhowlett: ? 

what would improve readability is coding convention that specifies {} must always used (as is enforced in all new language these days).

src/confluent_kafka/src/Consumer.c
mhowlett: ok

src/confluent_kafka/src/Consumer.c
mhowlett: ... yep. whoever implemented that was a thinker.

src/confluent_kafka/src/Consumer.c
mhowlett: yep, it's all enforced in librdkafka. thanks for pointing that out ...

src/confluent_kafka/src/Consumer.c
mhowlett: good point

src/confluent_kafka/admin/__init__.py
```diff
@@ -87,28 +83,37 @@ def __str__(self):
...
+        BROKER = RESOURCE_BROKER  #: Broker resource. Resource name is broker id.
 
     def __init__(self, restype, name,
                  set_config=None, described_configs=None, error=None):
         """
         :param ConfigResource.Type restype: Resource type.
-        :param str name: Resource name, depending on restype.
-                          For RESOURCE_BROKER the resource name is the broker id.
-        :param dict set_config: Configuration to set/overwrite. Dict of str, str.
+        :param str name: The resource name, which depends on restype.
```
edenhill, line 114: Adding "The" doesn't really add anything here (and elsewhere), and is added inconsistently.

src/confluent_kafka/admin/__init__.py
```diff
@@ -187,21 +193,21 @@ class AdminClient (_AdminClientImpl):
...
     change in future versions of the client.
 
     See examples/adminapi.py for example usage.
 
-    For more information see the Java Admin API documentation:
-    https://docs.confluent.io/current/clients/javadocs/org/apache/kafka/clients/admin/package-frame.html
+    For more information see the `Java Admin API documentation
+    <https://docs.confluent.io/current/clients/javadocs/org/apache/kafka/clients/admin/package-frame.html>`_.
 
-    Requires broker version v0.11.0.0 or later.
```
edenhill, line 204: nit: the release was actually named v0.11.0.0

src/confluent_kafka/admin/__init__.py
```diff
@@ -291,22 +297,23 @@ def _make_futures(futmap_keys, class_check, make_result_fn):
...
+        :param float operation_timeout: The operation timeout in seconds,
                   controlling how long the CreateTopics request will block
                   on the broker waiting for the topic creation to propagate
                   in the cluster. A value of 0 returns immediately. Default: 0
-        :param float request_timeout: Set the overall request timeout in seconds,
+        :param float request_timeout: The overall request timeout in seconds,
                   including broker lookup, request transmission, operation time
                   on broker, and response. Default: `socket.timeout.ms*1000.0`
-        :param bool validate_only: Tell broker to only validate the request,
+        :param bool validate_only: If true, the request is validated only,
```
edenhill, line 311: This reads weird now. "the request is only validated without creating the topic"

src/confluent_kafka/admin/__init__.py
```diff
@@ -387,23 +394,22 @@ def create_partitions(self, new_partitions, **kwargs):
...
                   per call since these resource requests must be sent to the
                   broker specified in the resource.
 
-        :param list(ConfigResource) resources: Resources to get configuration for.
-        :param float request_timeout: Set the overall request timeout in seconds,
+        :param list(ConfigResource) resources: Resources to get the configuration for.
+        :param float request_timeout: The overall request timeout in seconds,
                   including broker lookup, request transmission, operation time
                   on broker, and response. Default: `socket.timeout.ms*1000.0`
-        :param bool validate_only: Tell broker to only validate the request,
```
edenhill, line 403: :+1:

src/confluent_kafka/admin/__init__.py
```diff
@@ -439,14 +443,16 @@ def alter_configs(self, resources, **kwargs):
...
                   broker specified in the resource.
 
-        :param list(ConfigResource) resources: Resources to update configuration for.
-        :param float request_timeout: Set the overall request timeout in seconds,
+        :param list(ConfigResource) resources: Resources to update configuration of.
+        :param float request_timeout: The overall request timeout in seconds,
                   including broker lookup, request transmission, operation time
                   on broker, and response. Default: `socket.timeout.ms*1000.0`.
-        :param bool validate_only: Tell broker to only validate the request,
+        :param bool validate_only: If true, the request is validated only,
```
edenhill, line 450: dito weirdness

src/confluent_kafka/src/Consumer.c
```diff
@@ -63,8 +67,12 @@ static void Consumer_dealloc (Handle *self) {
 
                 CallState_begin(self, &cs);
 
-                rd_kafka_destroy_flags(self->rk,
-                                       RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE);
+                /* If application has not called c.close() then
```
edenhill, line 70: That's why we were passing NO_CONSUMER_CLOSE, why are you changing this?

src/confluent_kafka/src/Consumer.c
```diff
@@ -1039,13 +1123,20 @@ static PyMethodDef Consumer_methods[] = {
 	  "  :param callable on_revoke: callback to provide handling of "
 	  "offset commits to a customized store on the start of a "
 	  "rebalance operation.\n"
+	  "  :param callable on_lost: callback to provide handling in "
+	  "the case a partition assignment has been lost. If not "
```
edenhill, line 1127: the partition assignment

src/confluent_kafka/src/Consumer.c
```diff
@@ -1039,13 +1123,20 @@ static PyMethodDef Consumer_methods[] = {
...
+	  "specified, lost partition events will be delivered to "
+	  "on_revoke, if specified. Partitions that have been lost may "
+	  "already be owned by other members in the group and therefore "
+	  "committing offsets, for example, may fail.\n"
```
edenhill, line 1131: Should it say that this is only used for coop-sticky?

src/confluent_kafka/src/Consumer.c
```diff
@@ -1082,20 +1173,23 @@ static PyMethodDef Consumer_methods[] = {
 	  METH_VARARGS|METH_KEYWORDS,
 	  ".. py:function:: consume([num_messages=1], [timeout=-1])\n"
 	  "\n"
-	  "  Consume messages, calls callbacks and returns list of messages "
-	  "(possibly empty on timeout).\n"
+	  "  Consume a list of messages (possibly empty on timeout). Callbacks may "
```
edenhill, line 1176: "Consume a list of messages" sounds like it is consuming from a list, should probably say that it returns a list of messages.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1082,20 +1173,23 @@ static PyMethodDef Consumer_methods[] = {
...
 	  "\n"
 	  "  The application must check the returned :py:class:`Message` "
 	  "object's :py:func:`Message.error()` method to distinguish "
-	  "between proper messages (error() returns None), or an event or "
-	  "error for each :py:class:`Message` in the list (see error().code() "
-	  "for specifics).\n"
+	  "between proper messages (error() returns None) and errors "
+      "for each :py:class:`Message` in the list (see error().code() "
+	  "for specifics). If the enable.partition.eof configuration "
+      "property is set to True, partition EOF events will also be "
```
edenhill, line 1184: configuration property

src/confluent_kafka/src/Consumer.c
```diff
@@ -1106,10 +1200,33 @@ static PyMethodDef Consumer_methods[] = {
...
+          "to the current partition assignment. This list must not contain "
+          "duplicate entries, or any entry corresponding to an already "
+          "assigned partition. When a COOPERATIVE assignor (i.e. incremental "
+          "rebalancing) is being used, this method may be used in the on_assign "
+          "callback to update the current assignment and specify start offsets. "
+          "The application should pass a list of partitions identical to the "
+          "list passed to the callback, even if the list is empty. Note that if "
+          "you do not call incremental_assign in your on_assign handler, this "
+          "will be done automatically and start offsets will be the last committed "
+          "offsets, or determined via the auto offset reset policy if there "
```
edenhill, line 1224: refer to the auto.offset.reset config property here

src/confluent_kafka/src/Consumer.c
```diff
@@ -1106,10 +1200,33 @@ static PyMethodDef Consumer_methods[] = {
...
+          "are none. This method may also be used outside the context of a "
+          "rebalance callback.\n"
```
edenhill, line 1226: But may it be used for non-cooperative assignors?

src/confluent_kafka/src/Consumer.c
```diff
@@ -1281,9 +1419,9 @@ static PyMethodDef Consumer_methods[] = {
 	  "\n"
 	  "  Actions performed:\n"
 	  "\n"
-	  "  - Stops consuming\n"
-	  "  - Commits offsets - except if the consumer property 'enable.auto.commit' is set to False\n"
-	  "  - Leave consumer group\n"
+	  "  - Stop consuming.\n"
```
edenhill, line 1422: "Actions performed (by this method)" would need this list to be in third-person form, thus "Stops", "Commits", "Leaves", etc.

tests/README.md
```diff
@@ -1,73 +1,101 @@
...
+"Unit" tests are the ones directly in the `./test` directory. These tests do
+not require an active Kafka cluster.
+
+You can run them selectively like so:
+
+    $ pytest -s -v tests/test_Producer.py
+
+Or run them all with:
+
+    $ pytest -s -v tests/test_*
```
edenhill, line 28: Will this pick up temp files, e.g., `test_foo.py~` ?

tests/README.md
```diff
@@ -1,73 +1,101 @@
...
+
+### The New Way
+
+The newer integration tests utilise `pytest` and define the `kafka_cluster` fixture (in `./tests/integration/conftest.py`) which uses [trivup](https://github.com/edenhill/trivup) to bring up a Kafka Cluster and Schema Registry instance automatically.
+
+You can run these tests selectively like so:
+
+    pytest -v -s ./tests/integration/consumer/test_consumer_error.py
+
+WIP: If you would like to avoid creating / destroying a cluster each time you run a test, you can update `conftest.py` to use `create_byo_cluster` instead of `create_trivup_cluster` (note: currently does not support Schema Registry, Kafka cluster only).
```
edenhill, line 72: It'd be nice if this was controlled by an environment variable

tests/docker/bin/cluster_up.sh
```diff
@@ -37,11 +37,14 @@ docker-compose -f $PY_DOCKER_COMPOSE_FILE up -d
 
 echo "Setting throttle for throttle test..."
 docker-compose -f $PY_DOCKER_COMPOSE_FILE exec kafka sh -c "
-        /usr/bin/kafka-configs  --zookeeper zookeeper:2181 \
+        /usr/bin/kafka-configs  --bootstrap-server kafka:9092 \
                 --alter --add-config 'producer_byte_rate=1,consumer_byte_rate=1,request_percentage=001' \
                 --entity-name throttled_client --entity-type clients"
 
+echo "Awaiting schema-registry..."
```
edenhill, line 44: Since the name is passed to await_http, why not just let await_http print the awaiting thingie?

tests/integration/cluster_fixture.py
```diff
@@ -270,3 +270,26 @@ def client_conf(self, conf=None):
 
     def stop(self):
         self._cluster.stop(cleanup=True)
+
+
+class ByoFixture(KafkaClusterFixture):
+    """
+    A Kafka cluster fixture that assumes an already running cluster.
```
edenhill, line 277: Should this set up itself from env variables, e.g `$BROKERS`?

tests/integration/conftest.py
```diff
@@ -20,17 +20,26 @@
...
+def create_trivup_cluster():
+    return TrivupFixture({'with_sr': True,
+                          'debug': True,
+                          'cp_version': 'latest',
+                          'broker_conf': ['transaction.state.log.replication.factor=1',
+                                          'transaction.state.log.min.isr=1']})
+
+
+def create_byo_cluster():
+    return ByoFixture({'bootstrap.servers': 'localhost:9092'})
```
edenhill, line 37: We should avoid these hardcoded ports, I often run multiple clusters in parallel (for parallel testing).
trivup solves this by allocating dynamic ports, and I think we should base these bootstraps on env BROKERS instead of hardcoded, and error out if BROKERS is not set.
It is much less error prone, much more flexible, and more user friendly, than these hidden fragile assumptions.

tests/integration/consumer/test_cooperative_rebalance_1.py
```diff
@@ -0,0 +1,88 @@
...
+    msg3 = consumer.poll(1)
+    assert msg3 == None
+
+    # Exceed MaxPollIntervalMs => lost partitions.
+    time.sleep(8)
+    
+    consumer.close()
+
+    assert 2 == assign_count
+    assert 1 == lost_count
```
edenhill, line 87: Do we really know that the lost is from maxpoll timeout and not from the close?
Better specifically poll after sleep to make sure the lost is seen, then do close.

tests/requirements.txt
```diff
@@ -4,3 +4,8 @@ pytest>=6.0.0;python_version>="3.0"
 pytest-timeout
 requests-mock
 trivup>=0.8.2
+fastavro
+avro;python_version<="3.0"
+avro-python3;python_version>="3.0"
+jsonschema
+protobuf
```
edenhill, line 11: These deps should not be here, they're already on the avro/protobuf/etc sub-packages and those dependencies will be automatically resolved if you install the sub-packages with `pip` rather than `setup.py`.

This requirements file is for the **test code** only

edenhill (changes requested):
There should be dry-run API tests of the new APIs in tests/test_Consumer.py

Issue #834, opened 2020-04-15
New AvroDeserializer class requires a reader schema


mparry:
Description
===========

The new `AvroDeserializer` class added in 1.4.0 requires a reader schema but this seems unnecessary and means that it's not possible to use these new classes in the same way as the old `AvroConsumer` - i.e. in `AvroConsumer` the default values for `reader_key_schema` and `reader_value_schema` are `None`. 

Specifically, we have use cases where we are just interested in consuming (and, say, displaying) whatever messages are forthcoming on a topic and it's sufficient - and indeed preferable - to just use the corresponding writer schema from the registry.

I would be happy to provide a PR addressing this, unless the reader schema requirement is deliberate for some reason.

(A couple of other minor issues I noticed while looking at `schema_registry/avro.py`: i) the error messages about `to_dict` not being callable have the arguments the wrong way around; ii) the attempt to raise a `ValueError` about unrecognized properties will fail; the second call to `format()` should be `join()` instead.)

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): 1.4.0

rnpridgeon:
Thanks for the feedback @mparry. 

Reader Schema requirement:

This was actually an item of debate during the design process. The question being should the Deserliazer be polymorphic(since we can access the writer schema from the Schema Registry). Ultimately the decision was made that the deserializer should only handle a single type, that which is defined in the reader schema. 

That said we marked these features as experimental for exactly this reason. To see how our users intend to use the API. I'll link any subseuqent issues related to the same thing here to document interest in handling "Generic" Avro records.

Added a fix for the error message/exception raising to the v1.4.1rc branch. Thanks for catching that!

mparry:
The benefit that I'm interested in isn't so much polymorphism (though sure, I can see that results too) but rather that the Python client needs no advance knowledge of the schema. In a dynamic language, this is quite a nice feature!

Clearly if you want to use the data in depth then you probably need to know _something_ about the schema but I think there are various use cases where this isn't true. To expand on my original example, we have tools that simply monitor and display the data being published on a topic, and it's quite undesirable to have to manually propagate the schema downstream to there, when the registry serves this purpose rather well.

[...]

PR #1000, opened 2020-11-29
Make reader schema optional in AvroDeserializer

97nitt:
This PR was created in response to the following issue:
https://github.com/confluentinc/confluent-kafka-python/issues/834

It makes the reader schema optional in `AvroDeserializer`. When it is not provided, the writer schema is used for deserialization. Tests were updated to verify the deserializer works both with and without a reader schema.

*NOTE*: this PR introduces breaking changes in _both_ the `AvroSerializer` and `AvroDeserializer`. The order of constructor arguments for `AvroDeserializer` had to changed to accommodate the now optional `schema_str` argument. I reordered the constructor arguments for `AvroSerializer` just for consistency between the two classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ tests/docker/conf/tls/*
 .DS_Store
 .idea
 tmp-KafkaCluster
+.venv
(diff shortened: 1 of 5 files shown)


mhowlett (approved):
thanks @97nitt - LGTM

PR #1010, opened 2021-01-04
Soak testing client

edenhill:


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -587,6 +593,18 @@ static PyMethodDef Message_methods[] = {
 	  "  :rtype: (int, int)\n"
 	  "\n"
 	},
+	{ "latency", (PyCFunction)Message_latency, METH_NOARGS,
+          "Retrieve the time it took to produce the message, from calling "
+          "produce() to the time the acknowledgement was received from "
+          "the broker.\n"
+          "Must only be used with the producer for successfully produced "
```
mhowlett, line 600: > Must only be used with the producer for successfully produced messages.

doesn't really make sense i don't think, and i don't think you need to state this. should just return None if not applicable.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -587,6 +593,18 @@ static PyMethodDef Message_methods[] = {
 	  "  :rtype: (int, int)\n"
 	  "\n"
 	},
+	{ "latency", (PyCFunction)Message_latency, METH_NOARGS,
+          "Retrieve the time it took to produce the message, from calling "
+          "produce() to the time the acknowledgement was received from "
```
mhowlett, line 598: is this more correct? "to the time the delivery report was made available to the application"

tests/soak/README.md
```diff
@@ -0,0 +1,13 @@
+# Soak testing client
+
+The soak testing clients should be run for prolonged periods
+of time, typically 2+ weeks, to vet out any resource leaks, etc.
+
+The soak testing client is made up of a producer, producing messages to
+the configured topic, and a consumer, consuming the same messages back.
+
+DataDog reporting supported by setting datadog.api_key a and datadog.app_key
```
mhowlett, line 9: is supported

mhowlett (approved):

src/confluent_kafka/src/confluent_kafka.c
edenhill: Why doesnt it make sense to state this her?
Otherwise people will try to use it out of scope (consumer side, or for failed messages)  and wonder why it doesnt work

src/confluent_kafka/src/confluent_kafka.c
edenhill: No, it's the time the ack is received. The time in dr queue is not included

PR #1042, opened 2021-02-28
additions to requirements.txt when trying proto producer

dbwrangler:


dbwrangler:
[clabot:check]

examples/requirements.txt
```diff
@@ -7,3 +7,5 @@ pyrsistent==0.16.1;python_version<"3.0"
 pyrsistent;python_version>"3.0"
 jsonschema
 protobuf
+requests
```
edenhill, line 10: I'm thinking it should perhaps depend on confluent-kafka[avro] (et.al) instead to pull in these secondary dependencies?

dbwrangler: You mean update requirements with confluent-kafka[avro] instead? TBH would be nice to not have any avro references (other diff is fastavro, which seems to be a SchemaRegistryClient dependency),though I understand non-avro options are a relatively recent addition.

edenhill: The examples should depend on confluent-kafka[avro], not their respective secondary dependencies.

PR #1052, opened 2021-03-10
Added summary section to test documentation

mhowlett:
I fixed up the test documentation, only to then find/remember we already did this in the coop rebalancing PR. Anyway, a couple of extra bits I think worth adding.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,3 +1,15 @@
+# Overview
+
+Test summary:
+
+- `./tests/test_*` (pytest): Unit tests. These tests don't bring up any external processes.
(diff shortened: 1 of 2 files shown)


tests/README.md
```diff
@@ -1,3 +1,15 @@
+# Overview
+
+Test summary:
+
+- `./tests/test_*` (pytest): Unit tests. These tests don't bring up any external processes.
```
edenhill, line 5: external processes is a bit vague, Kafka cluster would be more helpful.

edenhill (approved):

PR #1061, opened 2021-03-12
fix test_message_serializer.py

mhowlett:
this should fix the travis failure on linux/2.7. given this is the old avro client which is going to be depreciated, i think this hack solution is good enough.

review this only after travis is green. if travis is not green, i will fix

### DIFF
--- a/tests/avro/test_message_serializer.py
+++ b/tests/avro/test_message_serializer.py
@@ -34,6 +34,7 @@ from tests.avro.mock_schema_registry_client import MockSchemaRegistryClient
 from confluent_kafka import avro
 from pyflakes.test.harness import skipIf
 
+
 class TestMessageSerializer(unittest.TestCase):
     def setUp(self):
         # need to set up the serializer
--- a/tests/avro/test_message_serializer.py
+++ b/tests/avro/test_message_serializer.py
@@ -23,7 +23,9 @@
 import struct
 
 import unittest
-from unittest.mock import patch
+
+if sys.version >= 3.0
+    from unittest.mock import patch
 
 from tests.avro import data_gen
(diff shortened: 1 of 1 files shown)


tests/avro/test_message_serializer.py
```diff
@@ -23,7 +23,9 @@
 import struct
 
 import unittest
-from unittest.mock import patch
+
+if sys.version >= 3.0
```
edenhill, line 27: need to import sys

tests/avro/test_message_serializer.py
```diff
@@ -76,7 +78,10 @@ def test_encode_record_with_schema(self):
             message = self.ms.encode_record_with_schema(topic, basic, record)
             self.assertMessageIsSame(message, record, schema_id)
 
-    def test_encode_record_with_schema_sets_writers_cache_once(self):
+    def test_encode_r   ecord_with_schema_sets_writers_cache_once(self):
```
edenhill, line 81: MIND THE GAP

tests/avro/test_message_serializer.py
```diff
@@ -76,7 +78,10 @@ def test_encode_record_with_schema(self):
...
+        if sys.version < 3.0
+            # unittest.mock.patch not available
+            return
```
edenhill, line 84: this should be a `@pytest.skipif()...` function attribute

edenhill (changes requested):

tests/avro/test_message_serializer.py
mhowlett: the note specifying not to review this yet was evidently well founded... thanks for the feedback though useful.

edenhill (approved):

PR #1065, opened 2021-03-17
Fix test

hrchu:
seek() accepts one TopicPartition only


edenhill:
[clabot:check]

edenhill:
Thank you!

PR #1066, opened 2021-03-17
test code enhancement

hrchu:
poll() consumes kwargs according to the spec.
https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.Consumer.poll


edenhill:
The positional arg calling convention is okay too and less verbose.
Since there is only one argument to poll() the kwarg is a bit meh.

Issue #935, opened 2020-08-13
Bug in protobuf serializer: unnecessary roundtrip to schema registry


CTCC1:
Description
===========
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/schema_registry/protobuf.py#L315
_known_subject is never populated.

The ProtobufSerializer has a cache for the subject created. However, it is never populated in the code. As such, the serializer does a round trip to the schema registry upon every serialization to lookup the schema by calling the lookup function here
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/schema_registry/protobuf.py#L323

How to reproduce
================
Try to serialize multiple times with the same serializer instance and look at the schema registry logs.

Checklist
=========
Please provide the following information:

 - [ ] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

MSeal:
Appears to be resolved with more recent releases

Issue #1053, opened 2021-03-11
ProtobufSerializer is slow


bobbymlp:
I intent to use confluent kafka in python to produce batches of messages, writing tens of thousands of messages in bursts and I know the that underlying librdkafka library has no problem but the ProtobufSerializer takes around 20-30ms per record to serialise them and the records are small

I note you write performance is  "on par with the Java client for larger message sizes (where the overhead of the Python interpreter has less impact)"

Do you expect protobuf serialisation to be slow and any advise on how to quickly send thousands of records using this library?

edenhill:
What's the approx message/object size for the 20-30ms / msg messages?

bobbymlp:
around 100 bytes

edenhill:
Okay, then I'm guessing it is not the actual serialization that  takes 20-30ms (unless you're on a 486), but that schema is registered with the schema-registry: however, unless each message has its own schema this should only happen once (per schema type).

bobbymlp:
Is this a bug in your code or my usage. I see known_subjects is checked but not updated

https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/schema_registry/protobuf.py

it is being updated on the avro version?

bobbymlp:
What is the eta to have this fix released? It is a very small change I can make the PR if you like

mhowlett:
please do a PR and i'll review

mbonnefoy:
Hi @mhowlett @bobbymlp,

I noticed the same problem while doing some performance profiling on the `ProtobufSerializer`.
I've opened a pull request for it.

Thanks,
Matthieu

rvinzent:
Any updates on this?

bobbymlp:
The issue was fixed in june here: https://github.com/confluentinc/confluent-kafka-python/pull/1133

But there hasn't been a release since, you can apply a patch locally if you really need this

pranavrth:
This was released with the above PR in v1.8.2. Closing this issue.

PR #1082, opened 2021-04-12
Add missing update to _known_subjects in ProtobufSerializer

mbonnefoy:
Fixes https://github.com/confluentinc/confluent-kafka-python/issues/935 https://github.com/confluentinc/confluent-kafka-python/issues/1053

Without this update, a schema lookup is performed for every `ProtobufSerializer` call.

### DIFF
--- a/src/confluent_kafka/schema_registry/protobuf.py
+++ b/src/confluent_kafka/schema_registry/protobuf.py
@@ -323,6 +323,8 @@ class ProtobufSerializer(object):
                 self._schema_id = self._registry.lookup_schema(
                     subject, self._schema).schema_id
 
+            self._known_subjects.add(subject)
+
         with _ContextStringIO() as fo:
             # Write the magic byte and schema ID in network byte order
             # (big endian)


mbonnefoy:
[clabot:check]

mbonnefoy:
Hey @mhowlett @edenhill ,

Just a friendly reminder.

Any chance one of you can sanity check this PR?
afaict, Protobuf serialization is not production ready without it.

Thanks,
Matthieu

src/confluent_kafka/schema_registry/protobuf.py
```diff
@@ -323,6 +323,8 @@ def __call__(self, message_type, ctx):
                 self._schema_id = self._registry.lookup_schema(
                     subject, self._schema).schema_id
 
+            self._known_subjects.add(subject)
```
mhowlett, line 326: is this correct in the case auto register is not enabled and the schema isn't in schema registry?

mbonnefoy: Hi @mhowlett,

I am not super familiar with the `auto_register:false` use case but I suspect that even if you don't auto register your schema you still want to avoid making a POST request on each [schema lookup](https://github.com/confluentinc/confluent-kafka-python/blob/d9a91a3d74c6447d53ba1c5733b0f99c2db2e259/src/confluent_kafka/schema_registry/schema_registry_client.py#L412-L414)?

The change suggested seems to be consistent with the [avro behaviour](https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/schema_registry/avro.py#L219-L223).

Cheers,
Matthieu

rvinzent:
Looks like this should close #935 as well

mbonnefoy:
> Looks like this should close #935 as well

Thanks @rvinzent; it is the same problem. I've added it to the description.

mbonnefoy:
Closing as it's been addressed in passing by https://github.com/confluentinc/confluent-kafka-python/pull/1133

PR #1085, opened 2021-04-14
Fixed some testing docs and consume in DeserializingConsumer

Egg3141592654:
This commit enables the consume method for the deserializing consumer. Encountering some issues running tests in WSL2, will create a VM and try to run tests fully there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ tests/docker/conf/tls/*
 .idea
 tmp-KafkaCluster
 .venv
+venv_test/*
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,4 @@ tests/docker/conf/tls/*
 .idea
 tmp-KafkaCluster
 .venv
-venv_test/*
+venv_test/
--- a/src/confluent_kafka/deserializing_consumer.py
+++ b/src/confluent_kafka/deserializing_consumer.py
@@ -173,7 +173,7 @@ class DeserializingConsumer(_ConsumerImpl):
         return deserialized_messages
 
 
-    def _parse_deserialize_message(message):
+    def _parse_deserialize_message(self, message):
         """
         Internal class method for deserializing and maintaining consistency between poll and consume classes.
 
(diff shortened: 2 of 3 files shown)


Egg3141592654:
> It looks like @Egg3141592654 hasn't signed our **C**ontributor **L**icense **A**greement, yet.
> 
> > The purpose of a CLA is to ensure that the guardian of a project's outputs has the necessary ownership or grants of rights over all contributions to allow them to distribute under the chosen licence.
> > [Wikipedia](http://en.wikipedia.org/wiki/Contributor_License_Agreement)
> 
> You can read and sign our full Contributor License Agreement [here](http://clabot.confluent.io/cla).
> 
> Once you've signed reply with `[clabot:check]` to prove it.
> 
> Appreciation of efforts,
> 
> clabot

[clabot:check]

src/confluent_kafka/deserializing_consumer.py
```diff
@@ -130,29 +130,83 @@ def poll(self, timeout=-1):
...
+        messages = super(DeserializingConsumer, self).consume(num_messages, timeout)
+
+        if messages is None:
+            return []
+
+        deserialized_messages = []
+
+        for message in messages:
+            if message.error() is not None:
+                raise ConsumeError(message.error(), kafka_message=message)
```
mhowlett, line 169: this implementation effectively causes message without error to be thrown away, which is probably not ideal.

thanks for the PR, currently re-considering the need for batch consume.

tests/README.md
```diff
@@ -18,7 +18,7 @@ A python3 env suitable for running tests:
 
     $ python3 -m venv venv_test
     $ source venv_test/bin/activate
-    $ pip install -r test/requirements.txt
+    $ pip install -r tests/requirements.txt
```
mhowlett, line 21: 👍

.gitignore
```diff
@@ -27,3 +27,4 @@ tests/docker/conf/tls/*
 .idea
 tmp-KafkaCluster
 .venv
+venv_test/*
```
mhowlett, line 30: 👍
but a newline would be nice.

edenhill: No need for `*`, the dir is sufficient.

src/confluent_kafka/deserializing_consumer.py
Egg3141592654: Good point, only did this to ensure that we kept parity with the singular API, but returning the messages with the error code is likely the better idea. I removed the exception throw.

raphaelauv:
Nice I really hope this will be part of 1.8.0 :)

PR #1087, opened 2021-04-14
Encode message indexes with varints to match java impl

### LINKED ISSUE #926: Java Python Protobuf MessageIndex Mismatch
Description
===========

I know the Protobuf Registry isn't fully supported here and the Python decoder ignores the indices following the schema-id for now but this seems like something that may bite someone in the future. There's a discrepancy between Java and Python Protobuf MessageIndex generation.

Java encodes and decodes Protobuf MessageIndexes as (Zig-Zag encoded) VarInts where Python treats them strictly as UVarInts which causes mismatches.

Java Implementation:
[MesageIndex Encoding](https://github.com/confluentinc/schema-registry/blob/1b19d5e35b9bb6272ddad1f95a5effe479de3a9b/protobuf-provider/src/main/java/io/confluent/kafka/schemaregistry/protobuf/MessageIndexes.java#L57)
[Varint Encoding](https://github.com/confluentinc/kafka/blob/master/clients/src/main/java/org/apache/kafka/common/utils/ByteUtils.java#L189)

Python Implementation:
[MessageIndex Decoding](https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/schema_registry/protobuf.py#L413)
[UVarInt Decoding](https://github.com/confluentinc/confluent-kafka-python/blob/5aa6439df081ead048932ee395085236ef0bca9d/confluent_kafka/schema_registry/protobuf.py#L357)

How to reproduce
================

Not entirely reproducible with the python protobuf decoder as-is. But publishing protobuf records of the second message defined inside of a multi-message .proto file from Java/Scala will result in a header like:

00 00 00 00 01 02 02 <serialized-payload>  
That is: 0-magic, schema-id 1, varint-index-size 1, varint-index 1

Where python will then decode this header as: 0-magic, schema-id 1, uvarint-index-size 2, uvarint-index 2, ???

Checklist
=========

 - [ ] Operating system:
 - [ ] Critical issue

gfredericks:
Fixes #926

Otherwise the python and java implementations are mutually
incompatible.

Is there anything else that needs to be done about this w.r.t. backwards compatibility?
This version will be incompatible with previous versions (w.r.t. reading/writing data) in the
same way that the previous versions are already incompatible with the java library.

That seems like a poor user experience (it's a "breaking change", in a sense), but I'm not
sure what a reasonable mitigation is, and doing nothing (retaining incompatibility with java)
is worse.

I should also mention that I was unable to run tests locally, so I'll be relying on the build
to tell me if I overlooked something.

### DIFF
--- a/src/confluent_kafka/schema_registry/protobuf.py
+++ b/src/confluent_kafka/schema_registry/protobuf.py
@@ -239,9 +239,9 @@ class ProtobufSerializer(object):
                               schema_type='PROTOBUF')
 
     @staticmethod
-    def _encode_uvarints(buf, ints):
+    def _encode_varints(buf, ints):
         """
-        Encodes each int as a uvarint onto buf
+        Encodes each int as a varint onto buf
 
         Args:
             buf (BytesIO): buffer to write to.
@@ -249,6 +249,10 @@ class ProtobufSerializer(object):
 
         """
         for value in ints:
+            if(value < 0):
+                value = (-value) << 1 - 1
+            else:
+                value <<= 1
             while (value & ~0x7f) != 0:
                 buf.write(_bytes((value & 0x7f) | 0x80))
                 value >>= 7
@@ -328,7 +332,7 @@ class ProtobufSerializer(object):
             # (big endian)
             fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))
             # write the record index to the buffer
-            self._encode_uvarints(fo, self._msg_index)
+            self._encode_varints(fo, self._msg_index)
             # write the record itself
             fo.write(message_type.SerializeToString())
             return fo.getvalue()
@@ -354,15 +358,15 @@ class ProtobufDeserializer(object):
         self._msg_class = MessageFactory().GetPrototype(descriptor)
 
     @staticmethod
-    def _decode_uvarint(buf):
+    def _decode_varint(buf):
         """
-        Decodes a single uvarint from a buffer.
+        Decodes a single varint from a buffer.
 
         Args:
             buf (BytesIO): buffer to read from
 
         Returns:
-            int: decoded uvarint
+            int: decoded varint
 
         Raises:
             EOFError: if buffer is empty
@@ -377,7 +381,10 @@ class ProtobufDeserializer(object):
                 value |= (i & 0x7f) << shift
                 shift += 7
                 if not (i & 0x80):
-                    return value
+                    if value & 1:
+                        return -(value + 1) >> 1
+                    else:
+                        return value >> 1
 
         except EOFError:
             raise EOFError("Unexpected EOF while reading index")
@@ -410,10 +417,10 @@ class ProtobufDeserializer(object):
             int: Protobuf Message index.
 
         """
-        size = ProtobufDeserializer._decode_uvarint(buf)
+        size = ProtobufDeserializer._decode_varint(buf)
         msg_index = [size]
         for _ in range(size):
-            msg_index.append(ProtobufDeserializer._decode_uvarint(buf))
+            msg_index.append(ProtobufDeserializer._decode_varint(buf))
 
         return msg_index
 
--- a/src/confluent_kafka/schema_registry/protobuf.py
+++ b/src/confluent_kafka/schema_registry/protobuf.py
@@ -158,6 +158,10 @@ class ProtobufSerializer(object):
     |                                     |          | Schema Registry subject names for Schema References  |
     |                                     |          | Defaults to reference_subject_name_strategy          |
     +-------------------------------------+----------+------------------------------------------------------+
+    |                                     |          | Specifies whether the Protobuf serializer should     |
+    | ``use.deprecated.format``           | bool     | serialize message indexes without zig-zag encoding.  |
+    |                                     |          | Defaults to False.                                   |
+    +-------------------------------------+----------+------------------------------------------------------+
 
     Schemas are registered to namespaces known as Subjects which define how a
     schema may evolve over time. By default the subject name is formed by
@@ -201,7 +205,8 @@ class ProtobufSerializer(object):
     _default_conf = {
         'auto.register.schemas': True,
         'subject.name.strategy': topic_subject_name_strategy,
-        'reference.subject.name.strategy': reference_subject_name_strategy
+        'reference.subject.name.strategy': reference_subject_name_strategy,
+        'use.deprecated.format': False
     }
 
     def __init__(self, msg_type, schema_registry_client, conf=None):
@@ -222,6 +227,10 @@ class ProtobufSerializer(object):
             'reference.subject.name.strategy')
         if not callable(self._ref_reference_subject_func):
             raise ValueError("subject.name.strategy must be callable")
+
+        self._use_deprecated_format = conf_copy.pop('use.deprecated.format')
+        if not isinstance(self._use_deprecated_format, bool):
+            raise ValueError("use.deprecated.format must be a boolean value")
 
         if len(conf_copy) > 0:
             raise ValueError("Unrecognized properties: {}"
@@ -257,6 +266,37 @@ class ProtobufSerializer(object):
                 buf.write(_bytes((value & 0x7f) | 0x80))
                 value >>= 7
             buf.write(_bytes(value))
+
+    @staticmethod
+    def _encode_uvarints(buf, ints):
+        """
+        Encodes each int as a uvarint onto buf
+
+        Args:
+            buf (BytesIO): buffer to write to.
+            ints ([int]): ints to be encoded.
+
+        """
+        for value in ints:
+            while (value & ~0x7f) != 0:
+                buf.write(_bytes((value & 0x7f) | 0x80))
+                value >>= 7
+            buf.write(_bytes(value))
+
+    @staticmethod
+    def _encode_index(buf, msg_idx, deprecated_format=False):
+        """
+        Writes message index in Schema Registry Protobuf format.
+
+        Args:
+            buf (BytesIO): byte buffer
+            msg_index: Tuple[int]
+            deprecated_format: boolean, optional
+        """
+        if deprecated_format:
+            self._encode_uvarints(buf, msg_index)
+        else:
+            self._encode_varints(buf, msg_index)
 
     def _resolve_dependencies(self, ctx, file_desc):
         """
@@ -332,7 +372,7 @@ class ProtobufSerializer(object):
             # (big endian)
             fo.write(struct.pack('>bI', _MAGIC_BYTE, self._schema_id))
             # write the record index to the buffer
-            self._encode_varints(fo, self._msg_index)
+            self._encode_index(fo, self._msg_index, self._use_deprecated_format)
             # write the record itself
             fo.write(message_type.SerializeToString())
             return fo.getvalue()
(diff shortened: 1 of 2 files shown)


rayokota (approved):
LGTM

edenhill:
Thank you for this!

We'll need to think out a compatibility strategy before merging it.

gfredericks:
I suppose we probably want to support the old and new behavior? Is it sufficient to just do that, configurably? If so,

A) what's the best configuration mechanism?
B) which behavior should be the default?
C) where does the configuration need to be documented?

rayokota:
@gfredericks , the Python client has the same issue.  I suggest we use another config `use.deprecated.format` to enable the old behavior.  We will do the same in the Python client.  WDYT?

gfredericks:
Sounds fine to me

this just amounts to adding that to the docstring and default conf for `ProtobufSerializer`, and then adding a new analogous `conf=None` arg to `ProtobufDeserializer? It doesn't look like it takes a config currently

where else does it need to be documented besides the docstrings of those classes?

rayokota:
@gfredericks , yes, sounds good.  The `use.deprecated.format` (with default `false`) should be used in both the serializer and deserializer.  I think the docstrings are sufficient.  Here is the text I am using in the dotnet client: `Specifies whether the Protobuf serializer should serialize message indexes without zig-zag encoding.`

gfredericks:
This doesn't affect me, but I just wanted to point out that any python-only or dotnet-only user that would naturally opt-in to this and plan to use it indefinitely could get themselves into a pretty gnarly situation if they blindly upgrade and start running this in production, since they will end up with published data in two formats and no deserializer that can read both.

(this is of course avoidable by reading the release notes carefully and setting the new config option, but I'm imagining many users won't do that)

Does that change the analysis at all?

If we want to take that concern seriously, I can think of two options

A) broadcast the change very loudly, somehow, so that users are more likely to be aware of it (rather than just have it in the docstring)
B) start by defaulting to `true`, and issue a warning (explaining the severity of the situation, that the default will change in the future and they could end up with incompatible data) if the user doesn't pass the config option; then in some future version switch the default to `false` (possibly even _requiring_ the config option to be passed, though that's a bit extreme)

rayokota:
@gfredericks , since out of the box this client should interoperate with the Java client (including Connect and ksqlDB), we'll have to do a good job communicating the change, such as in the release notes as you suggested.  

BTW, the issue only arises when trying to serialize/deserialize a message type that is not the first message in the Protobuf schema.

gfredericks:
Okay; I'll add a commit soon that supports both behaviors and the config options as described above

rayokota:
@gfredericks , have you had a chance to look at adding the `use.deprecated.format` flag?

gfredericks:
I haven't -- I could probably prioritize it for this week if that would be particularly strategic on your end.

rayokota:
@gfredericks , if you could prioritize this week, that would be awesome  :)

gfredericks:
I added a commit for backwards compatibility.

It looks like the travis build isn't wired up anymore?

I assume due to

> Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on. 

The upshot is that I don't know if the tests pass.

rayokota (approved):
Thanks @gfredericks !  LGTM

@edenhill , @mhowlett , can you take a look?

mhowlett:
> ... seems like a poor user experience (it's a "breaking change", in a sense), but I'm not sure what a reasonable mitigation is, and doing nothing (retaining incompatibility with java) is worse.

I think the only strictly correct (and good) thing to do is version bump the non-java clients to 2.0. The timing is actually probably pretty good for this as it's been a long time since a major version bump and there are a bunch of other breaking changes we could do simultaneously.

The second best thing I think is to merge this despite the breaking change, as per the rationale above and also my rational when merging the corresponding change in .NET (note: not in a release yet, could be backed out). It is a bit questionable though.

Based on the above, I think we should merge now either way, but I also want to have a broader discussion that includes @edenhill before doing that.

src/confluent_kafka/schema_registry/protobuf.py
```diff
@@ -158,6 +158,10 @@ class ProtobufSerializer(object):
     |                                     |          | Schema Registry subject names for Schema References  |
     |                                     |          | Defaults to reference_subject_name_strategy          |
     +-------------------------------------+----------+------------------------------------------------------+
+    |                                     |          | Specifies whether the Protobuf serializer should     |
+    | ``use.deprecated.format``           | bool     | serialize message indexes without zig-zag encoding.  |
+    |                                     |          | Defaults to False.                                   |
```
edenhill, line 163: It would be good to explain when this is needed, and also outline compatibility issues with this property. E.g., messages produced with version <= 1.7.0 will not be compatible with bub-bub-bub.

src/confluent_kafka/schema_registry/protobuf.py
```diff
@@ -238,6 +247,26 @@ def __init__(self, msg_type, schema_registry_client, conf=None):
...
+        Encodes each int as a varint onto buf
+
+        Args:
+            buf (BytesIO): buffer to write to.
+            ints ([int]): ints to be encoded.
+
+        """
+        for value in ints:
+            if(value < 0):
+                value = (-value) << 1 - 1
```
edenhill, line 262: This is not performing a shift since it is basically `(-value) << 0`.  Needs parens due to operator precedence.
Also; shouldn't this bug have been caught by a test?

src/confluent_kafka/schema_registry/protobuf.py
```diff
@@ -238,6 +247,26 @@ def __init__(self, msg_type, schema_registry_client, conf=None):
...
+        """
+        Encodes each int as a varint onto buf
+
+        Args:
+            buf (BytesIO): buffer to write to.
+            ints ([int]): ints to be encoded.
+
+        """
+        for value in ints:
+            if(value < 0):
```
edenhill, line 261: No need for parens

src/confluent_kafka/schema_registry/protobuf.py
```diff
@@ -342,13 +386,38 @@ class ProtobufDeserializer(object):
...
 
+        conf (dict): ProtobufDeserializer configuration.
+
+    ProtobufSerializer configuration properties:
+
+    +-------------------------------------+----------+------------------------------------------------------+
+    | Property Name                       | Type     | Description                                          |
+    +=====================================+==========+======================================================+
+    |                                     |          | Specifies whether the Protobuf serializer should     |
+    | ``use.deprecated.format``           | bool     | serialize message indexes without zig-zag encoding.  |
```
edenhill, line 397: deserialize.  But also needs an explanation why and when.

edenhill (changes requested):
Needs CHANGELOG entry: add a "Upgrade considerations" section outlining why this has changed, that it is a breaking change, and what the user needs to do , and when.

src/confluent_kafka/schema_registry/protobuf.py
gfredericks: I suspect in practice this code is never used because we'll never be encoding negative values?

do you think it's better to test the negatives, or change the contract of the function so it errors on negative values?

edenhill: It'd be best if it was fixed, and tests added.

rayokota:
@ideasculptor , a couple of comments:

1. Records created by the Python serializer are fine if they encoded the first message of the Protobuf schema (because the encoding for 0 is the same whether using zig-zag encoding or not).  It is only the records encoded with any other message than the first message of the schema that are problematic.  The first message is by far the more common use case.  (That's probably why this bug went unnoticed for so long.)
2. One of the purposes of this PR is so that the Protobuf deserializer can read records  serialized by a Java client that have already encoded messages other than the first message of the Protobuf schema.  Bumping the magic byte number won't help that.

Given that there are probably not many usages of encoding some message other than the first message of the schema by the Python serializer, it does not seem worth the effort to bump the magic byte number due to this bug.

ideasculptor:
That makes very little sense.  You will be changing the way that the python serializer encodes messages going forward.  You cannot do that without changing the version of the protocol since it will preclude anyone with existing records in their topics from ever updating to the latest serializer without losing the ability to decode old messages. This is not really an optional thing - confluent pushes their platform as an event log, and anyone using it as such is likely to have actual regulatory requirements about being able to access old data and not allowing data to be rewritten/modified.  If you break the python deserializer and/or fail to provide a mechanism for other languages to read records written by the old serializer, you will force customers into regulatory non-compliance - effectively forcing them to abandon your platform.

It makes FAR more sense to assume that users only produce messages to a given topic from one client language than it does to just assume that no one has messages that aren't at index 0 (I certainly have messages that aren't at index 0, and I have messages that are at index 0 that are encoded with  1,0 as the message index array because that shortcut of writing just 0 looked like recipe for disaster and I can afford one whole extra byte in my messages when encoding them in order to ensure correctness.  It also eliminates a special case conditional from the serialization code.  The special case only works correctly without an extra conditional when deserializing. You have to specifically include to in any serialization code, so I didn't bother.

If you assume that users know what language they used to encode messages with protocol version 0, then they can configure an updated deserializer to treat protocol 0 the correct way for THEIR topics.  If you leave the protocol version as 0, you screw everyone, since there is no way for the deserializer to programmatically determine how a given message was encoded and any topic that has both encodings will now have inaccessible messages for any given deserializer.  There will be no way to handle it without writing code to treat messages with a certain offset a different way - and that's assuming a naive user who updates their python package knows what the offsets were at the changeover point.

You seem to be completely ignoring the issue of java deserializers (or the golang deserializer or anything else that was implemented correctly and actually tested) needing to be able to read messages written by the python serializer.  If I know a topic was written by python and you bump the version, then I can treat protocol version 1 one way and protocol version 0 the broken way.  If you don't give me any clue, I have no way to deal with the switchover and I am forced to stick with the broken serializer forever so that I can implement a deserializer that will handle it correctly.  

The protocol version exists for a reason. Why would you resist using it when you've got a distinct incompatibility?  You are making a breaking change and not updating the version.  Not violating the prohibition on doing that is basically the first rule of commercial software development.  Confluent tout their platform as the centerpiece of an event-driven architecture, but an event-driven architecture means append only event streams that can be read from any location.  If you change the python serializer without updating the protocol version, you are breaking any event-driven system that has used kafka and python in its implementation.  That's a catastrophic architecture problem for anyone in that situation.  One that, frankly, if confluent goes through with, would cause me to look elsewhere for my event log needs because I would know that I cannot trust confluent to deliver a product that will be sufficiently backward compatibility to satisfy the regulatory requirements I am subject to.  I MUST be able to read old messages on topics.  The federal government literally requires it.

ideasculptor:
Frankly, if the problem is that you only allowed yourself 255 potential versions and you are reluctant to use one up, then you should also use the protocol bump to go to 2 bytes for the protocol version so that you don't have to burn yet another if and when you finally decide that is necessary.  If the limited version set is already causing a problem when you have an obviously breaking change happening, you've definitely made your set too small.  The sooner you fix it, the better. If you fix it now, then ANY non-0 value in the first byte can be assumed to mean that the protocol version will be carried in 2 bytes instead of 1.  If you wait, then you have to treat 0 and 1 as single byte versions and so on.  It only makes your problem more complicated to solve the longer you wait.  Either assume that 255 versions is likely to carry you for long enough even if you burn 1 to fix the python serializer or else fix the version size problem as well.

ideasculptor:
> @ideasculptor , a couple of comments:
> 
> 2. One of the purposes of this PR is so that the Python deserializer can read records  serialized by a Java client that have already encoded messages other than the first message of the Protobuf schema.  Bumping the magic byte number won't help that.
> 

It most certainly DOES help that.  It allows a deserializer to decode protocol 0 correctly with only the knowledge of which language was used to encode it.  If you don't bump the version, then knowledge of the language used to encode a message is useless for helping to decode it, since it could still go either way.  You don't need to just decode java serialized messages in python.  You also need to decode python serialized messages in java.  Just because that has been broken all along except for the special case of the first message in the schema doesn't mean that there aren't messages out there that may encounter the problem in future.  All it takes is someone wanting to use a new language on an existing topic to suddenly find out that they cannot. It's particularly problematic for someone wanting to use the rest of the confluent platform, since that is all implemented in java and java will remain unable to decode messages serialized by python except in the special case.

By bumping the protocol version and providing a configurable mechanism for handling version 0 in ALL future clients, including the rest of the confluent platform, you give anyone who encounters the problem in the future a solution.  Without it, you are hanging them out to dry.

gfredericks:
@edenhill you redid this on master, is that right? https://github.com/confluentinc/confluent-kafka-python/commit/2ac0d72b24b14e5246445ad9ce66ec9c8828ef4e

Should we close this PR then?

Issue #974, opened 2020-10-27
Implement support of Schema Registry's schema references 


FrancescoPessina:
Description
===========
When consuming an Avro message with a schema which has a schema reference, the consumer fails with message `confluent_kafka.avro.serializer.SerializerError: unable to fetch schema with id xx: Received bad schema (id xx) from registry: Schema parse failed`.

Schema references is a new feature introduced in Confluent Platform version 5.5.0 (see here https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/#avro-unions-with-schema-references)

How to reproduce
================
Given a schema containing a schema reference registered into the schema registry, such as:

Nested schema:
```
{
  "namespace": "org.schema_registry_test_app.avro",
  "type": "record",
  "name": "Result",
  "fields": [
    {
      "name": "up",
      "type": "string"
    },
    {
      "name": "down",
      "type": "string"
    }
  ]
}
```

Schema with reference:
```
{
  "namespace": "org.schema_registry_test_app.avro",
  "type": "record",
  "name": "AvroTest",
  "fields": [
    {
      "name": "id",
      "type": {
        "name": "Uuid",
        "type": "fixed",
        "size": 16
      }
    },
    {
      "name": "results",
      "type": {
        "type": "array",
        "items": "org.schema_registry_test_app.avro.Result"
      }
    }
  ]
}
```

Try to consume with `AvroConsumer` a message produced with `AvroTest` schema.

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version(1.5.0)` and `confluent_kafka.libversion()`):
 - [x] Apache Kafka broker version: 2.5
 - [ ] Client configuration: `{...}`
 - [x] Operating system: Linux (docker) and MacOS
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

slominskir:
I believe schema references actually are partially supported as the confluent-kafka-python API includes a `references` field in the [Schema](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/index.html#schema) object.  In your case you may just need to modify your API usage.   Apparently AVROConsumer is legacy and replaced with DeserializingConsumer (though  I don't see how subscribing to multiple topics with differing deserializers, but a single Consumer is supposed to work the new way).

[...]

PR #1088, opened 2021-04-14
Schema references support

slominskir:
Fixes https://github.com/confluentinc/confluent-kafka-python/issues/974

To use Serializer:

```
named_schemas = {}
ref_dict = loads(ref_schema_str)
parse_schema(ref_dict, _named_schemas=named_schemas)
ref_schema = SchemaReference("org.test.MyReferencedItem", "referenced-subject", "1")
schema = Schema(schema_str, "AVRO", [ref_schema])

avro_serializer = AvroSerializer(schema_registry_client, schema, None, None, named_schemas)
```

To use Deserializer:

```
named_schemas = {}
ref_dict = loads(ref_schema_str)
parse_schema(ref_dict, _named_schemas=named_schemas)

# schema can be None if using registry (writer schema)
ref_schema = SchemaReference("org.test.MyReferencedItem", "referenced-subject", "1")
schema = Schema(schema_str, "AVRO", [ref_schema])

avro_deserializer = AvroDeserializer(schema_registry_client, schema, None, False, named_schemas)
```


src/confluent_kafka/schema_registry/avro.py
```diff
@@ -133,14 +133,14 @@ class AvroSerializer(Serializer):
...
-                 '_subject_name_func', '_to_dict']
+                 '_subject_name_func', '_to_dict', '_named_schemas']
 
     # default configuration
     _default_conf = {'auto.register.schemas': True,
                      'subject.name.strategy': topic_subject_name_strategy}
 
-    def __init__(self, schema_registry_client, schema_str,
-                 to_dict=None, conf=None):
+    def __init__(self, schema_registry_client, schema,
```
mhowlett, line 142: just looking at this quickly.
can you comment on the pros/cons of passing a `schema` here vs `schema_str` as you see them?

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -169,10 +170,8 @@ def __init__(self, schema_registry_client, schema_str,
             raise ValueError("Unrecognized properties: {}"
                              .format(", ".join(conf_copy.keys())))
 
-        # convert schema_str to Schema instance
-        schema = _schema_loads(schema_str)
         schema_dict = loads(schema.schema_str)
-        parsed_schema = parse_schema(schema_dict)
+        parsed_schema = parse_schema(schema_dict, _named_schemas=self._named_schemas)
```
mhowlett, line 174: cool - fastavro seems like it has good support for referenced schemas, we should definitely support it.

again, the first thing i'm wondering is what type the API should expect. feel free to give some commentary on that, if you have a clear opinion.

src/confluent_kafka/schema_registry/avro.py
slominskir: Passing a Schema allows setting schema references.

src/confluent_kafka/schema_registry/avro.py
slominskir: Using a Schema object seems reasonable and is done in the Java API:  https://docs.confluent.io/platform/current/schema-registry/serdes-develop/serdes-avro.html#avro-serializer.

However it does break backwards compatibility and does require awkwardly indicating 'AVRO' Schema type.   Also there are confusingly lots of lower level APIs (AVRO and fastavro) leaking through the Kafka API resulting in situations like many different types of Schemas:

   - [AVRO Schema](https://avro.apache.org/docs/current/gettingstartedpython.html)
   - [fastavro Schema](https://fastavro.readthedocs.io/en/latest/schema.html) - a dictionary
   - [Kafka Schema](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.schema_registry.Schema)
   - [Kafka RegisteredSchema](https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#confluent_kafka.schema_registry.RegisteredSchema)
   - Kafka SchemaReference - missing from docs?

There are alternatives such as using kwargs such that users could continue to use the String representation of the schema if they had no need for references (I believe kwargs is similar in spirit to Java method overloading).   It is also possible to leverage the dynamic typing of Python to allow the schema arg to be either a String or a Schema (though this seems like it would be confusing).

slominskir:
The failing test is only with Python 2.7 and is due to fastavro API change.   The 2.7 version of the tests use fastavro 0.24.2 whereas in the 3.6 version of the tests fastavro 1.4.0 is used.  The issue is the signature of the method parse_schemas changed and the parameter _named_schemas was renamed named_schemas.   How is this kind of difference usually handled?

avanish-appdirect:
Can we also have a single generic schema object? or make this code this to use both the schema object  Confluent-kafka-python `SchemaRegistryClient` method have different schema object for different method for example:  `get_latest_version` return you schema with references as array of json as below
```
[{'name': 'com.appdirect.common.event.EventMetadata', 'subject': 'com.appdirect.common.event.EventMetadata', 'version': 5}, {'name': 'com.appdirect.appwise.schema.notify.WebPushNotificationPayload', 'subject': 'com.appdirect.appwise.schema.notify.WebPushNotificationPayload', 'version': 2}]
```
While `get_schema` method return you schema with references as array of `SchemaReference` object as below
```
[<confluent_kafka.schema_registry.schema_registry_client.SchemaReference object at 0x1037a9790>, <confluent_kafka.schema_registry.schema_registry_client.SchemaReference object at 0x1037a97f0>]
```

As it it more like to fetch schema with subject name rather then schema_id because we can map our schema with topic and reuse topic name to fetch the schema.

Currently I get below error the is when I pass schema object where references are array of json object i.e `get_latest_version` :
```
  File "/Users/lib/python3.9/site-packages/confluent_kafka/serializing_producer.py", line 172, in produce
    value = self._value_serializer(value, ctx)
  File "/Users/lib/python3.9/site-packages/confluent_kafka/schema_registry/avro.py", line 219, in __call__
    registered_schema = self._registry.lookup_schema(subject,
  File "/Userslib/python3.9/site-packages/confluent_kafka/schema_registry/schema_registry_client.py", line 407, in lookup_schema
    request['references'] = [{'name': ref.name,
  File "/Users/lib/python3.9/site-packages/confluent_kafka/schema_registry/schema_registry_client.py", line 407, in <listcomp>
    request['references'] = [{'name': ref.name,
AttributeError: 'dict' object has no attribute 'name'

During handling of the above exception, another exception occurred:

  File "/Users/a/python-avro-producer/arvo_producer.py", line 70, in <module>
    main()
  File "/Users/a/python-avro-producer/arvo_producer.py", line 64, in main
    producer.produce(topic=topic, value=value, on_delivery=delivery_report)
  File "/Users/lib/python3.9/site-packages/confluent_kafka/serializing_producer.py", line 174, in produce
    raise ValueSerializationError(se)
confluent_kafka.error.ValueSerializationError: KafkaError{code=_VALUE_SERIALIZATION,val=-161,str="'dict' object has no attribute 'name'"}
```

mhowlett:
ok, i'm going to try to facilitate getting this PR in between other things... 

@slominskir - two things: 1. can you resolve the conflicts :-). 2. the suggested change to the AvroSerializer constructor is breaking. can you make it so that a string or Schema is accepted? (do you consider that idiomatic Python?)

there may be other things, that is just what comes to mind first.

slominskir:
@mhowlett - I'll take a look at this again.   Accepting either a string or a Schema sounds reasonable (cleanest way may be via `@overload` annotation https://peps.python.org/pep-3124)

mhowlett:
cool, thanks. I'll give it a careful look over / play after you've done that.
i may not get to this immediately, but please feel free to @ me if it's taking more than a couple of weeks ...

slominskir:
@mhowlett - The CI doesn't seem to be working, but otherwise this PR is ready for first pass inspection.    A few points to consider:
1. I tried to keep changes to a minimum and therefore users must manually register any references with the Schema Registry, parse schemas to populate named_schemas, and create a references array with all references (in order) and potentially include nested references too.   Changes to the code will likely balloon if we try to handle this in the serializer as we may determine registry caching would need to be updated to capture subject and version alternative schema lookups and/or determine Schema/SchemaReference/RegisteredSchema objects need to be re-worked.
2. In the deserializer case when the writer schema is fetched from the registry the code currently will resolve only one-level deep of associated references. (recursive code probably should be handled carefully).
3. Not super excited about the `@overload` approach to supporting either a string or Schema.   Not sure if there is a better way though.   A factory method could be used, but that probably makes most sense for string case allowing the constructor to take the Schema.   Pulling in a module that provides multiple dispatch seems like overkill.  Using a Union might be simplest.

slominskir:
Worth mentioning that without https://github.com/confluentinc/confluent-kafka-python/pull/1304 we can't do recursive named_schemas resolution of nested references (another reason this PR only goes one deep).    We may eventually want something like:

```
    def __parse_schema_recursive(self, schema, named_schemas):
        print(f"Handling schema {schema.schema_str}")
        for ref in schema.references:
            print(f"Handling reference: {ref.name}")
            ref_reg_schema = self._registry.get_version(ref.subject, ref.version)
            self.__parse_schema_recursive(ref_reg_schema.schema, named_schemas)
            ref_dict = loads(ref_reg_schema.schema.schema_str)
            parse_schema(ref_dict, named_schemas=named_schemas)
```

But first SchemaRegistryClient.get_version() must return a RegisteredSchema containing a Schema with a non-empty references array.

src/confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -671,19 +671,22 @@ class Schema(object):
...
+    __slots__ = ['schema_str', 'schema_type', 'references', 'named_schemas', '_hash']
 
-    def __init__(self, schema_str, schema_type, references=[]):
+    def __init__(self, schema_str, schema_type, references=[], named_schemas={}):
         super(Schema, self).__init__()
 
         self.schema_str = schema_str
         self.schema_type = schema_type
         self.references = references
+        self.named_schemas = named_schemas
```
anchitj, line 689: @slominskir Any reason why do the user needs to provide `named_schemas`. Can't we fetch it from schema registry like you've done [here](https://github.com/confluentinc/confluent-kafka-python/blob/c973f3ca2a91fd730b511cc153de58a41ae0a965/src/confluent_kafka/schema_registry/avro.py#L385-L395)? This seems to be leaking implementation detail of the library which can be avoided.

slominskir: If I'm remembering correctly in some cases the user may already have the `named_schemas` object so it would be inefficient to require the serde to look it up.   Mostly though, it's probably because I was keeping changes to a minimum and trying to make it possible to use schema references using the odd mix of existing APIs.

For the serializer - you'd need to update code to iterate over the references array and populate named_schemas, probably here: https://github.com/confluentinc/confluent-kafka-python/blob/c973f3ca2a91fd730b511cc153de58a41ae0a965/src/confluent_kafka/schema_registry/avro.py#L206-L207

However, the serializer does not have schema strings for schema references and in the case of it not existing in the registry yet you can't look it up either.  Perhaps require references exist in the registry before a parent schema can be registered (users would need to manually loop over references themselves registering them in order).  Else it could be a separate array added to the Schema object that contains reference schema strings in the same index that corresponds to the SchemaReference array.

Depending on the approach above, the auto register code may need to be updated to loop over references and register them.  This would require the reference schema strings, probably here: https://github.com/confluentinc/confluent-kafka-python/blob/c973f3ca2a91fd730b511cc153de58a41ae0a965/src/confluent_kafka/schema_registry/avro.py#L259-L264

For the deserializer - In many cases the reader and writer schema are the same so if you do move this behavior into the serde you can probably check if reader and writer schema_id match and just use the one lookup that's already being done.

anchitj: Thanks for the context. We are working on to include the schema reference support. There are some more changes needed which I'll push in this PR itself.

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -67,6 +67,24 @@ def _schema_loads(schema_str):
...
+    Resolves named schemas referenced by the provided schema recursively.
+    :param schema: Schema to resolve named schemas for.
+    :param schema_registry_client: SchemaRegistryClient to use for retrieval.
+    :param named_schemas: Dict of named schemas resolved recursively.
+    :return: named_schemas dict.
+    """
+    if named_schemas is None:
+        named_schemas = {}
+    for ref in schema.references:
+        referenced_schema = schema_registry_client.get_version(ref.subject, ref.version)
```
pranavrth, line 81: Can we do a batch call to schema registry?

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -315,13 +339,31 @@ class AvroDeserializer(Deserializer):
...
 
     def __init__(self, schema_registry_client, schema_str=None, from_dict=None, return_record_name=False):
+        schema = None
+        if schema_str is not None:
+            if isinstance(schema_str, str):
+                schema = _schema_loads(schema_str)
+            elif isinstance(schema_str, Schema):
+                schema = schema_str
+            else:
+                raise ValueError('You must pass either str or Schema')
```
pranavrth, line 353: either schema string or schema object

src/confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -686,12 +686,14 @@ class Schema(object):
     Args:
         schema_str (str): String representation of the schema.
 
+        schema_type (str): The schema type: AVRO, PROTOBUF or JSON.
+
         references ([SchemaReference]): SchemaReferences used in this schema.
 
-        schema_type (str): The schema type: AVRO, PROTOBUF or JSON.
-    """
+        named_schemas (dict): Named schemas
```
pranavrth, line 693: Not added in the args.

Remove this file as there are no changes to this file.

pranavrth (changes requested):
Check all the comments mentioned in JSON schema reference [PR](https://github.com/confluentinc/confluent-kafka-python/pull/1514).

src/confluent_kafka/schema_registry/schema_registry_client.py
anchitj: Thanks, removed this.

src/confluent_kafka/schema_registry/avro.py
anchitj: Updated.

src/confluent_kafka/schema_registry/avro.py
anchitj: I checked on this, it doesn't looks like it's available directly through schema registry. https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string-%20subject)-versions

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -155,15 +173,21 @@ class AvroSerializer(Serializer):
...
 
-    def __init__(self, schema_registry_client, schema_str,
-                 to_dict=None, conf=None):
+    def __init__(self, schema_registry_client, schema_str, to_dict=None, conf=None):
+        if isinstance(schema_str, str):
+            schema = _schema_loads(schema_str)
+        elif isinstance(schema_str, Schema):
+            schema = schema_str
+        else:
+            raise TypeError('You must pass either string or schema object')
```
pranavrth, line 189: Schema strib or schema object.

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -200,9 +224,9 @@ def __init__(self, schema_registry_client, schema_str,
             raise ValueError("Unrecognized properties: {}"
                              .format(", ".join(conf_copy.keys())))
 
-        schema = _schema_loads(schema_str)
         schema_dict = loads(schema.schema_str)
-        parsed_schema = parse_schema(schema_dict)
+        self._named_schemas = _resolve_named_schema(schema, schema_registry_client)
+        parsed_schema = parse_schema(schema_dict, named_schemas=self._named_schemas)
```
pranavrth, line 229: Aren't these two lines doing the same thing?

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -315,13 +340,31 @@ class AvroDeserializer(Deserializer):
...
 
     def __init__(self, schema_registry_client, schema_str=None, from_dict=None, return_record_name=False):
+        schema = None
+        if schema_str is not None:
+            if isinstance(schema_str, str):
+                schema = _schema_loads(schema_str)
+            elif isinstance(schema_str, Schema):
+                schema = schema_str
+            else:
+                raise TypeError('You must pass either string or schema object')
```
pranavrth, line 354: Same

pranavrth (changes requested):
Need to understand deserializer part. Will do over the call.

src/confluent_kafka/schema_registry/avro.py
anchitj: `_resolve_named_schema` is parsing all the referenced schemas, but not the top level one. Line number 229 parses the top level one using the `named_schemas`.

src/confluent_kafka/schema_registry/avro.py
anchitj: Updated

src/confluent_kafka/schema_registry/avro.py
emasab: Also, parse_schema adds the parsed schema to the named_schemas dict

tests/integration/schema_registry/test_avro_serializers.py
```diff
@@ -185,3 +257,92 @@ def test_avro_record_serialization_custom(kafka_cluster):
...
+    sr.register_schema("user", Schema(User.schema_str, 'AVRO'))
+    sr.register_schema("award_properties", Schema(AwardProperties.schema_str, 'AVRO'))
+    sr.register_schema("award", Schema(Award.schema_str, 'AVRO', [award_properties_schema_ref]))
+
+    references = [user_schema_ref, award_schema_ref]
+    schema = Schema(AwardedUser.schema_str, 'AVRO', references)
+    return awarded_user, schema
+
+
+def _reference_common(kafka_cluster, awarded_user, serializer_schema, deserializer_schema):
```
pranavrth, line 284: name

pranavrth (changes requested):

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -370,10 +413,12 @@ def __call__(self, data, ctx):
             writer_schema = self._writer_schemas.get(schema_id, None)
 
             if writer_schema is None:
-                schema = self._registry.get_schema(schema_id)
-                prepared_schema = _schema_loads(schema.schema_str)
+                registered_schema = self._registry.get_schema(schema_id)
+                if self._named_schemas is None:
```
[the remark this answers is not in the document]
anchitj: Updated

tests/integration/schema_registry/test_avro_serializers.py
```diff
@@ -185,3 +257,92 @@ def test_avro_record_serialization_custom(kafka_cluster):
     user2 = msg.value()
 
     assert user2 == user
+
+
+def _get_reference_data_and_register_schemas(kafka_cluster):
```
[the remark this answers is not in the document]
anchitj: Done

tests/integration/schema_registry/test_avro_serializers.py
anchitj: Done

tests/integration/schema_registry/test_avro_serializers.py
```diff
@@ -185,3 +257,92 @@ def test_avro_record_serialization_custom(kafka_cluster):
     user2 = msg.value()
 
     assert user2 == user
+
+
+def register_avro_schemas_and_build_awarded_user_schema(kafka_cluster):
```
pranavrth, line 262: Move these functions to top.

Add _ as this is private

tests/integration/schema_registry/test_avro_serializers.py
```diff
@@ -185,3 +257,92 @@ def test_avro_record_serialization_custom(kafka_cluster):
...
+    consumer = kafka_cluster.consumer(value_deserializer=value_deserializer)
+    consumer.assign([TopicPartition(topic, 0)])
+
+    msg = consumer.poll()
+    awarded_user2 = msg.value()
+
+    assert awarded_user2 == awarded_user
+
+
+def _test_avro_reference(kafka_cluster):
```
pranavrth, line 329: Remove _

Issue #864, opened 2020-05-13
Docs Schema referenced schema appears incorrect


Amertz08:
Description
===========

The `Schema` object docs show `references` should be `[Schema]` but I think it is actually supposed to be `[SchemaReference]`.

If you use `Schema` then attempt to use `SchemaRegistry.register_schema` it references attributes that do not exist.

```python
# CP 5.5 adds new fields (for JSON and Protobuf).
if len(schema.references) > 0 or schema.schema_type != 'AVRO':
    request['schemaType'] = schema.schema_type
    request['references'] = [{'name': ref.name,
    'subject': ref.subject,
    'version': ref.version}
     for ref in schema.references]
```

How to reproduce
================

```python
import json

from confluent_kafka.schema_registry import SchemaRegistryClient, Schema, SchemaReference
from fastavro.schema import fullname

if __name__ == "__main__":
    subject = "some-subject"
    registry_url = "http://registry.blah"
    client = SchemaRegistryClient({"url": registry_url})

    with open("child.avsc", "r") as f:
        schema_str = f.read()
        child = Schema(schema_str, "AVRO")

    child_id = client.register_schema(subject, child)
    registered = client.lookup_schema(subject, child)

    # name = fullname(json.loads(registered.schema.schema_str))
    # referenced = SchemaReference(
    #     name=name,
    #     subject=subject,
    #     version=registered.version
    # )

    with open("parent.avsc", "r") as f:
        schema_str = f.read()
        parent = Schema(schema_str, "AVRO", [child])
        # parent = Schema(schema_str, "AVRO", [referenced])

    schema_id = client.register_schema("webhooks", parent)
    print(schema_id)
```

Also it appears that the following logic might be incorrect.
```python
if len(schema.references) > 0 or schema.schema_type != 'AVRO':
```

If I make the changes commented out in my example I can get passed the attribute error but I get an error from the API

```
confluent_kafka.schema_registry.error.SchemaRegistryError: Unrecognized field: schemaType (HTTP status code 422, SR code 422)
```

That being said it would be nice to have an example of how to register nested Avro schemas somewhere in the docs. We are having a rough time trying to figure it out.

Checklist
=========
Please provide the following information:

confluent-kafka[avro]==1.4.1
librdkafk ('1.4.0', 17039615)
Mac OSX
confluentinc/cp-schema-registry:5.3.1

[...]

PR #1092, opened 2021-04-19
Docs say Schema when they mean SchemaReference

slominskir:
Fixes https://github.com/confluentinc/confluent-kafka-python/issues/864


mhowlett:
yep, that's right. see also usage here: https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/schema_registry/schema_registry_client.py#L377

mhowlett (approved):

PR #1094, opened 2021-04-20
Add note about Windows environment variables

slominskir:
The environment variable names are different with Visual Studio it seems then with gcc or whatever the example given is using.


edenhill (approved):

edenhill:
[clabot:check]

edenhill:
Thank you!

PR #1096, opened 2021-04-21
Add error_cb to confluent_cloud.py example

edenhill:


examples/confluent_cloud.py
```diff
@@ -48,42 +48,76 @@
...
 import uuid
 
-from confluent_kafka import Producer, Consumer
+from confluent_kafka import Producer, Consumer, KafkaError, KafkaException
 
 
+def error_cb(err):
+    """ The error callback is used for generic client errors. These
+        errors are generally to be considered informational as the client will
+        automatically try to recover from all errors, and no extra action
```
mhowlett, line 57: > will automatically try to recover from all errors

what about fatal errors? probably should mention those, especially as I assume we will make idempotent producer the default before too long like Java, at which point they will apply to this example.

examples/confluent_cloud.py
```diff
@@ -48,42 +48,76 @@
...
+        is typically required by the application.
+        For this example however, we terminate the application if the client
+        is unable to connect to any broker (_ALL_BROKERS_DOWN) and on
+        authentication errors (_AUTHENTICATION). """
```
mhowlett, line 61: 👍

examples/confluent_cloud.py
edenhill: I think we'll want to take a bigger grab and introduce the error attributes, treating _AUTHENTICATION as a fatal error, rather than having certain error codes and fatal attribute.

mhowlett: oh, we don't have these in python yet. ok.

don't think auth is fatal in all cases.

examples/confluent_cloud.py
```diff
@@ -48,42 +48,76 @@
...
+
+    print("Client error: {}".format(err))
+    if err.code() == KafkaError._ALL_BROKERS_DOWN or \
+       err.code() == KafkaError._AUTHENTICATION:
+        # Any exception raised from this callback will be re-raised from the
```
mhowlett, line 66: amazing, so it does

PR #1099, opened 2021-04-21
Added venv info to example README.md

mhowlett:


examples/README.md
```diff
@@ -20,3 +20,37 @@ Additional examples for [Confluent Cloud](https://www.confluent.io/confluent-clo
...
+
+## venv setup
+
+It's usually a good idea to install Python dependencies in a virtual environment to avoid
+conflicts between projects.
+
+To setup a venv with the latest release version of confluent-kafka and dependencies of all examples installed:
+
+```
+> python3 -m venv venv_examples
```
edenhill, line 32: Never seen '>' denote the standard bash PS1 prompt, that's usually the PS2 (multiline) prompt. Use `$ ` instead

examples/README.md
```diff
@@ -20,3 +20,37 @@ Additional examples for [Confluent Cloud](https://www.confluent.io/confluent-clo
...
+
+To setup a venv that uses the current source tree version of confluent_kafka, you
+need to have a C compiler and librdkafka installed
+([from a package](https://github.com/edenhill/librdkafka#installing-prebuilt-packages), or
+[from source](https://github.com/edenhill/librdkafka#build-from-source)). Then:
+
+```
+> python3 -m venv venv_examples
+> source venv_examples/bin/activate
+> python setup.py build
```
edenhill, line 46: For this case it is better to do `python setup.py develop` instead of build and install.

edenhill (changes requested):

examples/README.md
mhowlett: thanks

mhowlett:
please don't unnecessarily slow down workflow by "requesting changes" for trivial things, then forgetting to follow up with an approval

mhowlett:
but whoops, i did screw up, new commit coming.

edenhill (approved):

PR #1114, opened 2021-05-06
Headers in SerializationContext

ldom:


edenhill:
It'd be good with a simple test to make sure they're passed correctly.

And there's also a bunch of tests that instantiate a SerializationContext() without passing this argument, which are likely to fail now. Perhaps make headers optional (None) or update those tests, or both.

ldom:
I'll do both

tests/schema_registry/test_avro_serializer.py
```diff
@@ -111,7 +111,7 @@ def test_avro_serializer_record_subject_name_strategy(load_avsc):
                                      conf={'subject.name.strategy':
                                            record_subject_name_strategy})
 
-    ctx = SerializationContext('test_subj', MessageField.VALUE)
+    ctx = SerializationContext('test_subj', MessageField.VALUE, [])
```
edenhill, line 114: From a test perspective we yield better coverage if some of these use the default value (not specified), and some use an empty list

mhowlett (approved):

edenhill (changes requested):
There needs to be at least one test that verifies that passed headers are in-fact available on the ctx, preferably a negative test too (no headers in, no headers on ctx)

ldom:
I've added a few tests. Some are an overkill but there is never too many tests.

src/confluent_kafka/serialization/__init__.py
```diff
@@ -56,10 +56,13 @@ class SerializationContext(object):
         field (MessageField): Describes what part of the message is
             being serialized.
 
+        headers (List of header tuples): Message headers
```
edenhill, line 59: Should say that it may be None

tests/schema_registry/test_avro_serializer.py
```diff
@@ -176,3 +176,66 @@ def test_avro_serializer_subject_name_strategy_default(load_avsc):
...
+    test_serializer = AvroSerializer(test_client, 'int',
+                                     conf={'subject.name.strategy':
+                                           record_subject_name_strategy})
+
+    headers = [('k1', 'v1'),]
+
+    ctx = SerializationContext('test_subj', MessageField.VALUE, headers)
+    assert ('k1', 'v1') in ctx.headers
+    headers.extend([('k2', 'v2'),])
+    assert ('k2', 'v2') in ctx.headers
```
edenhill, line 240: nit: This is a side-effect of not copying the headers, which may not even be desirable, so not sure this is something we should test for.

tests/schema_registry/test_avro_serializer.py
```diff
@@ -176,3 +176,66 @@ def test_avro_serializer_subject_name_strategy_default(load_avsc):
...
+    assert len(headers) == 2
```
edenhill, line 241: I prefer as exact as possible matching in tests, and order matters for the headers, so suggest doing an exact match between ctx.headers and what is expected in there.

edenhill (changes requested):
Good stuff!

tests/schema_registry/test_avro_serializer.py
ldom: removed

ldom: not really useful anyway

src/confluent_kafka/serialization/__init__.py
```diff
@@ -56,10 +56,13 @@ class SerializationContext(object):
         field (MessageField): Describes what part of the message is
             being serialized.
 
+        headers (optional, defaults to None, List of header tuples): Message headers
```
edenhill, line 59: I think the paren may only hold the type, so reframe as:
```python
  headers (list): List of message header tuples. Defaults to None.
```

edenhill:
Looks good!

Just need to sort out the flake8 errors:
https://travis-ci.org/github/confluentinc/confluent-kafka-python/jobs/769668688#L803

(run flake8 locally to find all style issues)

PR #1120, opened 2021-05-13
Header in serializing context - UPDATED

ldom:


tests/schema_registry/test_avro_serializer.py
```diff
@@ -176,3 +180,17 @@ def test_avro_serializer_subject_name_strategy_default(load_avsc):
...
+def test_serialization_context_headers_optional(load_avsc):
+    """
+    Ensures SerializationContext headers parameter is optional
+    """
+    conf = {'url': TEST_URL}
+    test_client = SchemaRegistryClient(conf)
+    _ = AvroSerializer(test_client, 'int',
+                       conf={'subject.name.strategy': record_subject_name_strategy})
+
+    ctx = SerializationContext('test_subj', MessageField.VALUE)
```
jliunyu, line 194: Can we verify the default headers at test_avro_serializer_topic_record_subject_name_strategy() instead of adding a new function?

jliunyu (approved):

PR #1133, opened 2021-06-01
Enhance serializers with use.latest.version config

rayokota:
Also add skip.known.types config for Protobuf serializer

### DIFF
--- a/tests/schema_registry/test_avro_serializer.py
+++ b/tests/schema_registry/test_avro_serializer.py
@@ -74,6 +74,29 @@ def test_avro_serializer_config_auto_register_schemas_false(mock_schema_registry
     assert test_client.counter['POST'].get('/subjects/{}'.format(subject)) == 1
 
 
+def test_avro_serializer_config_use_latest_version(mock_schema_registry):
+    """
+    Ensures auto.register.schemas=False does not register schema
+    """
+    conf = {'url': TEST_URL}
+    test_client = mock_schema_registry(conf)
+    topic = "test-use-latest-version"
+    subject = topic + '-key'
+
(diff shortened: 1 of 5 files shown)


tests/schema_registry/test_avro_serializer.py
```diff
@@ -74,6 +74,29 @@ def test_avro_serializer_config_auto_register_schemas_false(mock_schema_registry
...
+
+    test_serializer("test",
+                    SerializationContext("test-use-latest-version",
+                                         MessageField.KEY))
+
+    register_count = test_client.counter['POST'].get('/subjects/{}/versions'
+                                                     .format(subject), 0)
+    assert register_count == 0
+    # Ensure latest was requested
+    assert test_client.counter['GET'].get('/subjects/{}/versions/latest'.format(subject)) == 1
```
mhowlett, line 97: there should really be more test coverage of this stuff.
but i'm ok with the effort/benefit tradeoff.

mhowlett (approved):
thanks, LGTM.

Atheuz:
Fixes bug where a schema lookup is performed for every ProtobufSerializer call. Made Protobuf and schema registry impractical to use for anything but the smallest producers as it would make the lookup, on my network this would take 0.2 seconds per produce. Adds up quick.

mhowlett:
@Atheuz - to clarify - are you saying this PR fixed such a bug, or the bug remains in 1.8.2?

Atheuz:
@mhowlett this PR fixed the bug and as such is fixed in the 1.8.2 release. It just wasn't stated anywhere here, so I wanted to add a comment.

PR #1134, opened 2021-06-10
changed log_cb to logger in docstring

mhowlett:


edenhill (changes requested):
Need a test case to verify it works. Just copy one of the tests in test_log.py

mhowlett:
feel free to fix

Issue #1097, opened 2021-04-21
Support for connect_cb


hallfox:
We have an experimental patch which adds support for setting `connect_cb` as a configuration, however looking back we found [this issue from 2017
](https://github.com/confluentinc/confluent-kafka-python/issues/219#issuecomment-315983155) which says that adding it would be hard.

And it is! By doing so we discovered a deadlock that would occur since the `connect_cb` is called in a different thread than the other callback configs due to in part the GIL being held during calls to `rd_kafka_assign`. Trying to solve this, we released the GIL for the duration of `rd_kafka_assign` and things seem to be working well now.

That context is here because I'm curious to know:
1. Is supporting `connect_cb` hard for other specific reasons?
2. Does the assign call need to hold the GIL for other safety reasons?

Basically I'm wondering if there are some invariants I may be stepping on with these changes. Thanks.

edenhill:
What's your use of the connect_cb?

hallfox:
We need to make a connection through a proxy address.

edenhill:
To answer your original questions:

> Is supporting connect_cb hard for other specific reasons?

No, it should be straight forward as long as the callback doesn't do anything overly complex since it is called from an internal broker thread.

> Does the assign call need to hold the GIL for other safety reasons?

No, no need. But the deadlock may occur in many of the librdkafka C calls (any call that may need to aquire a broker lock), so all of them need to release the GIL if we were to add connect_cb support.
Before going down that route it would be good to understand the cost of releasing and re-acquiring the GIL. The underlying mutex itself is cheap, but maybe CPython does other things as well?

hallfox:
> Before going down that route it would be good to understand the cost of releasing and re-acquiring the GIL.

So we did already try going down this route and have a patch basically ready to go, however we haven't done a broader analysis considering the performance impact of this. Given that this might be adding some fairly costly operations in a hot path, what do you think makes sense? Maybe releasing the GIL only under certain conditions?

edenhill:
Since the librdkafka API does not make any guarantees what internal locks it may or may not acquire (and that would be a tough guarantee to maintain as the code evolves), we pretty much need to release the GIL on all rdkafka API calls to avoid a deadlock.

hallfox:
We've already done this and have some anecdotal evidence that it works, should I open a PR and continue this discussion there?

edenhill:
[...]

PR #1140, opened 2021-06-16
Add support for connect_cb

hallfox:
This PR adds the ability to set the `connect_cb` function to a Python function object. Through many iterations it has caused us to run into a few deadlocking situations because of the GIL, which is why several `rdkafka` calls now release it. We are looking for some feedback, but in particular this PR is still missing

- [ ] Windows support
- [ ] Docs
- [ ] Tests

Notably as discussed in #1097 this patch also releases/acquires the GIL in potentially hot paths, so the overall performance impact here is worth some attention.

### DIFF
--- a/src/confluent_kafka/src/Consumer.c
+++ b/src/confluent_kafka/src/Consumer.c
@@ -271,7 +271,9 @@ static PyObject *Consumer_assign (Handle *self, PyObject *tlist) {
 
 	self->u.Consumer.rebalance_assigned++;
 
+        Py_BEGIN_ALLOW_THREADS
 	err = rd_kafka_assign(self->rk, c_parts);
+        Py_END_ALLOW_THREADS
 
 	rd_kafka_topic_partition_list_destroy(c_parts);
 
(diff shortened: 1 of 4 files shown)


src/confluent_kafka/src/Consumer.c
```diff
@@ -271,7 +271,9 @@ static PyObject *Consumer_assign (Handle *self, PyObject *tlist) {
 
 	self->u.Consumer.rebalance_assigned++;
 
+        Py_BEGIN_ALLOW_THREADS
```
jliunyu, line 274: Looks the indentation doesn't match with the previous code.

src/confluent_kafka/src/Consumer.c
```diff
@@ -271,7 +271,9 @@ static PyObject *Consumer_assign (Handle *self, PyObject *tlist) {
...
 	err = rd_kafka_assign(self->rk, c_parts);
+        Py_END_ALLOW_THREADS
```
jliunyu, line 276: ditto

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -224,6 +224,7 @@ typedef struct {
 	PyObject *error_cb;
 	PyObject *throttle_cb;
 	PyObject *stats_cb;
+        PyObject *connect_cb;
```
jliunyu, line 227: ditto

src/confluent_kafka/src/Consumer.c
```diff
@@ -298,7 +300,10 @@ static PyObject *Consumer_unassign (Handle *self, PyObject *ignore) {
 
 	self->u.Consumer.rebalance_assigned++;
 
+        Py_BEGIN_ALLOW_THREADS
```
jliunyu, line 303: ditto

jliunyu (changes requested):
@hallfox, thanks a lot for your change! 

Can you please add test case for the code?

Issue #729, opened 2019-11-29
Error if assign_cb raises Exception


coldeasy:
Description
===========
A `SystemError` exception is raised when calling `poll` or `close` on a consumer if the `on_assign` callback raises an Exception.
```python
    def log(self, level, msg, *args, **kwargs):
        """
        Log 'msg % args' with the integer severity 'level'.

        To pass exception information, use the keyword argument exc_info with
        a true value, e.g.

        logger.log(level, "We have a %s", "mysterious problem", exc_info=1)
        """
>       if not isinstance(level, int):
E       SystemError: <built-in function isinstance> returned a result with an error set
```

I think this is due to the cpython code not clearing the python exception during callbacks and subsequent callbacks executing python code.
Reading the docs, I get the impression that the librdkafka callbacks which execute python code should clear exceptions `PyErr_Clear` and restore them when python code is calling cpython functions (e.g. check and set the error when `consumer.poll` is called). If this is right, we could probably replace the `crashed` counter with the exception info.

How to reproduce
================
```python
    import logging
    logging.basicConfig(level=logging.DEBUG)
    from confluent_kafka import Producer
    from confluent_kafka import Consumer
    prod = Producer(
        {
            'bootstrap.servers': servers,
            'debug': 'all'
        },
        logger=logging.getLogger('')
    )
    prod.produce('test', b'data')
    assert prod.flush(10) == 0
    cons = Consumer(
        {
            'bootstrap.servers': servers,
            'group.id': 'test',
            'auto.offset.reset': 'smallest',
            'debug': 'all'
        },
        logger=logging.getLogger('')
    )

    def _on_assign(self, *args, **kwargs):
        raise Exception("Oops")

    cons.subscribe(['test'], on_assign=_on_assign, on_revoke=_on_assign)
    msg = cons.poll(10)
    assert msg is not None
    cons.close()
```

I can provide the debug logs if necessary.

Checklist
=========
Please provide the following information:

 - [x] `confluent_kafka.version(), confluent_kafka.libversion() == (('1.2.2', 16908800), ('1.2.2', 16909055))`
 - [x] 0.10.2.1
 - [ ] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

coldeasy:
This is the stack trace that's printed to stderr
```
[...]

Issue #865, opened 2020-05-13
SystemError when a callback raises


spenczar:
Description
===========
I'd like to raise an exception when the `error_cb` callback is triggered. If I do so, _two_ exceptions are raised. First the one I asked for, and second is a `SystemError`.

I'm working on a command-line tool. I really want to crash in case of an error.

The `SystemError` is making it difficult for me to write tests around my error reporting. It also makes the error messages much harder to interpret for users of my command-line tool.

How to reproduce
================
I think this is a minimal reproducer - it will try to connect to a broker that doesn't exist so it will trigger the callback:
```python
#!/usr/bin/env python
import confluent_kafka


def demo():
    consumer = confluent_kafka.Consumer({
        "bootstrap.servers": "nonexistent",
        "group.id": "demo",
        "error_cb": error_cb,
    })
    consumer.subscribe(["test"])
    return consumer.consume(1, timeout=1)


def error_cb(error):
    raise(confluent_kafka.KafkaException(error))


if __name__ == "__main__":
    demo()
```

If I run this, this is what I get:
```
Traceback (most recent call last):
  File "minimal_reproducer.py", line 17, in error_cb
    raise(confluent_kafka.KafkaException(error))
cimpl.KafkaException: KafkaError{code=_RESOLVE,val=-193,str="nonexistent:9092/bootstrap: Failed to resolve 'nonexistent:9092': Temporary failure in name resolution (after 9ms in state CONNECT)"}

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "minimal_reproducer.py", line 21, in <module>
    demo()
  File "minimal_reproducer.py", line 13, in demo
    return consumer.consume(1, timeout=1)
  File "minimal_reproducer.py", line 17, in error_cb
    raise(confluent_kafka.KafkaException(error))
SystemError: <class 'cimpl.KafkaException'> returned a result with an error set
```

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):
```
>>> import confluent_kafka
>>> confluent_kafka.version()
('1.4.1', 17039616)
>>> confluent_kafka.libversion()
('1.4.0', 17039615)
```
 - [x] Apache Kafka broker version:
n/a
 - [x] Client configuration: `{...}`
in reproducer above
 - [x] Operating system:
Ubuntu 20.04, Linux 5.4.0
[...]

PR #1146, opened 2021-06-21
Unset and reset error state to avoid triggering SystemErrors

cnweaver:
Currently `error_cb` attempts to handle the case of the callback raising a python error, but this handling breaks down when the callback is invoked multiple times. The problem is that since raising an error simply consists of setting the error state in the interpreter, if this is not cleared the interpreter will treat every subsequent return from C code back to python as having raised an error, which is itself an error if that code (not having actually raised an error) attempts to return a value (including `None` to indicate successful completion). 

The error callback may be invoked multiple times by functions like `Consumer_consume` when retries are performed due to network issues. Given a callback of the form:

	def handle_error(kafka_error: confluent_kafka.KafkaError):
		logger.warning(f"internal kafka error: {kafka_error}")
		if some_condition:
			raise SomeError

Once `some_condition` has caused `SomeError` to be raised, subsequent invocations of the callback will fail in ways likely surprising to the user: A `SystemError` complaining that logger.warning "returned a result with an error set". (`logger.warning` could be any code which ends up calling into a C implmentation, including `print`.) Although this error will chain back to the original `SomeError` as the cause, it alters the control flow of the subsequent callback invocations, potentially bypasing logic that was important to the user. Additionally, the `SystemError`s obscure the original error type, so that attempting to handle it in the obvious way (`try. . . except SomeError`) will not work. 

The proposed solution is that confluent-kafka, instead of leaving the error state set after the callback to propagate when `Consumer_consume`, et al. return, should record any error directly after the callback completes, unset the error state, and then restore any error state before returning to python. `CallState_crash` and `CallState_end` seem to be well-positioned to do this. One possible source of complexity is that repeated invocations of the callback may raise multiple errors. The approach taken in this patch is simply to preserve and propagate the last error raised. It would certainly be possile to preserve all errors, but it does not seem obvious how to usefully communicate multiple errors back to the user.

### DIFF
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -2120,6 +2120,9 @@ void CallState_begin (Handle *h, CallState *cs) {
 	cs->thread_state = PyEval_SaveThread();
 	assert(cs->thread_state != NULL);
 	cs->crashed = 0;
+	cs->err_type = NULL;
+	cs->err_value = NULL;
+	cs->err_traceback = NULL;
 #ifdef WITH_PY_TSS
         PyThread_tss_set(&h->tlskey, cs);
 #else
@@ -2140,8 +2143,15 @@ int CallState_end (Handle *h, CallState *cs) {
 
 	PyEval_RestoreThread(cs->thread_state);
 
-	if (PyErr_CheckSignals() == -1 || cs->crashed)
+	if (PyErr_CheckSignals() == -1 || cs->crashed) {
+		if (cs->err_type) { /* Restore any stored error */
+			PyErr_Restore(cs->err_type, cs->err_value, cs->err_traceback);
+			cs->err_type = NULL;
+			cs->err_value = NULL;
+			cs->err_traceback = NULL;
+		}
 		return 0;
+	}
 
 	return 1;
 }
@@ -2177,6 +2187,22 @@ void CallState_resume (CallState *cs) {
  */
 void CallState_crash (CallState *cs) {
 	cs->crashed++;
+	/* Obtain and clear the current error state */
+	PyObject *err_type, *err_value, *err_traceback;
+	PyErr_Fetch(&err_type, &err_value, &err_traceback);
+	if (err_type) {
+		/* If there was a previously stored error, discard it */
+		if (cs->err_type)
+			Py_DECREF(cs->err_type);
+		if (cs->err_value)
+			Py_DECREF(cs->err_value);
+		if (cs->err_traceback)
+			Py_DECREF(cs->err_traceback);
+		/* Save the new error */
+		cs->err_type = err_type;
+		cs->err_value = err_value;
+		cs->err_traceback = err_traceback;
+	}
 }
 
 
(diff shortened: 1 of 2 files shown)


cnweaver:
[clabot:check]

spenczar:
I think this fixes #729 and #865.

austenwho:
@cnweaver Do you think this issue could be causing issues in instances where the user does NOT raise an exception? I was encountering random SystemError exceptions from builtins, like `list.append()` (which my code was not performing) and also in a call to built-in `sleep()`

In my case I am not using the error handler callback, just the on_assign and on_revoke callbacks. I'm hopeful that your fix might solve my issue as well. 

At first I believed this to be a race condition where the TopicPartitions list was being appended to while I was in the process of reading from the list, I'd get the following SystemError:

Got system error trying to reference partition object: SystemError("<method 'append' of 'list' objects> returned a result with an error set")

Note: My code was not altering the list at all, just iterating over the list of partitions to create a log set of what partitions were being assigned/revoked.

Wrapping the iteration over the TopicPartitions list in a try/except SystemError loop kludged the way through this issue, and I attempted to mitigate the issue by adding a pause in execution of one second with sleep(1), which seemed to work briefly (this is an intermittent issue that is difficult to reproduce, unfortunately), I then started getting the following SystemError instead:

SystemError: <built-in function sleep> returned a result with an error set

It seems that every step I take to mitigate this issue just results in SystemError being produced somewhere else and leaking through...

cnweaver:
It sounds possible, although I think I would need to better understand exactly how your code was structured to really be certain. Was the loop over the list located inside your callback, or somewhere else and to some degree concurrent with the callback?

austenwho:
@cnweaver Actually, I think it was myself that needed to fully understand your explanation of what was going on. You were right that the subsequent raising of SystemError is very confusing and misleading.

In my callbacks I was using an exception as control flow (a thing I don't particularly like, but is Pythonic) to handle resetting poller state when assignments/revocations happened. I misunderstood at first what you meant about _raising_ an exception and _causing_ an exception. I was doing the former, intentionally. I was not doing the latter, and that's where my confusion was.

I was able to fix my problem, specifically, by moving the exception I was raising inside the callback to a flag that would be checked outside the callback's stack, and then raised. So far this seems to work and never taints the Cimpl for subsequent calls to these callbacks.

My solution, which seems to be working without issue so far (all 30 minutes or so):
```python
class MyClass:
    callback_state: PollerException = None

    def on_assign(consumer: Consumer, partitions: List[TopicPartitions]):
        # do stuff with partitions
        # Here I used to simply raise DerivedPollerException(), which was the issue
        self.callback_state = DerivedPollerException("...")

    def check_callback_state():
        if self.callback_state:
            state = self.callback_state
            self.callback_state = None
            raise state

    def poll():
        while True:
            self.check_callback_state()
            # do normal stuff
```

mjuric:
Hi, @cnweaver -- is @austenwho's proposed workaround sufficient?

@confluentinc, is there any way to help test or review this PR so it could potentially get merged?

cnweaver:
@austenwho 's solution is a viable alternative, I think. If it is to be adopted, though, `CallState_crash` should probably still be updated to swallow errors if any are raised, and documentation should be updated to inform users that errors raised from their callbacks will not be propagated, so they should avoid raising intentionally (and instead use a pattern like @austenwho has outlined).

edenhill:
Huhm, this shouldn't be happening as all the callback handlers will call `rd_kafka_yield()` if an exception was raised. The yield will make librdkafka return immediately from whatever poll() call it is running, which in turn should return back to Python and propagate the exception to the user.

PR #1162, opened 2021-07-09
avro: fix primitive and union schema parsing bugs

### LINKED ISSUE #989: Avro Desrialization Error when Schema is primitive type.
Description
===========

When attempting to use AvroDeserializer with a schema that is a primitive type for example "string" a value deserialization message is received.

Versions:
```
>>> confluent_kafka.version()
('1.5.0', 17104896)
>>> confluent_kafka.libversion()
('1.5.0', 17105151)
```
Confluent Platform: 6.0.0
MacOS

How to reproduce
================
Here is the producer code
``` python
from confluent_kafka.schema_registry import SchemaRegistryClient
from confluent_kafka.schema_registry.avro import AvroSerializer
from confluent_kafka.serializing_producer import SerializingProducer


def delivery_report(err, msg):
    if err is not None:
        print("Delivery failed for User record {}: {}".format(msg.key(), err))
        return
    print(
        "User record {} successfully produced to {} [{}] at offset {}".format(
            msg.key(), msg.topic(), msg.partition(), msg.offset()
        )
    )


# The AVRO Schema
schema_str = '{"type": "string"}'
value = "test"
# Schema Registry Client Initiliazation
schema_registry_configuration = {"url": "http://127.0.0.1:8081"}
schema_registry_client = SchemaRegistryClient(schema_registry_configuration)
# Producer Client Initiliazation
avro_serializer_configutation = {"auto.register.schemas": True}
producer_client_configuration = {
    "bootstrap.servers": "localhost:9092",
    "value.serializer": AvroSerializer(
        schema_str=schema_str, schema_registry_client=schema_registry_client, conf=avro_serializer_configutation
    ),
}
producer = SerializingProducer(producer_client_configuration)
# Produce a record with appropriate headers
producer.produce(topic="test", value=value, on_delivery=delivery_report)
# Flush producer
producer.flush()

```

Here is the consumer code:
```python
from confluent_kafka.schema_registry import SchemaRegistryClient
from confluent_kafka.schema_registry.avro import AvroDeserializer
from confluent_kafka.deserializing_consumer import DeserializingConsumer

# The AVRO Schema
schema_str = '{"type": "string"}'

schema_registry_configuration = {
    "url": "http://localhost:8081",
}
schema_registry_client = SchemaRegistryClient(schema_registry_configuration)
serializer = AvroDeserializer(schema_str=schema_str, schema_registry_client=schema_registry_client)
conf = {
    "client.id": "test_3",
    "bootstrap.servers": "localhost:9092",
    "auto.offset.reset": "earliest",
    "value.deserializer": serializer,  # Serializer used for message values.
    "group.id": "test-group",
}
consumer = DeserializingConsumer(conf)
try:
    consumer.subscribe(["test"])
    while True:
        message = consumer.poll(timeout=5.0)
        # id created to track logic through logs
        if message is None:
            continue
        else:
            print(message.value())
        consumer.commit(asynchronous=True)
except Exception as e:
    raise (e)
    print(e)

```

Running 
`python3 producer.py`
Then:
`python3 consumer.py`
We retrieve the following error from the consumer

```
  File "/Users/.../.pyenv/versions/3.8.1/lib/python3.8/site-packages/confluent_kafka/deserializing_consumer.py", line 137, in poll
    value = self._value_deserializer(value, ctx)
  File "/Users/.../.pyenv/versions/3.8.1/lib/python3.8/site-packages/confluent_kafka/schema_registry/avro.py", line 317, in __call__
    writer_schema = parse_schema(loads(
  File "/Users/.../.pyenv/versions/3.8.1/lib/python3.8/json/__init__.py", line 357, in loads
    return _default_decoder.decode(s)
  File "/Users/.../.pyenv/versions/3.8.1/lib/python3.8/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/.../.pyenv/versions/3.8.1/lib/python3.8/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 11 (char 10)

During handling of the above exception, another exception occurred:

  File "consumer.py", line 56, in <module>
    raise (e)
  File "consumer.py", line 39, in <module>
    message = consumer.poll(timeout=5.0)
  File "/Users/.../.pyenv/versions/3.8.1/lib/python3.8/site-packages/confluent_kafka/deserializing_consumer.py", line 139, in poll
    raise ValueDeserializationError(exception=se, kafka_message=msg)
confluent_kafka.error.ValueDeserializationError: KafkaError{code=_VALUE_DESERIALIZATION,val=-159,str="Expecting ',' delimiter: line 1 column 11 (char 10)"}
```
Checklist
=========

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):
 - [x] Apache Kafka broker version:
 - [x] Client configuration: `{...}`
 - [x] Operating system:
 - [x] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Critical issue

dylrich:
Fixes #989 

This commit cleans up a few bugs in the _schema_loads function:

We use primitive types retrieved from the Confluent Registry and encountered an issue where _schema_loads would cause json deserialization errors by double quoting valid primitive declarations. Previous tests included incorrectly specified primitive declarations, according to the Avro spec primitive declarations are valid JSON documents, but they had been specified as strings of their type name with no quoting. I fixed the tests as well as the issue in _schema_loads

Somewhat separately, there was also an issue with Avro union types. _schema_loads was incorrectly causing json serialization errors for unions because it included them on accident with its special-casing of primitive declarations. I added a check for json arrays to exclude them from the special casing. I also had to add a check later to ensure the _schema_name property was special-cased to None for unions. This should have no impact on names in the registry because _schema_name isn't used at all for the recommended subject name strategy with unions.

### DIFF
--- a/src/confluent_kafka/schema_registry/avro.py
+++ b/src/confluent_kafka/schema_registry/avro.py
@@ -62,8 +62,8 @@ def _schema_loads(schema_str):
     schema_str = schema_str.strip()
 
     # canonical form primitive declarations are not supported
-    if schema_str[0] != "{":
-        schema_str = '{"type":"' + schema_str + '"}'
+    if schema_str[0] != "{" and schema_str[0] != "[":
+        schema_str = '{"type":' + schema_str + '}'
 
     return Schema(schema_str, schema_type='AVRO')
 
@@ -187,11 +187,20 @@ class AvroSerializer(Serializer):
         schema = _schema_loads(schema_str)
         schema_dict = loads(schema.schema_str)
         parsed_schema = parse_schema(schema_dict)
-        # The Avro spec states primitives have a name equal to their type
-        # i.e. {"type": "string"} has a name of string.
-        # This function does not comply.
-        # https://github.com/fastavro/fastavro/issues/415
-        schema_name = parsed_schema.get('name', schema_dict['type'])
+
+        if isinstance(parsed_schema, list):
+            # if parsed_schema is a list, we have an Avro union and there
+            # is no valid schema name. This is fine because the only use of
+            # schema_name is for supplying the subject name to the registry
+            # and union types should use topic_subject_name_strategy, which
+            # just discards the schema name anyway
+            schema_name = None
+        else:
+            # The Avro spec states primitives have a name equal to their type
+            # i.e. {"type": "string"} has a name of string.
+            # This function does not comply.
+            # https://github.com/fastavro/fastavro/issues/415
+            schema_name = parsed_schema.get("name", schema_dict["type"])
 
         self._schema = schema
         self._schema_name = schema_name
--- a/tests/schema_registry/test_avro_serializer.py
+++ b/tests/schema_registry/test_avro_serializer.py
@@ -34,7 +34,7 @@ def test_avro_serializer_config_auto_register_schemas():
     """
     conf = {'url': TEST_URL}
     test_client = SchemaRegistryClient(conf)
-    test_serializer = AvroSerializer(test_client, 'string',
+    test_serializer = AvroSerializer(test_client, '"string"',
                                      conf={'auto.register.schemas': False})
     assert not test_serializer._auto_register
 
@@ -60,7 +60,7 @@ def test_avro_serializer_config_auto_register_schemas_false(mock_schema_registry
     topic = "test-auto-register"
     subject = topic + '-key'
 
-    test_serializer = AvroSerializer(test_client, 'string',
+    test_serializer = AvroSerializer(test_client, '"string"',
                                      conf={'auto.register.schemas': False})
 
     test_serializer("test",
@@ -83,7 +83,7 @@ def test_avro_serializer_config_use_latest_version(mock_schema_registry):
     topic = "test-use-latest-version"
     subject = topic + '-key'
 
-    test_serializer = AvroSerializer(test_client, 'string',
+    test_serializer = AvroSerializer(test_client, '"string"',
                                      conf={'auto.register.schemas': False, 'use.latest.version': True})
 
     test_serializer("test",
@@ -104,7 +104,7 @@ def test_avro_serializer_config_subject_name_strategy():
 
     conf = {'url': TEST_URL}
     test_client = SchemaRegistryClient(conf)
-    test_serializer = AvroSerializer(test_client, 'int',
+    test_serializer = AvroSerializer(test_client, '"int"',
                                      conf={'subject.name.strategy':
                                            record_subject_name_strategy})
 
@@ -119,7 +119,7 @@ def test_avro_serializer_config_subject_name_strategy_invalid():
     conf = {'url': TEST_URL}
     test_client = SchemaRegistryClient(conf)
     with pytest.raises(ValueError, match="must be callable"):
-        AvroSerializer(test_client, 'int',
+        AvroSerializer(test_client, '"int"',
                        conf={'subject.name.strategy': dict()})
 
 
@@ -148,7 +148,7 @@ def test_avro_serializer_record_subject_name_strategy_primitive(load_avsc):
     """
     conf = {'url': TEST_URL}
     test_client = SchemaRegistryClient(conf)
-    test_serializer = AvroSerializer(test_client, 'int',
+    test_serializer = AvroSerializer(test_client, '"int"',
                                      conf={'subject.name.strategy':
                                            record_subject_name_strategy})
 
@@ -181,7 +181,7 @@ def test_avro_serializer_topic_record_subject_name_strategy_primitive(load_avsc)
     """
     conf = {'url': TEST_URL}
     test_client = SchemaRegistryClient(conf)
-    test_serializer = AvroSerializer(test_client, 'int',
+    test_serializer = AvroSerializer(test_client, '"int"',
                                      conf={'subject.name.strategy':
                                            topic_record_subject_name_strategy})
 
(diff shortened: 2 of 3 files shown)


jliunyu:
Then change looks good to me. 

But we need to wait for CI test to merge the code.

beaal:
@dylrich can you please add https://github.com/confluentinc/confluent-kafka-python/issues/989 to the Linked Issues?

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -187,11 +187,20 @@ def __init__(self, schema_registry_client, schema_str,
         schema = _schema_loads(schema_str)
         schema_dict = loads(schema.schema_str)
         parsed_schema = parse_schema(schema_dict)
-        # The Avro spec states primitives have a name equal to their type
-        # i.e. {"type": "string"} has a name of string.
-        # This function does not comply.
-        # https://github.com/fastavro/fastavro/issues/415
-        schema_name = parsed_schema.get('name', schema_dict['type'])
+        
+        # if parsed_schema is a list, we have an Avro union and there
```
edenhill, line 191: If

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -62,8 +62,8 @@ def _schema_loads(schema_str):
     schema_str = schema_str.strip()
 
     # canonical form primitive declarations are not supported
-    if schema_str[0] != "{":
-        schema_str = '{"type":"' + schema_str + '"}'
+    if schema_str[0] != "{" and schema_str[0] != "[":
+        schema_str = '{"type":' + schema_str + '}'
```
edenhill, line 66: The code previously added quotes around schema_str, but now does not.

src/confluent_kafka/schema_registry/avro.py
edenhill: Put this comment within the if clause.

tests/schema_registry/test_avro_serializer.py
```diff
@@ -34,7 +34,7 @@ def test_avro_serializer_config_auto_register_schemas():
     """
     conf = {'url': TEST_URL}
     test_client = SchemaRegistryClient(conf)
-    test_serializer = AvroSerializer(test_client, 'string',
+    test_serializer = AvroSerializer(test_client, '"string"',
```
edenhill, line 37: This seems like a breaking change. Let's keep the old behaviour.

edenhill (changes requested):

src/confluent_kafka/schema_registry/avro.py
dylrich: Yes, I believe this was actually a bug. According to the Avro spec, schemas should be valid JSON documents, so putting quotes around the schema string duplicates double quotes for canonical form primitives. We encountered this issue when pulling schemas from our actual registry instance. This is a breaking change, but I don't believe the previous behavior was correct. If you don't want to break compatibility we could check for the first character being a double quote and conditionally add quotes based on that check?

dylrich:
I no longer work with Kafka, so I can't help test fixes for this issue anymore. However, I still believe this is a bug with the current client. If there's no interest in this patch as is I will close the pull request and leave the patch in a comment in case someone else needs to monkey patch this bug.

msinto93:
What's stopping this PR from being merged? Issue #989 is a clear bug which this PR will fix. I notice the PR author has said they can no longer contribute to this, so if it is not possible to merge in its current state then I am happy to take it over.

mhowlett:
@dylrich - leave it open, we'll get to it. thanks for the PR :-)

src/confluent_kafka/schema_registry/avro.py
mhowlett: I just tested registering the schema `string`, with and without quotes in Confluent Schema Registry. It rejects the version without quotes (because it is not a valid avro schema) and accepts the version with quotes. I don't see how there would be any scenario in where there is working code relying on a `schama_str` for a primitive type doesn't have quotes, so this fixes a bug and doesn't introduce an incompatibility.

src/confluent_kafka/schema_registry/avro.py
mhowlett: 👍 would be good, but not deal breaker for me for merging if it's the only holding things up.

tests/schema_registry/test_avro_serializer.py
mhowlett: i'm quite sure this is strictly a bug fix, not a breaking change, see previous comment.

mhowlett (approved):
i've now considered this carefully, and it looks good to me, thanks for the fix. i'll delay merging though as a matter of protocol as @edenhill 's issue with it is technically still unresolved.

edenhill:
Thank you!

raphaelauv:
There is still no new released version with that fix , so if you need it you can monkey_patch

```python
def custom(schema_str):
    schema_str = schema_str.strip()
    # canonical form primitive declarations are not supported
    if schema_str[0] != "{" and schema_str[0] != "[":
        schema_str = '{"type":' + schema_str + '}'
    return Schema(schema_str, schema_type='AVRO')

import confluent_kafka.schema_registry.avro as monkey_patch
monkey_patch._schema_loads = custom

from confluent_kafka.schema_registry.avro import AvroDeserializer

...
```

PR #1167, opened 2021-07-20
Implement message headers in callbacks

cnweaver:
Adding this support to `Message_new0` seems desirable to give consistent support everywhere, but might have broader implications which are not obvious to me, in which case I am open to suggestions about how else this feature should be implemented instead.


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -792,6 +792,12 @@ PyObject *Message_new0 (const Handle *handle, const rd_kafka_message_t *rkm) {
         else
                 self->latency = -1;
 
+#ifdef RD_KAFKA_V_HEADERS
+    rd_kafka_headers_t* hdrs;
+    if (rd_kafka_message_headers(rkm, &hdrs) == RD_KAFKA_RESP_ERR_NO_ERROR)
+        self->c_headers = rd_kafka_headers_copy(hdrs);
```
edenhill, line 798: Object creation is very costly in Python, and headers are object-rich, which results in a performance degradation.
We thus don't want to have headers available by default. In the Go and Dotnet clients we have configuration to enable headers in delivery reports, which we could do here too.
But generally I think it is better for the application to pass and bind variables to its delivery callback/lambda with whatever extra per-messsage information it needs, rather than serializing and deserializing the same information through the Kafka client.

cnweaver: Thanks, binding a variable into the callback is a good (simpler) solution for my particular use; I hadn't realized it was an option for dumb reasons having to do with another intervening library and not reading documentation carefully enough. In terms of speed it's no different for me (one extra object created either way), but I can certainly understand that people dealing with high throughput and using lots of headers might want to skip getting back headers they don't need. I can try to look at one of the other clients and add a similar switch to this patch, but I think this is no longer a blocking concern for my own work.

PR #1193, opened 2021-09-02
Don't call `PyEval_InitThreads()` when using Python >= 3.9

chrisnovakovic:
`PyEval_InitThreads()` is a no-op and is deprecated as of Python 3.9 (and will be removed in Python 3.11) - only call it when using earlier Python versions.


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2542,7 +2542,9 @@ static struct PyModuleDef cimpl_moduledef = {
 static PyObject *_init_cimpl (void) {
 	PyObject *m;
 
-        PyEval_InitThreads();
+#if PY_VERSION_HEX < 0x03090000
```
edenhill, line 2545: Seems like it is only needed in 3.6 and earlier, so perhaps change this to `< 0x..3.7.0`

chrisnovakovic: I don't know enough about the CPython API to know whether I could get away with that, but that's good to hear, thanks - I'll update it in a second.

edenhill: Should be fine:
https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads

chrisnovakovic:
[clabot:check]

src/confluent_kafka/src/confluent_kafka.c
chrisnovakovic: Great, thanks - I've fixed the commit message too.

edenhill:
Can you force-push to this PR again to trigger CI jobs?

chrisnovakovic:
Rebased on master.

PR #1205, opened 2021-09-29
Update Jenkinsfile Nodelabel

jkao97:
Update Jenkinsfile Nodelabel to use new build image.

Job is still active but doesn't seem to run correctly anymore: https://jenkins.confluent.io/job/system-test-python-client/job/master/. -> still making change in case we want to fix, if not we should deprecate/remove job.


edenhill (approved):
LGTM

jkao97:
@edenhill can you help merge/give tools admin to this repo. It seems I do not have write access to merge.

Issue #1197, opened 2021-09-13
`UnknownType` error when producing using a schema without namespace


ffissore:
Description
===========
When a schema has no namespace defined and it defines a type which it later refers, fastavro raises exception `UnknownType`

How to reproduce
================
Given this schema
```
{
    "name": "HelloWorldValue",
    "type": "record",
    "fields": [
        {
            "name": "hello",
            "type": "string"
        },
        {
            "name": "innerhello",
            "type": [
                "null",
                "HelloWorldValue"
            ],
            "default": null
        }
    ]
}
```
load it with `avro.load` and pass it to a `produce` call, like so
```
producer.produce(
    topic=topic,
    value=dict(hello='world'),
    value_schema=broken_schema
)
```
`fastavro` will raise exception `fastavro._schema_common.UnknownType: .HelloWorldValue`

This happens because of the different way `avro-python` and `fastavro` work. In `MessageSerializer._get_encoder_func`, `RecordSchema.to_json()` is called, which produces a string where the inner reference to `HelloWorldValue` starts with a dot: `.HelloWorldValue`. When `fastavro` parses such string before encoding the record, it fails to lookup type `.HelloWorldValue`.

Adding a namespace makes it work however schema registry doesn't allow us to add it as it's not a `FULL` compatible change.

Another workaround is to change how `MessageSerializer._get_encoder_func` calls `to_json`, like so:
```
schema = writer_schema.to_json(names=Names(default_namespace=""))
```
However, it's not possible to customise that function from the outside.

Checklist
=========
Please provide the following information:

 - [X] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): this was tested with both confluent-kafka 1.6.0 and latest fastavro 1.4.4, and with confluent-kafka 1.0.1 and fastavro 0.21.17
 - [ ] Apache Kafka broker version:
 - [ ] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

ffissore:
To be clearer, this is the result of calling `to_json()` without arguments
```
{'type': 'record', 'name': 'HelloWorldValue', 'fields': [{'type': 'string', 'name': 'hello'}, {'type': ['null', '.HelloWorldValue'], 'name': 'innerhello', 'default': None}]}
```
[...]

PR #1208, opened 2021-09-30
Use `str(Schema)` rather than `Schema.to_json`

### LINKED ISSUE #1156: schema parser causes conflict with fastavro
Description
===========
The code in the reproduction section below uses Avro schema encoder in order to encode a message, which succeeds only if we transform the resulting schema encoding by getting rid of the `MappingProxyType`. The transformation basically leaves everything unchanged except altering `MappingProxyType` to `dict` instances.

1. Is there a problem in the way I am calling the standard library which causes mapping proxy to be used, which in turn causes `fastavro` to throw? Can this be fixed by something as a user, or is this really a bug in the Confluent Kafka library?
2. In addition, the output `schemaId` from `registryClient.get_latest_schema()` is marked in the docs to return `str` but returns `int`. If I understand correctly, this is the intended input into the `schema_id` parameter of `serializer.encode_record_with_schema_id()` (and it works correctly if I call it), which is also marked as `int`. Is that a typo in the docs? In other words, it seems either `registryClient.get_latest_schema()` should return an integer, or `serializer.encode_record_with_schema_id()` should take a string, or I am doing something incorrectly :) Which one is it?

Thank you very much.

Environment
===========
I am running Python 3.9 with Confluent Kafka 1.7.0, `avro-python3` 1.10.0 and `fastavro` 1.4.1. Currently running on Windows, but also intend to deliver and support on Linux. 

    confluent_kafka.version() == ('1.7.0', 17235968)
    confluent_kafka.libversion() == ('1.7.0', 17236223)

How to reproduce
================

    from confluent_kafka      import Producer
    from confluent_kafka.avro import CachedSchemaRegistryClient, MessageSerializer
    
    from fastavro.schema import parse_schema
    from fastavro.validation import validate
    
    from types  import MappingProxyType
    from typing import Any
    
    import sys
    
    def transformMap(item: Any) -> Any:
        if type(item) in {dict, MappingProxyType}:
            return {k:transformMap(v) for k,v in item.items()}
        elif type(item) is list:
            return [transformMap(v) for v in item]
        else:
            return item
    
    def main(argv = None):
        msgType = 'InstrumentIdMsg'
        idFigi  = 'BBG123456789'
        head = {'sDateTime': 1, 'msgType': msgType,      'srcSeq': 1,
                'rDateTime': 1, 'src':     'Brownstone', 'reqID':  None,
                'sequence':  1}
        msgMap = {'head': head, 'product':  'Port', 'idIsin': None, 'idFigi': idFigi,
                  'idBB': None, 'benchmark': None,  'idCusip': None,'idCins': None}
    
        registryClient = CachedSchemaRegistryClient(url = 'http://local.KafkaRegistry.com:8081')
        schemaId, schema, version = registryClient.get_latest_schema(msgType)
    
        serializer = MessageSerializer(registry_client = registryClient)
    
        schemaMap = schema.to_json()

        # NOTE:
        # schemaMap cannot be used since it uses mappingproxy
        # which causes validate() and parse_schema() to throw
        schemaDict = transformMap(schemaMap)
    
        isValid = validate(datum = msgMap, schema = schemaDict, raise_errors = True)
        parsed_schema = parse_schema(schema = schemaDict)
    
        msg = serializer.encode_record_with_schema_id(schema_id = schemaId,
                                                      record    = msgMap)
    
        producer = Producer({'bootstrap.servers': 'kafkaServer:9092'})
        producer.produce(key = idFigi,
                         topic = 'TOPIC_NAME',
                         value = msg)
        return 0
    
    
    if __name__ == '__main__':
        sys.exit(main())

Checklist
=========

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):
 - [x] Client configuration: `{...}`
 - [x] Operating system:
 - [ ] Critical issue

ffissore:
Use `str(Schema)` rather than `Schema.to_json` to prevent `fastavro` from raising exception `TypeError: unhashable type: 'mappingproxy'`

Fixes #1156

When you try to produce a message with a schema such as
```json
{
    "name": "ComplexValue",
    "type": "record",
    "fields": [
        {
            "name": "primitive_field",
            "type": "string"
        },
        {
            "name": "optional_recursive_field",
            "type": [
                "null",
                "ComplexValue"
            ],
            "default": null
        },
        {
            "name": "logical_type_field",
            "type": [
                "null",
                {
                    "type": "long",
                    "logicalType": "timestamp-millis"
                }
            ],
            "default": 1
        }
    ]
}
```
`fastavro` raises `E   TypeError: unhashable type: 'mappingproxy'`

This is caused by the wrong json produced by `Schema.to_json`. Since `avro-python3` version 1.9.0 ([according to this JIRA](https://issues.apache.org/jira/browse/AVRO-2411)), the recommended way of serialising a schema to json is to use `str(schema)` rather than `schema.to_json()`

The code in this PR calls `str` and `json.loads` to have a proper `dict` instance that `fastavro` will be able to handle.

### DIFF
--- a/src/confluent_kafka/avro/serializer/message_serializer.py
+++ b/src/confluent_kafka/avro/serializer/message_serializer.py
@@ -20,6 +20,7 @@
 # derived from https://github.com/verisign/python-confluent-schemaregistry.git
 #
 import io
+import json
 import logging
 import struct
 import sys
@@ -79,7 +80,7 @@ class MessageSerializer(object):
     # Encoder support
     def _get_encoder_func(self, writer_schema):
         if HAS_FAST:
-            schema = writer_schema.to_json()
+            schema = json.loads(str(writer_schema))
             parsed_schema = parse_schema(schema)
             return lambda record, fp: schemaless_writer(fp, parsed_schema, record)
         writer = avro.io.DatumWriter(writer_schema)
@@ -175,8 +176,8 @@ class MessageSerializer(object):
         if HAS_FAST:
             # try to use fast avro
             try:
-                fast_avro_writer_schema = parse_schema(writer_schema_obj.to_json())
-                fast_avro_reader_schema = parse_schema(reader_schema_obj.to_json())
+                fast_avro_writer_schema = parse_schema(json.loads(str(writer_schema_obj)))
+                fast_avro_reader_schema = parse_schema(json.loads(str(reader_schema_obj)))
                 schemaless_reader(payload, fast_avro_writer_schema)
 
                 # If we reach this point, this means we have fastavro and it can
--- a/src/confluent_kafka/avro/serializer/message_serializer.py
+++ b/src/confluent_kafka/avro/serializer/message_serializer.py
@@ -177,7 +177,10 @@ class MessageSerializer(object):
             # try to use fast avro
             try:
                 fast_avro_writer_schema = parse_schema(json.loads(str(writer_schema_obj)))
-                fast_avro_reader_schema = parse_schema(json.loads(str(reader_schema_obj)))
+                if reader_schema_obj is not None:
+                    fast_avro_reader_schema = parse_schema(json.loads(str(reader_schema_obj)))
+                else:
+                    fast_avro_reader_schema = None
                 schemaless_reader(payload, fast_avro_writer_schema)
 
                 # If we reach this point, this means we have fastavro and it can
--- a/tests/avro/adv_schema.avsc
+++ b/tests/avro/adv_schema.avsc
@@ -56,6 +56,17 @@
                 "type" : "map",
                 "values" : "basicPerson"
             }
+        },
+        {
+            "name": "timestamp",
+            "type": [
+                "null",
+                {
+                    "type": "long",
+                    "logicalType": "timestamp-millis"
+                }
+            ],
+            "default": null
         }
     ]
 }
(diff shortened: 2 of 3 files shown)


ffissore:
I reviewed the first commit to include changes on the decoding side, which was also affected but that silently failed because wrapped in a `try...except`

I also added a `None` check, since both `self.reader_key_schema` and `self.reader_value_schema` can be `None` and the `reader_schema` arg [is optional](https://fastavro.readthedocs.io/en/latest/reader.html?highlight=schemaless#fastavro._read_py.schemaless_reader)

PS: travis build fails for an unrelated reason, something about libcrypto

ffissore:
This will also fix #1197: by using recent dependencies, with the confluent-kafka code fixed as per this PR, our tests pass with no other changes required

ffissore:
any luck in restoring CI @edenhill ? having this change available in a release would allow us to update the library and benefit from all the good stuff that was added

ffissore:
I extended the "advanced schema" in the test suite to include a logical type. If you run the test without the fix, you'll get `TypeError: unhashable type: 'mappingproxy'`

ffissore:
Force pushed the make the changes compatible with latest `master`

edenhill:
Thank you!

gsalluzzo-tc:
Hi there,
any idea on when the next release will be shipped to PyPI?
Of course I am interested in the fix for this issue.

Thanks!

mindflayer:
Hi there,
any idea on when the next release will be shipped to PyPI?
Of course I am interested in the fix for this issue.

Thanks!

PR #1212, opened 2021-10-03
Update avro dependency

jairhenrique:
https://avro.apache.org/docs/1.10.2/gettingstartedpython.html

> Notice for Python 3 users
> A package called "avro-python3" had been provided to support Python 3 previously, but the codebase was consolidated into the "avro" package and that supports both Python 2 and 3 now. The avro-python3 package will be removed in the near future, so users should use the "avro" package instead. They are mostly API compatible, but there's a few minor difference (e.g., function name capitalization, such as avro.schema.Parse vs avro.schema.parse).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ SCHEMA_REGISTRY_REQUIRES = ['requests']
 
 AVRO_REQUIRES = ['fastavro>=0.23.0,<1.0;python_version<"3.0"',
                  'fastavro>=1.0;python_version>"3.0"',
-                 'avro==1.10.0'
+                 'avro>=1.11.1,<2',
                  ] + SCHEMA_REGISTRY_REQUIRES
 
 JSON_REQUIRES = ['pyrsistent==0.16.1;python_version<"3.0"',
--- a/src/confluent_kafka/avro/requirements.txt
+++ b/src/confluent_kafka/avro/requirements.txt
@@ -1,3 +1,3 @@
 fastavro>=0.23.0
 requests
-avro==1.10.0
+avro>=1.11.1,<2
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -5,6 +5,6 @@ pytest-timeout
 requests-mock
 trivup>=0.8.3
 fastavro
-avro==1.10.0
+avro>=1.11.1,<2
 jsonschema
 protobuf


jairhenrique:
@edenhill

edenhill:
Good find!

We have some issues with CLA signing at the moment so we can't merge this just yet, will ping you when CLA bot is back up again.

jairhenrique:
> Good find!
> 
> We have some issues with CLA signing at the moment so we can't merge this just yet, will ping you when CLA bot is back up again.

@edenhill is it ok now?

setup.py
```diff
@@ -27,7 +27,7 @@
 
 AVRO_REQUIRES = ['fastavro>=0.23.0,<1.0;python_version<"3.0"',
                  'fastavro>=1.0;python_version>"3.0"',
-                 'avro==1.10.0'
+                 'avro==1.11.1',
```
choogeboom, line 30:
```suggestion
                 'avro>=1.11.1,<2.0.0',
```
Unless there's some reason we need to use 1.11.1, It's much better if you use a range strategy for specifying dependencies so that users of this library have more freedom to update their dependencies independently.

For example, I'm currently in a scenario where I am stuck using `avro==1.10.0` because that is the currently pinned version of the library. Unfortunately, there is a bug in version 1.10.0 that was fixed in 11.1.1. So now I need to wait for this library to update its dependencies in order to have my code function properly.

jairhenrique: :+1:

choogeboom:
@edenhill Hey, this PR is critical for my work to use any version of `confluent-kafka` past version 1.5.0, since there is a [bug in avro v1.10.0](https://issues.apache.org/jira/browse/AVRO-3218) that is fixed in avro 1.11.1. Can we get this PR merged and released?

If there's anything I can do to accelerate this process, please let me know.

choogeboom:
Hey @edenhill this is a pretty trivial PR, but it would solve a lot of problems for me (and my team at Wayfair) if it were merged/released. Can you _pleeeeeeaaaaasee_ put it on your short list?

setup.py
```diff
@@ -27,7 +27,7 @@
 
 AVRO_REQUIRES = ['fastavro>=0.23.0,<1.0;python_version<"3.0"',
                  'fastavro>=1.0;python_version>"3.0"',
-                 'avro==1.10.0'
+                 'avro>=1.11.1,<2',
```
edenhill, line 30: Also need to update tests/requirements.txt and src/confluent_kafka/avro/requirements.txt

edenhill (changes requested):

edenhill (approved):
LGTM, let's wait for CI to pass before merging.

Thank you!

choogeboom:
@edenhill Hey, I've been waiting over 2 months for a release with this commit. Any idea when the next release is going to be?

PR #1217, opened 2021-10-06
single schema object for to be returned for all method

avanishtiwari38:
Currently `SchemaRegistryClient` method have different schema object for different method for example:  `get_latest_version` return you schema with references as array of json as below
```
[{'name': 'com.test.common.event.Eventdata', 'subject': 'com.test.common.event.Eventdata', 'version': 5}, {'name': 'com.test.schema.user.Details', 'subject': 'com.test.schema.user.Details', 'version': 2}]
```
While `get_schema` method return you schema with references as array of `SchemaReference` object as below
```
[<confluent_kafka.schema_registry.schema_registry_client.SchemaReference object at 0x1037a9790>, <confluent_kafka.schema_registry.schema_registry_client.SchemaReference object at 0x1037a97f0>]
```

Which get it difficult to use like need to make different function one for to get schema id and other to get schema object with id that has `SchemaReference` object.

As it is more like to fetch schema with subject name rather than `schema_id` because we can map our schema with topic and reuse topic name to fetch the schema.

So this PR will make a single universal schema object that will be returned by any method call.

### DIFF
--- a/src/confluent_kafka/schema_registry/schema_registry_client.py
+++ b/src/confluent_kafka/schema_registry/schema_registry_client.py
@@ -677,7 +677,7 @@ class Schema(object):
 
         self.schema_str = schema_str
         self.schema_type = schema_type
-        self.references = self.schema_references(references)
+        self.references = [SchemaReference(name=ref['name'], subject=ref['subject'], version=ref['version']) for ref in references]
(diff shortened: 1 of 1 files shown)


rayokota:

PR #1228, opened 2021-10-26
Correct headers for serializer/deserializer docs.

wildcardops:
Deserializers were under the heading of Serializer. Serializers were under the heading of Deserializer. Aligned serializers and deserializers under matching headers.


edenhill:
[clabot:check]

PR #1240, opened 2021-11-23
Add reduce method to KafkaError and change tp_name

OneRaynyDay:
According to CPython documentation: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_name
> For example, a type named T defined in module M in subpackage Q in package P should have the tp_name initializer "P.Q.M.T".

Here, the module we're talking about is `cimpl`, and the package is `confluent_kafka`, so therefore it should be `confluent_kafka.cimpl.KafkaError` and such for the rest of the classes defined.

Also, I add __reduce__ for `KafkaError` so we can properly pickle it:

```
>>> import confluent_kafka
>>> confluent_kafka.cimpl.KafkaError(1).__reduce__()
(<class 'confluent_kafka.cimpl.KafkaError'>, (1, 'Broker: Offset out of range', 0, 0, 0))
```

### DIFF
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -71,6 +71,13 @@ typedef struct {
 
 static void cfl_PyErr_Fatal (rd_kafka_resp_err_t err, const char *reason);
 
+static PyObject *
+KafkaError_reduce (PyObject *self, PyObject *Py_UNUSED(ignored))
+{
+    KafkaError *error = (KafkaError*)self;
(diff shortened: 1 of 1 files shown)


OneRaynyDay:
Also happy to sign a CLA if necessary.

joaoe:
> According to CPython documentation: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_name

Nice catch. I think that affects all classes implemented in cimpl.

joaoe:
I also added another PR which also includes support for "reducing" Message objects, which end up benefiting from this pull request, if the Message has error set.

PR #1250, opened 2021-12-10
1.8.0: Use str(Schema) rather than Schema.to_json

ffissore:
This is the same as #1208 but targets branch `rel180`


ffissore:
Force pushed to make the changes compatible with latest `rel180` branch

edenhill:
The rel180 branch is dead. I merged your other PR instead.
Thank you

PR #1272, opened 2022-01-22
updated readme code example

phacic:
Updated producer code example to call `.produce()` before calling `.poll()` because calling `.poll()` first without produce hangs

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,14 @@ def delivery_report(err, msg):
         print('Message delivered to {} [{}]'.format(msg.topic(), msg.partition()))
 
 for data in some_data_source:
-    # Trigger any available delivery report callbacks from previous produce() calls
-    p.poll(0)
-
     # Asynchronously produce a message, the delivery report callback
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,13 +41,14 @@ def delivery_report(err, msg):
         print('Message delivered to {} [{}]'.format(msg.topic(), msg.partition()))
 
 for data in some_data_source:
-    # Trigger any available delivery report callbacks from previous produce() calls
-    p.poll(0)
```
edenhill, line 45: poll(0) should not block (0 is the timeout).
Are you sure you're seeing this blocking? If so, for how long?

phacic: It's an indefinite block

edenhill: I can't reproduce that.

Can you add a print statement before and after the poll(0) call in the original position and one just prior to flush() below, then reproduce and provide the output.

phacic: I guess i was missing the `(0)` thanks for pointing it out

PR #1273, opened 2022-01-24
Update Jenkinsfile to use maven settings helper

gracechensd:
Use the maven settings helper function [withMavenSettings()](https://github.com/confluentinc/jenkins-common/pull/415) which avoids including hardcoded credentials in our vault secrets.

### DIFF
--- a/tests/system/Jenkinsfile
+++ b/tests/system/Jenkinsfile
@@ -8,16 +8,19 @@ def config = jobConfig {
 }
 
 def job = {
+    def mavenSettingsFile = "${env.WORKSPACE_TMP}/maven-global-settings.xml"
+
     configureGitSSH("github/confluent_jenkins", "private_key")
-    def mavenSettingsFile = "/home/jenkins/.m2/settings.xml"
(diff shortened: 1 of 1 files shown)


tests/system/Jenkinsfile
```diff
@@ -9,15 +9,16 @@ def config = jobConfig {
 
 def job = {
     configureGitSSH("github/confluent_jenkins", "private_key")
+    def mavenSettingsFile = "/home/jenkins/.m2/settings.xml"
```
msch3, line 12: Is this always going to be the case? Would there be a better location in this file for this constant?

gracechensd: Honestly I don't feel super good about it being in a non-temp location so I might want to change it to WORKSPACE_TMP regardless of what it was before..
As for location in this file, it'll have to stay within the `job {}` closure if we want to use/anticipate there being any Jenkins env variables

msch3 (approved):

edenhill (approved):
LGTM

Issue #1078, opened 2021-04-01
Backwards incompatible change in v1.6.1


Incognito:
Description
===========

The argument order in the constructor signature for AvroSerializer was altered in v1.6.1 

```
-    def __init__(self, schema_str, schema_registry_client,
+    def __init__(self, schema_registry_client, schema_str,
```

[link to diff](https://github.com/confluentinc/confluent-kafka-python/compare/v1.6.0...v1.6.1#diff-cefb407efbfa0dc0896f90edd1f27281b539205e51f8f06ed4794635ebfad736R274)

It's an unexpected change for teams who want a simple bugfix, and perhaps can't even modify some internals if the package is in-directed elsewhere (like some wrapper package for example)

I think any one of these options would make this less painful for package consumers:

- revoke tag 1.6.1 and release a backwards-compatible 1.6.2 (it may be too late, especially if people already upgraded)
- Update all docs to exclusively use kwargs as a best-practice to avoid this issue in the future.
- Enforce kwargs in the next major
- Add release linting tools to the CI pipeline that inspects changes and prevents backwards-incompatible changes in patch releases (I could try to help with it if you want to use some AST parsing).

How to reproduce
================

Upgrade to the latest patch level.

dangreenisrael:
Kind of related: A static analysis tool that could detect backwards incompatible changes in CI would be awesome. Although, building it one off does seem out of scope for this project.

rterry123:
Thanks for brining up this issue. The docs show the original argument order and I had no idea what I was doing wrong.

mhowlett:
apologies - it was me who let that through in #1000. the rational was this is an improvement, AvroSerializer is designed to be used with SerializingProducer, and that class is still marked as experimental (apologies also that that is dragging out so long).

the docs are auto-generated from the code, and are currently at v1.5. we evidently need to get that automated as part of the release process as well.

leaving this issue open for the list of suggestions.

Incognito:
Thanks for the background, I think it's a common thing that happens in lots of software packages.

There's another few possibilities I didn't mention above (I don't like these options as much).

- There could be tests for everything that is considered public API, but it also means that the person doing the change has to be in the mindset of "this test isn't to check that it functions, its to check that it is backwards compatible", usually this doesn't scale so I don't like it too much.
- Two packages, one for stable code and one for experimental code. I haven't read all the code in confluent-kafka-python so I don't how how realistic that is, it also adds development overhead.
[...]

PR #1282, opened 2022-02-14
Fix example for #1078

rmoff:
This was raised in the community [here](https://stackoverflow.com/questions/71035716/deserializing-json-data-from-kafka-stream-kafka-connect-to-postgresql/71036423?noredirect=1#comment125632127_71036423). I've made what I _think_ is the correct fix, but please carefully review to verify :)


jliunyu:
@rmoff , thanks for providing this fix. The change looks good to me.

jliunyu (approved):

PR #1286, opened 2022-02-22
ACL operations

emasab:
Python bindings for the CreateAcls, DescribeAcls and DeleteAcls operations.

needs the 1.9.0 version here
https://github.com/edenhill/librdkafka/pull/3741

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,8 +27,7 @@ SCHEMA_REGISTRY_REQUIRES = ['requests']
 
 AVRO_REQUIRES = ['fastavro>=0.23.0,<1.0;python_version<"3.0"',
                  'fastavro>=1.0;python_version>"3.0"',
-                 'avro==1.10.0;python_version<"3.0"',
-                 'avro-python3==1.10.0;python_version>"3.0"'
+                 'avro==1.10.0'
                  ] + SCHEMA_REGISTRY_REQUIRES
 
 JSON_REQUIRES = ['pyrsistent==0.16.1;python_version<"3.0"',
--- a/src/confluent_kafka/admin/__init__.py
+++ b/src/confluent_kafka/admin/__init__.py
@@ -1,7 +1,34 @@
+# Copyright 2022 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """
 Kafka admin client: create, view, alter, and delete topics and resources.
 """
-from ..cimpl import (KafkaException, # noqa
+import concurrent.futures
+
+# Unused imports are keeped to be accessible using this public module
+from ._config import (ConfigSource,  # noqa: F401
+                      ConfigEntry,
+                      ConfigResource)
+from ._resource import (ResourceType,  # noqa: F401
+                        ResourcePatternType)
+from ._acl import (AclOperation,  # noqa: F401
+                   AclPermissionType,
+                   AclBinding,
+                   AclBindingFilter)
+from ..cimpl import (KafkaException,  # noqa: F401
+                     KafkaError,
                      _AdminClientImpl,
                      NewTopic,
                      NewPartitions,
@@ -16,180 +43,6 @@ from ..cimpl import (KafkaException, # noqa
                      RESOURCE_TOPIC,
                      RESOURCE_GROUP,
                      RESOURCE_BROKER)
-
-import concurrent.futures
-import functools
-
-from enum import Enum
-
-
-class ConfigSource(Enum):
-    """
-    Enumerates the different sources of configuration properties.
-    Used by ConfigEntry to specify the
-    source of configuration properties returned by `describe_configs()`.
-    """
-    UNKNOWN_CONFIG = CONFIG_SOURCE_UNKNOWN_CONFIG  #: Unknown
-    DYNAMIC_TOPIC_CONFIG = CONFIG_SOURCE_DYNAMIC_TOPIC_CONFIG  #: Dynamic Topic
-    DYNAMIC_BROKER_CONFIG = CONFIG_SOURCE_DYNAMIC_BROKER_CONFIG  #: Dynamic Broker
-    DYNAMIC_DEFAULT_BROKER_CONFIG = CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG  #: Dynamic Default Broker
-    STATIC_BROKER_CONFIG = CONFIG_SOURCE_STATIC_BROKER_CONFIG  #: Static Broker
-    DEFAULT_CONFIG = CONFIG_SOURCE_DEFAULT_CONFIG  #: Default
-
-
-class ConfigEntry(object):
-    """
-    Represents a configuration property. Returned by describe_configs() for each configuration
-    entry of the specified resource.
-
-    This class is typically not user instantiated.
-    """
-
-    def __init__(self, name, value,
-                 source=ConfigSource.UNKNOWN_CONFIG,
-                 is_read_only=False,
-                 is_default=False,
-                 is_sensitive=False,
-                 is_synonym=False,
-                 synonyms=[]):
-        """
-        This class is typically not user instantiated.
-        """
-        super(ConfigEntry, self).__init__()
-
-        self.name = name
-        """Configuration property name."""
-        self.value = value
-        """Configuration value (or None if not set or is_sensitive==True)."""
-        self.source = source
-        """Configuration source."""
-        self.is_read_only = bool(is_read_only)
-        """Indicates whether the configuration property is read-only."""
-        self.is_default = bool(is_default)
-        """Indicates whether the configuration property is using its default value."""
-        self.is_sensitive = bool(is_sensitive)
-        """
-        Indicates whether the configuration property value contains
-        sensitive information (such as security settings), in which
-        case .value is None."""
-        self.is_synonym = bool(is_synonym)
-        """Indicates whether the configuration property is a synonym for the parent configuration entry."""
-        self.synonyms = synonyms
-        """A list of synonyms (ConfigEntry) and alternate sources for this configuration property."""
-
-    def __repr__(self):
-        return "ConfigEntry(%s=\"%s\")" % (self.name, self.value)
-
-    def __str__(self):
-        return "%s=\"%s\"" % (self.name, self.value)
-
-
-@functools.total_ordering
-class ConfigResource(object):
-    """
-    Represents a resource that has configuration, and (optionally)
-    a collection of configuration properties for that resource. Used by
-    describe_configs() and alter_configs().
-
-    Parameters
-    ----------
-    restype : `ConfigResource.Type`
-       The resource type.
-    name : `str`
-       The resource name, which depends on the resource type. For RESOURCE_BROKER, the resource name is the broker id.
-    set_config : `dict`
-        The configuration to set/overwrite. Dictionary of str, str.
-    """
-
-    class Type(Enum):
-        """
-        Enumerates the different types of Kafka resources.
-        """
-        UNKNOWN = RESOURCE_UNKNOWN  #: Resource type is not known or not set.
-        ANY = RESOURCE_ANY  #: Match any resource, used for lookups.
-        TOPIC = RESOURCE_TOPIC  #: Topic resource. Resource name is topic name.
-        GROUP = RESOURCE_GROUP  #: Group resource. Resource name is group.id.
-        BROKER = RESOURCE_BROKER  #: Broker resource. Resource name is broker id.
-
-    def __init__(self, restype, name,
-                 set_config=None, described_configs=None, error=None):
-        """
-        :param ConfigResource.Type restype: Resource type.
-        :param str name: The resource name, which depends on restype.
-                         For RESOURCE_BROKER, the resource name is the broker id.
-        :param dict set_config: The configuration to set/overwrite. Dictionary of str, str.
-        :param dict described_configs: For internal use only.
-        :param KafkaError error: For internal use only.
-        """
-        super(ConfigResource, self).__init__()
-
-        if name is None:
-            raise ValueError("Expected resource name to be a string")
-
-        if type(restype) == str:
-            # Allow resource type to be specified as case-insensitive string, for convenience.
-            try:
-                restype = ConfigResource.Type[restype.upper()]
-            except KeyError:
-                raise ValueError("Unknown resource type \"%s\": should be a ConfigResource.Type" % restype)
-
-        elif type(restype) == int:
-            # The C-code passes restype as an int, convert to Type.
-            restype = ConfigResource.Type(restype)
-
-        self.restype = restype
-        self.restype_int = int(self.restype.value)  # for the C code
-        self.name = name
-
-        if set_config is not None:
-            self.set_config_dict = set_config.copy()
-        else:
-            self.set_config_dict = dict()
-
-        self.configs = described_configs
-        self.error = error
-
-    def __repr__(self):
-        if self.error is not None:
-            return "ConfigResource(%s,%s,%r)" % (self.restype, self.name, self.error)
-        else:
-            return "ConfigResource(%s,%s)" % (self.restype, self.name)
-
-    def __hash__(self):
-        return hash((self.restype, self.name))
-
-    def __lt__(self, other):
-        if self.restype < other.restype:
-            return True
-        return self.name.__lt__(other.name)
-
-    def __eq__(self, other):
-        return self.restype == other.restype and self.name == other.name
-
-    def __len__(self):
-        """
-        :rtype: int
-        :returns: number of configuration entries/operations
-        """
-        return len(self.set_config_dict)
-
-    def set_config(self, name, value, overwrite=True):
-        """
-        Set/overwrite a configuration value.
-
-        When calling alter_configs, any configuration properties that are not included
-        in the request will be reverted to their default values. As a workaround, use
-        describe_configs() to retrieve the current configuration and overwrite the
-        settings you want to change.
-
-        :param str name: Configuration property name
-        :param str value: Configuration value
-        :param bool overwrite: If True, overwrite entry if it already exists (default).
-                               If False, do nothing if entry already exists.
-        """
-        if not overwrite and name in self.set_config_dict:
-            return
-        self.set_config_dict[name] = value
 
 
 class AdminClient (_AdminClientImpl):
@@ -213,6 +66,7 @@ class AdminClient (_AdminClientImpl):
 
     Requires broker version v0.11.0.0 or later.
     """
+
     def __init__(self, conf):
         """
         Create a new AdminClient using the provided configuration dictionary.
@@ -275,6 +129,39 @@ class AdminClient (_AdminClientImpl):
                 fut.set_exception(e)
 
     @staticmethod
+    def _make_acls_result(f, futmap):
+        """
+        Map create ACL binding results to corresponding futures in futmap.
+        For create_acls the result value of each (successful) future is None.
+        For delete_acls the result value of each (successful) future is the list of deleted AclBindings.
+        """
+        try:
+            results = f.result()
+            futmap_values = list(futmap.values())
+            len_results = len(results)
+            len_futures = len(futmap_values)
+            if len_results != len_futures:
+                raise RuntimeError(
+                    "Results length {} is different from future-map length {}".format(len_results, len_futures))
+            for i, result in enumerate(results):
+                fut = futmap_values[i]
+                if isinstance(result, KafkaError):
+                    fut.set_exception(KafkaException(result))
+                else:
+                    fut.set_result(result)
+        except Exception as e:
+            # Request-level exception, raise the same for all the AclBindings or AclBindingFilters
+            for resource, fut in futmap.items():
+                fut.set_exception(e)
+
+    @staticmethod
+    def _create_future():
+        f = concurrent.futures.Future()
+        if not f.set_running_or_notify_cancel():
+            raise RuntimeError("Future was cancelled prematurely")
+        return f
+
+    @staticmethod
     def _make_futures(futmap_keys, class_check, make_result_fn):
         """
         Create futures and a futuremap for the keys in futmap_keys,
@@ -283,19 +170,14 @@ class AdminClient (_AdminClientImpl):
         futmap = {}
         for key in futmap_keys:
             if class_check is not None and not isinstance(key, class_check):
-                raise ValueError("Expected list of {}".format(type(class_check)))
-            futmap[key] = concurrent.futures.Future()
-            if not futmap[key].set_running_or_notify_cancel():
-                raise RuntimeError("Future was cancelled prematurely")
+                raise ValueError("Expected list of {}".format(repr(class_check)))
+            futmap[key] = AdminClient._create_future()
 
         # Create an internal future for the entire request,
         # this future will trigger _make_..._result() and set result/exception
         # per topic,future in futmap.
-        f = concurrent.futures.Future()
+        f = AdminClient._create_future()
         f.add_done_callback(lambda f: make_result_fn(f, futmap))
-
-        if not f.set_running_or_notify_cancel():
-            raise RuntimeError("Future was cancelled prematurely")
 
         return f, futmap
 
@@ -363,13 +245,13 @@ class AdminClient (_AdminClientImpl):
 
         return futmap
 
-    def list_topics(self, **kwargs):
+    def list_topics(self, *args, **kwargs):
 
-        return super(AdminClient, self).list_topics(**kwargs)
+        return super(AdminClient, self).list_topics(*args, **kwargs)
 
-    def list_groups(self, **kwargs):
+    def list_groups(self, *args, **kwargs):
 
-        return super(AdminClient, self).list_groups(**kwargs)
+        return super(AdminClient, self).list_groups(*args, **kwargs)
 
     def create_partitions(self, new_partitions, **kwargs):
         """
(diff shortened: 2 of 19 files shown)


src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+
+        cfl_PyObject_GetAttr(py_obj, "name", &py_name, NULL, 1);
+        cfl_PyObject_GetAttr(py_obj, "principal", &py_principal, NULL, 1);
+        cfl_PyObject_GetAttr(py_obj, "host", &py_host, NULL, 1);
+        int name_is_none = py_name == Py_None;
+        int principal_is_none = py_principal == Py_None;
+        int host_is_none = py_host == Py_None;
+
+        if(cfl_PyObject_GetInt(py_obj, "restype_int", &restype, 0, 1)
+            && (name_is_none || cfl_PyObject_GetString(py_obj, "name", &resname, NULL, 1))
```
edenhill, line 1030: This isn't a hot path so it doesn't matter in practice, but it would be more efficient and make the code more readable by adding a "allow_None" parameter to cfl_PyObject_GetString() that makes it return the value as NULL with positive return value in case the attr is None.

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+
+
+/**
+ * @brief Convert C acl_result_t array to py list.
+ */
+static PyObject *
+Admin_c_acl_result_to_py (const rd_kafka_acl_result_t **c_result,
+                            size_t cnt) {
+        PyObject *result;
+        size_t ti;
```
edenhill, line 1110: nit: typically just named `i`

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+
+        result = PyList_New(cnt);
+
+        for (ti = 0 ; ti < cnt ; ti++) {
+                PyObject *error;
+                const rd_kafka_error_t *c_error = rd_kafka_acl_result_error(c_result[ti]);
+
+                error = KafkaError_new_or_None(
+                        rd_kafka_error_code(c_error),
+                        rd_kafka_error_name(c_error));
```
edenhill, line 1120: this should be error_string - i.e., the human readable error description.  _name is just the enum name.

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+        result = PyList_New(cnt);
+
+        for (ti = 0 ; ti < cnt ; ti++) {
+                PyObject *error;
+                const rd_kafka_error_t *c_error = rd_kafka_DeleteAcls_result_response_error(c_result_responses[ti]);
+
+                if (c_error) {
+                        error = KafkaError_new_or_None(
+                                rd_kafka_error_code(c_error),
+                                rd_kafka_error_name(c_error));
```
edenhill, line 1149: _string

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+}
+
+
+/**
+ * @brief Convert C DeleteAcls result response array to py list.
+ */
+static PyObject *
+Admin_c_DeleteAcls_result_responses_to_py (const rd_kafka_DeleteAcls_result_response_t **c_result_responses,
+                            size_t cnt) {
+        const rd_kafka_AclBinding_t **c_matching_acls;
```
edenhill, line 1135: move these to the scope where they're used, the loop

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+                if (c_error) {
+                        error = KafkaError_new_or_None(
+                                rd_kafka_error_code(c_error),
+                                rd_kafka_error_name(c_error));
+                        PyList_SET_ITEM(result, ti, error);
+                } else {
+                        c_matching_acls = rd_kafka_DeleteAcls_result_response_matching_acls(
+                                                                        c_result_responses[ti],
+                                                                        &c_matching_acls_cnt);
+                        PyList_SET_ITEM(result, ti,
```
edenhill, line 1155: So here we're mixing two types in the same result list; KafkaErrors and AclBindings. Is that intended? Seems like it makes it harder to use.

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+        int cnt, i = 0;
+        struct Admin_options options = Admin_options_INITIALIZER;
+        PyObject *AclBinding_type = NULL;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        rd_kafka_AclBinding_t **c_objs = NULL;
+        CallState cs;
+        rd_kafka_queue_t *rkqu;
+        char errstr[128];
+
+        static char *kws[] = {"acls",
```
edenhill, line 1178:
```suggestion
        static const char *kws[] = {"acls",
```

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+        int r;
+        struct Admin_options options = Admin_options_INITIALIZER;
+        PyObject *AclBindingFilter_type = NULL;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        rd_kafka_AclBindingFilter_t *c_obj = NULL;
+        CallState cs;
+        rd_kafka_queue_t *rkqu;
+        char errstr[128];
+        
+        static char *kws[] = {"acl_binding_filter",
```
edenhill, line 1301:
```suggestion
        static const char *kws[] = {"acl_binding_filter",
```

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+        int cnt, i = 0;
+        struct Admin_options options = Admin_options_INITIALIZER;
+        PyObject *AclBindingFilter_type = NULL;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        rd_kafka_AclBindingFilter_t **c_objs = NULL;
+        CallState cs;
+        rd_kafka_queue_t *rkqu;
+        char errstr[128];
+        
+        static char *kws[] = {"acls",
```
edenhill, line 1403:
```suggestion
        static const char *kws[] = {"acls",
```

edenhill (changes requested):
This PR is great, fantastic work!

src/confluent_kafka/admin/__init__.py
```diff
@@ -363,13 +255,13 @@ def delete_topics(self, topics, **kwargs):
 
         return futmap
 
-    def list_topics(self, **kwargs):
+    def list_topics(self, *args, **kwargs):
```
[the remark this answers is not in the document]
emasab: It's compatible, it was to give a different error message. It think I had seen a test where it was called with one positional arg but now I can't find it. I remove it

src/confluent_kafka/src/Admin.c
emasab: Do I change it in the other places too? (Admin_create_topics ...). Then I have to cast to  char ** when calling PyArg_ParseTupleAndKeywords

src/confluent_kafka/admin/_config.py
```diff
@@ -0,0 +1,170 @@
...
+from ..cimpl import (CONFIG_SOURCE_UNKNOWN_CONFIG,
+                     CONFIG_SOURCE_DYNAMIC_TOPIC_CONFIG,
+                     CONFIG_SOURCE_DYNAMIC_BROKER_CONFIG,
+                     CONFIG_SOURCE_DYNAMIC_DEFAULT_BROKER_CONFIG,
+                     CONFIG_SOURCE_STATIC_BROKER_CONFIG,
+                     CONFIG_SOURCE_DEFAULT_CONFIG)
+from ._resource import ResourceType
+
+
+class ConfigSource(Enum):
```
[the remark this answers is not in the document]
emasab: yes, this was refactored from __init__.py, no changes here

src/confluent_kafka/admin/_resource.py
```diff
@@ -0,0 +1,43 @@
+from enum import Enum
+from ..cimpl import (RESOURCE_UNKNOWN,
```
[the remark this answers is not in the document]
emasab: yes, this was refactored from `__init__.py`, no changes here

src/confluent_kafka/src/Admin.c
```diff
@@ -976,6 +976,537 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+/**
+ * @brief Convert py AclBinding to C
+ */
+static rd_kafka_AclBinding_t*
+Admin_py_to_c_AclBinding (const PyObject *py_obj_arg,
+                        char *errstr,
+                        size_t errstr_size) {
+        int restype, resource_pattern_type, operation, permission_type;
+        char *resname = NULL, *principal = NULL, *host = NULL;
+        rd_kafka_AclBinding_t* ret = NULL;
```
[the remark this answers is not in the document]
emasab: I agree

tests/requirements.txt
```diff
@@ -5,6 +5,6 @@ pytest-timeout
 requests-mock
 trivup>=0.8.3
 fastavro
-avro
+avro==1.10.0 # latest version contains an inconsistent definition of SchemaParseException
```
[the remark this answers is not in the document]
emasab: I've found that it was present only in the latest version and was causing an error in python3.8 if I'm correct, I could check again if it's present to describe it better

tests/test_Admin.py
```diff
@@ -1,21 +1,76 @@
...
+    ConfigResource(ResourceType.GROUP, "mygroup")
+    ConfigResource(ResourceType.TOPIC, "")
     with pytest.raises(ValueError):
         ConfigResource("doesnt exist", "hi")
     with pytest.raises(ValueError):
-        ConfigResource(confluent_kafka.admin.RESOURCE_TOPIC, None)
+        ConfigResource(ResourceType.TOPIC, None)
+
+
+@pytest.mark.skipif(libversion()[1] < 0x000b0500,
```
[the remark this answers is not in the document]
emasab: Ok, I've seen it but I thought that maybe it was there because it could be dynamically linked with an older version

src/confluent_kafka/admin/__init__.py
```diff
@@ -479,6 +371,88 @@ def alter_configs(self, resources, **kwargs):
...
+
+        super(AdminClient, self).create_acls(acls, f, **kwargs)
+
+        return futmap
+
+    def describe_acls(self, acl_binding_filter, **kwargs):
+        """
+        Get acl bindings matching an acl binding filter.
+
+        :param AclBindingFilter acl_binding_filter: a filter with attributes that
```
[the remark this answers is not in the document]
emasab: I've described it using more sentences and it should be clearer now

tests/test_Admin.py
```diff
@@ -252,10 +307,153 @@ def test_alter_configs_api():
...
+
+    with pytest.raises(ValueError):
+        a.create_acls([acl_binding_filter1],
+                      request_timeout=-5)
+
+    with pytest.raises(TypeError):
+        a.delete_acls([acl_binding_filter1],
+                      unknown_operation="it is")
+
+
```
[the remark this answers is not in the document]
emasab: I think that this is raised by PyArg_ParseTupleAndKeywords, in fact it's the same TypeError also in the other tests

src/confluent_kafka/admin/_acl.py
```diff
@@ -0,0 +1,268 @@
...
+                         For RESOURCE_BROKER, the resource name is the broker id.
+        :param ResourcePatternType resource_pattern_type: The resource pattern, relative to the name.
+        :param str principal: The principal this AclBinding refers to.
+        :param str host: The host that the call is allowed to come from.
+        :param AclOperation operation: The operation/s specified by this binding.
+        :param AclPermissionType The permission type for the specified operation.
+        """
+        super(AclBinding, self).__init__()
+
+        self._set_attrs(restype, name, resource_pattern_type, principal, host,
```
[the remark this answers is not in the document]
emasab: It was to avoid repeating code because the validation checks are different, but I've found another way by adding the methods _not_none_args, _string_args and _forbidden_enums and overriding in the subclass

tests/requirements.txt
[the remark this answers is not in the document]
emasab: The problem turned out to be super weird. It goes like this:

1. The test/requirements.txt installs the latest version avro==1.11.0 that has and uses avro.errors.SchemaParseException
2. **tox** installs the current packages with dependencies like this:
`pip install . .[avro] .[schema-registry] .[json] .[protobuf]`
3. they have as dependency for python3 avro-python3==1.10.0 that is deprecated, but doesn't have an inconsistent definition of SchemaParseException. If you install it normally it doesn't have errors.py
4. pip installs avro-python3 in the avro folder **but doesn't delete the previous files**, it overwrites schema.py and keeps errors.py ending with a mix of avro-python3==1.10.0 and avro==1.11.0
5. the code in load.py
```    
try:
        # avro >= 1.11.0
        from avro.errors import SchemaParseException
except ImportError:
        # avro < 1.11.0
        from avro.schema import SchemaParseException
```
thinks that it has avro 1.11.0 and imports SchemaParseException, but in the code it's avro.schema.SchemaParseException that is used, so

6. this test fails
```
    def test_schema_load_parse_error(self):
        with pytest.raises(avro.ClientError) as excinfo:
            avro.load(data_gen.get_schema_path("invalid_scema.avsc"))
        assert 'Schema parse failed:' in str(excinfo.value)
```

The issue can be reproduced with these commands:
```
python -m venv venv_test
. venv_test/bin/activate
pip install avro
pip install avro-python3==1.10.0
```
there you find `venv_test/lib/python3.8/site-packages/avro/errors.py` but also SchemaParseException in schema.py.

I've kept the version specification of 1.10.0 and removed the avro-python3 dependency in favor of avro

src/confluent_kafka/src/Admin.c
emasab: It's intended, this way _make_acls_result can be used in create_acls and delete_acls, just in create_acls it's KafkaError or None, here is KafkaError or AclBinding. Also _make_acls_result is similiar to _make_topics_result. If it's better to change it, it could be converted into a tuple(result, error) in both create_acls and delete_acls.

emasab:
Thanks Magnus, you spotted a lot of improvements! Most of them are solved, some have a question to answer. Let me now!

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
edenhill: I think it should be backwards compatible, but if not we need to back this out.
Can you verify that calling both position and named arguments, and without parameters, works as expected?

src/confluent_kafka/admin/_acl.py
```diff
@@ -0,0 +1,231 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+import functools
+from .. import cimpl as _cimpl
+from ._resource import ResourceType, ResourcePatternType
+
+try:
+    string_type = basestring  # noqa
```
edenhill, line 21: Please provide noqa number

src/confluent_kafka/admin/_acl.py
```diff
@@ -0,0 +1,231 @@
...
+    ALL = _cimpl.ACL_OPERATION_ALL  #: ALL operation
+    READ = _cimpl.ACL_OPERATION_READ  #: READ operation
+    WRITE = _cimpl.ACL_OPERATION_WRITE  #: WRITE operation
+    CREATE = _cimpl.ACL_OPERATION_CREATE  #: CREATE operation
+    DELETE = _cimpl.ACL_OPERATION_DELETE  #: DELETE operation
+    ALTER = _cimpl.ACL_OPERATION_ALTER  #: ALTER operation
+    DESCRIBE = _cimpl.ACL_OPERATION_DESCRIBE  #: DESCRIBE operation
+    CLUSTER_ACTION = _cimpl.ACL_OPERATION_CLUSTER_ACTION  #: CLUSTER_ACTION operation
+    DESCRIBE_CONFIGS = _cimpl.ACL_OPERATION_DESCRIBE_CONFIGS  #: DESCRIBE_CONFIGS operation
+    ALTER_CONFIGS = _cimpl.ACL_OPERATION_ALTER_CONFIGS  #: ALTER_CONFIGS  operation
```
edenhill, line 41: double space

src/confluent_kafka/admin/_acl.py
```diff
@@ -0,0 +1,231 @@
...
+    IDEMPOTENT_WRITE = _cimpl.ACL_OPERATION_IDEMPOTENT_WRITE  #: IDEMPOTENT_WRITE operation
+
+    def __lt__(self, other):
+        if self.__class__ != other.__class__:
```
edenhill, line 45: This breaks sub-classing, right? Should we use isinstance() instead?

src/confluent_kafka/src/Admin.c
edenhill: Oh, okay, leave it as is then.

src/confluent_kafka/src/Admin.c
```diff
@@ -977,6 +1033,353 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+
+        rd_kafka_queue_destroy(rkqu); /* drop reference from get_background */
+        rd_kafka_AclBinding_destroy_array(c_objs, cnt);
+        free(c_objs);
+        Py_DECREF(AclBinding_type); /* from lookup() */
+        rd_kafka_AdminOptions_destroy(c_options);
+
+        Py_RETURN_NONE;
+err:
+        rd_kafka_AclBinding_destroy_array(c_objs, i);
```
edenhill, line 1141: Needs a null check.

src/confluent_kafka/src/Admin.c
```diff
@@ -977,6 +1033,353 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+        free(c_objs);
+        if(AclBinding_type) Py_DECREF(AclBinding_type);
+        rd_kafka_AdminOptions_destroy(c_options);
```
edenhill, line 1144: needs a null check

src/confluent_kafka/src/Admin.c
```diff
@@ -977,6 +1033,353 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+static PyObject *Admin_describe_acls (Handle *self, PyObject *args, PyObject *kwargs) {
+        PyObject *acl_binding_filter, *future;
+        int r;
+        struct Admin_options options = Admin_options_INITIALIZER;
+        PyObject *AclBindingFilter_type = NULL;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        rd_kafka_AclBindingFilter_t *c_obj = NULL;
+        CallState cs;
+        rd_kafka_queue_t *rkqu;
+        char errstr[128];
```
edenhill, line 1171: Makes this 512 just to be safe we don't truncate lengthy error strings.

src/confluent_kafka/src/Admin.c
```diff
@@ -977,6 +1033,353 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+static PyObject *Admin_create_acls (Handle *self, PyObject *args, PyObject *kwargs) {
+        PyObject *acls_list, *future;
+        int cnt, i = 0;
+        struct Admin_options options = Admin_options_INITIALIZER;
+        PyObject *AclBinding_type = NULL;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        rd_kafka_AclBinding_t **c_objs = NULL;
+        CallState cs;
+        rd_kafka_queue_t *rkqu;
+        char errstr[128];
```
edenhill, line 1048: 512

src/confluent_kafka/src/Admin.c
```diff
@@ -977,6 +1033,353 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+        CallState_end(self, &cs);
+
+        rd_kafka_queue_destroy(rkqu); /* drop reference from get_background */
+        rd_kafka_AclBinding_destroy(c_obj);
+        Py_DECREF(AclBindingFilter_type); /* from lookup() */
+        rd_kafka_AdminOptions_destroy(c_options);
+        Py_RETURN_NONE;
+err:
+        if(AclBindingFilter_type) Py_DECREF(AclBindingFilter_type);
+        rd_kafka_AdminOptions_destroy(c_options);
```
edenhill, line 1247: null check

src/confluent_kafka/src/Admin.c
```diff
@@ -977,6 +1033,353 @@ static PyObject *Admin_alter_configs (Handle *self, PyObject *args,
...
+
+        rd_kafka_queue_destroy(rkqu); /* drop reference from get_background */
+        rd_kafka_AclBinding_destroy_array(c_objs, cnt);
+        free(c_objs);
+        Py_DECREF(AclBindingFilter_type); /* from lookup() */
+        rd_kafka_AdminOptions_destroy(c_options);
+
+        Py_RETURN_NONE;
+err:
+        rd_kafka_AclBinding_destroy_array(c_objs, i);
```
edenhill, line 1366: null checks

src/confluent_kafka/src/Admin.c
```diff
@@ -1280,6 +1695,122 @@ Admin_c_ConfigResource_result_to_py (const rd_kafka_ConfigResource_t **c_resourc
...
+/**
+ * @brief Convert C AclBinding to py
+ */
+static PyObject *
+Admin_c_AclBinding_to_py (const rd_kafka_AclBinding_t *c_acl_binding) {
+        
+        PyObject *args, *kwargs, *AclBinding_type, *acl_binding;
+
+        AclBinding_type = cfl_PyObject_lookup("confluent_kafka.admin",
+                                        "AclBinding");
```
edenhill, line 1707: Error check?

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -42,28 +42,28 @@
  *  0xMMmmRRPP
  *  MM=major, mm=minor, RR=revision, PP=patchlevel (not used)
  */
-#define CFL_VERSION     0x01080300
```
edenhill, line 45: bump to 1.10.0

edenhill (changes requested):
Fantastic work, left some minor things that should be addressed.

setup.py
```diff
@@ -27,8 +27,7 @@
 
 AVRO_REQUIRES = ['fastavro>=0.23.0,<1.0;python_version<"3.0"',
                  'fastavro>=1.0;python_version>"3.0"',
-                 'avro==1.10.0;python_version<"3.0"',
-                 'avro-python3==1.10.0;python_version>"3.0"'
+                 'avro==1.10.0'
```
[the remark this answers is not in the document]
emasab: Yes in the [avro-python3 PyPI page](https://pypi.org/project/avro-python3/) it ways so and that avro-python3 is deprecated

src/confluent_kafka/src/Admin.c
emasab: Yes. In this case the error happens in the response conversion to python. It should return a KafkaError to be set in the future. The KafkaError has an error code of type _rd_kafka_resp_err_t_ but I'm not sure if one of the internal errors can be used or a new type of error should be added.

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
emasab: verified that in `integration_test.py` when testing `list_groups` [this call ](https://github.com/emasab/confluent-kafka-python/blob/c7030656c9fcb38bf141fe0f550cd95b0d27ffc5/tests/integration/integration_test.py#L1233) fails without `*args` and passes with. Then in the same file when testing `list_topics` I've added [this](https://github.com/emasab/confluent-kafka-python/blob/c7030656c9fcb38bf141fe0f550cd95b0d27ffc5/tests/integration/integration_test.py#L1165) new call that fails without and passes with.

src/confluent_kafka/src/Admin.c
edenhill: Huhm, I can't quite visualize the usage.
Could you show an example of what error handling of these results would look like on the application (or test) side with the current approach?

emasab: the recipient of this is a private method [_make_acls_result](https://github.com/confluentinc/confluent-kafka-python/blob/83669187e0980343e6c454ce3ac74484916be152/src/confluent_kafka/admin/__init__.py#L132), what the user sees  instead is a future that either has a result with a list of `AclBinding` or an exception of type `KafkaException` wrapping the `KafkaError`.
An example of the usage is in the [tests](https://github.com/confluentinc/confluent-kafka-python/blob/83669187e0980343e6c454ce3ac74484916be152/tests/integration/admin/test_basic_operations.py#L65)

examples/docker/Dockerfile.alpine
```diff
@@ -30,7 +30,7 @@ FROM alpine:3.12
 
 COPY . /usr/src/confluent-kafka-python
 
-ENV LIBRDKAFKA_VERSION v1.8.2
+ENV LIBRDKAFKA_VERSION v1.10.0-RC1
```
edenhill, line 33: Use v1.9.0-RC10 here to avoid breaking builds. There's hopefully nothing required from a future librdkafka

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -42,28 +42,28 @@
...
+#define CFL_VERSION_STR "1.10.0"
 
 /**
  * Minimum required librdkafka version. This is checked both during
  * build-time (just below) and runtime (see confluent_kafka.c).
  * Make sure to keep the MIN_RD_KAFKA_VERSION, MIN_VER_ERRSTR and #error
  * defines and strings in sync.
  */
-#define MIN_RD_KAFKA_VERSION 0x01060000
+#define MIN_RD_KAFKA_VERSION 0x010a00ff
```
edenhill, line 54: It requires librdkafka 1.9.0 where ACLs were added, right?
(the current change will break builds)

tests/integration/admin/test_basic_operations.py
```diff
@@ -0,0 +1,280 @@
+import confluent_kafka
```
edenhill, line 1: Add license & copyright header

edenhill (changes requested):
Just a couple of points, then good to go!

edenhill (approved):
LGTM

PR #1287, opened 2022-02-22
Deserializing.consume method docstring update

deepaktammali:
The docstring of _DeserializingConsumer.consume_ method currently says
`       :py:func:`Consumer.consume` not implemented, use
        :py:func:`DeserializingConsumer.poll` instead`

but the method _raises_ **NotImplementedError** 

I assume it should be 
`       :py:func:`DeserializingConsumer.consume` not implemented, use
        :py:func:`DeserializingConsumer.poll` instead`

This PR updates the same.
Thanks :)


deepaktammali:
I understood it was referring to the ConsumerImpl

PR #1306, opened 2022-03-23
Add new option oauth_cb to SchemaRegistryClient

bodepd:
This patch adds an option to supply an oauth callback to the
SchemaRegistryClient so that it accepts oauth tokens in the
same way that kafka brokers can receive tokens.

The function passes has the same return types as the same option
for kafka so that users can reusing their existing token service
for both kafka as well as the schema registry client.

If a user supplies this new option together with BASIC auth configuration,
it results in a ValueError.


bodepd:
I noticed some of the other patches have a CLA notification. Is there a CLA that I need to sign?

dylanmei:
@bodepd This is a valuable contribution. Is there a reason you've closed it?

PR #1340, opened 2022-05-16
tidy up json serde

mhowlett:


src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -45,7 +45,7 @@ def __exit__(self, *args):
 
 class JSONSerializer(Serializer):
     """
-    Serializes Python objects to JSON with Confluent Schema Registry framing.
+    Serializer that produces JSON formatted data with Confluent Schema Registry framing.
```
edenhill, line 48: "produces" is unfortunate wording here since I don't think it actually Kafka-produces anything.
I liked the previous version better.

mhowlett: I will change 'produces' to 'outputs'.

I was trying to avoid referring to what the serializer takes as input, as it gets complex to explain (and is explained below).

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -47,16 +45,15 @@ def __exit__(self, *args):
 
 class JSONSerializer(Serializer):
     """
-    JsonSerializer serializes objects in the Confluent Schema Registry binary
-    format for JSON.
+    Serializer that outputs JSON formatted data with Confluent Schema Registry framing.
```
edenhill, line 48: nit: formatted -> encoded

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -47,16 +45,15 @@ def __exit__(self, *args):
...
 
-    JsonSerializer configuration properties:
+    Configuration properties:
 
     +---------------------------+----------+--------------------------------------------------+
     | Property Name             | Type     | Description                                      |
     +===========================+==========+==================================================+
-    |                           |          | Registers schemas automatically if not           |
-    | ``auto.register.schemas`` | bool     | previously associated with a particular subject. |
+    |                           |          | If True, automatically register unrecognized     |
```
edenhill, line 55: I think the former wording was clearer. Now we have to wonder what an unrecognized schema is.

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -68,14 +65,15 @@ class JSONSerializer(Serializer):
     +-------------------------------------+----------+----------------------------------------+
     |                           |          | Callable(SerializationContext, str) -> str       |
     |                           |          |                                                  |
-    | ``subject.name.strategy`` | callable | Instructs the JsonSerializer on how to construct |
-    |                           |          | Schema Registry subject names.                   |
+    | ``subject.name.strategy`` | callable | Defines how Schema Registry subject names are    |
```
edenhill, line 68: Perhaps note how this callable should behave. E.g., return one of these strings: "inherit", "foo", ..

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -68,14 +65,15 @@ class JSONSerializer(Serializer):
...
+    |                           |          | constructed.                                     |
     |                           |          | Defaults to topic_subject_name_strategy.         |
     +---------------------------+----------+--------------------------------------------------+
 
-    Schemas are registered to namespaces known as Subjects which define how a
-    schema may evolve over time. By default the subject name is formed by
-    concatenating the topic name with the message field separated by a hyphen.
+    Schemas are registered against subject names in Confluent Schema Registry
+    that define a scope in which the schemas can be evolved. By default, the subject
+    name is formed by concatenating the topic name with the message field separated
```
edenhill, line 75: Message field is vague. Maybe just add `message field ("key" or "value") ...`

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -96,13 +94,16 @@ class JSONSerializer(Serializer):
...
-        Registry. See documentation below for additional details on Subjects
-        and schema registration.
+    Notes:
+        The ``title`` annotation, referred to elsewhere as a record name
+        is not strictly required by the JSON Schema specification. It is
+        however required by this serializer in order to register the schema
+        with Confluent Schema Registry.
+
+        Prior to serialization, all values must first be converted to
+        a dict instance. This may handled manually prior to calling
```
edenhill, line 104: 'may be'

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -96,13 +94,16 @@ class JSONSerializer(Serializer):
...
-        annotation(record name) is used to register the Schema with the Schema
-        Registry. See documentation below for additional details on Subjects
-        and schema registration.
+    Notes:
+        The ``title`` annotation, referred to elsewhere as a record name
+        is not strictly required by the JSON Schema specification. It is
+        however required by this serializer in order to register the schema
+        with Confluent Schema Registry.
+
+        Prior to serialization, all values must first be converted to
```
edenhill, line 103: What are these "values"?
It could be the message value, but then how do I serialize the key?
Or it could be a dict value? As a user I wouldn't know.

Since everything is an object in Python, and every object can (potentially) be serialized, I think we should stick with `object` being the input to a serializer (and output from a deserializer), and not use other ambiguous terms.

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -114,31 +115,27 @@ class JSONSerializer(Serializer):
...
-                 conf=None):
+    def __init__(self, schema_str, schema_registry_client, to_dict=None, conf=None):
         self._registry = schema_registry_client
         self._schema_id = None
-        # Avoid calling registry if schema is known to be registered
         self._known_subjects = set()
 
         if to_dict is not None and not callable(to_dict):
-            raise ValueError("to_dict must be callable with the signature"
-                             " to_dict(object, SerializationContext)->dict")
```
edenhill, line 137: We should look at automating style-checking, e.g., through `autopep8`

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -172,25 +169,23 @@ def __init__(self, schema_str, schema_registry_client, to_dict=None,
...
     def __call__(self, obj, ctx):
         """
-        Serializes an object to the Confluent Schema Registry's JSON binary
-        format.
+        Serializes an object to JSON, prepending it with Confluent Schema Registry
+        framing.
 
         Args:
-            obj (object): object instance to serialize.
+            obj (object): The object instance to serialize.
```
edenhill, line 176: super nit: Less 'the's are clearer.

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -172,25 +169,23 @@ def __init__(self, schema_str, schema_registry_client, to_dict=None,
...
-        Note:
-            None objects are represented as Kafka Null.
-
         Raises:
             SerializerError if any error occurs serializing obj
 
         Returns:
-            bytes: Confluent Schema Registry formatted JSON bytes
-
+            bytes: None if obj is None, else a byte array containing the JSON
```
edenhill, line 185: I think it is worthwhile to explain why it is okay to return None and what happens in the producer (namely the Kafka Null tombstone).

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -260,36 +255,37 @@ def __init__(self, schema_str, from_dict=None):
...
+        Deserialize a JSON formatted record with Confluent Schema Registry framing to
+        a dict, or object instance according to from_dict, if from_dict is specified.
 
         Args:
-            value (bytes): Confluent Schema Registry formatted JSON bytes
+            value (bytes): A JSON serialized record with Confluent Schema Regsitry framing.
 
-            ctx (SerializationContext): Metadata pertaining to the serialization
-                operation.
+            ctx (SerializationContext): Metadata pertaining to the serialization operation.
```
edenhill, line 264: old nit: "pertaining".. as a non-native english speaker this is a word I never see or use.

src/confluent_kafka/schema_registry/json_schema.py
```diff
@@ -260,36 +255,37 @@ def __init__(self, schema_str, from_dict=None):
 
     def __call__(self, value, ctx):
```
edenhill, line 256: Maybe we could find a better name than `value` since it conflicts with the Kafka notion of keys and values.
I.e., can this deserializer only deserialize message values?

edenhill (approved):
Nits and nits. Also great!

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: that is precisely the word i was looking for but didn't come to mind. thanks.

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: yes and yes.

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: i will note the namespace where standard naming strategies are defined (generally shouldn't use a custom one)

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: good point.

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: yes, good plan.

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: I don't think so - the official python docs have lots of 'the's in then, also the official .NET docs.

src/confluent_kafka/schema_registry/json_schema.py
mhowlett: let's go with 'data'

PR #1341, opened 2022-05-16
Format args in the correct order

kevswanberg:
The args in this format statement within the print statement are in the wrong order


mhowlett (approved):
thanks!

PR #1345, opened 2022-05-19
fix: test of invalid schema had old return status code

emasab:
changed return status code

### DIFF
--- a/tests/integration/schema_registry/test_api_client.py
+++ b/tests/integration/schema_registry/test_api_client.py
@@ -21,6 +21,22 @@ import pytest
 
 from confluent_kafka.schema_registry import Schema
 from confluent_kafka.schema_registry.error import SchemaRegistryError
+from tests.integration.conftest import kafka_cluster_fixture
+
+
+@pytest.fixture(scope="module")
+def kafka_cluster_cp_7_0_1():
+    """
+    Returns a Trivup cluster with CP version 7.0.1.
+    SR version 7.0.1 is the last returning 500 instead of 422
+    for the invalid schema passed to test_api_get_register_schema_invalid
+    """
+    for fixture in kafka_cluster_fixture(
+        brokers_env="BROKERS_7_0_1",
+        sr_url_env="SR_URL_7_0_1",
+        trivup_cluster_conf={'cp_version': '7.0.1'}
+    ):
+        yield fixture
 
 
 def _subject_name(prefix):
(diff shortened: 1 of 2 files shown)


tests/integration/schema_registry/test_api_client.py
```diff
@@ -176,9 +176,9 @@ def test_api_get_register_schema_invalid(kafka_cluster, load_file):
 
     with pytest.raises(SchemaRegistryError, match="Invalid schema") as e:
         sr.lookup_schema(subject, schema2)
-    # Not as documented but the caused by is correct.
-    assert e.value.http_status_code == 500
-    assert e.value.error_code == 500
+
+    assert e.value.http_status_code == 422
+    assert e.value.error_code == 42201
```
mhowlett, line 181: 42201 is subject not found error, shouldn't this be 42202?

emasab: this is what's returned when the test passes. That error code 42201 is for an invalid schema [stated here](https://docs.confluent.io/platform/current/schema-registry/develop/api.html) 42202 is for invalid version. In this case it's invalid schema because of the `"type" : "array"`, attribute in the avsc

mhowlett (approved):
given my input, what you do with it is up to you..

tests/integration/schema_registry/test_api_client.py
emasab: I think that the [documentation stayed the same](https://docs.confluent.io/platform/6.0.0/schema-registry/develop/api.html#post--subjects-(string-%20subject)-versions) but in version 7.0.2 they have fixed it and started returning 422 instead of internal server error.

emasab:
@mhowlett thank you! I've sent the [commit](https://github.com/confluentinc/confluent-kafka-python/pull/1345/commits/7a5d5082523a98fa3d5eaef5cead7ef82d30543b) to test with different schema registry versions. We can see if keep it or discard it.

tests/integration/schema_registry/test_api_client.py
```diff
@@ -21,6 +21,22 @@
...
 from confluent_kafka.schema_registry.error import SchemaRegistryError
+from tests.integration.conftest import kafka_cluster_fixture
+
+
+@pytest.fixture(scope="module")
+def kafka_cluster_confluent_7_0_1():
+    """
+    Returns a Trivup cluster with CP version 7.0.1.
+    SR version 7.0.1 is the last returning 500 instead of 422
+    for the invalid schema passed to test_api_get_register_schema_invalid
```
edenhill, line 32: Good comment!

edenhill (approved):
Great stuff!

edenhill:
The Travis-CI jobs haven't triggerd for this PR, for some reason. We'll need to sort that out before merging

edenhill:
(Ignore the close&reopen, trying to retrigger CI)

PR #1364, opened 2022-06-07
ACL documentation fixes

emasab:


Vikash08Mishra:
@emasab As per ACLBinding documentation [here](https://github.com/confluentinc/confluent-kafka-python/blob/c0cb24e8423da79098cb6875ee078412cbcfc22e/src/confluent_kafka/admin/_acl.py#86)(line-86), operations is plural but it's an enum so only one value can be passed. Is it a documentation issue as internal code also refers to single enum value in constructor validation ?
Though IMO operations should be a list of enums - single principal can have multiple operations permission. I understand we can add multiple ACLBinding as a workaround but it would be good to have these created in single call from librdkafka to Kafka servers. Same is supported by raw kafka scripts as well where we can define multiple '--operations' in single call. Let me know your thoughts.

PR #1370, opened 2022-06-14
Fix for duplicate acls

emasab:
In create_acls and delete_acls it returns a dict with AclBinding or AclBindingFilters as keys.
If sent parameters are not unique the returned dict has less entries than the sent ones and it causes and Exception.

### DIFF
--- a/examples/adminapi.py
+++ b/examples/adminapi.py
@@ -159,10 +159,6 @@ def example_create_acls(a, args):
         print(f"Invalid input: {e}")
         return
 
-    if fs is None:
-        print("Unknown error: fs is None")
-        return
-
     # Wait for operation to finish.
     for res, f in fs.items():
         try:
@@ -251,10 +247,6 @@ def example_delete_acls(a, args):
         fs = a.delete_acls(acl_binding_filters, request_timeout=10)
     except ValueError as e:
         print(f"Invalid input: {e}")
-        return
-    
-    if fs is None:
-        print("Unknown error: fs is None")
         return
 
     # Wait for operation to finish.
--- a/examples/adminapi.py
+++ b/examples/adminapi.py
@@ -152,7 +152,11 @@ def example_create_acls(a, args):
         )
     ]
 
-    fs = a.create_acls(acl_bindings, request_timeout=10)
+    try:
+        fs = a.create_acls(acl_bindings, request_timeout=10)
+    except ValueError as e:
+        print(f"Invalid input: {e}")
+        return
 
     # Wait for operation to finish.
     for res, f in fs.items():
@@ -237,7 +241,11 @@ def example_delete_acls(a, args):
         )
     ]
 
-    fs = a.delete_acls(acl_binding_filters, request_timeout=10)
+    try:
+        fs = a.delete_acls(acl_binding_filters, request_timeout=10)
+    except ValueError as e:
+        print(f"Invalid input: {e}")
+        return
 
     # Wait for operation to finish.
     for res, f in fs.items():
--- a/examples/adminapi.py
+++ b/examples/adminapi.py
@@ -152,10 +152,15 @@ def example_create_acls(a, args):
         )
     ]
 
+    fs = None
     try:
         fs = a.create_acls(acl_bindings, request_timeout=10)
     except ValueError as e:
         print(f"Invalid input: {e}")
+        return
+
+    if fs is None:
+        print("Unknown error: fs is None")
         return
 
     # Wait for operation to finish.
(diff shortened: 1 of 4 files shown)


examples/adminapi.py
```diff
@@ -237,7 +241,11 @@ def example_delete_acls(a, args):
         )
     ]
 
-    fs = a.delete_acls(acl_binding_filters, request_timeout=10)
+    try:
+        fs = a.delete_acls(acl_binding_filters, request_timeout=10)
```
mhowlett, line 245: same comment

jliunyu:
I removed the previous dummy change from the RC branch, please make sure it won't be merged again.

examples/adminapi.py
```diff
@@ -152,7 +152,16 @@ def example_create_acls(a, args):
...
 
-    fs = a.create_acls(acl_bindings, request_timeout=10)
+    fs = None
+    try:
+        fs = a.create_acls(acl_bindings, request_timeout=10)
+    except ValueError as e:
+        print(f"Invalid input: {e}")
+        return
+
+    if fs is None:
```
mhowlett, line 162: is this possible? I don't reckon, no need to check.

emasab: done

examples/adminapi.py
```diff
@@ -152,7 +152,11 @@ def example_create_acls(a, args):
         )
     ]
 
-    fs = a.create_acls(acl_bindings, request_timeout=10)
+    try:
+        fs = a.create_acls(acl_bindings, request_timeout=10)
```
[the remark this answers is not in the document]
edenhill: nit: `try` does not create a new scope.

examples/adminapi.py
```diff
@@ -152,7 +152,12 @@ def example_create_acls(a, args):
         )
     ]
 
-    fs = a.create_acls(acl_bindings, request_timeout=10)
+    fs = None
+    try:
+        fs = a.create_acls(acl_bindings, request_timeout=10)
+    except ValueError as e:
+        print(f"Invalid input: {e}")
```
edenhill, line 159: The error string itself should tell what the error is, which isn't necessarily "Invalid input", but could be other things (in the future). So recommend just printing the error, or something like: `f"create_acls() failed: {e}"`

src/confluent_kafka/admin/__init__.py
```diff
@@ -181,6 +181,10 @@ def _make_futures(futmap_keys, class_check, make_result_fn):
 
         return f, futmap
 
+    @staticmethod
+    def _has_duplicated(items):
```
edenhill, line 185: _has_duplicates()

src/confluent_kafka/admin/__init__.py
```diff
@@ -380,6 +384,8 @@ def create_acls(self, acls, **kwargs):
         :raises TypeException: Invalid input.
         :raises ValueException: Invalid input.
         """
+        if AdminClient._has_duplicated(acls):
+            raise ValueError("pass only different ACL bindings")
```
edenhill, line 388: "different" is a bit ambigious, better say: "duplicate ACL bindings not allowed"

src/confluent_kafka/admin/__init__.py
```diff
@@ -444,6 +450,8 @@ def delete_acls(self, acl_binding_filters, **kwargs):
         :raises TypeException: Invalid input.
         :raises ValueException: Invalid input.
         """
+        if AdminClient._has_duplicated(acl_binding_filters):
+            raise ValueError("pass only different ACL binding filters")
```
edenhill, line 454: dito

edenhill:
There are also some test failures due to trivup not starting, see if the a new run (after fixing the other issues) solves it.

examples/adminapi.py
```diff
@@ -237,7 +242,12 @@ def example_delete_acls(a, args):
         )
     ]
 
-    fs = a.delete_acls(acl_binding_filters, request_timeout=10)
+    fs = None
+    try:
+        fs = a.delete_acls(acl_binding_filters, request_timeout=10)
+    except ValueError as e:
+        print(f"Invalid input: {e}")
```
edenhill, line 249: Same comment as for create_acls()

examples/adminapi.py
```diff
@@ -237,7 +242,12 @@ def example_delete_acls(a, args):
         )
     ]
 
-    fs = a.delete_acls(acl_binding_filters, request_timeout=10)
+    fs = None
```
edenhill, line 245: There is no need for this since there is no case where fs has not been assigned when it is used later.

Same comment for the other locations

edenhill (changes requested):

edenhill (approved):
LGTM

examples/adminapi.py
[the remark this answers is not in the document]
mhowlett: whoah, what a language

PR #1383, opened 2022-07-04
Remove trailing whitespace

jayvdb:


tests/integration/schema_registry/gen/NestedTestProto_pb2.py
```diff
@@ -401,7 +401,7 @@
   ))
 _sym_db.RegisterMessage(ComplexType)
 
-NestedMessage = _reflection.GeneratedProtocolMessageType('NestedMessage', (_message.Message,), dict(
```
mhowlett, line 404: nit: this was auto-generated, and shouldn't be modified. but it doesn't matter.

mhowlett (approved):
thanks!

Issue #1219, opened 2021-10-08
Python 3.10 wheel build for confluent-kafka 1.7.0


Gyllsdorff:
Description
===========
Python 3.10 have been released but the latest `confluent-kafka` release `1.7.0` don't have a [prebuilt wheel](https://pypi.org/project/confluent-kafka/1.7.0/#files) for it.

It is easy to build it from source but it slows down some of the image build procedures quite a bit. 

How to reproduce
================

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`)
 - [x] Apache Kafka broker version:
 - [x] Client configuration: `{...}`: None
 - [x] Operating system: Linux
 - [x] Provide client logs (with `'debug': '..'` as necessary): none
 - [x] Provide broker log excerpts: none
 - [ ] Critical issue: No

grutts:
Is there a plan to build a wheel for 3.10?

david-walker-gfs:
Interested in this too. Seems 3.9 doesn't yet exist either.

ahtik:
@david-walker-gfs, 3.9 wheels are available, https://pypi.org/project/confluent-kafka/1.7.0/#files

johng:
This would be nice

david-walker-gfs:
> @david-walker-gfs, 3.9 wheels are available, https://pypi.org/project/confluent-kafka/1.7.0/#files

Thanks for the link!

edenhill:
v1.8.2 is now released, with py 3.6-3.9 wheels.  We'll look into v3.10 wheels for the next release.

grutts:
This project releases about 3 times per year. Would you consider releasing a patch (v1.8.3) earlier than the typical release frequency, with 3.10 wheels included?

edenhill:
The slow release cadence is due to severe CI issues during the latter part of 2021. We're soon moving to another solution which will allow us to release more often.

grutts:
Good news! Thanks

wkrt7:
When do you plan to release the wheel for Python 3.10?

arroyofnxio:
Hi @edenhill! Is there any way the community can help to prioritize this issue?

saabeilin:
Upvoting this as well ;-)

grutts:
[...]

PR #1384, opened 2022-07-04
setup.py: Add /usr/local for macos brew support

jayvdb:


mhowlett:
brew support of what? - can you explain a bit more the purpose

jayvdb:
Sure. This allows building with https://formulae.brew.sh/formula/librdkafka , which has headers and libs install into /usr/local

mhowlett:
we do binary wheels for python, right. why do you want to use the brew installed librdkafka? is it for M1 support?

jayvdb:
We do have one dev on an M1, and we had a requirement to switch to python 3.10 before the recent release which added wheels, so I had this patch so our CI (and me on a non-M1 mac) could easily build on mac before https://github.com/confluentinc/confluent-kafka-python/issues/1219 was resolved recently.

The instructions at https://github.com/confluentinc/confluent-kafka-python/blob/master/INSTALL.md#install-from-source-on-mac-os-x dont work without this patch.

mhowlett:
i wonder if this can it be done with env variables?

[ i'm not an expert here as you can tell from my n00b questions :-), so wary of screwing something up ]

jayvdb:
It can be done with env variables, but that then depends on whether other tools are all going to play nice. e.g. the package manager like flit/poetry/pipenv/pdm/etc, and pip.

It is also possible to create a custom site.cfg for a venv, e.g. https://github.com/numpy/numpy/blob/main/site.cfg.example

But they both require the user to know what they are doing.

A better approach is to use https://pypi.org/project/pkgconfig/ , which means it will find the correct paths to add for any platform, but then you need to add that as a setup_requires .  I've done that a few times for packages using external C deps, so I don't mind doing that if you prefer it.

edenhill:
If include_dirs (et.al) take precedence over standard paths then this PR introduces a problem with promoting /usr/local over system installed packages, which would be unexpected and undesired.
I think it would be better to instruct people how to set proper env vars for non-standard source builds instead.

edenhill:
I'd rather we didn't alter these defaults path at all, but instead provide instructions for how to build with non-default include/library paths.

Also note that we'll soon be providing m1 wheels.

jayvdb:
@edenhill , /usr/local is the correct path on macos, based on the instructions at https://github.com/confluentinc/confluent-kafka-python/blob/master/INSTALL.md#install-from-source-on-mac-os-x .  brew never installs into /usr , except for a few binary blob casks from non-OSS vendors that they can't control.

ivanklimberg:
It would be really helpful if this gets merged

setup.py
```diff
@@ -44,6 +44,8 @@
     librdkafka_libname = 'rdkafka'
 
 module = Extension('confluent_kafka.cimpl',
+                   include_dirs=['/usr/local/include/'],
+                   library_dirs=['/usr/local/lib'],
```
asottile-sentry, line 48: this isn't portable (especially for M1 where brew's librdkafka will be in `/opt` and not `/usr/local`)

you can usually utilize `CPATH="$(brew --prefix librdkafka)/include" LDFLAGS="-L$(brew --prefix librdkafka)/lib" ...` to build a wheel

PR #1401, opened 2022-07-26
V1.9.2rc integration tests upgrade

emasab:
- fix .env with newer docker-compose
  docker-compose in newer version was using the .env file as its configuration file and cluster-up wasn't working
- fixed basic auth integration test
  the test was using the wrong SR instance, the one without authentication
- protoc test fix with newer Python and protobuf versions
  the old compiled py files were giving an error with the latest python protobuf. Recompiling them would give an import error with   newer Python versions because of the relative import not explictlly declared. Changing them to absolute import is the way   [suggested](https://github.com/protocolbuffers/protobuf/issues/1491#issuecomment-263772124) by protobuf developers.


emasab:
Tests output
[tests.zip](https://github.com/confluentinc/confluent-kafka-python/files/9191738/tests.zip)

mhowlett (approved):
some good cleanup as well 👍

PR #1402, opened 2022-07-26
Support setting principal and SASL extensions in oauth_cb, handle failures

### LINKED ISSUE #1394: No support for setting SASL extensions or Principal with OAuthBearer tokens
Description
===========
librdkafka provides support for setting SASL extensions when setting a OAuthBearer token via `rd_kafka_oauthbearer_set_token()`. This is used in confluent-kafka-go [here](https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/handle.go#L313), but not in confluent-kafka-python, as seen [here](https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/src/confluent_kafka.c#L1553).

In addition, the principal name also cannot be passed through to `rd_kafka_oauthbearer_set_token()` in confluent-kafka-python.

Having feature parity between Go and Python would be desirable for us, as we look to support custom OAuth/OIDC callbacks for our customers to integrate their authorization servers with. Unfortunately the built-in KIP-768 OIDC callback only allows for Client ID+Secret authentication, whereas we also wish to support signed JWT client assertions as per [RFC-7523](https://datatracker.ietf.org/doc/html/rfc7523#section-2.2), hence the need to continue using custom callbacks.

Proposal
===========
It would be possible to add the principal and extensions as additional return values to the user-provided callback to `oauth_cb`, and the library handling both the old `(token_str, expiry_time)` and new `(token_str, expiry_time, principal_str, extensions_dict)`, ensuring to convert the dict into a C char pointer array before passing to `rd_kafka_oauthbearer_set_token`.

A class, similar to `OAuthBearerToken` from confluent-kafka-go, would be a more future-proof way to return the values from the callback, but this won't be backwards compatible without some added complexity, which may be acceptable.

We are happy to raise a PR for this, but we'd like to understand what method would be preferable, adding more (optional) return values to the callback function, or using a class object as the return value whilst supporting the old return value for backwards compatibility.

Manicben:
Fixes #1394

Extends the return values of the `oauth_cb` to include a principal string, and a dict of SASL extensions, which must be strings matching `[A-Za-z]+`. These return values are optional, meaning that existing callbacks still work.

When an exception is raised in the callback, `rd_kafka_oauthbearer_set_token_failure` is called to ensure another refresh is scheduled 10s later. This is for the case where the callback raising an exception is due to a token not being able to be fetched, which may succeed later. Without this, raising exceptions in the callback result in the client never re-fetching a token unless fully restarted.

Suggestions regarding the latter part are appreciated, an alternative would be to allow a return value of `None` to signal token fetch failure and the need to retry later. However raising an exception is a bit more Pythonic, so maybe raising a certain exception type should be the way to signal a failure.

### DIFF
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -1522,6 +1522,62 @@ static void log_cb (const rd_kafka_t *rk, int level,
         CallState_resume(cs);
 }
 
+/**
+ * @brief Translate Python \p key and \p value to C types and set on
+ *        provided \p extensions char* array at the provided index.
+ *
+ * @returns 1 on success or 0 if an exception was raised.
+ */
+static int py_extensions_to_c (char **extensions, Py_ssize_t idx,
+                               PyObject *key, PyObject *value) {
+        PyObject *ks, *ks8, *vo8 = NULL;
+        const char *k;
+        const char *v;
+        Py_ssize_t ksize = 0;
+        Py_ssize_t vsize = 0;
+
+        if (!(ks = cfl_PyObject_Unistr(key))) {
+                PyErr_SetString(PyExc_TypeError,
+                                "expected extension key to be unicode "
+                                "string");
+                return 0;
+        }
+
+        k = cfl_PyUnistr_AsUTF8(ks, &ks8);
+        ksize = (Py_ssize_t)strlen(k);
+
+        if (cfl_PyUnistr(_Check(value))) {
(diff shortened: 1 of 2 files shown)


mhowlett:
thanks for the PR @Manicben - we'll try to get to this soon.

emasab:
Thanks @Manicben . I'm reviewing it today

emasab (approved):
LGTM, I've just added a commit to remove the global variables

PR #1410, opened 2022-08-22
Add metadata to TopicPartition

milindl:
This PR builds on #1126 (retains commit history).

Since there are comments on the prior PR, I'll both add those comments here and address them here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## v1.10.0
 
-- Add metadata to TopicPartition type and commit() (TODO(milind) - fill in PR number here once created).
+- Add metadata to TopicPartition type and commit() (#1410).
 
 
 ## v1.9.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ This is a feature release:
  - OAUTHBEARER OIDC support
  - KIP-140 Admin API ACL support
 
+v1.7.1 is a maintenance release with the following fixes and enhancements:
+
 ### Fixes
 
  - The warnings for `use.deprecated.format` (introduced in v1.8.2)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,31 +40,7 @@ This is a feature release:
 
  - OAUTHBEARER OIDC support
  - KIP-140 Admin API ACL support
-
-v1.7.1 is a maintenance release with the following fixes and enhancements:
-
-### Fixes
-
- - The warnings for `use.deprecated.format` (introduced in v1.8.2)
-   had its logic reversed, which result in warning logs to be emitted when
-   the property was correctly configured, and the log message itself also
-   contained text that had it backwards.
-   The warning is now only emitted when `use.deprecated.format` is set
-   to the old legacy encoding (`True`). #1265
- - Use `str(Schema)` rather than `Schema.to_json` to prevent fastavro
-   from raising exception `TypeError: unhashable type: 'mappingproxy'`.
-   (@ffissore, #1156, #1197)
- - Fix the argument order in the constructor signature for
-   AvroDeserializer/Serializer: the argument order in the constructor
-   signature for AvroDeserializer/Serializer was altered in v1.6.1, but
-   the example is not changed yet. (@DLT1412, #1263)
- - Fix the json deserialization errors from `_schema_loads` for
-   valid primitive declarations. (@dylrich, #989)
-
-confluent-kafka-python is based on librdkafka v1.9.0, see the
-[librdkafka release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.9.0)
-for a complete list of changes, enhancements, fixes and upgrade considerations.
-
+=======
 
 ## v1.8.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## v1.10.0
 
-- Add metadata to TopicPartition type and commit() (#1126).
+- Add metadata to TopicPartition type and commit() (TODO(milind) - fill in PR number here once created).
 
 
 ## v1.9.2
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Confluent's Python client for Apache Kafka
 
+
+## v1.10.0
```
milindl, line 4: Need help with the CHANGELOG.md version and message, the current one is sort of a placeholder.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -843,18 +855,23 @@ static int TopicPartition_init (PyObject *self, PyObject *args,
...
 	static char *kws[] = { "topic",
 			       "partition",
 			       "offset",
+			       "metadata",
 			       NULL };
 
-	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "s|iL", kws,
-					 &topic, &partition, &offset))
-		return -1;
+	if (!PyArg_ParseTupleAndKeywords(args, kwargs, "s|iLs", kws,
```
milindl, line 866: > We typically try to keep the happy-path (non-error path) unindented, so restore the original:

Done

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1106,10 +1129,15 @@ rd_kafka_topic_partition_list_t *py_to_c_parts (PyObject *plist) {
 			return NULL;
 		}
 
-		rd_kafka_topic_partition_list_add(c_parts,
-						  tp->topic,
-						  tp->partition)->offset =
-			tp->offset;
+		rd_kafka_topic_partition_t *rktpar =
```
milindl, line 1132: >  Since this variable is only used in the for-loop, put it there instead.

Done

tests/integration/consumer/test_consumer_topicpartition_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limit
+#
+
+from confluent_kafka import TopicPartition, KafkaException
+
+def commit_and_check(consumer, topic, metadata):
+    consumer.commit(offsets=[TopicPartition(topic, 0, 1, metadata)], asynchronous=False)
+    offsets = consumer.committed([TopicPartition(topic, 0)], timeout=100)
+    assert len(offsets) == 1
```
milindl, line 24: > add: assert len(offsets) == 1
Done

tests/integration/consumer/test_consumer_topicpartition_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+    offsets = c.committed([TopicPartition(topic, 0)])
+    assert len(offsets) == 1
+    assert offsets[0].metadata is None
+
+    # Commit with only ASCII metadata.
+    metadata = 'hello world'
+    commit_and_check(c, topic, metadata)
+
+    # Commit with Unicode characters in metadata.
+    metadata = 'नमस्ते दुनिया'
```
milindl, line 45: > Add some binary metadata variations to this test.

Added unicode ones, explained in another comment

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -816,15 +816,27 @@ static int TopicPartition_clear (TopicPartition *self) {
...
+	if (self->metadata) {
+                free(self->metadata);
+		self->metadata = NULL;
+	}
 	return 0;
 }
 
 static void TopicPartition_setup (TopicPartition *self, const char *topic,
 				  int partition, long long offset,
+				  const char *metadata,
```
milindl, line 828: This is the biggest change I've done - making metadata a string rather than a binary object (`bytes` or something like that). While the prior request started with string and then changed to a binary object, I reverted it back to a utf-8 encoded string. 

The reasoning behind this is as such:
1. any integration test with the null byte (or for that matter, any byte > 127) were failing. The reason for the failure was coming from outside the python client (I checked that we were encoding the metadata into the PyObject correctly). It _seemed_ like the broker was treating the metadata as a string. 
2. On investigating further, I found that librdkafka code (src/rdkafka_request.c:rd_kafka_buf_write_topic_partitions and the methods it calls) treats the metadata as a c string (calls strlen on it and so on).
3. To further test this I modified "0099-commit_metadata.c" in librdkafka repository to be generic binary data (with \x00 in the middle) rather than an ascii string. The test started failing. 
4. And as a final verification, I looked at OffsetCommit in the kafka protocol. It expects the metadata to be a nullable string. 

So - based on this I've kept the metadata as a C string.

Please tell me if there's something wrong with this and if there's some way to pass binary data correctly.

This addresses most of the comments on the original PR as well.

CHANGELOG.md
mhowlett: I'm not sure what the adopted process is here with this client. My thought is to use `## vNext`, but that's just an idea.

src/confluent_kafka/src/confluent_kafka.c
mhowlett: In the OffsetCommit case, metadata is certainly a string. The only concern I can think of is that `TopicPartition` is a general-use data structure, so if in the future there is a separate need for topic partition metadata that needs to be binary, then our data structure isn't appropriate.

We could perhaps consider naming the parameter `string_metadata` to make this future proof. I'm not really into that idea.

I think i'm in favor of reverting again to a byte array, and doing a check and throwing an exception on the Python binding side if the array contains a 0 or is an otherwise invalid string. The downside of this API is it isn't as strong for the only current (and likely future) use case. And that's not good. But it is more future proof, and I think that is the stronger requirement to optimize on.

CHANGELOG.md
milindl: Sounds reasonable to me, changed it like so.

src/confluent_kafka/src/confluent_kafka.c
milindl: Reverted to the bytes type again, and I throw an exception for null bytes.

edenhill: Don't put any restrictions on the contents of the metadata, keep it binary and let the broker reject it if it is invalid.

milindl:
Updated PR to use strings as the metadata type. Reading with "s" ensures that we get a utf-8 encoded version of the unicode object or string, as well as an error for a NULL character in the middle of the string, which is what we want.

src/confluent_kafka/src/confluent_kafka.c
edenhill: Variable declarations need to go at the head of the block/scope to support older compilers, so move the rktpar declaration to line 1097, but keep the assignment here.

tests/integration/consumer/test_consumer_topicpartition_metadata.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2021 Confluent Inc.
```
edenhill, line 4: 2022

edenhill (changes requested):

src/confluent_kafka/src/confluent_kafka.c
milindl: Done

tests/integration/consumer/test_consumer_topicpartition_metadata.py
milindl: Done

edenhill (approved):
LGTM!

Good work!

PR #1412, opened 2022-08-22
Update protobuf.py to fix malformed table

stevebang:
Fix indentation issue of malformed table


mhowlett (approved):

mhowlett:
this is already fixed in a branch where I'm doing a lot of improvements to the API docs. 

you can feel free merge this, though it will collide with the branch i'm going to merge later so it will actually make my life marginally harder. but it'll get this in master quicker...

stevebang:
@mhowlett Glad to see that you were planning to fix this. I'll merge it only if you think I should (to get it into master sooner). Otherwise, I'm fine letting you fix it in your PR. This came up while trying to troubleshoot a doc PR build failure. [We don't think this caused the build to fail.]

stevebang:
Merging because the table doesn't display because of the error. Hopefully, a minor inconvenience to your PR.

PR #1425, opened 2022-09-13
CLIENTS-2607 - Updated docs to emphasize using logger instead of log_…

pranavrth:
These changes are related to the users getting confused for logger callback property. In the docs we refer to librdkafka properties first then explain logger properties of python client. Due to this, users tries to use log_cb property for log callback and fail.

In this PR, I tried to change our docs to reflect to use logger instead of log_cb configuration property.

Changes in the PR:

1. Updated configuration property doc to make sure that we specify python client specific properties before librdkafka properties.
2. added test cases to make sure the custom logger is logging in the correct format


pranavrth:
Docs would look like this:-

Screenshot 2022-09-13 at 2 10 18 PM

Screenshot 2022-09-13 at 2 07 18 PM

docs/index.rst
```diff
@@ -778,3 +774,9 @@ The Python bindings also provide some additional configuration properties:
     mylogger.addHandler(logging.StreamHandler())
     producer = confluent_kafka.Producer({'bootstrap.servers': 'mybroker.com'}, logger=mylogger)
 
+.. note::
+   In Python client, ``logger`` configuration property is used for log handler instead of ``log_cb``.
```
mhowlett, line 778: maybe: 

In Python client, the ``logger`` configuration property is used for log handler, not ``log_cb``.

tests/test_log.py
```diff
@@ -114,3 +115,47 @@ def test_logging_constructor():
...
+
+
+def test_producer_logger_logging_in_given_format():
+    """Test that asserts that logging is working by matching the log message"""
+
+    stringBuffer = StringIO()
+    logger = logging.getLogger('Producer logger')
+    logger.setLevel(logging.DEBUG)
+    handler = logging.StreamHandler(stringBuffer)
+    handler.setFormatter(logging.Formatter('The logger is - %(name)s'))
```
mhowlett, line 127: seems better to also include `%(message)s` in this.

tests/test_log.py
```diff
@@ -114,3 +115,47 @@ def test_logging_constructor():
...
+    p = confluent_kafka.Producer(
+        {"bootstrap.servers": "test", "logger": logger, "debug": "msg"})
+    val = 1
+    while val > 0:
+        val = p.flush()
+    logMessage = stringBuffer.getvalue().strip()
+    stringBuffer.close()
+
+    assert logMessage.startswith('The logger is - Producer logger')
+    assert logMessage.endswith('The logger is - Producer logger')
```
mhowlett, line 139: rather than check ends with, maybe check for a substring that you know for sure is going to be in the librdkafka emitted log.

docs/index.rst
pranavrth: Done

tests/test_log.py
pranavrth: Checking the log is for the INIT call.

pranavrth:
Merged the changes in another PR - https://github.com/confluentinc/confluent-kafka-python/pull/1447.

Closing this one.

PR #1426, opened 2022-09-13
Serialization Stabilization

mhowlett:
This branch contains only already-reviewed changes related to stabilizing the serde API. It includes:
- polish of the JSON Schema, Protobuf and Avro serde classes.
- various documentation improvements.
- Explicitly deprecating AvroProducer and AvroConsumer (and removing from examples and readme).
- SerializingProducer and DeserializingConsumer are still market as experimental. I expect we will probably ultimately choose to remove these, but we shouldn't until 2.0. The examples have been updated to not use them in this PR.


ffissore:
Sorry to join the party so late but, question: why are you deprecating AvroProducer and AvroConsumer? I couldn't find an issue with the reasoning behind this deprecation

Issue #1427, opened 2022-09-15
bool(Producer(...)) returns False


vladz-sternum:
Description
===========
It's not necessarily a bug, the Producer class has a False-y truth value when there's no enqueued items. This is caused by the combination of the default `nb_bool` implementation (https://github.com/python/cpython/blob/main/Objects/typeobject.c#L7915), which checks for the existence of `__bool__` and if it doesn't exist, it returns the result of `__len__`. This works well with normal containers, like lists/dictionaries, but feels off for a kafka producer. Moreover it behaves strangely for idiomatic python like:

```python
def default_producer():
    ...

def send_messages(producer, ...):
    # producer will evaluate to False if there's no queued messages
    producer = producer or default_producer()
```

It's not obvious if this is the intended behaviour, and "fixing" it would be a breaking change, but I thought it's quite weird. Perhaps just updating the docs might be enough.

How to reproduce
================
```
>>> confluent_kafka.version()
('1.8.2', 17302016)
>>> confluent_kafka.libversion()
('1.8.2', 17302271)
>>> from confluent_kafka import Producer, Consumer
>>> bool(Producer({}))
%5|1663241720.611|CONFWARN|rdkafka#producer-2| [thrd:app]: No `bootstrap.servers` configured: client will not be able to connect to Kafka cluster
False
>>> bool(Consumer({'group.id': 'asd'}))
True
```

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):
 - [ ] Apache Kafka broker version:
 - [x] Client configuration: `{...}`
 - [ ] Operating system:
 - [ ] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
nice observation! 

My view is that using `__len__` the way `Producer` does probably wasn't a good idea. I don't think we should ever change this though because it is breaking, and would impact more people than is justified by the change. Changing the defn of `nb_bool` would also be breaking, but seems unlikely to affect anyone (?) famous last words though, we've done breaking changes under similar justification before and mucked people up... overall i'm in favor of doing it. I think small API niceties like this add up and amount to happy users.

what do you think @edenhill ?

pranavrth:
Changes merged and released with [v2.0.2](https://github.com/confluentinc/confluent-kafka-python/releases/tag/v2.0.2)

PR #1445, opened 2022-10-12
Implement nb_bool for Producer

vladz-sternum:
This addresses  #1427
All tests passed locally.

This implements the nb_bool method for the consumer, so that the default (which uses __len__) will not be used. This avoids situations where producers with no enqueued items would have a False truth value.

### DIFF
--- a/src/confluent_kafka/src/Producer.c
+++ b/src/confluent_kafka/src/Producer.c
@@ -824,6 +824,23 @@ static PySequenceMethods Producer_seq_methods = {
 	(lenfunc)Producer__len__ /* sq_length */
 };
 
+static int Producer__bool__ (Handle *self) {
+        return 1;
+}
+
+static PyNumberMethods Producer_num_methods = {
+     0, // nb_add
+     0, // nb_subtract
+     0, // nb_multiply
+     0, // nb_remainder
+     0, // nb_divmod
+     0, // nb_power
+     0, // nb_negative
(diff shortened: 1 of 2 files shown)


src/confluent_kafka/src/Producer.c
```diff
@@ -879,7 +896,7 @@ PyTypeObject ProducerType = {
 	0,                         /*tp_setattr*/
 	0,                         /*tp_compare*/
 	0,                         /*tp_repr*/
-	0,                         /*tp_as_number*/
+	&Producer_num_methods,                         /*tp_as_number*/
```
pranavrth, line 899: Try to keep comment in line with other comments.

pranavrth (changes requested):
Other changes look good. Please rebase it with master.

src/confluent_kafka/src/Producer.c
vladz-sternum: I aligned the nearby `seq_methods` as well, let me know if you want it reverted.

pranavrth:
LGTM!

@edenhill - Any comments?

edenhill (approved):
LGTM, but it would be good with an entry in the CHANGELOG.md too so people know they can start using it like this.

pranavrth (approved):

vladz-sternum:
@pranavrth  I have rebased my branch

PR #1447, opened 2022-10-13
CLIENTS-2607: Updated documentation to emphasize the use of logger instead of log_cb

pranavrth:
These changes are related to the users getting confused for logger callback property. In the docs we refer to librdkafka properties first then explain logger properties of python client. Due to this, users tries to use log_cb property for log callback and fail.

In this PR, I tried to change our docs to reflect to use logger instead of log_cb configuration property.

Changes in the PR:

1. Updated configuration property doc to make sure that we specify python client specific properties before librdkafka properties.
2. added test cases to make sure the custom logger is logging in the correct format


docs/index.rst
```diff
@@ -720,12 +720,8 @@ providing a dict of configuration properties to the instance constructor, e.g.
   consumer = confluent_kafka.Consumer(conf)
 
 
-The supported configuration values are dictated by the underlying
-librdkafka C library. For the full range of configuration properties
-please consult librdkafka's documentation:
```
edenhill, line 725: Since it is mostly the librdkafka configuration properties that are important and typically changed, I think it made more sense to have the link to CONFIGURATION.md up front here.

edenhill (changes requested):

docs/index.rst
pranavrth: I moved it down as people used to read the CONFIGURATION.md before reading the python specific ones. I could find multiple issues where the user was using "log_cb" property instead of "logger" for python client.

Also, I felt that it makes sense to have python client specific properties in the front as it is the python client.

edenhill: Okay, fair enough,.

docs/index.rst
```diff
@@ -778,3 +774,8 @@ The Python bindings also provide some additional configuration properties:
     mylogger.addHandler(logging.StreamHandler())
     producer = confluent_kafka.Producer({'bootstrap.servers': 'mybroker.com'}, logger=mylogger)
 
+.. note::
+   In Python client, the ``logger`` configuration property is used for log handler, not ``log_cb``.
```
edenhill, line 778: In the Python client

edenhill (changes requested):

docs/index.rst
pranavrth: Done

edenhill (approved):
lgtm

PR #1454, opened 2022-10-26
Default new topic params 

mahajanadhitya:
Expected Behaviour : KIP 464 
Making num_partitions optional

### DIFF
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -392,7 +392,7 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
         rd_kafka_NewTopic_t **c_objs;
         rd_kafka_queue_t *rkqu;
         CallState cs;
-        
+
         /* topics is a list of NewTopic objects. */
         if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO|Off", kws,
                                          &topics, &future,
@@ -462,7 +462,7 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                                 i++;
                                 goto err;
                         }
-                        int partitions = sizeof(newt->replica_assignment)/sizeof(newt->replica_assignment[0]);
+                        
                         if (!Admin_set_replica_assignment(
                                     "CreateTopics", (void *)c_objs[i],
                                     newt->replica_assignment,
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -392,6 +392,7 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
         rd_kafka_NewTopic_t **c_objs;
         rd_kafka_queue_t *rkqu;
         CallState cs;
+        
         /* topics is a list of NewTopic objects. */
         if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO|Off", kws,
                                          &topics, &future,
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -454,7 +454,7 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                 }
 
                 if (newt->replica_assignment) {
-                        if (newt->replication_factor != -1 || newt->num_partitions != -1) {
+                        if (newt->replication_factor != -1) {
                                 PyErr_SetString(PyExc_ValueError,
                                                 "replication_factor/num_partitions and "
                                                 "replica_assignment are "
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -392,7 +392,8 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
         rd_kafka_NewTopic_t **c_objs;
         rd_kafka_queue_t *rkqu;
         CallState cs;
-
+        int default_partitions = 3;
+        int default_replications = 1;
         /* topics is a list of NewTopic objects. */
         if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO|Off", kws,
                                          &topics, &future,
@@ -462,11 +463,11 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                                 i++;
                                 goto err;
                         }
-
+                        int partitions = sizeof(newt->replica_assignment)/sizeof(newt->replica_assignment[0]);
                         if (!Admin_set_replica_assignment(
                                     "CreateTopics", (void *)c_objs[i],
                                     newt->replica_assignment,
-                                    newt->num_partitions, newt->num_partitions,  // problem 
+                                    partitions,partitions,  // problem 
                                     "num_partitions")) {
                                 i++;
                                 goto err;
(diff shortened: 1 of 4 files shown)


src/confluent_kafka/src/Admin.c
```diff
@@ -454,19 +455,19 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                 }
 
                 if (newt->replica_assignment) {
-                        if (newt->replication_factor != -1) {
+                        if (newt->replication_factor != -1 || newt->num_partitions !=-1) {
```
mhowlett, line 458: space after the `!=`. i suspect you need to run `flake8`

src/confluent_kafka/src/Admin.c
```diff
@@ -454,19 +455,19 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
...
                                 PyErr_SetString(PyExc_ValueError,
-                                                "replication_factor and "
+                                                "replication_factor/num_partitions and "
```
mhowlett, line 460: @edenhill might get upset with line lengths > 80 characters, though i don't believe this is strictly a convention in this code base.

src/confluent_kafka/src/Admin.c
```diff
@@ -454,19 +455,19 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
...
                                                 "replica_assignment are "
                                                 "mutually exclusive");
                                 i++;
                                 goto err;
                         }
-
+                        int partitions = sizeof(newt->replica_assignment)/sizeof(newt->replica_assignment[0]);
```
mhowlett, line 466: unsure what you are trying to do here, i don't think you need to do anything.

src/confluent_kafka/src/Admin.c
```diff
@@ -481,6 +482,19 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                                 goto err;
                         }
                 }
+                if(!newt->replica_assignment){
```
mhowlett, line 485: `else`

src/confluent_kafka/src/Admin.c
```diff
@@ -392,7 +392,8 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
         rd_kafka_NewTopic_t **c_objs;
         rd_kafka_queue_t *rkqu;
         CallState cs;
-
+        int default_partitions = 3;
+        int default_replications = 1;
```
mhowlett, line 396: the default values don't come from the client, they are new values on the broker.

mhowlett, line 396: the default values don't come from the client, they are new config properties on the broker.

src/confluent_kafka/src/Admin.c
mhowlett: you don't need anything in this block afaict, it's all done broker side.

src/confluent_kafka/src/AdminTypes.c
```diff
@@ -73,12 +73,12 @@ static int NewTopic_init (PyObject *self0, PyObject *args,
                                "replica_assignment",
                                "config",
                                NULL };
-
```
mhowlett, line 76: i wouldn't remove this new line, it's better style to have it imo.

src/confluent_kafka/src/AdminTypes.c
```diff
@@ -73,12 +73,12 @@ static int NewTopic_init (PyObject *self0, PyObject *args,
...
+        self->num_partitions = -1;
         self->replication_factor = -1;
         self->replica_assignment = NULL;
         self->config = NULL;
 
-        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "si|iOO", kws,
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "s|iiOO", kws,
```
mhowlett, line 81: 👍

src/confluent_kafka/src/Admin.c
pranavrth: `flake8` doesn't work for C code.

src/confluent_kafka/src/Admin.c
edenhill: This only works for statically sized arrays, e.g., `int foo[4];`, in fact NewTopic->replica_assignment isn't an array at all but a single PyObject that represents a Python list.

Also; variable declarations are only allowed at the beginning of scopes and blocks, to support older MSVC compilers.

src/confluent_kafka/src/Admin.c
```diff
@@ -481,6 +482,19 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                                 goto err;
                         }
                 }
+                if(!newt->replica_assignment){
+                        if(newt->replication_factor!=-1 && newt->num_partitions!=-1){
```
edenhill, line 486: We're not strictly enforcing code style in this repo, but try to maintain the same code style as for the other code, which has much more spaces around parens, operators, etc, to improve readability.

edenhill (changes requested):
Maintain the existing code style (more spaces!).
Add a test case.
The default partition count and replications are defined by the broker and should not be enforced here.

edenhill:
Also add CHANGELOG entry.

src/confluent_kafka/src/Admin.c
mahajanadhitya: Noted , used Python length function in C API for that!

src/confluent_kafka/src/Admin.c
mahajanadhitya: Understood!

src/confluent_kafka/src/Admin.c
mahajanadhitya: I have made the suggested changes

src/confluent_kafka/src/Admin.c
```diff
@@ -454,19 +454,20 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
...
                                 PyErr_SetString(PyExc_ValueError,
-                                                "replication_factor and "
+                                                "replication_factor/num_partitions and "
                                                 "replica_assignment are "
                                                 "mutually exclusive");
                                 i++;
                                 goto err;
                         }
-
+
```
edenhill, line 465: Keep the line as it were, just an empty line without whitespaces

src/confluent_kafka/src/Admin.c
```diff
@@ -481,6 +482,16 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
                                 goto err;
                         }
                 }
+                if(!newt->replica_assignment){
```
edenhill, line 485: Maintain current code style, space before paren, space after paren:
`if (!newt->...) {`

src/confluent_kafka/src/Admin.c
```diff
@@ -481,6 +482,16 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
...
+                        if((newt->replication_factor != -1 && newt->num_partitions != -1) || (newt->replication_factor == -1 && newt->num_partitions == -1)){
```
edenhill, line 486: Keep lines below 80 columns.
add spaces.

src/confluent_kafka/src/Admin.c
```diff
@@ -481,6 +482,16 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
...
+                                continue;
+                        }
+                        else{
```
edenhill, line 489: maintain current code style:
`} else {`

src/confluent_kafka/src/Admin.c
```diff
@@ -481,6 +482,16 @@ static PyObject *Admin_create_topics (Handle *self, PyObject *args,
...
+                                i++;
+                                goto err;
```
edenhill, line 491: Recommend switch the if-statement to avoid an else block:
```c
if (..whatever is wrong) {
  i++;
  goto err;
}
```

General recommendation is to keep the level of indent blocks down and typically only do indents for the error cases, e.g.:
```c
if (some_error) {
  handle_it..
  and stuff;
}

good_things();

if (some other_error)
  goto err;

more good things();

```

edenhill (changes requested):
Mostly style comments now.
Don't forget to update the CHANGELOG.md

pranavrth:
Please solve the build issues and the above mentioned comments by Magnus and Matt.

PR #1455, opened 2022-10-27
CLIENTS-1765: Added support for consumer.memberid() #1154

pranavrth:
Added consumer.memberid() for getting member id assigned to the consumer in a consumer group (#1154)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 
 ## v1.10.0
 
-- Add metadata to TopicPartition type and commit() (#1410).
+ - Added metadata to TopicPartition type and commit() (#1410).
+ - Added consumer.memberid() for getting member id assigned to 
+   the consumer in a consumer group (#1154).
 
 
 ## v1.9.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 
 ## v1.10.0
 
- - Added metadata to TopicPartition type and commit() (#1410).
- - Added consumer.memberid() for getting member id assigned to 
+ - Added metadata to `TopicPartition` type and `commit()` (#1410).
+ - Added `consumer.memberid()` for getting member id assigned to 
    the consumer in a consumer group (#1154).
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Confluent's Python client for Apache Kafka
 
 
-## v1.10.0
+## v1.9.3
 
  - Added metadata to `TopicPartition` type and `commit()` (#1410).
  - Added `consumer.memberid()` for getting member id assigned to 
--- a/src/confluent_kafka/src/Consumer.c
+++ b/src/confluent_kafka/src/Consumer.c
@@ -995,7 +995,7 @@ static PyObject *Consumer_memberid (Handle *self, PyObject *args,
         if (!memberid)
                 Py_RETURN_NONE;
 
-        if (strlen(memberid) == 0) {
+        if (!*memberid) {
                 rd_kafka_mem_free(self->rk, memberid);
                 Py_RETURN_NONE;
         }
--- a/src/confluent_kafka/src/Consumer.c
+++ b/src/confluent_kafka/src/Consumer.c
@@ -995,6 +995,11 @@ static PyObject *Consumer_memberid (Handle *self, PyObject *args,
         if (!memberid)
                 Py_RETURN_NONE;
 
+        if (strlen(memberid) == 0) {
+                rd_kafka_mem_free(self->rk, memberid);
+                Py_RETURN_NONE;
+        }
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 
 ## v1.10.0
 
-- Add metadata to TopicPartition type and commit() (#1410).
+ - Added metadata to TopicPartition type and commit() (#1410).
```
edenhill, line 6: hint: use backticks to make code stand out more, it looks nicer and makes it easier for the reader to identify what is code, e.g.:

Added metadata to `TopicPartition` type and `commit()`

..

Added `conusmer.memberid()`..

tests/integration/consumer/test_consumer_memberid.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2021 Confluent Inc.
```
edenhill, line 4: 2022 :)

tests/integration/consumer/test_consumer_memberid.py
```diff
@@ -0,0 +1,42 @@
...
+    consumer_conf = {'group.id': 'test'}
+
+    topic = "testmemberid"
+
+    kafka_cluster.create_topic(topic)
+
+    consumer = kafka_cluster.consumer(consumer_conf)
+
+    assert consumer is not None
+    assert len(consumer.memberid()) == 0
```
edenhill, line 33: I think you want to check for None here, since that's what the API is supposed to return when there is no memberid. so: `.. is None`

tests/integration/consumer/test_consumer_memberid.py
```diff
@@ -0,0 +1,42 @@
...
+    kafka_cluster.seed_topic(topic, value_source=[b'memberid'])
+
+    consumer.subscribe([topic])
+    msg = consumer.poll(10)
+    assert msg is not None
+    assert msg.value() == b'memberid'
+    assert isinstance(consumer.memberid(), str) is True
+    assert len(consumer.memberid()) > 0
+    consumer.close()
```
edenhill, line 42: You could even add a check that the consumer-is-closed exception is raised if you call memberid() after close() here, I think there is something like:
```python

with pytest.raises(some_confluent_exception):
    consumer.memberid()
```

edenhill (changes requested):
Good stuff! Left some minor comments

src/confluent_kafka/src/Consumer.c
```diff
@@ -980,6 +980,28 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+                                "Consumer closed");
+                return NULL;
+        }
+
+        memberid = rd_kafka_memberid(self->rk);
+
+        if (!memberid)
+                Py_RETURN_NONE;
+
+        memberidobj = Py_BuildValue("s", memberid);
```
edenhill, line 998: librdkafka will return an empty string rather than null if it hasn't received a memberid from the coordinator yet, but I think it makes more sense to return None from the python bindings in that case, so recommend adding that.
E.g.,;

edenhill: Nevermind, it looks like it'll actually return NULL

pranavrth: It is returning empty string.

CHANGELOG.md
pranavrth: Done

tests/integration/consumer/test_consumer_memberid.py
pranavrth: Done

tests/integration/consumer/test_consumer_memberid.py
pranavrth: Added exception test

src/confluent_kafka/src/Consumer.c
```diff
@@ -980,6 +980,33 @@ static PyObject *Consumer_poll (Handle *self, PyObject *args,
...
+                                "Consumer closed");
+                return NULL;
+        }
+
+        memberid = rd_kafka_memberid(self->rk);
+
+        if (!memberid)
+                Py_RETURN_NONE;
+
+        if (strlen(memberid) == 0) {
```
edenhill, line 998: nit: it is better to check if a string is empty by simply inspecting the first byte, which is O(1), rather than strlen() which is O(n). This doesn't matter at all here, of course, but it is good practice so mentioning it anyway:

```c
if (!*memberid) { 
```

edenhill (approved):
LGTM!

src/confluent_kafka/src/Consumer.c
pranavrth: Thanks for the suggestion. I still think in Java terms I think. :)

edenhill (approved):
LGTM!

PR #1456, opened 2022-10-27
Fixed build issue related to zlib version and download url

pranavrth:
Fixed build issue related to zlib version and download url.

- Pointed librdkafka src version to master instead of some release version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 env:
  global:
   - LIBRDKAFKA_VERSION=v1.9.2
-  - LIBRDKAFKA_SRC_CODE_VERSION=master
+  - LIBRDKAFKA_SRC_VERSION=master
 
 jobs:
  include:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,6 +1,7 @@
 env:
  global:
   - LIBRDKAFKA_VERSION=v1.9.2
+  - LIBRDKAFKA_SRC_CODE_VERSION=master
```
edenhill, line 4: tiny nit: I'd drop the "_CODE", a bit redundant.

.travis.yml
```diff
@@ -63,7 +64,7 @@ install:
  - if [[ $BUILD_WHEELS != 1 ]]; then pip install -r tests/requirements.txt ; fi
  - if [[ $MK_DOCS == y ]]; then pip install -r docs/requirements.txt; fi
  # Install librdkafka and confluent_kafka[avro] if not building wheels
- - if [[ $BUILD_WHEELS != 1 ]]; then pip install -U protobuf && tools/bootstrap-librdkafka.sh --require-ssl ${LIBRDKAFKA_VERSION} tmp-build ; fi
+ - if [[ $BUILD_WHEELS != 1 ]]; then pip install -U protobuf && tools/bootstrap-librdkafka.sh --require-ssl ${LIBRDKAFKA_SRC_CODE_VERSION} tmp-build ; fi
```
edenhill, line 67: I think this is a good permanent change which makes sure that the Python client is always buildable with the latest master of librdkafka (which should always be buildable as well). This will give us early error detection rather than later when bumping the released version.

edenhill (changes requested):

.travis.yml
pranavrth: Done

edenhill (approved):
LGTM

Feel free to merge when CIs pass

PR #1463, opened 2022-11-08
Fix consumer example

anchitj:
Manually store offsets using `store_offsets` to enable at-least once semantics and preventing consumers from committing offsets in case of error in processing.

Referenced from https://github.com/confluentinc/confluent-kafka-dotnet/pull/1918/files

### DIFF
--- a/examples/consumer.py
+++ b/examples/consumer.py
@@ -52,7 +52,7 @@ if __name__ == '__main__':
     # Consumer configuration
     # See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
     conf = {'bootstrap.servers': broker, 'group.id': group, 'session.timeout.ms': 6000,
-            'auto.offset.reset': 'earliest', 'enable.auto.offset.store': 'false'}
+            'auto.offset.reset': 'earliest', 'enable.auto.offset.store': False}
 
     # Check to see if -T option exists
     for opt in optlist:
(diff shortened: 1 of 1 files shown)


examples/consumer.py
```diff
@@ -102,6 +102,9 @@ def print_assignment(consumer, partitions):
                                  (msg.topic(), msg.partition(), msg.offset(),
                                   str(msg.key())))
                 print(msg.value())
+                # Store the offset associated with msg to a local cache. Stored offsets are committed to Kafka by a background thread every AutoCommitIntervalMs
```
edenhill, line 105: It is called auto.commit.interval.ms in Python.

Also try to keep under 80 columns in width.

examples/consumer.py
```diff
@@ -102,6 +102,9 @@ def print_assignment(consumer, partitions):
...
+                # Explicitly storing offsets after processing gives at-least once semantics, the default behavior does not.
+                c.store_offsets(message=msg)
```
edenhill, line 107: Not sure message=msg is helpful, it suggests this is not the default usage and that there are other parameters, which there aren't. Just pass the msg as is without named parameters.

edenhill (changes requested):

examples/consumer.py
anchitj: Thanks, updated. Broken down the comment to multiple lines to reduce the width.

examples/consumer.py
anchitj: Makes sense, updated now.

edenhill (approved):
LGTM

PR #1467, opened 2022-11-18
CLIENTS-2885: Wheel generation python 3.11

pranavrth:
Changes to generate python 3.11 wheel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,13 +51,6 @@ jobs:
     env: BUILD_WHEELS=1
     script: tools/wheels/build-wheels.sh ${LIBRDKAFKA_VERSION#v} wheelhouse
 
-  - name: "Wheels: MacOSX x64"
-    if: tag is present
-    os: osx
-    language: shell
-    env: BUILD_WHEELS=1
-    script: tools/wheels/build-wheels.sh ${LIBRDKAFKA_VERSION#v} wheelhouse
-
 install:
  # Install interceptors
  - tools/install-interceptors.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
  - Added metadata to `TopicPartition` type and `commit()` (#1410).
  - Added `consumer.memberid()` for getting member id assigned to 
    the consumer in a consumer group (#1154).
- - Added Python 3.11 wheel
+ - Added Python 3.11 wheels
 
 
 ## v1.9.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Confluent's Python client for Apache Kafka
 
 
-## v1.9.3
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -51,13 +51,6 @@ jobs:
     env: BUILD_WHEELS=1
     script: tools/wheels/build-wheels.sh ${LIBRDKAFKA_VERSION#v} wheelhouse
 
-  - name: "Wheels: MacOSX x64"
```
edenhill, line 54: Why is this removed?

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
 # Confluent's Python client for Apache Kafka
 
 
-## v1.9.3
+## Next Version
 
  - Added metadata to `TopicPartition` type and `commit()` (#1410).
  - Added `consumer.memberid()` for getting member id assigned to 
    the consumer in a consumer group (#1154).
+ - Added Python 3.11 wheel
```
edenhill, line 9: "wheels."

edenhill (changes requested):
Has this been tested for all platforms by pushing a tag (.e.g, v2.0.0wheeltest1)?
Things usually break when upgrading cibuild or delvewheel

.travis.yml
pranavrth: These wheels are generated with Semaphore as well. Travis requires credits OS wheel generation.

pranavrth:
I have pushed v1.8.99 in Test PyPi. I also ran test-wheels.py on the wheels.

edenhill:
Don't push tags that resemble real versions (vA.B.C) - people will end up finding them (you'd be surprised) - always use a suffix that makes it clear it is a test tag, e.g., v1.9.3tst1 or similar.

pranavrth:
It wasn't allowing to push v1.9.3tst4 due to [PEP-440](https://peps.python.org/pep-0440/). I will remember your point for next test release. Shall I delete v1.8.99?

edenhill:
Seems like PEP-440 proposes using `devN` suffix.

Yes, delete that tag.

pranavrth:
Deleted.

1 Test is failing with the latest changes in librdkafka.

edenhill:
The transaction APIs will now return ERR__CONFLICT if there is an another ongoing or started API call.

We still return ERR__STATE if an API call is not valid in the current state, but the conflict check is done first, thus the change of error returns.
I recommend updating the test in python.

edenhill:
I don't see a tag build on travis, did you push a tag?

pranavrth:
I added v1.9.3tst5 tag to an older commit. Build for this is - https://app.travis-ci.com/github/confluentinc/confluent-kafka-python/builds/257637307

edenhill (approved):
LGTM!

PR #1471, opened 2022-11-23
removed inline variables

mahajanadhitya:


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1621,7 +1621,7 @@ static void oauth_cb (rd_kafka_t *rk, const char *oauthbearer_config,
         Py_ssize_t rd_extensions_size = 0;
         char err_msg[2048];
         rd_kafka_resp_err_t err_code;
-
```
edenhill, line 1624: Put all the variable declarations together, then an empty line for the rest of the code.
i.e.;
```c
        Py_ssize_t rd_extensions_size = 0;
        char err_msg[2048];
        rd_kafka_resp_err_t err_code;
        int i;

        cs = CallState_get(h);
```

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1660,7 +1660,7 @@ static void oauth_cb (rd_kafka_t *rk, const char *oauthbearer_config,
                                                   sizeof(err_msg));
         Py_DECREF(result);
         if (rd_extensions) {
-                for(int i = 0; i < rd_extensions_size; i++) {
+                for(i = 0; i < rd_extensions_size; i++) {
```
edenhill, line 1663: Since `i` is only used in this scope/block, declare i on the line above instead of at the head of the function

edenhill (changes requested):
This PR has a bunch of other unrelated stuff that's already on master.
Rebase your branch on master so it only includes the c89 stuff.

Issue #1262, opened 2022-01-12
can't install v1.8.2 via pip on python2.7


chriselion:
Description
===========
Running `pip install confluent-kafka==1.8.2` on python 2.7 results in an error during the installation process. This is due to py2-incompatible code in setup.py:
https://github.com/confluentinc/confluent-kafka-python/blob/2ac0d72b24b14e5246445ad9ce66ec9c8828ef4e/setup.py#L12-L13

How to reproduce
================
```
docker run --rm -it python:2.7-slim-buster bash
pip install confluent-kafka==1.8.2
```
This produces the output:
```
# pip install confluent-kafka==1.8.2
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Collecting confluent-kafka==1.8.2
  Downloading confluent-kafka-1.8.2.tar.gz (104 kB)
     |████████████████████████████████| 104 kB 1.6 MB/s 
    ERROR: Command errored out with exit status 1:
     command: /usr/local/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-HN9px5/confluent-kafka/setup.py'"'"'; __file__='"'"'/tmp/pip-install-HN9px5/confluent-kafka/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-HN9px5/confluent-kafka/pip-egg-info
         cwd: /tmp/pip-install-HN9px5/confluent-kafka/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-HN9px5/confluent-kafka/setup.py", line 12, in <module>
        with open(os.path.join(work_dir, 'README.md'), encoding='utf-8') as f:
    TypeError: 'encoding' is an invalid keyword argument for this function
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```
[...]

PR #1487, opened 2022-12-25
Drop support for Python 2

Viicos:
_Originally posted by @chriselion in https://github.com/confluentinc/confluent-kafka-python/issues/1262#issuecomment-1012652420_

Python 3.2 seems to be the minimum required version, as [`functools.total_ordering`](https://docs.python.org/3/library/functools.html#functools.total_ordering) is used in some places.

If this is ok, I can go through the codebase and remove obsolete code related to Python2 compatibility. This would make it possible for me to catch anything that would require a higher version of Python


Viicos:
We could even raise the required version to `3.8`? And remove `six` integration

Viicos:
Handled since then.

PR #1492, opened 2023-01-04
Document that offsets_for_times accepts milliseconds

timmc-edx:
Other confluent-kafka-python APIs accept seconds since epoch, but this one uses milliseconds, so it's easy to think it accepts seconds. That would generally result in replaying all events.

### DIFF
--- a/src/confluent_kafka/src/Consumer.c
+++ b/src/confluent_kafka/src/Consumer.c
@@ -1427,6 +1427,7 @@ static PyMethodDef Consumer_methods[] = {
           " timestamp is greater than or equal to the given timestamp in the\n"
           " corresponding partition. If the provided timestamp exceeds that of the\n"
           " last message in the partition, a value of -1 will be returned.\n"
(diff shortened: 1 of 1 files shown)


timmc-edx:
I'd have to go to Legal to check if I can sign that CLA, which doesn't seem worth it for a one-line doc change. If someone who has already signed wants to make a similar change instead, go for it.

timmc-edx:
I was able to get Legal approval to sign the CLA.

timmc-edx:
@pranavrth (Picking you at random) -- can I get someone to take a look at this small change? I suspect this PR was simply missed due to the CLA issues, which I've since resolved.

timmc-edx:
@pranavrth Does this repo accept contributions?

timmc-edx:
Tried to get a review from contact at Confluent as well, but no luck. I have to assume this repo doesn't actually take contributions -- closing out PR.

PR #1498, opened 2023-01-16
gnu90 compatibility

mahajanadhitya:


emasab (changes requested):
LGTM: this way it compiles correctly with CFLAGS=-std=gnu90, GCC defaults to gnu90 before 5.1.0, CentOS 7 has this GCC version and it supported up to 2024.

Just create a second PR with a `feature/gnu90-compatibility` branch for running the CI

emasab:
#1512 was merged

PR #1517, opened 2023-02-03
Added the binding to Python Client for KIP 339 to incrementally update the config

mahajanadhitya:


src/confluent_kafka/admin/_config.py
```diff
@@ -177,3 +204,28 @@ def set_config(self, name, value, overwrite=True):
...
+        if name is None:
+            raise TypeError("Configuration name is needed")
+        if not isinstance(name, str):
+            raise TypeError("Configuration name must be a string")
+
+        if not isinstance(operation, AlterConfigOpType):
+            raise TypeError("Operation must be of type AlterConfigOpType")
+
+        if not isinstance(value, str) and operation != AlterConfigOpType.DELETE:
+            raise ValueError("The provided value should be a string for: " + operation.name)
```
pranavrth, line 228: TypeError

src/confluent_kafka/admin/_config.py
```diff
@@ -177,3 +204,28 @@ def set_config(self, name, value, overwrite=True):
         if not overwrite and name in self.set_config_dict:
             return
         self.set_config_dict[name] = value
+
```
pranavrth, line 207: I feel we should have different class for Incremental Alter configs. We are going to Remove Alter Configs in future. This implementation is confusing for now and will be difficult to reason about in future.

src/confluent_kafka/admin/_config.py
```diff
@@ -177,3 +204,28 @@ def set_config(self, name, value, overwrite=True):
...
+        Applies an incremental operation on specified key, while keeping
+        all the others unchanged.
+
+        :param str name: Configuration property name
+        :param AlterConfigOpType operation: Alter config operation
+        :param str value: Configuration value (optional if operation is DELETE)
+        """
+        if name is None:
+            raise TypeError("Configuration name is needed")
```
pranavrth, line 220: ValueError

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
 
         return 1;
 }
```
pranavrth, line 330: 2 empty lines.

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                        return 0;
+                }
+                k = cfl_PyUnistr_AsUTF8(ks, &ks8);
+
+                Py_ssize_t iter = 0;
+                PyObject *internal_ko,*internal_vo;
+                PyObject *internal_ks, *internal_ks8;
+                PyObject *internal_vs = NULL, *internal_vs8 = NULL;
+                const char *internal_k;
+                const char *internal_v;
```
pranavrth, line 354: Move these declarations to the top of the block.

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
 
         return 1;
 }
+static int
+Admin_incremental_config_to_c(void *c_obj, PyObject *dict){
```
pranavrth, line 332: I think `dict` should be first parameter and `c_obj` should be second

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                        } else if(!strcmp(internal_k,"value")) {
+                                if (!(internal_vs = cfl_PyObject_Unistr(internal_vo)) || !(internal_v = cfl_PyUnistr_AsUTF8(internal_vs, &internal_vs8))) {
+                                        PyErr_Format(PyExc_ValueError,
+                                                "expected value name to be unicode "
+                                                "string");
+                                        Py_XDECREF(internal_vs);
+                                        Py_XDECREF(internal_vs8);
+                                        Py_DECREF(internal_ks);
+                                        Py_XDECREF(internal_ks8);
+                                        return 0;
```
pranavrth, line 379: 1. We should decrease ref count for ks and ks8 as well. 
2. Move these to single `goto err` block.

src/confluent_kafka/admin/_config.py
```diff
@@ -177,3 +204,28 @@ def set_config(self, name, value, overwrite=True):
...
+            raise TypeError("Configuration name must be a string")
+
+        if not isinstance(operation, AlterConfigOpType):
+            raise TypeError("Operation must be of type AlterConfigOpType")
+
+        if not isinstance(value, str) and operation != AlterConfigOpType.DELETE:
+            raise ValueError("The provided value should be a string for: " + operation.name)
+        if value is None and operation != AlterConfigOpType.DELETE:
+            raise ValueError("Value is needed for operation: " + operation.name)
+        self.incremental_config[name] = {"operation_type": operation, "value": value}
```
pranavrth, line 231: Why are we keeping this as `dict` instead of `list of tuples`? like `[(k1, o1, v1), ...]` instead of `{k1: {operation_type: o1, value: v1}, ...}`?

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                                     "expected config name to be unicode "
+                                     "string");
+                        return 0;
+                }
+                k = cfl_PyUnistr_AsUTF8(ks, &ks8);
+
+                Py_ssize_t iter = 0;
+                PyObject *internal_ko,*internal_vo;
+                PyObject *internal_ks, *internal_ks8;
+                PyObject *internal_vs = NULL, *internal_vs8 = NULL;
```
pranavrth, line 352: I think we can do better than internal in naming? Internal is confusing. Maybe `inner_dict` prefix would be more understandable?

I would say better overall naming.

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                                        return 0;
+                                }
+
+                                v = internal_v;
+                        } else {
+                                PyErr_Format(PyExc_ValueError,"Unknown property:%s , allowed are operation_type and value",internal_k);
+                                Py_XDECREF(internal_vs);
+                                Py_XDECREF(internal_vs8);
+                                Py_DECREF(internal_ks);
+                                Py_XDECREF(internal_ks8);
```
pranavrth, line 388: same

src/confluent_kafka/src/Admin.c
```diff
@@ -328,7 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+
+                                return 0;
+                        }
+
+                        Py_XDECREF(internal_vs);
+                        Py_XDECREF(internal_vs8);
+                        Py_DECREF(internal_ks);
+                        Py_XDECREF(internal_ks8);
```
pranavrth, line 396: same

src/confluent_kafka/src/Admin.c
```diff
@@ -911,7 +998,153 @@ static PyObject *Admin_describe_configs (Handle *self, PyObject *args,
...
 
+static PyObject *Admin_incremental_alter_configs(Handle *self,PyObject *args,PyObject *kwargs) {
+        PyObject *resources, *future;
+        PyObject *validate_only_obj = NULL;
+        static char *kws[] = { "resources",
+                               "future",
+                               /* options */
+                               "validate_only",
+                               "request_timeout",
+                               "broker",
```
pranavrth, line 1009: I think I didn't see `broker` option in librdkafka. Can you explain what it does?

src/confluent_kafka/src/Admin.c
```diff
@@ -911,7 +998,153 @@ static PyObject *Admin_describe_configs (Handle *self, PyObject *args,
...
+                               NULL };
+        struct Admin_options options = Admin_options_INITIALIZER;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        PyObject *ConfigResource_type;
+        int cnt, i;
+        rd_kafka_ConfigResource_t **c_objs;
+        rd_kafka_queue_t *rkqu;
+        CallState cs;
+
+        /* topics is a list of NewPartitions_t objects. */
```
pranavrth, line 1019: I think this comment doesn't belong here.

src/confluent_kafka/admin/__init__.py
```diff
@@ -522,6 +523,35 @@ def alter_configs(self, resources, **kwargs):
 
         return futmap
 
+    def incremental_alter_configs(self, resources, **kwargs):
+        """
+        Update configuration properties for the specified resources.
+        Updates are incremental, i.e only the values mentioned are changed
+        and rest remain as is.
+
+        :param list(ConfigResource) resources: Resources to update configuration of.
```
pranavrth, line 532: Instead of `list(ConfigResource)`, can we create a class `IncrementalAlterConfigsRequest` having this list as a field and take that as input. So that its extensible in future if required.

src/confluent_kafka/src/Admin.c
```diff
@@ -911,7 +998,153 @@ static PyObject *Admin_describe_configs (Handle *self, PyObject *args,
...
+        if (validate_only_obj &&
+            !cfl_PyBool_get(validate_only_obj, "validate_only",
+                            &options.validate_only))
+                return NULL;
+        c_options = Admin_options_to_c(self, RD_KAFKA_ADMIN_OP_INCREMENTALALTERCONFIGS,
+                                       &options, future);
+        if (!c_options)
+                return NULL; /* Exception raised by options_to_c() */
+
+        /* Look up the ConfigResource class so we can check if the provided
```
pranavrth, line 1044: Comments are incorrect at multiple places. Please check all the comments.

src/confluent_kafka/src/Admin.c
```diff
@@ -911,7 +998,153 @@ static PyObject *Admin_describe_configs (Handle *self, PyObject *args,
...
+                                        "Expected list of "
+                                        "ConfigResource objects");
+                        goto err;
+                }
+
+                if (!cfl_PyObject_GetInt(res, "restype_int", &restype, 0, 0))
+                        goto err;
+
+                if (!cfl_PyObject_GetString(res, "name", &resname, NULL, 0, 0))
+                        goto err;
```
pranavrth, line 1087: Can be a single condition.

pranavrth (changes requested):
Did first round of Review.

src/confluent_kafka/admin/_config.py
emasab: It can be overwritten too

src/confluent_kafka/src/Admin.c
emasab: It's also in AlterConfigs, it allows to select a specific broker instead or the controller. It's possible to set it in all the APIs.
```c
                /* Use explicitly specified broker_id, if available. */
                broker_id = (int32_t)rd_kafka_confval_get_int(
                    &rko->rko_u.admin_request.options.broker);

                if (broker_id != -1) {
                        rd_kafka_dbg(rk, ADMIN, name,
                                     "%s using explicitly "
                                     "set broker id %" PRId32
                                     " rather than %" PRId32,
                                     name, broker_id,
                                     rko->rko_u.admin_request.broker_id);
                        rko->rko_u.admin_request.broker_id = broker_id;
                } else {
                        /* Default to controller */
                        broker_id = RD_KAFKA_ADMIN_TARGET_CONTROLLER;
                }
```

src/confluent_kafka/admin/__init__.py
emasab: In J. it cannot be extended, as they have a Map currently. If there's need to extend it they'll create a new API too.

src/confluent_kafka/src/Admin.c
emasab: Added a condition in librdkafka to revert to the default if altering broker configs and documented this option here.

src/confluent_kafka/admin/_config.py
emasab: Same considerations as we had in librdkafka apply here, as the ConfigResource contains a list of ConfigEntry differently from J.

src/confluent_kafka/src/Admin.c
```diff
@@ -329,6 +329,98 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+
+                if (!(ks = cfl_PyObject_Unistr(ko))) {
+                        PyErr_Format(PyExc_ValueError,
+                                     "expected config name to be unicode "
+                                     "string");
+                       goto err;
+                }
+                k = cfl_PyUnistr_AsUTF8(ks, &ks8);
+
+                const char *local_v;
```
pranavrth, line 357: Position not right for this i think

src/confluent_kafka/src/Admin.c
```diff
@@ -329,6 +329,98 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                Py_ssize_t iter = 0;
+                PyObject *local_ko,*local_vo;
+
+                if (!(ks = cfl_PyObject_Unistr(ko))) {
+                        PyErr_Format(PyExc_ValueError,
+                                     "expected config name to be unicode "
+                                     "string");
+                       goto err;
+                }
+                k = cfl_PyUnistr_AsUTF8(ks, &ks8);
```
pranavrth, line 355: I think this logic of converting a PyObject * string to char * should be moved to a function and reused. This makes the code unreadable in general. General comment. Not necessary to implement right now.

pranavrth (changes requested):
We should write integration test as well.

src/confluent_kafka/admin/_config.py
pranavrth: I was asking this so that we have a new constructor to have `set_incremental_config` parameter as well. We can add it here itself with default to none. Like what is happening with `set_config`.

src/confluent_kafka/admin/__init__.py
```diff
@@ -522,6 +523,39 @@ def alter_configs(self, resources, **kwargs):
...
+        :returns: A dict of futures for each resource, keyed by the ConfigResource.
+                  The future result() method returns None or throws a KafkaException.
+
+        :rtype: dict(<ConfigResource, future>)
+
+        :raises KafkaException: Operation failed locally or on broker.
+        :raises TypeError: Invalid type.
+        :raises ValueError: Invalid value.
+        """
+        f, futmap = AdminClient._make_futures(resources, ConfigResource,
```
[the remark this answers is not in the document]
emasab: Validation is in the C code, there are some unit tests for checking that it throws TypeError, the rest of the validation is in C

src/confluent_kafka/admin/_config.py
emasab: Done

src/confluent_kafka/src/Admin.c
```diff
@@ -328,93 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
-                                        "string");
-                                return 0;
+                if (!PyList_Check(vo) || (config_count = (int)PyList_Size(vo)) < 1) {
+                        PyErr_SetString(PyExc_ValueError,
+                                        "expected non-empty list of entry values");
+                        goto err;
+                }
+
+                for (i = 0; i < config_count; i++) {
+                        PyObject *config_op_value, *config_value,
```
pranavrth, line 361: config_op_value and config_value are confusing. Better use `config_op_and_value` instead of `config_op_value`.

src/confluent_kafka/src/Admin.c
```diff
@@ -328,93 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                                 *op_type;
+                        Py_ssize_t config_value_cnt;
+                        config_op_value = PyList_GET_ITEM(vo, i);
+                        rd_kafka_AlterConfigOpType_t op;
+                        const char *v = NULL;
+
+                        if (!PyList_Check(config_op_value) ||
+                            (config_value_cnt = (int)PyList_Size(config_op_value)) < 2) {
```
pranavrth, line 369: `config_value_cnt = (int)PyList_Size(config_op_value)) != 2` instead of `config_value_cnt = (int)PyList_Size(config_op_value)) < 2` as we are expecting exactly 2 values.

src/confluent_kafka/src/Admin.c
```diff
@@ -328,93 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                if (!PyList_Check(vo) || (config_count = (int)PyList_Size(vo)) < 1) {
+                        PyErr_SetString(PyExc_ValueError,
+                                        "expected non-empty list of entry values");
+                        goto err;
+                }
+
+                for (i = 0; i < config_count; i++) {
+                        PyObject *config_op_value, *config_value,
+                                 *op_type;
+                        Py_ssize_t config_value_cnt;
```
pranavrth, line 363: We dont need this vairable.

src/confluent_kafka/src/Admin.c
```diff
@@ -328,93 +328,94 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                        config_op_value = PyList_GET_ITEM(vo, i);
+                        rd_kafka_AlterConfigOpType_t op;
+                        const char *v = NULL;
+
+                        if (!PyList_Check(config_op_value) ||
+                            (config_value_cnt = (int)PyList_Size(config_op_value)) < 2) {
+                                PyErr_SetString(PyExc_ValueError,
+                                                "expected operation type and value");
```
pranavrth, line 371: Better message.

pranavrth (changes requested):

src/confluent_kafka/admin/_config.py
```diff
@@ -125,12 +125,15 @@ class ConfigResource(object):
...
-                 set_config=None, described_configs=None, error=None):
+                 set_config=None, described_configs=None, error=None,
+                 set_incremental_config=None):
         """
         :param ConfigResource.Type restype: Resource type.
         :param str name: The resource name, which depends on restype.
                          For RESOURCE_BROKER, the resource name is the broker id.
         :param dict set_config: The configuration to set/overwrite. Dictionary of str, str.
+        :param dict set_incremental_config: The configuration to alter incrementally.
+               Dictionary of str to [([AlterConfigOpType, value]|[AlterConfigOpType.DELETE])].
```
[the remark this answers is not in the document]
emasab: Check latest version of the documentation

src/confluent_kafka/admin/_config.py
```diff
@@ -225,7 +237,9 @@ def set_incremental_config(self, name, operation, value=None):
...
 
         if not isinstance(value, str) and operation != AlterConfigOpType.DELETE:
-            raise ValueError("The provided value should be a string for: " + operation.name)
+            raise TypeError("The provided value should be a string for: " + operation.name)
         if value is None and operation != AlterConfigOpType.DELETE:
             raise ValueError("Value is needed for operation: " + operation.name)
-        self.incremental_config[name] = {"operation_type": operation, "value": value}
+        if name not in self.incremental_config:
+            self.incremental_config[name] = []
+        self.incremental_config[name].append([operation, value])
```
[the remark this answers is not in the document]
emasab: Leaving as is after internal discussion because of characteristics the single element tuple

src/confluent_kafka/admin/_config.py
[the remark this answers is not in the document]
pranavrth: Its more clear but still confusing as `list(AlterConfigOpType, str)|list(AlterConfigOpType.DELETE)` make it feel that this list can have more than 2 values. `list([AlterConfigOpType, str] | [AlterConfigOpType.DELETE])` seems more clear? or Adding an example would help? What do you think?

src/confluent_kafka/src/Admin.c
```diff
@@ -352,28 +352,30 @@ Admin_incremental_config_to_c(PyObject *dict, void *c_obj){
...
+                        config_op_and_value = PyList_GET_ITEM(vo, i);
+                        int op;
                         const char *v = NULL;
 
-                        if (!PyList_Check(config_op_value) ||
-                            (config_value_cnt = (int)PyList_Size(config_op_value)) < 2) {
-                                PyErr_SetString(PyExc_ValueError,
-                                                "expected operation type and value");
+                        if (!PyList_Check(config_op_and_value) ||
+                            (int)PyList_Size(config_op_and_value) != 2) {
```
pranavrth, line 370: Typecasting is not needed anymore.

src/confluent_kafka/admin/_config.py
```diff
@@ -18,6 +18,32 @@
...
+    #: Set the value of the configuration entry.
+    SET = _cimpl.ALTER_CONFIG_OP_TYPE_SET
+
+    #: Revert the configuration entry
+    #: to the default value (possibly null).
+    DELETE = _cimpl.ALTER_CONFIG_OP_TYPE_DELETE
+
+    #: (For list-type configuration entries only.)
+    #:  Add the specified values
+    #:  to the current value
```
pranavrth, line 36:
```suggestion
    #:  to the current list of values
```

src/confluent_kafka/admin/_config.py
```diff
@@ -18,6 +18,32 @@
...
+    #:  of the configuration entry.
+    APPEND = _cimpl.ALTER_CONFIG_OP_TYPE_APPEND
+
+    #: (For list-type configuration entries only.)
+    #:  Removes the specified values
+    #:  from the current value
```
pranavrth, line 42:
```suggestion
    #:  from the current list of values
```

src/confluent_kafka/admin/_config.py
```diff
@@ -18,6 +18,32 @@
 from ._resource import ResourceType
 
 
+class AlterConfigOpType(Enum):
```
pranavrth, line 21: We can just call `configuration` instead of `configuration entry`

src/confluent_kafka/admin/_config.py
```diff
@@ -133,6 +162,16 @@ def __init__(self, restype, name,
...
+        self.incremental_config = dict()
+        if incremental_configs is not None:
+            for name, op_type_values in incremental_configs.items():
+                for op_type_value in op_type_values:
+                    op_type = op_type_value[0]
+                    if op_type != AlterConfigOpType.DELETE:
+                        value = op_type_value[1]
+                        self.add_incremental_config(name, op_type, value)
+                    else:
+                        self.add_incremental_config(name, op_type)
```
pranavrth, line 174: nit: extra line break

src/confluent_kafka/src/Admin.c
```diff
@@ -329,6 +329,98 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
         return 1;
 }
 
+
+static int
+Admin_incremental_config_to_c(PyObject *dict, void *c_obj){
```
pranavrth, line 334: why are we not using `rd_kafka_ConfigResource_t *c_obj` here?

tests/test_Admin.py
```diff
@@ -320,6 +320,41 @@ def test_alter_configs_api():
             f.result(timeout=1)
 
 
+def test_incremental_alter_configs_api():
+    a = AdminClient({"socket.timeout.ms": 10})
+
+    with pytest.raises(TypeError):
+        a.incremental_alter_configs(None)
+
+    with pytest.raises(ValueError):
```
pranavrth, line 329: This should be `TypeError`

tests/test_Admin.py
```diff
@@ -320,6 +320,41 @@ def test_alter_configs_api():
...
+        a.incremental_alter_configs("something")
+
+    with pytest.raises(ValueError):
+        a.incremental_alter_configs([])
+
+    resources = [ConfigResource(ResourceType.BROKER, "3"),
```
pranavrth, line 335: Let's test constructor as well.

examples/adminapi.py
```diff
@@ -260,6 +260,35 @@ def example_delete_acls(a, args):
...
+    configuration properties with their previous values.
+
+    Input Format : ResourceType1 ResourceName1 Key=Operation:Value;Key2=Operation2:Value2
+    ResourceType2 ResourceName2 ...
+
+    Example: TOPIC T1 compression.type=SET:lz4;cleanup.policy=ADD:compact TOPIC T2 compression.type=SET:gzip ...
+    """
+    resources = []
+    for restype, resname, configs in zip(args[0::3], args[1::3], args[2::3]):
+        resource = ConfigResource(restype, resname)
```
pranavrth, line 274: Let's create this using constructor to help with the example on how to use the constructor as well?

pranavrth (changes requested):
Great work. Almost there! Just few nits.

src/confluent_kafka/admin/_config.py
emasab: It's better to specify to distinguish from the config resource

tests/test_Admin.py
emasab: string is interpreted as iterable and _make_futures raises
`raise ValueError("Expected list of {}".format(repr(class_check)))`
changing it would be breaking for the users.
I could pass the error to raise as parameter for the legacy functions by changing `_make_futures` for newer methods

examples/adminapi.py
emasab: In this case given that it's not a static list but it's created dynamically I think using the constructor make the example more difficult to read. I add a comment about using the constructor

src/confluent_kafka/admin/_config.py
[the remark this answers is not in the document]
emasab: Doesn't seem more clear to me. I've added an example in the example file.

src/confluent_kafka/admin/_config.py
pranavrth: We are not using ConfigEntry anywhere in Incremental alter configs implementation. Wouldn't it confuse users from docs perspective? They are not aware of ConfigEntry in the first place.

tests/test_Admin.py
pranavrth: Shall we add a new `_make_futures` for this use case where we raise the correct error and add a FIXME to fix Error in major update in old APIs?

examples/adminapi.py
```diff
@@ -260,6 +260,49 @@ def example_delete_acls(a, args):
...
+        for name, operation_and_value in [conf.split('=') for conf in configs.split(';')]:
+            operation, value = operation_and_value.split(':')
+            operation = AlterConfigOpType[operation]
+            resource.add_incremental_config(name, operation, value)
+        resources.append(resource)
+
+    #
+    # When config entries are fixed it's more readable to use the
```
pranavrth, line 282: Fixed where?
We are not using config entries so its confusing I think.

src/confluent_kafka/admin/_config.py
pranavrth: Now that we are allowing multiple values for a config, does it make sense to take input as the above mentioned way? or in for of `list(ConfigEntry)` like we do in librdkafka? Just a thought.

pranavrth (changes requested):

src/confluent_kafka/admin/_config.py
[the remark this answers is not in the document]
pranavrth: Let's add an example here in the doc as well.

src/confluent_kafka/admin/_config.py
emasab: I could use `configuration property`, it seems more used here:
```python
        self.name = name
        """Configuration property name."""
        self.value = value
        """Configuration value (or None if not set or is_sensitive==True)."""
        self.source = source
        """Configuration source."""
        self.is_read_only = bool(is_read_only)
        """Indicates whether the configuration property is read-only."""
        self.is_default = bool(is_default)
        """Indicates whether the configuration property is using its default value."""
```

src/confluent_kafka/src/Admin.c
```diff
@@ -331,92 +331,87 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
-                if (!PyList_Check(vo) || (config_count = (int)PyList_Size(vo)) < 1) {
-                        PyErr_SetString(PyExc_ValueError,
-                                        "expected non-empty list of entry values");
+                r = PyObject_IsInstance(config_entry, ConfigEntry_type);
+                if (r == -1)
+                        goto err; /* Exception raised by IsInstance() */
+                else if (r == 0) {
+                        PyErr_SetString(PyExc_TypeError,
+                                        "Expected list of "
+                                        "ConfigEntry objects");
```
pranavrth, line 369:
```suggestion
                                        "Expected elements of the field 'incremental_configs' to be of type ConfigEntry.");
```

src/confluent_kafka/src/Admin.c
```diff
@@ -331,92 +331,87 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
-                        const char *v = NULL;
-
-                        if (!PyList_Check(config_op_value) ||
-                            (config_value_cnt = (int)PyList_Size(config_op_value)) < 2) {
-                                PyErr_SetString(PyExc_ValueError,
-                                                "expected operation type and value");
-                                goto err;
-                        }
+                if (!cfl_PyObject_GetAttr(config_entry, "incremental_operation",
+                                          &incremental_operation, NULL, 1, 0))
```
pranavrth, line 374: I guess `incremental_operation` is Enum and `op` is the int value for it. Can we name them to suggest that?

src/confluent_kafka/src/Admin.c
```diff
@@ -331,92 +331,87 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
-                        if (error) {
-                                PyErr_Format(PyExc_ValueError,
-                                        "%s config %s failed: %s",
-                                        op, k, rd_kafka_error_string(error));
-                                rd_kafka_error_destroy(error);
-                                goto err;
-                        }
+                if (op != RD_KAFKA_ALTER_CONFIG_OP_TYPE_DELETE &&
+                    !cfl_PyObject_GetString(config_entry, "value", &value, NULL, 1, 0))
+                        goto err;
```
pranavrth, line 386: We should let librdkafka handle it for consistent behaviour across all the clients? If not, then we should set proper error here.

src/confluent_kafka/src/Admin.c
```diff
@@ -331,92 +331,87 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
-                        rd_kafka_AlterConfigOpType_t op;
-                        const char *v = NULL;
-
-                        if (!PyList_Check(config_op_value) ||
-                            (config_value_cnt = (int)PyList_Size(config_op_value)) < 2) {
-                                PyErr_SetString(PyExc_ValueError,
-                                                "expected operation type and value");
-                                goto err;
-                        }
+                if (!cfl_PyObject_GetAttr(config_entry, "incremental_operation",
```
pranavrth, line 373: We should decrease ref count for `incremental_operation` as new reference is returned by `cfl_PyObject_GetAttr` function.

src/confluent_kafka/src/Admin.c
```diff
@@ -1053,6 +1048,14 @@ static PyObject *Admin_incremental_alter_configs(Handle *self,PyObject *args,PyO
...
         }
 
+        ConfigEntry_type = cfl_PyObject_lookup("confluent_kafka.admin",
+                                                  "ConfigEntry");
+        if (!ConfigEntry_type) {
+                Py_DECREF(ConfigResource_type);
+                rd_kafka_AdminOptions_destroy(c_options);
+                return NULL; /* Exception raised by find() */
+        }
+
```
pranavrth, line 1058: Scope of `ConfigEntry` is in the function `Admin_incremental_config_to_c`. Why are we getting the type here?

examples/adminapi.py
```diff
@@ -260,6 +261,37 @@ def example_delete_acls(a, args):
...
+    Input Format : ResourceType1 ResourceName1 Key=Operation:Value;Key2=Operation2:Value2
+    ResourceType2 ResourceName2 ...
+
+    Example: TOPIC T1 compression.type=SET:lz4;cleanup.policy=ADD:compact TOPIC T2 compression.type=SET:gzip ...
+    """
+    resources = []
+    for restype, resname, configs in zip(args[0::3], args[1::3], args[2::3]):
+        incremental_configs = []
+        for name, operation_and_value in [conf.split('=') for conf in configs.split(';')]:
+            operation, value = operation_and_value.split(':')
```
pranavrth, line 277: What is the expectation from the input for the delete case? The code gives error in the following case :-

`TOPIC T1 compression.type=DELETE`

I think we should check if `:` is there or not first. If not, then operation should be DELETE. If yes, then we can allow any operation (In case of DELETE, it would ignore value part)

pranavrth (changes requested):

src/confluent_kafka/src/Admin.c
```diff
@@ -331,92 +331,87 @@ static int Admin_set_replica_assignment (const char *forApi, void *c_obj,
...
+                PyErr_Format(PyExc_TypeError,
+                             "expected list type");
+                goto err;
+        }
 
-        while(PyDict_Next(dict, &pos, &ko, &vo)){
-                const char *k;
-                int config_count = 0;
-                Py_ssize_t i = 0;
+        if ((config_entry_count = (int)PyList_Size(incremental_configs)) < 1) {
```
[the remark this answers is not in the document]
emasab: That I know it's more correct if count is associated with singular than with plural

src/confluent_kafka/src/Admin.c
emasab: the error is set by the cfl_PyObject_GetString function there

src/confluent_kafka/src/Admin.c
emasab: to avoid repeating the lookup N times

examples/adminapi.py
emasab: the idea was that you passed an empty string after the ":", but I can do this change

src/confluent_kafka/src/Admin.c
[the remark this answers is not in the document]
pranavrth: I would say "count of config entries" makes more sense. But its fine in both case.

src/confluent_kafka/src/Admin.c
pranavrth: Shall we have a specific error message which suggest that value is necessary for non Delete operations? Also, we can also check empty string here if it makes sense to do that. Your call on both of these.

src/confluent_kafka/src/Admin.c
pranavrth: Makes sense. Did realize this to be in a for loop.

src/confluent_kafka/src/Admin.c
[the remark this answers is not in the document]
emasab: In this case it's better plural, but when count is suffix it's better singular

src/confluent_kafka/admin/_config.py
```diff
@@ -133,6 +164,8 @@ def __init__(self, restype, name,
         else:
             self.set_config_dict = dict()
 
+        self.incremental_configs = list(incremental_configs or [])
```
pranavrth, line 167: I don't know how correct this is. For example, in string case if `incremental_configs = "abcd"` then `self.incremental_configs = ['a', 'b', 'c', 'd']`. This is anyway incorrect input itself but the error message that the user will get will not be correct.

I would say, change it to `self.incremental_configs = incremental_configs or []`

pranavrth (changes requested):
I think this last change and we are good to go.

src/confluent_kafka/src/Admin.c
emasab: It's ok if you pass an empty string, the value is ignored for delete operations. If I overwrite the error there it could hide the underlying error, that could be missing attribute, value is None or not a unicode string type.

src/confluent_kafka/admin/_config.py
emasab: That won't copy the original list. In this case given the string is iterable it's ok following duck typing to treat it as a list an make it fail because of the type.

pranavrth: It will fail but the error won't be correct. Changing it to `self.incremental_configs = incremental_configs or []` is incorrect as well.

pranavrth: Anyways. Its a corner case which can be solved by reading the doc. So, its fine to keep it this way.

emasab: I don't think the error is incorrect, you have passed a str that is an iterable of str. The error says that the first element 'a' of the collection is not a ConfigEntry

PR #1522, opened 2023-02-13
Automatic style fixes

pranavrth:
Fixing styles with automatic style fix make command.

As this is the new project, we should use latest version of clang-format. Currently, using v15. We should keep clang-format version up-to date as much as possible.

### DIFF
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -52,55 +52,56 @@ typedef struct {
         PyException_HEAD
 #else
         PyObject_HEAD
-        /* Standard fields of PyBaseExceptionObject which we inherit from. */
-        PyObject *dict;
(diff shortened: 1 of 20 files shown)


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -807,342 +776,331 @@ PyObject *Message_new0 (const Handle *handle, const rd_kafka_message_t *rkm) {
...
+        int partition        = RD_KAFKA_PARTITION_UA;
+        long long offset     = RD_KAFKA_OFFSET_INVALID;
+        const char *metadata = NULL;
 
-	static char *kws[] = { "topic",
-			       "partition",
-			       "offset",
-			       "metadata",
-			       NULL };
+        static char *kws[] = {"topic", "partition", "offset", "metadata", NULL};
```
pranavrth, line 831: Fix braced list initialization

PR #1539, opened 2023-03-27
Fix typo consumer group state unknown

pranavrth:


milindl:
Shouldn't we keep the old one too? Otherwise it breaks the API with the next release.

pranavrth:
> Shouldn't we keep the old one too? Otherwise it breaks the API with the next release.

Yes, but we are planning to upgrade from 2.0 to 2.1. So, I think it would be fine. 
Also, this feature got released very recently and not in the consumer or producer path.

pranavrth:
> Shouldn't we keep the old one too? Otherwise it breaks the API with the next release.

Yes, but we are planning to upgrade from 2.0 to 2.1. So, I think it would be fine. 
Also, this feature got released very recently and not in the consumer or producer path.

pranavrth:
@emasab @milindl - We can't keep both the enum values as both will correspond to the same value i.e. 0 from librdkafka. When converting the response from C to python, there are two options to choose from and it can give only one response either UNKOWN or UNKNOWN not both.

We need to choose from changing the value right now or during the next major release. We don't have option to keep both the values.

If we are choosing to change the value in major release, we need to tag this change to the major release so that we don't forget. I have other changes as well in my mind to change in major release. Let's figure out how to tag them preferably in the code itself.

emasab (approved):

PR #1555, opened 2023-04-18
Add flag for subject version permanent delete

rsolovev:
The SchemaRegistryClient `delete_version()` function does not have `?permanent=true` functionality mentioned in Schema Registry API reference - https://docs.confluent.io/platform/current/schema-registry/develop/api.html#delete--subjects-(string-%20subject)-versions-(versionId-%20version)

https://github.com/confluentinc/confluent-kafka-python/blob/370d62e2029d80f12fdf04dca9afbbd664962625/src/confluent_kafka/schema_registry/schema_registry_client.py#L563-L585

Added flag support similar to `delete_subject()` implementation and taking into the account this documentation lines:
> The default is false. If the flag is not included, a soft delete is performed. You must perform a soft delete first, then the hard delete.


rayokota:
Subsumed by https://github.com/confluentinc/confluent-kafka-python/pull/1867

PR #1575, opened 2023-05-30
[KIP-554] User SCRAM credentials API

mahajanadhitya:


examples/adminapi.py
```diff
@@ -560,6 +562,77 @@ def example_alter_consumer_group_offsets(a, args):
...
+    mechanism_description = {}
+    mechanism_description[ScramMechanism.SCRAM_SHA_256] = "SCRAM-SHA-256"
+    mechanism_description[ScramMechanism.SCRAM_SHA_512] = "SCRAM-SHA-512"
+    mechanism_description[ScramMechanism.UNKNOWN] = "UNKNOWN"
+    for username, fut in futmap.items():
+        print("Username: {}".format(username))
+        try:
+            value = fut.result()
+            for scram_credential_info in value.scram_credential_infos:
+                print("    Mechanism: {} Iterations: {}".
```
pranavrth, line 579: better use `f"    Mechanism: {scram_credential_info.mechanism} Iterations: {scram_credential_info.iterations}"`

examples/adminapi.py
```diff
@@ -560,6 +562,77 @@ def example_alter_consumer_group_offsets(a, args):
...
+                      format(scram_credential_info.mechanism,
+                             scram_credential_info.iterations))
+        except KafkaException as e:
+            print("    Error: {}".format(e))
```
pranavrth, line 583: ditto

examples/adminapi.py
```diff
@@ -560,6 +562,77 @@ def example_alter_consumer_group_offsets(a, args):
...
+
+
+def example_alter_user_scram_credentials(a, args):
+    """
+    AlterUserScramCredentials
+    """
+    alterations = []
+    i = 0
+    op_cnt = 0
+    while i < len(args):
```
pranavrth, line 593: separate out args validation and business logic.

examples/adminapi.py
```diff
@@ -560,6 +562,77 @@ def example_alter_consumer_group_offsets(a, args):
...
+        op = args[i]
+
+        if op == "UPSERT":
+            if i + 5 >= len(args):
+                raise ValueError(
+                    f"Invalid number of arguments for alteration {op_cnt}, expected 5, got {len(args) - i - 1}")
+            user = args[i + 1]
+            mechanism = ScramMechanism[args[i + 2]]
+            iterations = int(args[i + 3])
+            salt = args[i + 4]
```
pranavrth, line 603: How this is going to work? If there is a `i+4` parameter which is a password, wouldn't it consider password as salt? and then there will be no password. Salt as last optional parameter makes more sense.

examples/adminapi.py
```diff
@@ -560,6 +562,77 @@ def example_alter_consumer_group_offsets(a, args):
...
+    AlterUserScramCredentials
+    """
+    alterations = []
+    i = 0
+    op_cnt = 0
+    while i < len(args):
+        op = args[i]
+
+        if op == "UPSERT":
+            if i + 5 >= len(args):
```
pranavrth, line 597: This condition is wrong as salt is optional.

pranavrth (changes requested):
Went through _scram.py file, examples folder and docs folder only. Going through other files.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
         :raises TypeException: Invalid input.
         """
         super(AdminClient, self).set_sasl_credentials(username, password)
+
+    @staticmethod
+    def _make_describe_user_scram_credentials_result(f, futmap):
```
pranavrth, line 822: Move this method with other staticmethods in the top.

pranavrth, line 822: All `_make_..._result` methods seems similar. Shall we make a generic method for all of these?`

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+        except Exception as e:
+            for _, fut in futmap.items():
+                fut.set_exception(e)
+
+    def describe_user_scram_credentials(self, users, **kwargs):
+        """
+        Describe user SASL/SCRAM credentials.
+        Duplicate users aren't allowed.
+
+        :param list(str) users: List of user names to describe.
```
pranavrth, line 849:
```suggestion
        :param list(str) users: List of user names to describe. Duplicate users are not allowed.
```

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                else:
+                    fut.set_result(value)
+        except Exception as e:
+            for _, fut in futmap.items():
+                fut.set_exception(e)
+
+    def describe_user_scram_credentials(self, users, **kwargs):
+        """
+        Describe user SASL/SCRAM credentials.
+        Duplicate users aren't allowed.
```
pranavrth, line 847: Remove it from here and add it to `users param doc`.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+
+        :param list(str) users: List of user names to describe.
+        :param float request_timeout: The overall request timeout in seconds,
+               including broker lookup, request transmission, operation time
+               on broker, and response. Default: `socket.timeout.ms*1000.0`
+
+        :returns: A dict of futures keyed by user name.
+                  The future result() method returns None or
+                  raises KafkaException
```
pranavrth, line 856: This is incorrect.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+
+        f, futmap = AdminClient._make_futures(users, None,
+                                              AdminClient._make_describe_user_scram_credentials_result)
+
+        super(AdminClient, self).describe_user_scram_credentials(users, f, **kwargs)
+
+        return futmap
+
+    @staticmethod
+    def _make_alter_user_scram_credentials_result(f, futmap):
```
pranavrth, line 877: ditto.

This is exactly same function as `_make_describe_user_scram_credentials_result`

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
         :raises TypeException: Invalid input.
         """
         super(AdminClient, self).set_sasl_credentials(username, password)
+
+    @staticmethod
+    def _make_describe_user_scram_credentials_result(f, futmap):
+        try:
+            results = f.result()
+            futmap_values = list(futmap.values())
+            results_values = list(results.values())
```
pranavrth, line 826: Do we really need these 2 variables?

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+               on broker, and response. Default: `socket.timeout.ms*1000.0`
+
+        :returns: A dict of futures keyed by user name.
+                  The future result() method returns None or
+                  raises KafkaException
+
+        :rtype: dict[str, future]
+
+        :raises TypeError: Invalid type.
+        :raises ValueError: Invalid value.
```
pranavrth, line 861: Also KafkaException.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+               including broker lookup, request transmission, operation time
+               on broker, and response. Default: `socket.timeout.ms*1000.0`
+
+        :returns: A dict of futures keyed by user name.
+                  The future result() method returns None or
+                  raises KafkaException
+
+        :rtype: dict[str, future]
+
+        :raises TypeError: Invalid type.
```
pranavrth, line 860:
```suggestion
        :raises TypeError: Invalid Input Type.
```

src/confluent_kafka/admin/__init__.py
pranavrth, line 861:
```suggestion
        :raises ValueError: Invalid Input value.
```

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+        :raises ValueError: Invalid value.
+        """
+        if not isinstance(users, list):
+            raise TypeError("Expected input to be list of String")
+        for user in users:
+            if not isinstance(user, string_type):
```
pranavrth, line 866: Add empty string check?

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                raise TypeError("Each value should be a string")
+
+        f, futmap = AdminClient._make_futures(users, None,
```
pranavrth, line 869: Use `_make_futures_v2` being added in KIP-339 PR.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+        :returns: A dict of futures keyed by user name.
+                  The future result() method returns None or
+                  raises KafkaException
+
+        :rtype: dict[str, future]
+
+        :raises TypeError: Invalid type.
+        :raises ValueError: Invalid value.
+        """
+        if not isinstance(users, list):
```
pranavrth, line 863: Move parameter checks to a separate local method.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+        :returns: A dict of futures keyed by user name.
+                  The future result() method returns None or
+                  raises KafkaException
+
+        :rtype: dict[str, future]
+
+        :raises TypeError: Invalid type.
+        :raises ValueError: Invalid value.
+        """
+        if not isinstance(alterations, list):
```
pranavrth, line 919: Move parameter checking to a separate local function.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+               on broker, and response. Default: `socket.timeout.ms*1000.0`
+
+        :returns: A dict of futures keyed by user name.
+                  The future result() method returns None or
+                  raises KafkaException
+
+        :rtype: dict[str, future]
+
+        :raises TypeError: Invalid type.
+        :raises ValueError: Invalid value.
```
pranavrth, line 917: Ditto.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                else:
+                    fut.set_result(value)
+        except Exception as e:
+            for _, fut in futmap.items():
+                fut.set_exception(e)
+
+    def alter_user_scram_credentials(self, alterations, **kwargs):
+        """
+        Alter user SASL/SCRAM credentials.
+        The pair (user, mechanism) must be unique among alterations.
```
pranavrth, line 902: Move this statement to the correct param description below.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+        :raises TypeError: Invalid type.
+        :raises ValueError: Invalid value.
+        """
+        if not isinstance(alterations, list):
+            raise TypeError("Expected input to be list")
+        if len(alterations) == 0:
+            raise ValueError("Expected at least one alteration")
+        for alteration in alterations:
+            if not isinstance(alteration, UserScramCredentialAlteration):
+                raise TypeError("Expected each element of list to be subclass of UserScramCredentialAlteration")
```
pranavrth, line 925:
```suggestion
                raise TypeError("Expected each element of list 'alterations' to be subclass of UserScramCredentialAlteration")
```

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                raise TypeError("'user' cannot be None")
+            if not isinstance(alteration.user, string_type):
+                raise TypeError("'user' must be a string")
+            if not alteration.user:
+                raise ValueError("'user' cannot be empty")
+
+            if isinstance(alteration, UserScramCredentialUpsertion):
+                if alteration.salt is not None and not alteration.salt:
+                    raise ValueError("'salt' can be None but cannot be empty")
+                if alteration.salt and not isinstance(alteration.salt, bytes):
```
pranavrth, line 936: I think user should send the salt and password as string. We can internally convert it to bytes.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                    raise TypeError("'salt' must be bytes")
+                if alteration.password is None:
+                    raise TypeError("'password' cannot be None")
+                if not isinstance(alteration.password, bytes):
+                    raise TypeError("'password' must be bytes")
```
pranavrth, line 941: ditto

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                if not alteration.password:
+                    raise ValueError("'password' cannot be empty")
+
+                if not isinstance(alteration.scram_credential_info, ScramCredentialInfo):
+                    raise TypeError("Expected credential_info to be ScramCredentialInfo Type")
+                if alteration.scram_credential_info.iterations < 1:
```
pranavrth, line 947: I think iterations can never be non positive. In this case, we can put this check in `ScramCredentialInfo` constructor.

Instead, all the class level checks can be put in the class constructor itself. Applicable to other classes as well.

pranavrth (changes requested):

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        PyObject *result = NULL;
+        size_t cnt, i;
+        const rd_kafka_AlterUserScramCredentials_result_response_t **responses =
+                rd_kafka_AlterUserScramCredentials_result_responses(result_event, &cnt);
+        result = PyDict_New();
+        for(i=0; i<cnt; i++){
+                const rd_kafka_AlterUserScramCredentials_result_response_t *response = responses[i];
+                const rd_kafka_error_t *error = rd_kafka_AlterUserScramCredentials_result_response_error(response);
+                const char *username = rd_kafka_AlterUserScramCredentials_result_response_user(response);
+                PyDict_SetItemString(result, username, KafkaError_new_or_None(rd_kafka_error_code(error),rd_kafka_error_string(error)));
```
pranavrth, line 3277: Since (user, mechanism) is unique in the request, only the last one is kept here in the response. This is incorrect.

src/confluent_kafka/admin/__init__.py
```diff
@@ -809,3 +817,146 @@ def set_sasl_credentials(self, username, password):
...
+                    raise ValueError("Iterations should be positive")
+                if not isinstance(alteration.scram_credential_info.mechanism, ScramMechanism):
+                    raise TypeError("Expected the mechanism to be ScramMechanism Type")
+            elif isinstance(alteration, UserScramCredentialDeletion):
+                if not isinstance(alteration.mechanism, ScramMechanism):
+                    raise TypeError("Expected the mechanism to be ScramMechanism Type")
+            else:
+                raise TypeError("Expected each element of list to be subclass of UserScramCredentialAlteration")
+
+        f, futmap = AdminClient._make_futures(set([alteration.user for alteration in alterations]), None,
```
pranavrth, line 957: Since the pair (user, mechanism) is unique in the request. Using set of users is incorrect here. Only one response would be allowed per user in this case. I see (in Admin.c) that only last response is retained.

pranavrth (changes requested):
Ongoing C files review.

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,361 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+                                goto err;
+                        }
+
+                        cfl_PyObject_GetInt(scram_credential_info,"iterations",&iterations,0,1);
+                        cfl_PyObject_GetAttr(scram_credential_info,"mechanism",&mechanism,NULL,0,0);
+                        if (!PyObject_IsInstance(mechanism, ScramMechanism_type)) {
+                                PyErr_SetString(PyExc_TypeError,
+                                        "mechanism should be of ScramMechanism type");
+                                goto err;
+                        }
```
pranavrth, line 1794: We can move this out as this is mandatory in both kind of alteration.

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,361 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+                        cfl_PyObject_GetInt(mechanism,"value",&mechanism_val,0,1);
+
+                        c_alterations[i] = rd_kafka_UserScramCredentialUpsertion_new(user_c,salt_c, salt_c_size, password_c,
+                                password_c_size, (rd_kafka_ScramMechanism_t) mechanism_val, iterations);
+
+                        Py_DECREF(salt);
+                        Py_DECREF(password);
+                        Py_DECREF(scram_credential_info);
```
pranavrth, line 1802:
```suggestion
                        Py_DECREF(scram_credential_info);
                        Py_DECREF(mechanism);
                        mechanism = NULL;
```

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,361 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+        Py_XDECREF(salt);
+        Py_XDECREF(password);
+        Py_XDECREF(scram_credential_info);
+        Py_XDECREF(mechanism);
+
+        Py_DECREF(UserScramCredentialAlteration_type); /* from lookup() */
+        Py_DECREF(UserScramCredentialUpsertion_type); /* from lookup() */
+        Py_DECREF(UserScramCredentialDeletion_type); /* from lookup() */
+        Py_DECREF(ScramCredentialInfo_type); /* from lookup() */
+        Py_DECREF(ScramMechanism_type); /* from lookup() */
```
pranavrth, line 1865:
```suggestion
        Py_XDECREF(UserScramCredentialAlteration_type); /* from lookup() */
        Py_XDECREF(UserScramCredentialUpsertion_type); /* from lookup() */
        Py_XDECREF(UserScramCredentialDeletion_type); /* from lookup() */
        Py_XDECREF(ScramCredentialInfo_type); /* from lookup() */
        Py_XDECREF(ScramMechanism_type); /* from lookup() */
```

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,361 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+
+        if(c_alterations) {
+                for(i=0;i<alteration_cnt;i++)
+                        rd_kafka_UserScramCredentialAlteration_destroy(c_alterations[i]);
```
pranavrth, line 1869: 1. Should be till `i` not till `alteration_cnt` as we can reach here in error case. In this case
2. Use this function instead - `rd_kafka_UserScramCredentialAlteration_destroy_array`

```suggestion
                rd_kafka_UserScramCredentialAlteration_destroy_array(c_alterations, i);
```

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,361 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+         */
+        CallState_begin(self, &cs);
+        rd_kafka_AlterUserScramCredentials(self->rk, c_alterations, alteration_cnt, c_options, rkqu);
+        CallState_end(self, &cs);
+
+        if(c_alterations) {
+                for(i=0;i<alteration_cnt;i++)
+                        rd_kafka_UserScramCredentialAlteration_destroy(c_alterations[i]);
+                free(c_alterations);
+        }
```
pranavrth, line 1843:
```suggestion
        rd_kafka_UserScramCredentialAlteration_destroy_array(c_alterations, alteration_cnt);
        free(c_alterations);
```

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,361 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+        return NULL;
+}
+
+
+const char alter_user_scram_credentials_doc[] = PyDoc_STR(
+        ".. py:function:: alter_user_scram_credentials(alterations)\n"
+        "\n"
+        "  Alters the credentials for a user.\n"
+        "  Supported : Upsertion , Deletion.\n");
```
pranavrth, line 1884: 1. Add parameters future and request_timeout
2. Add note not to use this function directly.

pranavrth (changes requested):
Checking response

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
         return NULL;
 }
 
+static PyObject *Admin_c_ScramMechanism_to_py(rd_kafka_ScramMechanism_t mechanism){
+        PyObject *result = NULL, *value_dict = NULL;
+        PyObject *args = NULL;
+        value_dict = PyDict_New();
+        cfl_PyDict_SetInt(value_dict, "value",(int) mechanism);
+        args = PyTuple_New(0);
+        result = PyObject_Call(cfl_PyObject_lookup("confluent_kafka.admin",
```
pranavrth, line 3204: I think we need to decrease refcount for the new reference returned by `cfl_PyObject_lookup` function.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+                                                   "ScramMechanism"), args, value_dict);
+        Py_DECREF(value_dict);
+        return result;
+}
+
+static PyObject *Admin_c_ScramCredentialInfo_to_py(const rd_kafka_ScramCredentialInfo_t *scram_credential_info){
+        PyObject *result = NULL;
+        PyObject *args = NULL;
+        result = PyDict_New();
+        PyDict_SetItemString(result,"mechanism",Admin_c_ScramMechanism_to_py(rd_kafka_ScramCredentialInfo_mechanism(scram_credential_info)));
```
pranavrth, line 3214: Calling multiple nested functions decrease readability. Create local variables and use them.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        cfl_PyDict_SetInt(result,"iterations", rd_kafka_ScramCredentialInfo_iterations(scram_credential_info));
+        args = PyTuple_New(0);
+        result = PyObject_Call(cfl_PyObject_lookup("confluent_kafka.admin",
```
pranavrth, line 3217: ditto. Do everywhere below vas well.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+                                                   "ScramCredentialInfo"), args, result);
+        return result;
+}
+
+static PyObject *Admin_c_UserScramCredentialsDescription_to_py(const rd_kafka_UserScramCredentialsDescription_t *description){
+        PyObject *result = NULL;
+        PyObject *args = NULL;
+        PyObject *scram_credential_infos = NULL;
+        int scramcredentialinfo_cnt;
```
pranavrth, line 3226: `scram_credential_info_cnt`

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        int i;
+        result = PyDict_New();
```
pranavrth, line 3228: We need to decrease refcount for this dict as well.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3195,89 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        cfl_PyDict_SetString(result, "user", rd_kafka_UserScramCredentialsDescription_user(description));
+
+        scramcredentialinfo_cnt = rd_kafka_UserScramCredentialsDescription_scramcredentialinfo_count(description);
+        scram_credential_infos = PyList_New(scramcredentialinfo_cnt);
+        for(i=0;i<scramcredentialinfo_cnt;i++){
+                PyList_SET_ITEM(scram_credential_infos,i,Admin_c_ScramCredentialInfo_to_py(rd_kafka_UserScramCredentialsDescription_scramcredentialinfo(description,i)));
+        }
+
+        PyDict_SetItemString(result,"scram_credential_infos",scram_credential_infos);
+        args = PyTuple_New(0);
```
pranavrth, line 3238: We need to decrease refcount for this args as well.

pranavrth (changes requested):
1. Check all the local variables in the Response parsing for Ref Count.
2. Don't nest multiple functions. create local variables for readability.

tests/test_Admin.py
```diff
@@ -697,3 +698,34 @@ def test_alter_consumer_group_offsets_api():
...
+        e = f.exception(timeout=1)
+        assert isinstance(e, KafkaException)
+        assert e.args[0].code() == KafkaError._TIMED_OUT
+
+    with pytest.raises(ValueError):
+        a.alter_user_scram_credentials([])
+    with pytest.raises(TypeError):
+        a.alter_user_scram_credentials([None])
+    with pytest.raises(TypeError):
+        a.alter_user_scram_credentials([UserScramCredentialAlteration("sam")])
```
pranavrth, line 731: There are alot more conditions can be tested as per error handling. If possible, add them.

tests/integration/admin/test_basic_operations.py
```diff
@@ -192,6 +196,55 @@ def verify_consumer_group_offsets_operations(client, our_topic, group_id):
...
+    futmap = admin_client.describe_user_scram_credentials([newuser])
+    assert isinstance(futmap, dict)
+    assert len(futmap) == 1
+    assert newuser in futmap
+    fut = futmap[newuser]
+    with pytest.raises(KafkaException) as ex:
+        result = fut.result()
+    assert ex.value.args[0] == KafkaError.RESOURCE_NOT_FOUND
+
+    futmap = admin_client.alter_user_scram_credentials([UserScramCredentialUpsertion(newuser,
```
pranavrth, line 213: Can we also test multiple Alterations in one request for a single user?

pranavrth (changes requested):

src/confluent_kafka/admin/_scram.py
```diff
@@ -0,0 +1,116 @@
...
+    Enumerates SASL/SCRAM mechanisms.
+    """
+    UNKNOWN = cimpl.SCRAM_MECHANISM_UNKNOWN  #: Unknown SASL/SCRAM mechanism
+    SCRAM_SHA_256 = cimpl.SCRAM_MECHANISM_SHA_256  #: SCRAM-SHA-256 mechanism
+    SCRAM_SHA_512 = cimpl.SCRAM_MECHANISM_SHA_512  #: SCRAM-SHA-512 mechanism
+
+    def __lt__(self, other):
+        if self.__class__ != other.__class__:
+            return NotImplemented
+        return self.value < other.value
```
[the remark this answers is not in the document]
emasab: To be able to sort a list of ScramMechanism, by id

src/confluent_kafka/src/Admin.c
emasab: Here it's inside ScramCredentialInfo while in deletion is directly in the alteration without iterations

examples/adminapi.py
emasab: salt isn't really optional in this command line example, it needs to be set to empty string to try the None value. Added a comment about that.

examples/adminapi.py
```diff
@@ -560,6 +562,77 @@ def example_alter_consumer_group_offsets(a, args):
...
 
+def example_describe_user_scram_credentials(a, args):
+    """
+    Describe User Scram Credentials
+    """
+    futmap = a.describe_user_scram_credentials(args)
+    mechanism_description = {}
+    mechanism_description[ScramMechanism.SCRAM_SHA_256] = "SCRAM-SHA-256"
+    mechanism_description[ScramMechanism.SCRAM_SHA_512] = "SCRAM-SHA-512"
+    mechanism_description[ScramMechanism.UNKNOWN] = "UNKNOWN"
```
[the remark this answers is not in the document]
emasab: A future for every described user, as they can have individual errors.

src/confluent_kafka/admin/__init__.py
emasab: KafkaException is raised only by the fut.result(), not by calling this method

src/confluent_kafka/admin/__init__.py
emasab: The response is one per user, all alterations about that user are executed or none

src/confluent_kafka/admin/__init__.py
emasab: Better use bytes as in Java and leave the resposability of chosing the encoding to the user.

src/confluent_kafka/admin/__init__.py
emasab: As done in other methods, classes can be data classes and validation can happen when they're used.

src/confluent_kafka/admin/_scram.py
```diff
@@ -0,0 +1,116 @@
...
+        The user name.
+    scram_credential_infos: list(ScramCredentialInfo)
+        SASL/SCRAM credential representations for the user.
+    """
+    def __init__(self, user: str, scram_credential_infos: list):
+        self.user = user
+        self.scram_credential_infos = scram_credential_infos
+
+
+class UserScramCredentialAlteration:
```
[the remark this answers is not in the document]
emasab: Subclassing ABC doesn't really make it non-instantiable if there isn't an abstract method.
It this case it should be more like a Kotlin sealed class, that cannot be subclassed outside the package,
but it's not possible in Python.

src/confluent_kafka/admin/__init__.py
emasab: For the moment I've unified just these two

CHANGELOG.md
```diff
@@ -1,10 +1,21 @@
 # Confluent's Python client for Apache Kafka
 
+## v2.2.0
+
+v2.2.0 is a feature release with the following features, fixes and enhancements:
+
+ - [KIP-554](https://cwiki.apache.org/confluence/display/KAFKA/KIP-554%3A+Add+Broker-side+SCRAM+Config+API):
+   User SASL/SCRAM credentials alteration and description (#1575).
```
pranavrth, line 8:
```suggestion
 - Added [KIP-554](https://cwiki.apache.org/confluence/display/KAFKA/KIP-554%3A+Add+Broker-side+SCRAM+Config+API):
   User SASL/SCRAM credentials alteration and description (#1575).
```

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,365 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+        ScramCredentialInfo_type = cfl_PyObject_lookup("confluent_kafka.admin",
+                                                  "ScramCredentialInfo");
+        if (!ScramCredentialInfo_type) {
+                        PyErr_SetString(PyExc_ImportError,
+                                "Not able to load ScramCredentialInfo type");
+                        goto err;
+        }
+
+        ScramMechanism_type = cfl_PyObject_lookup("confluent_kafka.admin",
+                                                  "ScramMechanism");\
```
pranavrth, line 1714: Remove `\`

```suggestion
                                                  "ScramMechanism");
```

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,365 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+                }
+
+                Py_DECREF(user);
+                user_c = cfl_PyUnistr_AsUTF8(u_user, &uo_user);
+
+                if(PyObject_IsInstance(alteration,UserScramCredentialUpsertion_type)){
+                        /* Upsertion Type*/
+                        cfl_PyObject_GetAttr(alteration,"scram_credential_info",&scram_credential_info,NULL,0,0);
+                        if (!PyObject_IsInstance(scram_credential_info, ScramCredentialInfo_type)) {
+                                PyErr_SetString(PyExc_TypeError,
```
pranavrth, line 1759: I think the idea to mention alteration index is good. We should add it to all the errors.

`Alteration %d: `

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,365 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+                                        "scram_credential_info should be of ScramCredentialInfo type");
+                                goto err;
+                        }
+
+                        cfl_PyObject_GetInt(scram_credential_info,"iterations",&iterations,0,1);
+                        cfl_PyObject_GetAttr(scram_credential_info,"mechanism", &mechanism, NULL, 0, 0);
+                        if (!PyObject_IsInstance(mechanism, ScramMechanism_type)) {
+                                PyErr_SetString(PyExc_TypeError,
```
pranavrth, line 1767: ditto and below as well.

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,365 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+         *
+         * We need to set up a CallState and release GIL here since
+         * the event_cb may be triggered immediately.
+         */
+        CallState_begin(self, &cs);
+        rd_kafka_AlterUserScramCredentials(self->rk, c_alterations, alteration_cnt, c_options, rkqu);
+        CallState_end(self, &cs);
+
+        if(c_alterations) {
+                rd_kafka_UserScramCredentialAlteration_destroy_array(c_alterations, i);
```
pranavrth, line 1844: I would use `alteration_cnt` here instead of `i` to make it future proof.

src/confluent_kafka/src/Admin.c
```diff
@@ -1531,6 +1528,365 @@ const char Admin_list_consumer_groups_doc[] = PyDoc_STR(
...
+        if (!ScramMechanism_type) {
+                        PyErr_SetString(PyExc_ImportError,
+                                "Not able to load ScramMechanism type");
+                        goto err;
+        }
+
+        c_options = Admin_options_to_c(self, RD_KAFKA_ADMIN_OP_ALTERUSERSCRAMCREDENTIALS,
+                                       &options, future);
+        if (!c_options)
+                return NULL; /* Exception raised by options_to_c() */
```
pranavrth, line 1724: I think we should use `goto err` here instead of `return NULL` as many objects are created before this.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3199,121 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        PyObject *args = NULL;
+        PyObject *ScramMechanism_type;
+        value_dict = PyDict_New();
+        cfl_PyDict_SetInt(value_dict, "value",(int) mechanism);
+        args = PyTuple_New(0);
+        ScramMechanism_type = cfl_PyObject_lookup("confluent_kafka.admin",
+                                                  "ScramMechanism");
+        result = PyObject_Call(ScramMechanism_type, args, value_dict);
+        Py_DECREF(value_dict);
+        Py_DECREF(ScramMechanism_type);
```
pranavrth, line 3213: Forgot to Py_DECREF for args.

```suggestion
        Py_DECREF(args);
        Py_DECREF(value_dict);
        Py_DECREF(ScramMechanism_type);
```

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3199,121 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        }
+
+        PyDict_SetItemString(kwargs,"scram_credential_infos",scram_credential_infos);
+        args = PyTuple_New(0);
+        UserScramCredentialsDescription_type = cfl_PyObject_lookup("confluent_kafka.admin",
+                                                                   "UserScramCredentialsDescription");
+        result = PyObject_Call(UserScramCredentialsDescription_type, args, kwargs);
+        Py_DECREF(args);
+        Py_DECREF(kwargs);
+        Py_DECREF(UserScramCredentialsDescription_type);
```
pranavrth, line 3269: Forgot to `Py_DECREF` for `scram_credential_infos `.

```suggestion
        Py_DECREF(args);
        Py_DECREF(kwargs);
        Py_DECREF(UserScramCredentialsDescription_type);
        Py_DECREF(scram_credential_infos );
```

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3199,121 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+                }
+        }
+        return result;
+}
+
+static PyObject *Admin_c_AlterUserScramCredentialsResult_to_py(const rd_kafka_AlterUserScramCredentials_result_t *result_event) {
+        PyObject *result = NULL;
+        size_t cnt, i;
+        const rd_kafka_AlterUserScramCredentials_result_response_t **responses =
+                rd_kafka_AlterUserScramCredentials_result_responses(result_event, &cnt);
```
pranavrth, line 3305: Let's move this to the below switch case to make it consistent with other functions. Also, I think its better to convert `c_responses` than to convert `result_event`.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3199,121 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+}
+
+static PyObject *Admin_c_DescribeUserScramCredentialsResult_to_py(const rd_kafka_DescribeUserScramCredentials_result_t *result_event) {
+        PyObject *result = NULL;
+        const rd_kafka_UserScramCredentialsDescription_t **descriptions;
+        size_t description_cnt;
+        size_t i;
+
+        result = PyDict_New();
+        descriptions = rd_kafka_DescribeUserScramCredentials_result_descriptions(result_event, &description_cnt);
```
pranavrth, line 3280: Let's move this to the below switch case to make it consistent with other functions. Also, I think its better to convert `descriptions` than to convert `result_event`.

src/confluent_kafka/src/Admin.c
```diff
@@ -2835,6 +3199,121 @@ static PyObject *Admin_c_DescribeConsumerGroupsResults_to_py(
...
+        }
+        return result;
+}
+
+static PyObject *Admin_c_AlterUserScramCredentialsResult_to_py(const rd_kafka_AlterUserScramCredentials_result_t *result_event) {
+        PyObject *result = NULL;
+        size_t cnt, i;
+        const rd_kafka_AlterUserScramCredentials_result_response_t **responses =
+                rd_kafka_AlterUserScramCredentials_result_responses(result_event, &cnt);
+        result = PyDict_New();
```
pranavrth, line 3306: Questions: Just trying to understand why are we using dict as a response here? Earlier we were using list for the response.

src/confluent_kafka/src/Admin.c
```diff
@@ -3138,7 +3617,20 @@ static void Admin_background_event_cb (rd_kafka_t *rk, rd_kafka_event_t *rkev,
 
                 break;
         }
-
+        case RD_KAFKA_EVENT_DESCRIBEUSERSCRAMCREDENTIALS_RESULT:
+        {
+                const rd_kafka_DescribeUserScramCredentials_result_t *result_event;
+                result_event = rd_kafka_event_DescribeUserScramCredentials_result(rkev);
+                result = Admin_c_DescribeUserScramCredentialsResult_to_py(result_event);
+                break;
```
pranavrth, line 3625: Can we have consistent names as we had earlier?

src/confluent_kafka/src/Admin.c
```diff
@@ -3138,7 +3617,20 @@ static void Admin_background_event_cb (rd_kafka_t *rk, rd_kafka_event_t *rkev,
...
+        }
+        case RD_KAFKA_EVENT_ALTERUSERSCRAMCREDENTIALS_RESULT:
+        {
+                const rd_kafka_AlterUserScramCredentials_result_t *result_event;
+                result_event = rd_kafka_event_AlterUserScramCredentials_result(rkev);
+                result = Admin_c_AlterUserScramCredentialsResult_to_py(result_event);
+                break;
```
pranavrth, line 3632: ditto.

src/confluent_kafka/src/Admin.c
pranavrth, line 1: Can we append `c_` for all the variables which are not Python object to maintain consistency?

pranavrth (changes requested):
Almost there. Some nits. Nothing major.

examples/adminapi.py
```diff
@@ -586,7 +671,11 @@ def example_alter_consumer_group_offsets(a, args):
         sys.stderr.write(
             ' alter_consumer_group_offsets <group> <topic1> <partition1> <offset1> ' +
             '<topic2> <partition2> <offset2> ..\n')
-
+        sys.stderr.write(' describe_user_scram_credentials [<user1> <user2> ..]\n')
+        sys.stderr.write(' alter_user_scram_credentials UPSERT <user1> <mechanism1> ' +
+                         '<iterations1> <password1> <salt1> ' +
```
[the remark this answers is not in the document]
emasab: It's difficult to give this documentation here. I've added a comment for the empty salt case in the code. In the end this is an example so they'll be reading the code.

CHANGELOG.md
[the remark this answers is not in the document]
emasab: When we've implemented KIP we've always put KIP number as first thing, see previous releases

src/confluent_kafka/admin/__init__.py
```diff
@@ -38,6 +38,13 @@
                      ConsumerGroupDescription,
                      MemberAssignment,
                      MemberDescription)
+from ._scram import (UserScramCredentialAlteration,  # noqa: F401
+                     UserScramCredentialUpsertion,
+                     UserScramCredentialDeletion,
+                     ScramCredentialInfo,
+                     ScramMechanism,
+                     UserScramCredentialsDescription)
+
```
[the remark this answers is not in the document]
emasab: It's exposed here. It's imported without renaming with _ . See the example where it's imported.

src/confluent_kafka/admin/__init__.py
```diff
@@ -342,6 +372,58 @@ def _check_alter_consumer_group_offsets_request(request):
...
+                if alteration.scram_credential_info.iterations < 1:
+                    raise ValueError("Iterations should be positive")
+                if not isinstance(alteration.scram_credential_info.mechanism, ScramMechanism):
+                    raise TypeError("Expected the mechanism to be ScramMechanism Type")
+            elif isinstance(alteration, UserScramCredentialDeletion):
+                if not isinstance(alteration.mechanism, ScramMechanism):
+                    raise TypeError("Expected the mechanism to be ScramMechanism Type")
+            else:
+                raise TypeError("Expected each element of list 'alterations' " +
+                                "to be subclass of UserScramCredentialAlteration")
```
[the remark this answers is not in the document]
emasab: Yeah, somone could be subclassing UserScramCredentialAlteration, but only these two subclasses are allowed. I think using the same TypeError is ok, NotImplementedError is used when an abstract method isn't implemented in a subclass.

src/confluent_kafka/admin/__init__.py
emasab: Will change it after merging KIP-339

src/confluent_kafka/src/Admin.c
emasab: The dict is a map from string to Error or None, then it's converted to a map from username to Future in Python code.

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
pranavrth: I think we should change the error message. This message is incorrect.

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
pranavrth: My question was if this is the right place to expose for all these classes. Should it be in `confluent_kafka` module or in `confluent_kafka.admin` module. I think admin makes more sense. So there is nothing to do here. Whats your thought?

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
emasab: ok

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
emasab: yeah, in Java too they are in `org.apache.kafka.clients.admin`

pranavrth (approved):
Great work Guys!. 

LGTM! :)

PR #1576, opened 2023-05-30
[KIP-396] List offsets implementation

mahajanadhitya:


emasab (changes requested):
Here are the comments, sent a PR where I've done most of the suggestions for testing them.

src/confluent_kafka/admin/__init__.py
```diff
@@ -1010,5 +1056,48 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
...
+            if not isinstance(kwargs['isolation_level'], _IsolationLevel):
+                raise TypeError("isolation_level argument should be an IsolationLevel")
+            kwargs['isolation_level'] = kwargs['isolation_level'].value
+
+        topic_partition_offsets_copy = []
+        for topic_partition, offset_spec in topic_partition_offsets.items():
+            topic_partition_offsets_copy.append(_TopicPartition(topic_partition.topic, int(topic_partition.partition),
+                                                int(offset_spec._value)))
+
+        topic_partition_keys = [_TopicPartition(topic_partition.topic, topic_partition.partition)
```
pranavrth, line 1096: Why are we not using `TopicPartition` from the input given by the user?

src/confluent_kafka/admin/_listoffsets.py
```diff
@@ -0,0 +1,149 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from .. import cimpl
+
+
+class OffsetSpec(ABC):
```
pranavrth, line 19: Need more understanding on this. We will discuss on a call.

pranavrth (changes requested):
First Round of Review. Checking test cases.

examples/adminapi.py
```diff
@@ -681,6 +682,57 @@ def example_alter_user_scram_credentials(a, args):
...
 
 
+def example_list_offsets(a, args):
+    topic_partition_offsets = {}
+    if len(args) < 1:
+        raise ValueError(
+            f"Invalid number of arguments for list offsets, expected at least 1, got {len(args)}")
+    i = 1
+    partition_i = 1
+    isolation_level = IsolationLevel[args[0]]
```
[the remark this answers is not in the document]
mahajanadhitya: Actually it is not but Emanuele has made so the request is more writable by the user. But I do feel like it is a overkill.

src/confluent_kafka/admin/__init__.py
```diff
@@ -1010,5 +1056,48 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
                                                  AdminClient._make_user_scram_credentials_result)
 
         super(AdminClient, self).alter_user_scram_credentials(alterations, f, **kwargs)
+        return futmap
+
+    def list_offsets(self, topic_partition_offsets, **kwargs):
+        """
+        Enables to find the beginning offset,
+        end offset as well as the offset matching a timestamp
+        or the offset with max timestamp in partitions.
```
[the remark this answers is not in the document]
mahajanadhitya: Please suggest something in things like this otherwise it remains just stuck I feel this is an apt. description.

src/confluent_kafka/admin/__init__.py
mahajanadhitya: Because it might have offset or other value set and we do not want that in our key so only topic and partition should be set and for the internal usecase the offset_spec is stored in the offset.

examples/adminapi.py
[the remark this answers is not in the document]
emasab: To try different values of that option. It's the first argument in other examples (LK) too

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
emasab: It's similar to the javadoc
https://kafka.apache.org/35/javadoc/org/apache/kafka/clients/admin/Admin.html#listOffsets(java.util.Map)

emasab (approved):
Thanks @mahajanadhitya @pranavrth . This looks good, let's wait for @pranavrth approval or for other comments.

examples/adminapi.py
```diff
@@ -681,6 +682,57 @@ def example_alter_user_scram_credentials(a, args):
             print("{}: Error: {}".format(username, e))
 
 
+def example_list_offsets(a, args):
+    topic_partition_offsets = {}
+    if len(args) < 1:
+        raise ValueError(
+            f"Invalid number of arguments for list offsets, expected at least 1, got {len(args)}")
```
philipnee, line 689: it can only be 0 here no? len(args) < 1.

src/confluent_kafka/admin/__init__.py
emasab: This has been changed because the rich compare method only checks the topic and partition

examples/adminapi.py
```diff
@@ -765,6 +767,57 @@ def example_alter_user_scram_credentials(a, args):
             print("{}: Error: {}".format(username, e))
 
 
+def example_list_offsets(a, args):
+    topic_partition_offsets = {}
+    if len(args) == 0:
```
pranavrth, line 772: There is no restriction like this in Java. Is empty map allowed there?

src/confluent_kafka/_model/__init__.py
```diff
@@ -132,3 +132,22 @@ def __init__(self, id, leader, replicas, isr):
...
+
+class IsolationLevel(Enum):
+    """
+    Enum for Kafka isolation levels.
+
+    Values
+    ------
+    READ_COMMITTED          : Skip offsets belonging to an aborted transaction.
+    READ_UNCOMMITTED        : Receive all the offsets.
+    """
```
pranavrth, line 145: Move them to field level. It doesn't render properly in docs.

src/confluent_kafka/admin/__init__.py
```diff
@@ -1103,5 +1158,43 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
...
+        :raises ValueError: Invalid input value.
+        """
+        AdminClient._check_list_offsets_request(topic_partition_offsets, kwargs)
+
+        kwargs['isolation_level'] = kwargs['isolation_level'].value
+
+        topic_partition_offsets_copy = []
+        for topic_partition, offset_spec in topic_partition_offsets.items():
+            topic_partition_offsets_copy.append(_TopicPartition(topic_partition.topic, int(topic_partition.partition),
+                                                int(offset_spec._value)))
```
pranavrth, line 1193:
```suggestion
        topic_partition_offsets_list = [_TopicPartition(topic_partition.topic, int(topic_partition.partition), int(offset_spec._value)) for topic_partition, offset_spec in topic_partition_offsets.items()]
```

src/confluent_kafka/admin/__init__.py
```diff
@@ -1103,5 +1158,43 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
...
+        :rtype: dict[TopicPartition, future]
+
+        :raises TypeError: Invalid input type.
+        :raises ValueError: Invalid input value.
+        """
+        AdminClient._check_list_offsets_request(topic_partition_offsets, kwargs)
+
+        kwargs['isolation_level'] = kwargs['isolation_level'].value
+
+        topic_partition_offsets_copy = []
```
pranavrth, line 1190: `topic_partition_offsets_list` `instead`.

src/confluent_kafka/admin/__init__.py
```diff
@@ -1103,5 +1158,43 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
...
+                  raises KafkaException
+
+        :rtype: dict[TopicPartition, future]
+
+        :raises TypeError: Invalid input type.
+        :raises ValueError: Invalid input value.
+        """
+        AdminClient._check_list_offsets_request(topic_partition_offsets, kwargs)
+
+        kwargs['isolation_level'] = kwargs['isolation_level'].value
```
pranavrth, line 1188: Generally, we should not change the original variable. I would suggest to change the variable to `isolation_level_value` and use it in the `Admin.c` file

src/confluent_kafka/admin/_listoffsets.py
```diff
@@ -0,0 +1,149 @@
...
+    ListOffsetsResultInfo
+    Result of a `AdminClient.list_offsets` call associated to a partition.
+
+    Parameters
+    ----------
+    offset: int
+        The offset returned by the list_offsets call.
+    timestamp: int
+        The timestamp in milliseconds corresponding to the offset.
+        Not available (-1) when querying for the earliest of latest offsets.
```
pranavrth, line 140:
```suggestion
        Not available (-1) when querying for the earliest or the latest offsets.
```

?

src/confluent_kafka/src/Admin.c
```diff
@@ -2869,6 +2878,85 @@ const char Admin_alter_consumer_group_offsets_doc[] = PyDoc_STR(
...
+        struct Admin_options options = Admin_options_INITIALIZER;
+        rd_kafka_AdminOptions_t *c_options = NULL;
+        rd_kafka_topic_partition_list_t *c_topic_partitions = NULL;
+        CallState cs;
+        rd_kafka_queue_t *rkqu;
+
+        static char *kws[] = {"topic_partitions",
+                             "future",
+                             /* options */
+                             "isolation_level",
```
pranavrth, line 2893:
```suggestion
                             "isolation_level_value",
```

src/confluent_kafka/src/Admin.c
```diff
@@ -4219,6 +4311,52 @@ Admin_c_GroupResults_to_py (const rd_kafka_group_result_t **c_result_responses,
         return all_groups_result;
 }
 
+/**
+ *
+ * @brief Convert C ListOffsetsResult response to pyobject.
+ *
+ */
+static PyObject *Admin_c_ListOffsetsResult_to_py (const rd_kafka_ListOffsets_result_t *c_result_event) {
```
pranavrth, line 4319: This function should take `rd_kafka_ListOffsetsResultInfo_t **` as input instead of `rd_kafka_ListOffsets_result_t *`. Conversion from `rd_kafka_ListOffsets_result_t` to `rd_kafka_ListOffsetsResultInfo_t **` should happen in the case block of `Admin_background_event_cb` as per convention that we are following.

src/confluent_kafka/src/Admin.c
```diff
@@ -4219,6 +4311,52 @@ Admin_c_GroupResults_to_py (const rd_kafka_group_result_t **c_result_responses,
...
+                } else {
+                        PyObject *args = NULL;
+                        PyObject *kwargs = NULL;
+                        kwargs = PyDict_New();
+                        cfl_PyDict_SetLong(kwargs,"offset", c_topic_partition->offset);
+                        cfl_PyDict_SetLong(kwargs,"timestamp", c_timestamp);
+                        cfl_PyDict_SetInt(kwargs,"leader_epoch",
+                                rd_kafka_topic_partition_get_leader_epoch(c_topic_partition));
+                        args = PyTuple_New(0);
+                        value = PyObject_Call(ListOffsetsResultInfo_type, args, kwargs);
```
pranavrth, line 4349: Please check in other places whether we should handle exception case here?

src/confluent_kafka/src/Admin.c
```diff
@@ -4552,6 +4690,13 @@ static void Admin_background_event_cb (rd_kafka_t *rk, rd_kafka_event_t *rkev,
                 break;
         }
 
+        case RD_KAFKA_EVENT_LISTOFFSETS_RESULT:
+        {
+                const rd_kafka_ListOffsets_result_t *c_list_offsets_result = rd_kafka_event_ListOffsets_result(rkev);
+                result = Admin_c_ListOffsetsResult_to_py(c_list_offsets_result);
+                break;
+        }
```
pranavrth, line 4698: Same

tests/integration/admin/test_list_offsets.py
```diff
@@ -0,0 +1,61 @@
...
+
+    requests = {topic_partition: OffsetSpec.max_timestamp()}
+    futmap = admin_client.list_offsets(requests, isolation_level=IsolationLevel.READ_UNCOMMITTED, request_timeout=30)
+    for _, fut in futmap.items():
+        result = fut.result()
+        assert isinstance(result, ListOffsetsResultInfo)
+        assert (result.offset == 1)
+
+    requests = {topic_partition: OffsetSpec.for_timestamp(base_timestamp + 150)}
+    futmap = admin_client.list_offsets(requests, isolation_level=IsolationLevel.READ_UNCOMMITTED, request_timeout=30)
```
pranavrth, line 52: Let's add a case for no Admin options.

tests/test_Admin.py
```diff
@@ -1014,3 +1014,19 @@ def test_alter_user_scram_credentials_api():
         a.alter_user_scram_credentials([UserScramCredentialDeletion("sam", "string type")])
     with pytest.raises(TypeError):
         a.alter_user_scram_credentials([UserScramCredentialDeletion("sam", 123)])
+
+
+def test_list_offsets_api():
```
pranavrth, line 1019: Add more test cases for all the validations.

pranavrth (changes requested):
Few more comments. 
Add Changelog.

src/confluent_kafka/admin/__init__.py
```diff
@@ -1103,5 +1155,46 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
...
+
+        :rtype: dict[TopicPartition, future]
+
+        :raises TypeError: Invalid input type.
+        :raises ValueError: Invalid input value.
+        """
+        AdminClient._check_list_offsets_request(topic_partition_offsets, kwargs)
+
+        if 'isolation_level' in kwargs:
+
```
pranavrth, line 1186: Remove extra line break

src/confluent_kafka/admin/__init__.py
```diff
@@ -1103,5 +1155,46 @@ def alter_user_scram_credentials(self, alterations, **kwargs):
...
+            kwargs['isolation_level_value'] = kwargs['isolation_level'].value
+            del kwargs['isolation_level']
```
pranavrth, line 1188: My point was that we should not touch the input variable. We can omit this line.

src/confluent_kafka/src/Admin.c
```diff
@@ -4219,6 +4309,57 @@ Admin_c_GroupResults_to_py (const rd_kafka_group_result_t **c_result_responses,
...
+                                goto raise;
+                }
+                PyDict_SetItem(result, c_part_to_py(c_topic_partition), value);
+                Py_DECREF(value);
+        }
+
+        Py_DECREF(ListOffsetsResultInfo_type);
+        return result;
+raise:
+        Py_DECREF(result);
```
pranavrth, line 4359: Use `Py_XDECREF` instead for raise cases.

when `i=0`, `result=NULL`, this statement will give segfault. Its safer to use XDECREF in general.

pranavrth (approved):
LGTM!. Minor nits.

src/confluent_kafka/admin/__init__.py
emasab: No it gives an error in that case because there's no `isolation_level` parameter

src/confluent_kafka/src/Admin.c
emasab: result cannot be NULL here, it's initialized before the for loop and there's no goto before

pranavrth: Still I prefer XDECREF. But its fine this way as well.

src/confluent_kafka/admin/__init__.py
pranavrth: Didn't get you. I am talking about deleting the key.

emasab: Yeah, if it's not deleted the error comes later when calling with kwargs

pranavrth: I don't understand why and where it gives issue. But its fine either way.

PR #1593, opened 2023-06-24
Added example with oauth/oidc example with CCloud logicalCluster and identityPoolId

sarwarbhuiyan:
The following example shows how to configure the producer with OAUTH. For Confluent Cloud where we need to pass extensions for "logicalCluster" and "identityPoolId" we need to pass some additional paramters.

### DIFF
--- a/examples/oauth_producer.py
+++ b/examples/oauth_producer.py
@@ -29,36 +29,29 @@ from confluent_kafka.serialization import StringSerializer
 import requests
 
 
-def _get_token(args, config):
-    """Note here value of config comes from sasl.oauthbearer.config below.
-    It is not used in this example but you can put arbitrary values to
-    configure how you can get the token (e.g. which token URL to use)
-    """
-    payload = {
-        'grant_type': 'client_credentials',
-        'scope': ' '.join(args.scopes)
-    }
-    resp = requests.post(args.token_url,
-                         auth=(args.client_id, args.client_secret),
-                         data=payload)
-    token = resp.json()
-    return token['access_token'], time.time() + float(token['expires_in'])
-
-
 def producer_config(args):
     logger = logging.getLogger(__name__)
-    return {
+    logger.setLevel(logging.DEBUG)
+    params = {
         'bootstrap.servers': args.bootstrap_servers,
-        'security.protocol': 'sasl_plaintext',
+        'retries': 1,
+        'message.send.max.retries': 1,
+        'delivery.timeout.ms': 30000,
+        'security.protocol': 'SASL_SSL',
         'sasl.mechanisms': 'OAUTHBEARER',
-        # sasl.oauthbearer.config can be used to pass argument to your oauth_cb
-        # It is not used in this example since we are passing all the arguments
-        # from command line
-        # 'sasl.oauthbearer.config': 'not-used',
-        'oauth_cb': functools.partial(_get_token, args),
-        'logger': logger,
+        'sasl.oauthbearer.method': 'oidc',
+        'sasl.oauthbearer.client.id': args.client_id,
+        'sasl.oauthbearer.client.secret': args.client_secret,
+        'sasl.oauthbearer.token.endpoint.url': args.token_url,
+        'sasl.oauthbearer.scope': ' '.join(args.scopes)
     }
+    if args.logical_cluster and args.identity_pool_id:
+        params['sasl.oauthbearer.extensions'] = 'logicalCluster='+args.logical_cluster+',identityPoolId='+args.identity_pool_id
 
+    if args.debug:
+       params['debug'] = args.debug
+
+    return params
 
 def delivery_report(err, msg):
     """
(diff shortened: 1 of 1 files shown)


pranavrth:
This example is more specific to Confluent Cloud. Can we move it to a new file?

pranavrth:
Just make sure that this example is tested.

examples/oauth_producer.py
```diff
@@ -29,36 +29,29 @@
...
-
-
 def producer_config(args):
     logger = logging.getLogger(__name__)
-    return {
+    logger.setLevel(logging.DEBUG)
+    params = {
         'bootstrap.servers': args.bootstrap_servers,
-        'security.protocol': 'sasl_plaintext',
+        'retries': 1,
```
pranavrth, line 37: This shouldn't be part of the example.

examples/oauth_producer.py
```diff
@@ -29,36 +29,29 @@
...
+        'message.send.max.retries': 1,
```
pranavrth, line 38: This shouldn't be part of the example as well.

examples/oauth_producer.py
```diff
@@ -131,5 +124,8 @@ def main(args):
                         help="Token URL.")
     parser.add_argument('--scopes', dest="scopes", required=True, nargs='+',
                         help="Scopes requested from OAuth server.")
-
+    parser.add_argument('--logical-cluster', dest="logical_cluster", required=False, help="Logical Cluster.")
+    parser.add_argument('--identity-pool-id', dest="identity_pool_id", required=False, help="Identity Pool ID.")
+    parser.add_argument('--debug', dest="debug", required=False, help="Comma separated list of following values broker,topic,msg")
```
pranavrth, line 129: Let's not add debug as part of the example.

examples/oauth_producer.py
```diff
@@ -29,36 +29,29 @@
...
 def producer_config(args):
     logger = logging.getLogger(__name__)
-    return {
+    logger.setLevel(logging.DEBUG)
+    params = {
         'bootstrap.servers': args.bootstrap_servers,
-        'security.protocol': 'sasl_plaintext',
+        'retries': 1,
+        'message.send.max.retries': 1,
+        'delivery.timeout.ms': 30000,
```
pranavrth, line 39: This shouldn't be part of the example as well.

examples/oauth_producer.py
```diff
@@ -131,5 +124,8 @@ def main(args):
                         help="Token URL.")
     parser.add_argument('--scopes', dest="scopes", required=True, nargs='+',
                         help="Scopes requested from OAuth server.")
-
+    parser.add_argument('--logical-cluster', dest="logical_cluster", required=False, help="Logical Cluster.")
```
pranavrth, line 127: If we move this to CC specific example, does it make sense to mark these are required?

pranavrth (changes requested):

examples/oauth_producer.py
sarwarbhuiyan: The original example is using an outdated callback based method so this example applies there too. Also, since the only difference between CC and AK/CP is this one param, it doesn't make sense to split it into another file. There's a wider task of updating all the examples to be using the recommended approach. 

The purpose of examples is to educate.

sarwarbhuiyan:
the example was tested against confluent cloud

examples/oauth_producer.py
pranavrth: We can improve old example as well and create a new example to cater mainly CC usecase. The idea is to make it easy for non CC user to use SASL as well. With everything specific to CC, user might get confused.

homestealth:
@pranavrth there seems to be two issues here:
1. the original client example for oauth seems to be using the incorrect callbackhandler. it should be using the librd one - can we fix this asap?
2. ccloud example is missing (for sasl extensions) - this is ok in a separate file, but we should create it

examples/oauth_producer.py
sarwarbhuiyan: It's not just specific to CC though. If you don't include the logicalClusterId and identityPoolId in the cli params, it's just just works against any AK or CP.

sarwarbhuiyan: I just don't want code that'll be copy pasted for a 2 line difference. The cli works with both CP and CC.

sarwarbhuiyan:
> @pranavrth there seems to be two issues here:
> 
> 1. the original client example for oauth seems to be using the incorrect callbackhandler. it should be using the librd one - can we fix this asap?
> 2. ccloud example is missing (for sasl extensions) - this is ok in a separate file, but we should create it

@homestealth the ccloud example is here in my PR with two additional parameters that does the trick and since this is a cli example I've added optional parameters to the cli which make it work. It seems overkill to have a separate file for the exact same code to be copied.

pranavrth:
I think it is fine to keep only one file but we should add a comment stating the two new configs are for CC case only.

Please also sign the CLA.

pranavrth:
@emasab - Any comment from your side before we can merge this?

dylanmei:
This is a valuable example, since it is config-based and removes the callback. How can we close this out?

pranavrth:
@sarwarbhuiyan - Can you please address my last comment to add some documentation? Please sign the CLA as well. We should merge this PR.

jriester:
Hi team, two more customers have reported issues with our current example. GTS has been finding this pull and giving the customer the example here, any way we can move this to live?

PR #1595, opened 2023-06-26
Fix SchemaRegistryClient docs

pko-trackunit:
Fixed few mismached URLs and missing default value for a parameter.

### DIFF
--- a/src/confluent_kafka/schema_registry/schema_registry_client.py
+++ b/src/confluent_kafka/schema_registry/schema_registry_client.py
@@ -437,7 +437,7 @@ class SchemaRegistryClient(object):
             SchemaRegistryError: if subjects can't be found
 
         See Also:
-            `GET subjects API Reference <https://docs.confluent.io/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions>`_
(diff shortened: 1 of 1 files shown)


naxinf:
Hi @pko-trackunit, thanks for raising the PR. I double checked and the two doc URLs you fixed and the one for `lookup_schema()` (should be https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string- subject)) are indeed incorrect

> Also, it feels like `lookup_schema()` method should not be using an endpoint that "Register a new schema under the specified subject". Unless I am misunderstanding what "lookup" means.

src/confluent_kafka/schema_registry/schema_registry_client.py
```diff
@@ -437,7 +437,7 @@ def get_subjects(self):
             SchemaRegistryError: if subjects can't be found
 
         See Also:
-            `GET subjects API Reference <https://docs.confluent.io/current/schema-registry/develop/api.html#get--subjects-(string-%20subject)-versions>`_
+            `GET subjects API Reference <https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get--subjects>`_
```
naxinf: Let's omit the `/platform` part (which will be added during redirection) of the URl to stay consistent with the rest of URLs in docs

fangnx (approved):

PR #1626, opened 2023-08-15
Added list_schemas method to output all schemas registered to a Schema Registry instance.

wadefacer-hcp:
Using this library I have found myself wishing there was a direct way to leverage the {BASE_URL}/schemas end point to pull all needed data for schema version control. I found the library currently lacking the functionality in a single API call so I am proposing changes to incorporate this functionality.


joaoe:
Hi.
I think you should clean and rebase your branch. Your commit includes 711 new files with your own virtual environment, I don't think you want to merge that in.

PR #1627, opened 2023-08-18
Support avro union schema serde

### LINKED ISSUE #1562: UnknownType error when deserializing using union schema
Product: Confluent Cloud / confluent_kafka
Using confluent-kafka 2.1.0     

We were trying to follow the instructions [here](https://www.confluent.io/blog/multiple-event-types-in-the-same-kafka-topic/) to configure the schema on one of our topics to be an Avro union. Using the control center (we use Confluent Cloud), we set the schema to ["Schema1", "Schema2"], adding Schema1 and Schema2 as references (not the actual names, they are long and irrelevant).
We were then able to produce events with both schemas to the topic. However, when we tried to consume and deserialize using the AvroDeserializer, we got "fastavro._schema_common.UnknownType: Schema1"

The pseudocode of what we're trying to do:
```
msg = consumer.poll()
msg_type = msg.headers()[MSG_TYPE]  # will return whether it will be Type1 or Type2
deserializer = AvroDeserializer(
          schema_str: get_schema_from_msg_type(msg_type), # will return Schema1 or Schema2
          schema_registry_client=schema_registry_client, 
          from_dict=some_from_dict
)
deserialized_value = deserializer(msg.value(), SerializationContext(msg.topic(), MessageField.VALUE, msg.headers())
```

I think the issue is here: https://github.com/confluentinc/confluent-kafka-python/blob/f709e7138d41a36e89efb02210f4c9dbe020f752/src/confluent_kafka/schema_registry/avro.py#L419
When the code calls parse_schema, it doesn't use `expand=True.` Thus, it sets the writer schema to ["Schema1","Schema2"], as was configured on the topic, but does not provide a reference to what Schema1 and Schema2 are. Thus, when it calls `schemaless_reader` on the next line, it has no idea what Schema1 and Schema2 are. I think the fix is to change the call to 
```
writer_schema = parse_schema(
       loads(prepared_schema.schema_str), 
       named_schemas=self._named_schemas),
       expand=True
```

sookyomkim:
fixes #1562 

This pr expand named schema fully to their true schemas to fix "UnknownType" error when using avro union schema like below.

```json
[
    "confluent.io.examples.serialization.avro.User",
    "confluent.io.examples.serialization.avro.Award"
]
```

### DIFF
--- a/src/confluent_kafka/schema_registry/avro.py
+++ b/src/confluent_kafka/schema_registry/avro.py
@@ -82,7 +82,7 @@ def _resolve_named_schema(schema, schema_registry_client, named_schemas=None):
             referenced_schema = schema_registry_client.get_version(ref.subject, ref.version)
             _resolve_named_schema(referenced_schema.schema, schema_registry_client, named_schemas)
             parse_schema(loads(referenced_schema.schema.schema_str),
-                         named_schemas=named_schemas, expand=True)
+                         named_schemas=named_schemas)
     return named_schemas
 
 
--- a/src/confluent_kafka/schema_registry/avro.py
+++ b/src/confluent_kafka/schema_registry/avro.py
@@ -425,7 +425,7 @@ class AvroDeserializer(Deserializer):
             obj_dict = schemaless_reader(payload,
                                          writer_schema,
                                          self._reader_schema,
-                                         self._return_record_name)
+                                         self._return_record_name, return_record_name_override=True)
 
             if self._from_dict is not None:
                 return self._from_dict(obj_dict, ctx)
--- a/src/confluent_kafka/schema_registry/avro.py
+++ b/src/confluent_kafka/schema_registry/avro.py
@@ -81,7 +81,8 @@ def _resolve_named_schema(schema, schema_registry_client, named_schemas=None):
         for ref in schema.references:
             referenced_schema = schema_registry_client.get_version(ref.subject, ref.version)
             _resolve_named_schema(referenced_schema.schema, schema_registry_client, named_schemas)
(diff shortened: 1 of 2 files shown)


src/confluent_kafka/schema_registry/avro.py
```diff
@@ -81,7 +81,8 @@ def _resolve_named_schema(schema, schema_registry_client, named_schemas=None):
         for ref in schema.references:
             referenced_schema = schema_registry_client.get_version(ref.subject, ref.version)
             _resolve_named_schema(referenced_schema.schema, schema_registry_client, named_schemas)
-            parse_schema(loads(referenced_schema.schema.schema_str), named_schemas=named_schemas)
+            parse_schema(loads(referenced_schema.schema.schema_str),
+                         named_schemas=named_schemas)
```
rayokota, line 85: I think we need `expand=True` here top

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -416,14 +418,14 @@ def __call__(self, data, ctx):
...
-                    prepared_schema.schema_str), named_schemas=self._named_schemas)
+                writer_schema = parse_schema(loads(prepared_schema.schema_str),
+                                             named_schemas=self._named_schemas, expand=True)
                 self._writer_schemas[schema_id] = writer_schema
 
             obj_dict = schemaless_reader(payload,
                                          writer_schema,
                                          self._reader_schema,
-                                         self._return_record_name)
+                                         self._return_record_name, return_record_name_override=True)
```
rayokota, line 428: Is there a reason why we need `return_record_name_override=True`?

rayokota (commented):
Thanks @sookyomkim , left some comments

rayokota:

PR #1630, opened 2023-08-22
Feature/bugfix 1616

### LINKED ISSUE #1616: How do I describe all users using the admin client?
Description
===========
The `describe_user_scram_credentials` method was added to the admin client recently but I'm unable to find a way to describe all users. The equivalent Kafka CLI command for what I'm trying to do is `kafka-configs.sh --describe --entity-type users`. I also see that librdkafka supports this functionality https://github.com/confluentinc/librdkafka/blob/7cc860d4d099bfe4c53f60950189ea139a660128/src/rdkafka.h#L8716-L8717 

How to reproduce
================
I tried passing in an empty user list but it doesn't work
```
admin_client.describe_user_scram_credentials([]) # just returns empty map
```

Checklist
=========

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): ('2.2.0', 33685504) and ('2.2.0', 33685759)
 - [x] Apache Kafka broker version: 3.3.1
 - [x] Operating system: Mac
 - [ ] Critical issue

mahajanadhitya:
Fixes #1616


src/confluent_kafka/admin/__init__.py
```diff
@@ -247,20 +247,21 @@ def _make_acls_result(f, futmap):
...
-            len_futures = len(futmap)
-            if len(results) != len_futures:
-                raise RuntimeError(
-                    f"Results length {len_results} is different from future-map length {len_futures}")
-            for username, value in results.items():
-                fut = futmap.get(username, None)
-                if fut is None:
+            if "__all__" in futmap:
+                fut = futmap["__all__"]
+                fut.set_result(results)
```
pranavrth, line 252: Can there be any error case?

src/confluent_kafka/admin/__init__.py
```diff
@@ -976,12 +977,13 @@ def describe_user_scram_credentials(self, users, **kwargs):
         :raises ValueError: Invalid input value.
         """
         AdminClient._check_describe_user_scram_credentials_request(users)
-
         f, futmap = AdminClient._make_futures_v2(users, None,
                                                  AdminClient._make_user_scram_credentials_result)
 
-        super(AdminClient, self).describe_user_scram_credentials(users, f, **kwargs)
+        if len(users) == 0:
+            futmap["__all__"] = AdminClient._create_future()
```
pranavrth, line 984: Shouldn't this be part of the above `_make_future` functions specific to this API?

examples/adminapi.py
```diff
@@ -603,18 +603,29 @@ def example_describe_user_scram_credentials(a, args):
...
+        fut = futmap["__all__"]
+        results = fut.result()
+        for username, response in results.items():
+            print("Username : {}\n".format(username))
+            if isinstance(response, KafkaError):
+                print(response)
+            else:
+                for scram_credential_info in response.scram_credential_infos:
+                    print(f"    Mechanism: {scram_credential_info.mechanism} " +
+                          f"Iterations: {scram_credential_info.iterations}")
```
pranavrth, line 628: Make the example easy for user to understand.

src/confluent_kafka/admin/__init__.py
mahajanadhitya: No error could be there.

src/confluent_kafka/admin/__init__.py
mahajanadhitya: Actually no it is generic function and not specific to api otherwise we would have to change the signature of the api. As this is the only case for describe_user_scram api.

src/confluent_kafka/admin/__init__.py
```diff
@@ -247,20 +247,21 @@ def _make_acls_result(f, futmap):
     def _make_user_scram_credentials_result(f, futmap):
```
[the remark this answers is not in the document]
mahajanadhitya: Since we are given a dict and we have to only set the result, so no change to alter would be there. Also __all__ is checked when futmap is empty , incase of alter we would always have non empty futmap so it does not affect the alter workflow.

src/confluent_kafka/admin/__init__.py
```diff
@@ -314,6 +326,16 @@ def _make_futures_v2(futmap_keys, class_check, make_result_fn):
 
         return f, futmap
 
+    @staticmethod
+    def _make_futures_v3(make_result_fn):
+        # Create an internal future for the entire request,
+        # this future will trigger _make_..._result() and set result/exception
+        f = AdminClient._create_future()
+        internalfuture = AdminClient._create_future()
```
pranavrth, line 334: better name it `internal_f` as everywhere f means future.

examples/adminapi.py
```diff
@@ -602,19 +602,32 @@ def example_describe_user_scram_credentials(a, args):
...
-            print("    Error: {}".format(e))
+            print("Failed to describe all user scram credentials : {}".format(e))
         except Exception as e:
-            print(f"    Unexpected exception: {e}")
+            raise
+    else:
+        futmap = a.describe_user_scram_credentials(args)
+        for username, fut in futmap.items():
+            print("Username: {}".format(username))
+            try:
```
pranavrth, line 622: Separate out both the cases in different functions. Or make it easier for user to understand both the cases.

pranavrth (changes requested):
Add test cases and changelog

src/confluent_kafka/admin/__init__.py
```diff
@@ -314,6 +326,16 @@ def _make_futures_v2(futmap_keys, class_check, make_result_fn):
 
         return f, futmap
 
+    @staticmethod
+    def _make_futures_v3(make_result_fn):
```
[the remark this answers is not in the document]
mahajanadhitya: suggest something please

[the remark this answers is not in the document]
mahajanadhitya: I have named it _make_internal_futures

emasab:
Made some change following what we were talking about, keeping same behavior if a list is passed

pranavrth (changes requested):
- Rebase the branch. 
- None and empty list both have same behavior. Reflect this in the docs.

src/confluent_kafka/admin/__init__.py
```diff
@@ -398,9 +422,13 @@ def _check_alter_consumer_group_offsets_request(request):
 
     @staticmethod
     def _check_describe_user_scram_credentials_request(users):
+        if users is None:
+            return
```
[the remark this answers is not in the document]
emasab: this just skips the checks

src/confluent_kafka/admin/__init__.py
```diff
@@ -955,34 +983,40 @@ def set_sasl_credentials(self, username, password):
...
 
-    def describe_user_scram_credentials(self, users, **kwargs):
+    def describe_user_scram_credentials(self, users=None, **kwargs):
         """
         Describe user SASL/SCRAM credentials.
 
         :param list(str) users: List of user names to describe.
-               Duplicate users aren't allowed.
+               Duplicate users aren't allowed. Can be None
+               to describe all user's credentials.
```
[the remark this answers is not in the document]
emasab: Cannot be empty, in that case we're keeping current behavior to give an error

src/confluent_kafka/admin/__init__.py
```diff
@@ -955,34 +983,40 @@ def set_sasl_credentials(self, username, password):
...
         :param float request_timeout: The overall request timeout in seconds,
                including broker lookup, request transmission, operation time
                on broker, and response. Default: `socket.timeout.ms*1000.0`
 
-        :returns: A dict of futures keyed by user name.
-                  The future result() method returns the
-                  :class:`UserScramCredentialsDescription` or
-                  raises KafkaException
+        :returns: In case None is passed it returns a single future.
```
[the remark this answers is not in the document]
emasab: In case an empty list is passed we keep current behavior an return a futmap

examples/adminapi.py
```diff
@@ -602,19 +602,46 @@ def example_describe_user_scram_credentials(a, args):
...
-            print("    Error: {}".format(e))
-        except Exception as e:
-            print(f"    Unexpected exception: {e}")
+            print("Failed to describe all user scram credentials : {}".format(e))
+        except Exception:
+            raise
+    else:
+        """
+        Case: Describe specified user scram credentials
+        Input: users is a list
```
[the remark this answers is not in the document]
emasab: it can be empty we keep current behavior

src/confluent_kafka/admin/__init__.py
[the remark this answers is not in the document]
pranavrth: Got your point. Does java do the same?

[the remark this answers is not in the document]
emasab: No, java it's for both null and empty. But we've released it and cannot change this behavior

src/confluent_kafka/admin/__init__.py
```diff
@@ -277,19 +277,19 @@ def _make_acls_result(f, futmap):
                 fut.set_exception(e)
 
     @staticmethod
-    def _make_user_scram_credentials_result(f, futmap):
+    def _make_futmap_result(f, futmap):
```
pranavrth, line 280: Can we use latest created `_make_futmap_result_list` function instead?

tests/test_Admin.py
```diff
@@ -869,10 +869,14 @@ def test_describe_user_scram_credentials_api():
     # Describe User Scram API
     a = AdminClient({"socket.timeout.ms": 10})
 
+    f = a.describe_user_scram_credentials()
```
pranavrth, line 872: Add an integration test for the fix that we are deploying i.e. describe all users case.

pranavrth (changes requested):
Fix CI.

src/confluent_kafka/admin/__init__.py
emasab: They're different in one the future returns a `dict` and in `_make_futmap_result_list` it returns a `list`

pranavrth (approved):
LGTM!.

PR #1649, opened 2023-10-11
Adding python 3.12 wheels

pranavrth:


pranavrth:
Will update the changelog and fix the styling issue once other PRs are merged.

tests/requirements.txt
```diff
@@ -1,9 +1,13 @@
+urllib3<2.0.0;python_version<="3.7"
+urllib3
 flake8
 pytest==4.6.9;python_version<="3.0"
 pytest>=6.0.0;python_version>="3.0"
 pytest-timeout
 requests-mock
 trivup>=0.8.3
+fastavro<1.8.0;python_version=="3.7"
```
milindl, line 9: question, not a comment: below it seems we support cp36, but here, no dependency will be installed for python 3.6, only 3.7 above, is this the intended behaviour?

milindl (approved):
lgtm!

tests/requirements.txt
emasab: does it work if you set ?
```suggestion
fastavro<1.8.0;python_version<="3.7"
```

milindl: Because there is just "fastavro" in the line below, it falls back to that and handles 3.6, so it's not a problem. I discussed with Pranav on a call earlier

tools/build-manylinux.sh
```diff
@@ -89,12 +89,13 @@ for PYBIN in /opt/python/cp*/bin/; do
     for PYTHON_VERSION in "${PYTHON_VERSIONS[@]}"; do
         if [[ $PYBIN == *"$PYTHON_VERSION"* ]]; then
             echo "## Installing $PYBIN"
-            "${PYBIN}/pip" install confluent_kafka -f /io/wheelhouse
-            "${PYBIN}/python" -c 'import confluent_kafka; print(confluent_kafka.libversion())'
-            "${PYBIN}/pip" install -r /io/tests/requirements.txt
-            "${PYBIN}/pytest" /io/tests/test_Producer.py
+            "${PYBIN}pip" -V
```
emasab, line 92: Would be better avoiding the double slash by removing final slash from `for PYBIN in /opt/python/cp*/bin/; do`

As in `# Compile wheels`

PR #1662, opened 2023-10-24
Fix calls with empty lists

emasab:
describe_topics and list_offsets should return an empty output if and empty input is passed as parameter.
Consistent with Java and user friendly for users using filter and map.


pranavrth (approved):
LGTM!. Just one comment - Can we add a test case in integration test for this usecase?

PR #1336, opened 2022-05-12
don't install LICENSE.txt to /usr/LICENSE.txt (or $PREFIX)

asottile-sentry:
data_files is intended to be used to place manual files in the output.  the intended mechanism for including the license in the distribution is the `license_file` or `license_files` options -- I picked the former since it has further compatibility

### before:

```console
$ CPATH="$(brew --prefix librdkafka)/include" LDFLAGS="-L$(brew --prefix librdkafka)/lib" bash -xc 'git rev-parse HEAD && rm -rf venv && virtualenv venv -qq && venv/bin/pip install -qq . && ls -al venv'
+ git rev-parse HEAD
e7e066120f05ca6f9c37b18d77f974f1a6cad222
+ rm -rf venv
+ virtualenv venv -qq
+ venv/bin/pip install -qq .
+ ls -al venv
total 44
drwxr-xr-x  8 asottile staff   256 May 12 09:30 .
drwxr-xr-x 25 asottile staff   800 May 12 09:30 ..
lrwxr-xr-x  1 asottile staff    93 May 12 09:30 .Python -> /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/Python3
-rw-r--r--  1 asottile staff    40 May 12 09:30 .gitignore
-rw-r--r--  1 asottile staff 32984 May 12 09:30 LICENSE.txt
drwxr-xr-x 20 asottile staff   640 May 12 09:30 bin
drwxr-xr-x  3 asottile staff    96 May 12 09:30 lib
-rw-r--r--  1 asottile staff   528 May 12 09:30 pyvenv.cfg
```

### after:

```console
$ CPATH="$(brew --prefix librdkafka)/include" LDFLAGS="-L$(brew --prefix librdkafka)/lib" bash -xc 'git rev-parse HEAD && rm -rf venv && virtualenv venv -qq && venv/bin/pip install -qq . && ls -al venv'
+ git rev-parse HEAD
4c6328c314a8c7366950161eaf161c0b89f49558
+ rm -rf venv
+ virtualenv venv -qq
+ venv/bin/pip install -qq .
+ ls -al venv
total 8
drwxr-xr-x  7 asottile staff 224 May 12 09:30 .
drwxr-xr-x 25 asottile staff 800 May 12 09:30 ..
lrwxr-xr-x  1 asottile staff  93 May 12 09:30 .Python -> /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/Python3
[...]

PR #1672, opened 2023-11-13
[CI] list the LICENSE as a license_file

emasab:
Data files are installed in the wrong directory when packaging this, and it is very uncommon to do so. data files should be files needed for the runtime only.


emasab:
Thanks @dirkmueller  @asottile-sentry ! Chose @dirkmueller one because `license_files` is available since setuptools 42 (Nov 2019) and it's older than 3 years. [Confluent Platinum end of support](https://docs.confluent.io/platform/current/clients/index.html#librdkafka-and-derived-clients)

emasab:
See #713 #1336.

emasab:
@asottile-sentry the author is still @dirkmueller. We cannot merge directly because our CI won't run on external code and that code could change any moment. We're acknowledging contributions in CHANGELOG.

asottile-sentry:
talk about "open source" only in name

emasab:
We're working on a way to run CI directly on external PRs, still after manual review of its code, so we don't have to do this.

PR #1681, opened 2023-11-24
Fix OAUTHBEARER/OIDC extensions invalid writes

emasab:
an additional byte must be added for the closing zeros


emasab:
Done in #1745

PR #1692, opened 2023-12-15
Fix segfault with describe_topics and flaky connection

lpsinger:
If you call describe_topics on a flaky connection, sometimes the admin client reply has the host set to a null pointer. When this occurs, instead of segfaulting, report the host as None.

### DIFF
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -3351,8 +3351,12 @@ Admin_c_AclBinding_to_py (const rd_kafka_AclBinding_t *c_acl_binding) {
                                 rd_kafka_AclBinding_resource_pattern_type(c_acl_binding));
         cfl_PyDict_SetString(kwargs, "principal",
                                      rd_kafka_AclBinding_principal(c_acl_binding));
-        cfl_PyDict_SetString(kwargs, "host",
-                                     rd_kafka_AclBinding_host(c_acl_binding));
+        if (rd_kafka_AclBinding_host(c_acl_binding)) {
+                cfl_PyDict_SetString(kwargs, "host",
+                                        rd_kafka_AclBinding_host(c_acl_binding));
+        } else {
+                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
+        }
         cfl_PyDict_SetInt(kwargs, "operation",
                                      rd_kafka_AclBinding_operation(c_acl_binding));
         cfl_PyDict_SetInt(kwargs, "permission_type",
@@ -3624,9 +3628,13 @@ static PyObject *Admin_c_MemberDescription_to_py(const rd_kafka_MemberDescriptio
                              "client_id",
                              rd_kafka_MemberDescription_client_id(c_member));
 
-        cfl_PyDict_SetString(kwargs,
-                             "host",
-                             rd_kafka_MemberDescription_host(c_member));
+        if (rd_kafka_MemberDescription_host(c_member)) {
+                cfl_PyDict_SetString(kwargs,
+                                "host",
+                                rd_kafka_MemberDescription_host(c_member));
+        } else {
+                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
+        }
 
         const char * c_group_instance_id = rd_kafka_MemberDescription_group_instance_id(c_member);
         if(c_group_instance_id) {
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -3351,12 +3351,8 @@ Admin_c_AclBinding_to_py (const rd_kafka_AclBinding_t *c_acl_binding) {
                                 rd_kafka_AclBinding_resource_pattern_type(c_acl_binding));
         cfl_PyDict_SetString(kwargs, "principal",
                                      rd_kafka_AclBinding_principal(c_acl_binding));
-        if (rd_kafka_AclBinding_host(c_acl_binding)) {
-                cfl_PyDict_SetString(kwargs, "host",
-                                        rd_kafka_AclBinding_host(c_acl_binding));
-        } else {
-                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
-        }
+        cfl_PyDict_SetString(kwargs, "host",
+                                     rd_kafka_AclBinding_host(c_acl_binding));
         cfl_PyDict_SetInt(kwargs, "operation",
                                      rd_kafka_AclBinding_operation(c_acl_binding));
         cfl_PyDict_SetInt(kwargs, "permission_type",
@@ -3628,13 +3624,9 @@ static PyObject *Admin_c_MemberDescription_to_py(const rd_kafka_MemberDescriptio
                              "client_id",
                              rd_kafka_MemberDescription_client_id(c_member));
 
-        if (rd_kafka_MemberDescription_host(c_member)) {
-                cfl_PyDict_SetString(kwargs,
-                                "host",
-                                rd_kafka_MemberDescription_host(c_member));
-        } else {
-                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
-        }
+        cfl_PyDict_SetString(kwargs,
+                             "host",
+                             rd_kafka_MemberDescription_host(c_member));
 
         const char * c_group_instance_id = rd_kafka_MemberDescription_group_instance_id(c_member);
         if(c_group_instance_id) {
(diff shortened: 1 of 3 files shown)


pranavrth:
Thanks for the PR.

Even though the changes seems correct, I think we should not change the base implementation which is being used in alot of APIs. Some of the values of other APIs must not be NULL. I think its better to handle the scenario in this way - https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/src/Admin.c#L3198-L3201.

I am also checking if it should be possible to get the host information as NULL or should we handle it as an error instead. But in the meantime you can rebase the branch and do the change mentioned above if you have time. Otherwise, you can wait for my revert.

lpsinger:
> But in the meantime you can rebase the branch and do the change mentioned above if you have time.

Done.

src/confluent_kafka/src/Admin.c
```diff
@@ -3351,8 +3351,12 @@ Admin_c_AclBinding_to_py (const rd_kafka_AclBinding_t *c_acl_binding) {
...
         cfl_PyDict_SetString(kwargs, "principal",
                                      rd_kafka_AclBinding_principal(c_acl_binding));
-        cfl_PyDict_SetString(kwargs, "host",
-                                     rd_kafka_AclBinding_host(c_acl_binding));
+        if (rd_kafka_AclBinding_host(c_acl_binding)) {
+                cfl_PyDict_SetString(kwargs, "host",
+                                        rd_kafka_AclBinding_host(c_acl_binding));
+        } else {
+                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
+        }
```
pranavrth, line 3359: Please revert this change:

1. This change is not related to `describe_topic`
2. We can't reproduce this
3. It shouldn't be null as per the protocol

src/confluent_kafka/src/Admin.c
```diff
@@ -3624,9 +3628,13 @@ static PyObject *Admin_c_MemberDescription_to_py(const rd_kafka_MemberDescriptio
...
-        cfl_PyDict_SetString(kwargs,
-                             "host",
-                             rd_kafka_MemberDescription_host(c_member));
+        if (rd_kafka_MemberDescription_host(c_member)) {
+                cfl_PyDict_SetString(kwargs,
+                                "host",
+                                rd_kafka_MemberDescription_host(c_member));
+        } else {
+                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
+        }
```
pranavrth, line 3637: Please revert this change:

1. This change is not related to describe_topic
2. We can't reproduce this
3. It shouldn't be null as per the protocol

src/confluent_kafka/src/Metadata.c
```diff
@@ -208,10 +208,18 @@ static PyObject *c_broker_to_py(Handle *self, PyObject *BrokerMetadata_type,
...
+                                                c_broker.host) == -1) {
+                        Py_DECREF(broker);
+                        return NULL;
+                }
+        } else {
+                if (cfl_PyObject_SetItemString(broker, "host",
+                                                Py_None) == -1) {
+                        Py_DECREF(broker);
+                        return NULL;
+                }
```
pranavrth, line 222: Please revert this change:

1. This function is used in deprecated API and will be removed in future versions
2. This change is not related to describe_topic
3. We can't reproduce this
4. It shouldn't be null as per the protocol

pranavrth (changes requested):
Provided minor comments.

Please update the CHANGELOG.md as well with the entry for the fix.

src/confluent_kafka/src/Metadata.c
lpsinger: Are you sure about that? Ultimately this function is called by `list_topics` in Metadata.c which sounds like it may be related to the segfaults I reported.

Were you able to isolate those segfaults to the one bit of code that you do want to change, in confluent_kafka.c?

emasab: The host comes from the Metadata response in `list_topics` and is not nullable there
https://github.com/apache/kafka/blob/e8cd661becb08279702f1670bb8f1d816537adea/clients/src/main/resources/common/message/MetadataResponse.json#L54

otherwise it can come from list_group (deprecated) that in librdkafka only does a list + describe and in that case it's taken from the host of the coordinator that returned the DescribeGroup response here:
https://github.com/confluentinc/librdkafka/blob/c95d2ee950bda779c63843d376aab1d90ed8dfa9/src/rdkafka.c#L4739

src/confluent_kafka/src/Admin.c
emasab: +1 : In this case it's the member host and is not nullable by the protocol

https://github.com/apache/kafka/blob/18f2589c1e087654c0cdd46d2473246f37f0b66a/clients/src/main/resources/common/message/DescribeGroupsResponse.json#L57

src/confluent_kafka/src/Admin.c
emasab: Not nullable too in DescribeAclsResponse and DeleteAclsResponse

https://github.com/apache/kafka/blob/d65d8867983e54a36178c3a1227b68c8a468d415/clients/src/main/resources/common/message/DescribeAclsResponse.json#L45
https://github.com/apache/kafka/blob/d65d8867983e54a36178c3a1227b68c8a468d415/clients/src/main/resources/common/message/DeleteAclsResponse.json#L49

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1620,7 +1620,10 @@ PyObject *c_Node_to_py(const rd_kafka_Node_t *c_node) {
 
         cfl_PyDict_SetInt(kwargs, "id", rd_kafka_Node_id(c_node));
         cfl_PyDict_SetInt(kwargs, "port", rd_kafka_Node_port(c_node));
-        cfl_PyDict_SetString(kwargs, "host", rd_kafka_Node_host(c_node));
+        if (rd_kafka_Node_host(c_node))
```
emasab, line 1623: Now I have a doubt that the fix should be here in CKPy, because this uses the Metadata response too for the replicas and isrs and it uses the Metadata host here (not nullable)
https://github.com/apache/kafka/blob/e8cd661becb08279702f1670bb8f1d816537adea/clients/src/main/resources/common/message/MetadataResponse.json#L54
And broker metadata must be present for all the indexed brokers

emasab: OK, here it can be None because it's possible not all the brokers are returned from the Metadata call. And in that case it just sets the broker id here
https://github.com/confluentinc/librdkafka/blob/c95d2ee950bda779c63843d376aab1d90ed8dfa9/src/rdkafka_aux.c#L281

In the `list_topics` API, instead, only the broker id list is returned in the `replicas` and `isrs` fields so the broker list in `ClusterMetadata` only contains complete `BrokerMetadata` objects.

emasab: It seems it's the expected behavior. I've created a PR for changing the documentation [here](https://github.com/apache/kafka/pull/15809).
An update: in Go and .NET it doesn't happen because it either uses `C.GoString` or `PtrToStringUTF8` that returns `nil` or `null`.

lpsinger:
@emasab, what else do you want me to do with this?

emasab:
@lpsinger after reviewing all the cases only in `c_Node_to_py` it's needed to check for NULL values because it takes the nodes from the Metadata request and some of then can be absent, so without host, when the broker is down and not listed there.

lpsinger:
> Could you leave only that change?

Done. Would you like me to squash the changes?

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1620,7 +1620,10 @@ PyObject *c_Node_to_py(const rd_kafka_Node_t *c_node) {
...
+                cfl_PyDict_SetString(kwargs, "host", rd_kafka_Node_host(c_node));
+        else
+                cfl_PyObject_SetItemString(kwargs, "host", Py_None);
```
pranavrth, line 1626: This function is not present. Please use `PyDict_SetItemString` instead as used in https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/src/Admin.c#L3198-L3201

pranavrth: Build itself is failing due to this.

lpsinger: Fixed.

pranavrth (approved):
Thanks for the contribution. LGTM!

emasab (approved):

lpsinger:
You're very welcome!

PR #1704, opened 2024-01-17
Set avro/json schema on use.latest.version=True

ifnesi:
The current `AvroSerializer` (`src/confluent_kafka/schema_registry/avro.py`) and `JSONSerializer` (`src/confluent_kafka/schema_registry/json_schema.py`) classes do not override the schema when `use.latest.schema` is set to `True`, it currently only returns the `schema_id`. This PR is to address that.

Please notice the same behaviour happens with `ProtobufSerializer` (`src/confluent_kafka/schema_registry/protobuf.py`), however not sure how `use.latest.schema` could work with Protobuf schemas as it needs to be "compiled" and the SR stores the "uncompiled" version of it. Maybe we should drop that configuration parameter out?


src/confluent_kafka/schema_registry/avro.py
```diff
@@ -228,27 +222,46 @@ def __init__(self, schema_registry_client, schema_str, to_dict=None, conf=None):
...
-        parsed_schema = parse_schema(schema_dict, named_schemas=self._named_schemas)
+        # Set common instance variables
+        # Called on __init__ and
+        # on __call__ if use.latest.version is set to True
+        self._set_instance_variables(schema_str)
+
+    def _set_instance_variables(self, schema_str, schema_id=None):
+        """
+        Function to be called upon __init__ and when serializing a message (__call__)
+        if use.latest.version is set True
```
pranavrth, line 233: Add explanation of what this function does as well. This can be written after the description.

src/confluent_kafka/schema_registry/json_schema.py
pranavrth, line 1: Same comments as Avro file.

pranavrth (changes requested):
Please update the PR with the following:
1. Rebase the PR
2. Test cases for the fix
3. CHANGELOG entry

Done review of the code part of the PR. I am going to test these changes once you have updated the PR with the requested changes and provide more comments if required.

I am checking the Protobuf part.

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -185,23 +185,17 @@ class AvroSerializer(Serializer):
...
     def __init__(self, schema_registry_client, schema_str, to_dict=None, conf=None):
-        if isinstance(schema_str, str):
-            schema = _schema_loads(schema_str)
-        elif isinstance(schema_str, Schema):
-            schema = schema_str
-        else:
+        if not isinstance(schema_str, (str, Schema)):
             raise TypeError('You must pass either schema string or schema object')
 
+        self._to_dict = to_dict
```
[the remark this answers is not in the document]
ifnesi: I will move the two validation IFs to the start of the method. We still need `self._to_dict = to_dict` because it is an instance variable required on other class methods

src/confluent_kafka/schema_registry/avro.py
```diff
@@ -228,27 +222,46 @@ def __init__(self, schema_registry_client, schema_str, to_dict=None, conf=None):
...
 
-        schema_dict = loads(schema.schema_str)
-        self._named_schemas = _resolve_named_schema(schema, schema_registry_client)
-        parsed_schema = parse_schema(schema_dict, named_schemas=self._named_schemas)
+        # Set common instance variables
+        # Called on __init__ and
+        # on __call__ if use.latest.version is set to True
+        self._set_instance_variables(schema_str)
+
+    def _set_instance_variables(self, schema_str, schema_id=None):
```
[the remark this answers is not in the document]
ifnesi: Done

src/confluent_kafka/schema_registry/avro.py
ifnesi: Done

rayokota (approved):
These changes LGTM.  

In the future, if we don't find a way to "compile" the string representation of the Protobuf schema, we can enhance SR to return the Protobuf schema in serialized form.  I'll be looking at this when I add CSFLE support for the Python client.  So let's not remove the `use.latest.version` config for Protobuf.

pranavrth (changes requested):
Thanks for fixing this. LGTM!

pranavrth (approved):

pranavrth:
@rayokota - You need to sign the CLA for the PR to get merged.

rayokota (approved):

rayokota:

PR #1721, opened 2024-03-09
Update registry rest client to support bearer authentication

gtato:
This change allows schema registry clients to specify a token in case the registry server supports token/oidc authentication.


iamed2:
When using Bearer auth from curl, we needed to provide additional headers for `Confluent-Identity-Pool-Id` and `target-sr-cluster`. Have you succeeded in using this branch to connect?

Additionally, needing to recreate the client whenever a token refresh is needed is cumbersome. Inserting a user-provided token-fetching callback as part of a custom `auth` for the Requests Session may be a more user-friendly option. A custom auth could look something like:

```python
class BearerAuth(requests.auth.AuthBase):
    def __init__(self, token_cb):
        self.token_cb = token_cb

    def __call__(self, request):
        """Called directly by requests to add the Authorization header"""
        token = token_cb()
        request.headers["Authorization"] = f"Bearer {token}"
        return request
```
and you can hook this in by setting `self.session.auth = BearerAuth(token_cb)`.

PR #1735, opened 2024-04-25
[KIP-848] Added support for testing with new 'consumer' group protocol.

pranavrth:
* Added support for testing with new 'consumer' group protocol.
* Fixed a incorrect testcase.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -8,7 +8,7 @@ execution_time_limit:
 global_job_config:
   env_vars:
     - name: LIBRDKAFKA_VERSION
-      value: v2.5.3-RC1
+      value: v2.5.3
   prologue:
     commands:
       - checkout
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,7 +35,7 @@ copyright = u'2016-2024, Confluent Inc.'
 # built documents.
 #
 # The short X.Y version.
-version = '2.5.3rc1'
+version = '2.5.3'
 # The full version, including alpha/beta/rc tags.
 release = version
 ######################################################################
--- a/examples/docker/Dockerfile.alpine
+++ b/examples/docker/Dockerfile.alpine
@@ -30,7 +30,7 @@ FROM alpine:3.12
 
 COPY . /usr/src/confluent-kafka-python
 
-ENV LIBRDKAFKA_VERSION v2.5.3-RC1
+ENV LIBRDKAFKA_VERSION v2.5.3
 ENV KAFKACAT_VERSION master
 
 
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ trove_classifiers = [
 setup(name='confluent-kafka',
       # Make sure to bump CFL_VERSION* in confluent_kafka/src/confluent_kafka.h
       # and version in docs/conf.py.
-      version='2.5.3rc1',
+      version='2.5.3',
       description='Confluent\'s Python client for Apache Kafka',
       author='Confluent Inc',
       author_email='support@confluent.io',
--- a/src/confluent_kafka/src/confluent_kafka.h
+++ b/src/confluent_kafka/src/confluent_kafka.h
@@ -43,7 +43,7 @@
  *  MM=major, mm=minor, RR=revision, PP=patchlevel (not used)
  */
 #define CFL_VERSION     0x02050300
-#define CFL_VERSION_STR "2.5.3rc1"
+#define CFL_VERSION_STR "2.5.3"
 
 /**
  * Minimum required librdkafka version. This is checked both during
--- a/tools/smoketest.sh
+++ b/tools/smoketest.sh
@@ -30,7 +30,7 @@ fi
 pyvers_tested=
 
 # Run tests with python3
-for py in 3.8 ; do
+for py in 3.9 ; do
     echo "$0: # Smoketest with Python$py"
 
     if ! python$py -V ; then
--- a/tools/test-manylinux.sh
+++ b/tools/test-manylinux.sh
@@ -31,12 +31,16 @@ fi
 echo "$0 running from $(pwd)"
 
 function setup_ubuntu {
+    export DEBIAN_FRONTEND=noninteractive
     # Ubuntu container setup
     apt-get update
-    apt-get install -y python3.8 curl
+    apt-get install -y -q software-properties-common
+    add-apt-repository ppa:deadsnakes/ppa
     # python3-distutils is required on Ubuntu 18.04 and later but does
     # not exist on 14.04.
-    apt-get install -y python3.8-distutils || true
+    apt-get install -y -q python3.9
+    apt-get install -y -q python3.9-distutils
+    apt-get install -y -q curl
 }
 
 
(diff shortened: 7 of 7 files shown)


tests/integration/cluster_fixture.py
```diff
@@ -20,10 +20,10 @@
 
 from trivup.clusters.KafkaCluster import KafkaCluster
 
-from confluent_kafka import Consumer, Producer, DeserializingConsumer, \
-    SerializingProducer
+from confluent_kafka import Producer, SerializingProducer
 from confluent_kafka.admin import AdminClient, NewTopic
 from confluent_kafka.schema_registry.schema_registry_client import SchemaRegistryClient
+from ..common import TestDeserializingConsumer, TestConsumer
```
emasab, line 26: Let's use absolute imports: `tests.common`

tests/integration/consumer/test_consumer_topicpartition_metadata.py
```diff
@@ -29,11 +29,11 @@ def commit_and_check(consumer, topic, metadata):
     assert offsets[0].metadata == metadata
 
 
-def test_consumer_topicpartition_metadata(kafka_cluster):
-    topic = kafka_cluster.create_topic("test_topicpartition")
+def test_consumer_topicpartition_metadata(kafka_single_broker_cluster):
+    topic = kafka_single_broker_cluster.create_topic("test_topicpartition")
```
emasab, line 33: To make integration tests for admin APIs work with KRaft we need to add some sleeps after a number of operations:
`time.sleep(1)`
after:
- create_topic
- create_acls
- delete_acls
- alter_configs
- incremental_alter_configs
- alter_user_scram_credentials

emasab: In `test_describe_operations `we can disable this test with `consumer`, as it returns it's a simple consumer group (needs to use the new RPC instead)
```Python
    if not TestUtils.use_group_protocol_consumer():
        verify_describe_groups(sasl_cluster, admin_client, our_topic)
```

emasab: In `test_list_offsets` we need to disable this check at the moment
```python
        # FIXME: This test is disabled until KAFKA-16310 is released
        #        with AK 3.8.0
        # requests = {topic_partition: OffsetSpec.max_timestamp()}
        # futmap = admin_client.list_offsets(requests, **kwargs)
        # for _, fut in futmap.items():
        #     result = fut.result()
        #     assert isinstance(result, ListOffsetsResultInfo)
        #     assert (result.offset == 1)
```

emasab: In `test_user_scram_credentials` with KRaft it's giving an error if more than one `UserScramCredentialUpsertion` is given for the same user, even if it's for a different SCRAM mechanism, so you can add an _if_ an do the upsertion in two steps when using kraft

tests/integration/conftest.py
```diff
@@ -26,26 +26,34 @@
...
 def create_trivup_cluster(conf={}):
     trivup_fixture_conf = {'with_sr': True,
                            'debug': True,
-                           'cp_version': '7.4.0',
-                           'version': '3.4.0',
-                           'broker_conf': ['transaction.state.log.replication.factor=1',
-                                           'transaction.state.log.min.isr=1']}
+                           'cp_version': '7.6.0',
+                           'kraft': TestUtils.use_kraft(),
+                           'version': 'trunk@f6c9feea76d01a46319b0ca602d70aa855057b07',
```
emasab, line 42: version needs to be trunk only with new protocol, otherwise 3.7.0
```python
def _broker_version():
    return (TestUtils.use_group_protocol_consumer() and
            'trunk@f6c9feea76d01a46319b0ca602d70aa855057b07' or
            '3.7.0')
```

tools/source-package-verification.sh
```diff
@@ -20,7 +20,11 @@ python setup.py build && python setup.py install
 if [[ $OS_NAME == linux && $ARCH == x64 ]]; then
     flake8 --exclude ./_venv,*_pb2.py
     make docs
-    python -m pytest --timeout 1200 --ignore=dest
+    if [[ $TEST_CONSUMER_GROUP_PROTOCOL == consumer ]]; then
+        python -m pytest --timeout 1200 --ignore=dest --ignore=tests/integration/admin
+    else
+        python -m pytest --timeout 1200 --ignore=dest
+    fi
```
emasab, line 27: Let's run the integration tests after previous changes:
```suggestion
    if [[ -z $TEST_CONSUMER_GROUP_PROTOCOL ]]; then
        flake8 --exclude ./_venv,*_pb2.py
        make docs
    fi
    python -m pytest --timeout 1200 --ignore=dest
```

emasab (changes requested):
Great, Pranav! Just a few changes more and we can enable the integration tests too

tests/integration/consumer/test_consumer_topicpartition_metadata.py
pranavrth: > In test_list_offsets we need to disable this check at the moment

This test is passing. Why do we need to disable this test?

>In test_user_scram_credentials with KRaft it's giving an error if more than one UserScramCredentialUpsertion is given for the same user, even if it's for a different SCRAM mechanism, so you can add an if an do the upsertion in two steps when using kraft

Why is it behaving differently in KRaft and Zookeeper? This is a bug in Kafka I think.

tests/integration/consumer/test_consumer_topicpartition_metadata.py
```diff
@@ -29,11 +29,11 @@ def commit_and_check(consumer, topic, metadata):
     assert offsets[0].metadata == metadata
 
 
-def test_consumer_topicpartition_metadata(kafka_cluster):
-    topic = kafka_cluster.create_topic("test_topicpartition")
+def test_consumer_topicpartition_metadata(kafka_single_broker_cluster):
```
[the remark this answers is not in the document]
pranavrth: I think I didn't understand this correctly. I tried with `session` and it was working as it is.

joaoe:
@pranavrth @emasab Hi. Can you guys explain why no one is maintaining this project and accepting MRs ? Thre's now over 75 open PRs, many to fix bugs. Thanks.

emasab (changes requested):

emasab (changes requested):
Some comments to improve tests reliability

tests/integration/schema_registry/test_proto_serializers.py
```diff
@@ -47,19 +47,19 @@
...
 ])
-def test_protobuf_message_serialization(kafka_cluster, pb2, data):
+def test_protobuf_message_serialization(kafka_single_broker_cluster, pb2, data):
     """
     Validates that we get the same message back that we put in.
 
     """
-    topic = kafka_cluster.create_topic("serialization-proto")
-    sr = kafka_cluster.schema_registry()
+    topic = kafka_single_broker_cluster.create_topic("serialization-proto")
```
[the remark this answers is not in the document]
pranavrth: Instead of this, I have created a new function `create_topic_and_wait_propogation` which can be used directly.

tests/integration/admin/test_list_offsets.py
```diff
@@ -62,12 +66,13 @@ def test_list_offsets(kafka_cluster):
...
-            result = fut.result()
-            assert isinstance(result, ListOffsetsResultInfo)
-            assert (result.offset == 1)
+        if TestUtils.use_kraft():
+            requests = {topic_partition: OffsetSpec.max_timestamp()}
+            futmap = admin_client.list_offsets(requests, **kwargs)
+            for _, fut in futmap.items():
+                result = fut.result()
+                assert isinstance(result, ListOffsetsResultInfo)
+                assert (result.offset == 1)
```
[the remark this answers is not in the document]
pranavrth: Enabled and working fine with 3.8.0

PR #1743, opened 2024-05-08
Added script to install confluent_kafka module and do a sanity test with multiple python versions.

pranavrth:


tools/test-released-wheels.sh
```diff
@@ -0,0 +1,54 @@
...
+    echo "Uninstalling confluent_kafka if already installed"
+    pip uninstall -y confluent_kafka > /dev/null 2>&1 || true
+
+    if [ "$2" = "test" ]; then
+        echo "Installing confluent_kafka from test PyPI"
+        pip install --index-url https://test.pypi.org/simple/ confluent_kafka==$1 > /dev/null 2>&1
+    else
+        echo "Installing confluent_kafka"
+        pip install confluent_kafka==$1 > /dev/null 2>&1
+    fi
```
anchitj, line 46: Any reason we are redirecting to `/dev/null` for all the commands? It can be removed for better debugging.

tools/test-released-wheels.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+set -e
+
+# Check if an argument is present
+if [ -z "$1" ]; then
+    echo "Please provide a version of confluent_kafka to install."
+    exit 1
+fi
+
```
anchitj, line 10: We should move $1 and $2 to params for better readability and use them instead
```
CONFLUENT_KAFKA_VERSION=$1
USE_TEST_PYPI=$2
```

tools/test-released-wheels.sh
```diff
@@ -0,0 +1,54 @@
...
+    pyenv install -s $version > /dev/null 2>&1
+
+    echo Using $version
+    pyenv shell $version > /dev/null 2>&1
+
+    echo Python version: `python -V`
+    echo Pip version: `pip -V`
+
+    echo "Uninstalling confluent_kafka if already installed"
+    pip uninstall -y confluent_kafka > /dev/null 2>&1 || true
```
anchitj, line 38: Should we add `venv` here to not make changes in the global space?

anchitj (changes requested):
Some comments

PR #1744, opened 2024-05-08
DGS-10484 Improve SR caching on Python client

kcorman0:
Adding to some SR calls which lack caching but could benefit from it


rayokota (approved):
Thanks @kcorman0 , LGTM

emasab (approved):
LGTM!

PR #1758, opened 2024-06-03
Logger with Admin Client

pranavrth:
- Fixed `logger` not working when provided as an argument to `AdminClient`
 - Updated `examples/adminapi.py` to include usage of the custom logger with `AdminClient`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ v2.4.1 is a maintenance release with the following fixes and enhancements:
  - Fixed invalid write in `OAUTHBEARER/OIDC` extensions copy (#1745)
  - Fixed an issue related to import error of `TopicCollection` and `TopicPartitionInfo` classes when importing through other module like mypy.
  - Fixed a segfault when `commit` or `store_offsets` consumer method is called incorrectly with errored Message object
- - Fixed `logger` not working when provided as an argument to `AdminClient`
+ - Fixed `logger` not working when provided as an argument to `AdminClient` instead of a configuration property
 
 confluent-kafka-python is based on librdkafka v2.4.1, see the
 [librdkafka release notes](https://github.com/confluentinc/librdkafka/releases/tag/v2.4.1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
 
 v2.4.1 is a maintenance release with the following fixes and enhancements:
 
+ - Added an example to show the usage of the custom logger with `AdminClient` 
  - Removed usage of `strcpy` to enhance security of the client (#1745)
  - Fixed invalid write in `OAUTHBEARER/OIDC` extensions copy (#1745)
  - Fixed an issue related to import error of `TopicCollection` and `TopicPartitionInfo` classes when importing through other module like mypy.
  - Fixed a segfault when `commit` or `store_offsets` consumer method is called incorrectly with errored Message object
+ - Fixed `logger` not working when provided as an argument to `AdminClient`
 
 confluent-kafka-python is based on librdkafka v2.4.1, see the
 [librdkafka release notes](https://github.com/confluentinc/librdkafka/releases/tag/v2.4.1)
--- a/examples/adminapi.py
+++ b/examples/adminapi.py
@@ -890,16 +890,8 @@ if __name__ == '__main__':
     operation = sys.argv[2]
     args = sys.argv[3:]
 
-    # Custom logger
-    logger = logging.getLogger('AdminClient')
-    logger.setLevel(logging.INFO)
-    handler = logging.StreamHandler()
-    handler.setFormatter(logging.Formatter('%(asctime)-15s %(levelname)-8s %(message)s'))
-    logger.addHandler(handler)
-
     # Create Admin client
-    a = AdminClient({'bootstrap.servers': broker},
-                    logger=logger)
+    a = AdminClient({'bootstrap.servers': broker})
 
     opsmap = {'create_topics': example_create_topics,
               'delete_topics': example_delete_topics,
@@ -928,6 +920,3 @@ if __name__ == '__main__':
         sys.exit(1)
 
     opsmap[operation](a, args)
-
-    # Log messages through custom logger if provided
-    a.poll(0)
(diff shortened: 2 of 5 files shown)


anchitj (commented):
Minor comments

examples/adminapi.py
```diff
@@ -890,8 +890,16 @@ def example_list_offsets(a, args):
...
+    # Custom logger
+    logger = logging.getLogger('AdminClient')
+    logger.setLevel(logging.INFO)
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter('%(asctime)-15s %(levelname)-8s %(message)s'))
+    logger.addHandler(handler)
+
     # Create Admin client
-    a = AdminClient({'bootstrap.servers': broker})
+    a = AdminClient({'bootstrap.servers': broker},
```
emasab, line 901: logger should be a property like in the producer and consumer, not an additional parameter
`'logger': logger,`

emasab (changes requested):

examples/adminapi.py
pranavrth: In python, its an additional parameter as well.

emasab: right, I delete it so I don't request this change

src/confluent_kafka/admin/__init__.py
```diff
@@ -109,16 +109,21 @@ class AdminClient (_AdminClientImpl):
...
         The AdminClient is a standard Kafka protocol client, supporting
         the standard librdkafka configuration properties as specified at
         https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
 
+        :param dict config: Configuration properties. At a minimum ``bootstrap.servers`` **should** be set\n"
+        :param Logger logger: Optional Logger instance to use as a custom log messages handler.
+
         At least 'bootstrap.servers' should be configured.
         """
+        if logger is not None:
```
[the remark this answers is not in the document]
pranavrth: We are not doing this anywhere else. Let's not do here as well.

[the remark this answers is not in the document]
emasab: This is checked later by librdkafka, whenever possible we delegate the validation there

src/confluent_kafka/admin/__init__.py
```diff
@@ -109,16 +109,19 @@ class AdminClient (_AdminClientImpl):
...
-        https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+        https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md
 
-        At least 'bootstrap.servers' should be configured.
+        :param dict conf: Configuration properties. At a minimum ``bootstrap.servers`` **should** be set\n"
+        :param Logger logger: Optional Logger instance to use as a custom log messages handler.
         """
+        if logger is not None:
+            conf['logger'] = logger
         super(AdminClient, self).__init__(conf)
```
emasab, line 125: You can just pass the logger here without changing the conf
```suggestion
        super(AdminClient, self).__init__(conf, logger=logger)
```

src/confluent_kafka/admin/__init__.py
```diff
@@ -109,16 +109,19 @@ class AdminClient (_AdminClientImpl):
     Requires broker version v0.11.0.0 or later.
     """
 
-    def __init__(self, conf):
+    def __init__(self, conf, logger=None):
```
emasab, line 112:
```suggestion
    def __init__(self, *args, **kwargs):
```

emasab (changes requested):

src/confluent_kafka/admin/__init__.py
pranavrth: This doesn't work in case the logger is `None`. I tried this only first.

In general, if the logger is provided in conf as well as in the parameter, parameter gets more priority. By doing this, the logger is always None if it is provided in conf.

src/confluent_kafka/admin/__init__.py
emasab, line 112:
```suggestion
    def __init__(self, conf, **args):
```

emasab (changes requested):

src/confluent_kafka/admin/__init__.py
emasab: Yes, was checking this an updated my comment

pranavrth:
> This is checked later by librdkafka, whenever possible we delegate the validation there

There is inbetween manipulation of the config properties in python binding as well. Due to that part, I have to check here first.

https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/src/confluent_kafka.c#L2215

src/confluent_kafka/admin/__init__.py
pranavrth: This will start taking all the properties in the conf as parameter. I don't know if we should support this.

pranavrth: I was thinking of only supporting logger property like this.

emasab: That's what happens with the Producer and Consumer too

examples/adminapi_logger.py
```diff
@@ -0,0 +1,47 @@
...
+logger = logging.getLogger('AdminClient')
+logger.setLevel(logging.DEBUG)
+handler = logging.StreamHandler()
+handler.setFormatter(logging.Formatter('%(asctime)-15s %(levelname)-8s %(message)s'))
+logger.addHandler(handler)
+
+# Create Admin client
+a = AdminClient({'bootstrap.servers': broker,
+                 'debug': 'all'},
+                logger=logger)
```
emasab, line 22:
```suggestion
                logger=logger)
# Alternatively, pass the logger as a key.
# When passing it as an argument, it overwrites the key.
#
# a = AdminClient({'bootstrap.servers': broker,
#                  'debug': 'all',
#                  'logger': logger})
```

CHANGELOG.md
```diff
@@ -4,10 +4,12 @@
 
 v2.4.1 is a maintenance release with the following fixes and enhancements:
```
emasab, line 5: Version can be changed to 2.5.0  when we merge `delete_records`

emasab (changes requested):
Thanks, just a few comments

CHANGELOG.md
pranavrth: Already done that in delete_records as that is a new feature.

emasab (approved):
Thanks Pranav!

PR #1771, opened 2024-07-01
Remove CentOS 6 and 7 and add CentOS 8 (built with AlmaLinux 8)

emasab:


pranavrth:
Please fix the build and add a changelog for the same.

PR #1790, opened 2024-08-05
Adding int ID for Transactional ID Resource Type

JohnPreston:
Works to build with https://github.com/confluentinc/librdkafka/pull/4803
As librdkafka in this case is installed in `/usr/local/lib` the build and install is done using

```shell
C_INCLUDE_PATH=/usr/local/lib LIBRARY_PATH=/usr/local/lib LD_LIBRARY_PATH=/usr/local/lib python setup.py build
C_INCLUDE_PATH=/usr/local/lib LIBRARY_PATH=/usr/local/lib LD_LIBRARY_PATH=/usr/local/lib python setup.py install
```

Tested with 2 different python projects using the locally compiled libs, and it works for both listing the ACLs and setting them.

### DIFF
--- a/src/confluent_kafka/admin/__init__.py
+++ b/src/confluent_kafka/admin/__init__.py
@@ -75,6 +75,7 @@ from ..cimpl import (KafkaException,  # noqa: F401
                      RESOURCE_TOPIC,
                      RESOURCE_GROUP,
                      RESOURCE_BROKER,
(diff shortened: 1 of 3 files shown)


pranavrth:
Created internal PR for the same - https://github.com/confluentinc/confluent-kafka-python/pull/1834

PR #1798, opened 2024-08-14
Retrieve schema from schema registry client in `JSONDeserialier` if `schema_str` not provided

### LINKED ISSUE #1541: JSON deserialization with a schema from schema registry.
Description
===========
Avro and Protobuf deserializers accept schema registry client to fetch appropriate schema from registry. JSON deserializer defines `schema_str` parameter as required and use its value to validate payload.
https://github.com/confluentinc/confluent-kafka-python/blob/f3055be4d2aa18ffc833cc88f7b08b173d4da433/src/confluent_kafka/schema_registry/json_schema.py#L269

https://github.com/confluentinc/confluent-kafka-python/blob/f3055be4d2aa18ffc833cc88f7b08b173d4da433/src/confluent_kafka/schema_registry/json_schema.py#L316
So where should this schema come from? Should it be fetched from registry by user code and supplied to `JSONDeserializer`? If so what's the purpose of having JSON serializer register schemas?

How to reproduce
================
\-

Checklist
=========

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`): 1.9.2 & 1.9.2
 - [x] Apache Kafka broker version: 3.3.1
 - [ ] Operating system: 
 - [ ] Critical issue

kleinschmidt:
Fixes #1541.

I think this is worth including because it brings the `JSONDeserializer` into better alignment with the avro/protobuf deserializers.  Currently the JSON deserializer just throws away the schema ID information, yet still _requires_ that the schema string is passed in as input.  This makes it pretty annoying to work with on the client side (you must either hard code your schema string into your application or look up the schema from the registry by ID which also needs to be hard coded or subject which requires hard coding the subject naming convention).

Thanks to the recent improvements in caching in the schema registry client this won't result in excessive network requests.  There's a bit of optimization on the table here (could cache the `json.loads` for the schema ID) but I think this is probably fine to get started with.

I haven't attempted to update tests here at all, just the source code and docstrings.

### DIFF
--- a/src/confluent_kafka/schema_registry/json_schema.py
+++ b/src/confluent_kafka/schema_registry/json_schema.py
@@ -296,16 +296,18 @@ class JSONDeserializer(Deserializer):
     framing.
 
     Args:
-        schema_str (str, Schema):
+        schema_str (str, Schema, optional):
             `JSON schema definition <https://json-schema.org/understanding-json-schema/reference/generic.html>`_
             Accepts schema as either a string or a :py:class:`Schema` instance.
             Note that string definitions cannot reference other schemas. For referencing other schemas,
-            use a :py:class:`Schema` instance.
+            use a :py:class:`Schema` instance.  If not provided, schemas will be
+            retrieved from schema_registry_client based on the schema ID in the
+            wire header of each message.
 
         from_dict (callable, optional): Callable(dict, SerializationContext) -> object.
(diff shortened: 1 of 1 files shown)


rayokota (approved):
Thanks @kleinschmidt , LGTM

PR #1822, opened 2024-09-26
Fix Basic Auth credentials parsing in SchemaRegistryClient

goshaQ:
### Description:

Hello, I noticed that `SchemaRegistryClient` does not handle a special character (colon) in passwords correctly, which should be allowed according to [RFC-7617](https://www.rfc-editor.org/rfc/rfc7617.html):

> For the user-id, recipients MUST support all characters defined in
>  the "UsernameCasePreserved" profile defined in Section 3.3 of
>  [RFC7613], with the exception of the colon (":") character.
>
>   For the password, recipients MUST support all characters defined in
>   the "OpaqueString" profile defined in Section 4.2 of [RFC7613].

### Changes:

- Updated the parsing logic for `basic.auth.user.info` to split on the first occurrence of a colon, ensuring the username and password are separated correctly, even when colons are present in the password.

### DIFF
--- a/src/confluent_kafka/schema_registry/schema_registry_client.py
+++ b/src/confluent_kafka/schema_registry/schema_registry_client.py
@@ -105,7 +105,7 @@ class _RestClient(object):
                                  " remove basic.auth.user.info from the"
                                  " configuration")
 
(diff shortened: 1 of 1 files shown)


rayokota (approved):
Thanks @goshaQ , LGTM

PR #1591, opened 2023-06-19
Many stability fixes with Handle objects

joaoe:
Fixes https://github.com/confluentinc/confluent-kafka-python/issues/1590 which is about the Producer class
Also realized that the same error affected the AdminClient class, so also fixed that.
And then cleaned up the Consumer code to use the same internal API.

PR #1823, opened 2024-09-26
Admin client impl fixes

siobhan-sterrett:
A handful of minor fixes to cimpl. Probably nothing that will impact real-life use-cases, since these are all buried inside `_AdminClientImpl`, but they were bothering me.

1. Swaps `"topic_names"` and `"future"` in `kws` array in `Admin_describe_topics()`

If a user tried to pass `topic_names` or `future` into `_AdminClientImpl.describe_topics()` as keyword arguments, they would end up swapped, which would lead to confusing errors.

2. Raises `TypeError` in `Admin_list_consumer_group_offset()` if `request` is not a list
3. Raises `TypeError` in `Admin_alter_consumer_group_offsets()` if `request` is not a list

In both of these cases, the library would raise `ValueError` if `request` was a list with anything other than 1 item... but if `request` wasn't a list, it would result in an internal error when it tried to access its first item.

4. Allows elements of `plist` in `py_to_c_parts()` to be instances of a subtype of `TopicPartition`

If a user passed a list of instances of a subtype of `TopicPartition`, the library would raise a `TypeError`, even if the instances were otherwise valid.

### DIFF
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -2794,10 +2794,15 @@ PyObject *Admin_alter_consumer_group_offsets (Handle *self, PyObject *args, PyOb
          * admin operation is finished, so we need to keep our own refcount. */
         Py_INCREF(future);
 
-        if (PyList_Check(request) &&
-            (requests_cnt = (int)PyList_Size(request)) != 1) {
-                PyErr_SetString(PyExc_ValueError,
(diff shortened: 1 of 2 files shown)


joaoe:
I think this might touch some code I changed here 
https://github.com/confluentinc/confluent-kafka-python/pull/1591

But this project is not accepting PRs. It's annoying.

PR #1825, opened 2024-10-01
fix: update schema registry cache after deleting subject

### LINKED ISSUE #1824: Bug: Schema Registry Cache not updated after deletion
Description
===========
When using `delete_subject()`, the schema cache should be updated but instead it still contains the deleted subject and related schemas, as a result when registering new schema with the same subject name using `register_schema()` 

```py
schema_id = self._cache.get_schema_id_by_subject(subject_name, schema)
if schema_id is not None:
     return schema_id
```
returns deleted schema_id

0xn4utilus:
Closes #1824


rayokota (approved):
Thanks @0xn4utilus , LGTM

PR #1830, opened 2024-10-09
[KIP-848 EA] Admin API for listing consumer groups now has

emasab:
an optional filter to return only groups of given types


pranavrth (approved):
LGTM!

pranavrth (approved):
LGTM!.

PR #1833, opened 2024-10-10
[KIP-460] ElectLeaders implementation public API changes and documentation improvement

emasab:
- ElectionType in confluent_kafka as in Java
- dictionary value is a KafkaError as it's not thrown
- ElectionType should be added to documentation .rst
- docstring changes
- tests without brokers with successful validation that leads to a timeout
- example elects leaders for all partitions in case no partition is specified


pranavrth (changes requested):
Just a minor nit.

pranavrth (approved):
LGTM!.

PR #1891, opened 2025-01-07
Changes for v2.8.0 release

pranavrth:


.semaphore/semaphore.yml
```diff
@@ -8,28 +8,28 @@ execution_time_limit:
...
+  #     jobs:
+  #       - name: Test
+  #         commands:
+  #           - sem-version python 3.9
+  #           - pip install tox
+  #           - tox -e cover
+  #           - mkdir test-output
+  #           - cp test-report.xml test-output
+  #           - test-results publish test-output
+  #           - artifact push workflow coverage.xml
```
anchitj, line 32: Any reason why are we commenting them instead of removing?

PR #1895, opened 2025-01-08
Keep previous (setup.py) schema registry optional dependency name for compatibility

emasab:
In rare cases this dependency was used alone, but its dependencies were also included in avro, json or protobuf extras, so this is a minor fix.
Optional dependencies cannot be named with a hyphen so when used in requirements.txt files or pip install they will be replaced with an underscore.

What
----

Checklist
------------------

Test & Review
------------
  - it can be tested by running `pip install .[schema-registry]` as happened in previous versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ include-package-data = false
 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements/requirements.txt"]}
 optional-dependencies.schemaregistry = { file = ["requirements/requirements-schemaregistry.txt"] }
+optional-dependencies.schema_registry = { file = ["requirements/requirements-schemaregistry.txt"] }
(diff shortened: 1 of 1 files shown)


pranavrth (approved):
LGTM!.

pyproject.toml
```diff
@@ -30,6 +30,7 @@ include-package-data = false
 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements/requirements.txt"]}
 optional-dependencies.schemaregistry = { file = ["requirements/requirements-schemaregistry.txt"] }
+optional-dependencies.schema_registry = { file = ["requirements/requirements-schemaregistry.txt"] }
```
pranavrth, line 33: Will this work with "schema-registry".

pranavrth (changes requested):

pyproject.toml
pranavrth: I have checked and its working.

pranavrth (approved):
LGTM!.

PR #1900, opened 2025-01-09
Migrate from setuptools to poetry

rayokota:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


pyproject.toml
```diff
@@ -1,6 +1,62 @@
 [build-system]
-requires = [ "setuptools>=62", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = [ "poetry-core>=2.0.0,<3.0.0", "wheel"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+packages = [{include = "confluent_kafka", from = "src"}]
+
+[tool.poetry.dependencies]
```
pranavrth, line 8: We are moving optional dependencies to required ones. Most of the users of this client don't use schema registry and we shouldn't install those dependencies by default.

PR #1911, opened 2025-01-24
docs: README.md

peppedilillo:
Corrects pip install syntax for package plus dependencies.


rayokota (approved):
Thanks @peppedilillo , LGTM

PR #1963, opened 2025-04-08
Fixed alpine based docker container creation script

pranavrth:


examples/docker/Dockerfile.alpine
```diff
@@ -30,13 +30,14 @@ FROM alpine:3.12
 
 COPY . /usr/src/confluent-kafka-python
 
-ENV LIBRDKAFKA_VERSION v2.8.0
-ENV KAFKACAT_VERSION master
+ENV LIBRDKAFKA_VERSION="v2.8.0"
+ENV KCAT_VERSION="master"
+ENV CKP_VERSION="master"
```
milindl, line 35: We can remove this if not needed

milindl (approved):
Approved with nit

examples/docker/Dockerfile.alpine
pranavrth: Will do it in another PR.

PR #1965, opened 2025-04-10
Implement Async Schema Registry client

rohitsanj:
## What

- Converted the existing sync Schema Registry clients to be made async and placed them in the `_async` module.
- The code was converted such that it could be made sync by simply applying some regular expression driven replace logic (like `await func(x, y, z)` -> `func(x, y, z)`, `AsyncXYZ` -> `XYZ` and so on). Going forward, all diffs must be solely made to the `_async` module, the sync clients will be derived from the async clients.
- The same treatment is applied to the tests, any new async test written must work with async keywords removed.

#1971 must be merged before this one.

### `unasync.py` script

Scenario 1: When changes are made to the _async directory but not to its _sync counterpart:

```bash
$ python tools/unasync.py --check

⚠️  Detected changes to a _sync directory that are uncommitted.

Files with differences:
  - src/confluent_kafka/schema_registry/_sync/serde.py

Please either:
1. Commit the changes in the generated _sync files, or
2. Revert the changes in the original _async files
```

Scenario 2: When changes across _async and _sync are consistent.

```
$ python tools/unasync.py --check

✅ All _sync directories are up to date!
```

----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


tests/common/_async/consumer.py
rohitsanj, line 1: Async implementation of the Kafka consumer -- used only in tests and not surfaced as a public API.

tests/common/_async/producer.py
rohitsanj, line 1: Async implementation of the Kafka producer -- used only in tests and not surfaced as a public API.

rayokota (dismissed):
Thanks @rohitsanj , LGTM

tests/common/_async/consumer.py
```diff
@@ -0,0 +1,105 @@
...
+        return self
+
+    async def __anext__(self):
+        return await self.poll(None)
+
+    async def poll(self, poll_timeout: int = -1):
+        poll_timeout = None if poll_timeout == -1 else poll_timeout
+        async with timeout(poll_timeout):
+            while True:
+                # Zero timeout here is what makes it non-blocking
```
rohitsanj, line 55: Remove this because it's wrong -- `poll(0)` could block. 

Edit: Not pressing, can be handled later.

PR #1974, opened 2025-04-18
fix: Handled ctx=None for AvroDeserializer and ProtobufDeserializer

### LINKED ISSUE #1939: Handling of null ctx in ProtobufDeserializer
The `ctx` argument in `ProtobufDeserializer` is marked as Optional.

https://github.com/confluentinc/confluent-kafka-python/blob/3f40f1906ee4c0f9a6aec10be8cee47bc9072230/src/confluent_kafka/schema_registry/protobuf.py#L815

In previous version of the library, calling `ProtobufDeserializer.__call__` with `ctx=None` worked. With the new version it fails with an `AttributeError`, around [here](https://github.com/confluentinc/confluent-kafka-python/blob/3f40f1906ee4c0f9a6aec10be8cee47bc9072230/src/confluent_kafka/schema_registry/__init__.py#L73)

I think this was introduced by this recent change https://github.com/confluentinc/confluent-kafka-python/pull/1852

There is a work around, consisting of passing a`subject.name.strategy` that ignores the context:

```python
ProtobufDeserializer(xxx,  {"subject.name.strategy": lambda *_: None})
```

I would argue that the code should be able to handle a null ctx, without having to change the config..

In particular in the context of a `ProtobufDeserializer` that doesn't have a schema registry set (and is using schema on read). Because in this case, the subject is not used, and therefore the ctx is not needed.

Steps to reproduce:

```python
import dataclasses

import pytest
from confluent_kafka.schema_registry import RegisteredSchema, Schema
from confluent_kafka.schema_registry.protobuf import (
    ProtobufDeserializer,
    ProtobufSerializer,
)
from confluent_kafka.serialization import MessageField, SerializationContext
from google.protobuf.wrappers_pb2 import DoubleValue


@dataclasses.dataclass
class InMemorySchemaRegistryClient:
    """In memory schema registry, for test"""

    schemas: dict = dataclasses.field(default_factory=dict)

    def register_schema(self, subject_name, schema, *_, **__) -> int:
        try:
            return self.schemas[schema].schema_id
        except KeyError:
            schema_id = len(self.schemas)
            self.schemas[schema] = RegisteredSchema(
                schema_id=schema_id,
                schema=Schema(schema, "PROTOBUF", []),
                subject=subject_name,
                version=1,
            )
            return schema_id

    # noinspection PyUnusedLocal
    def lookup_schema(self, subject_name, schema):
        return self.schemas.get(schema, None)


def test_end_to_end_kafka():
    context = SerializationContext("test-topic-1", MessageField.VALUE)
    serializer = ProtobufSerializer(
        DoubleValue, InMemorySchemaRegistryClient(), {"use.deprecated.format": False}
    )
    deserializer = ProtobufDeserializer(DoubleValue, {"use.deprecated.format": False})

    msg_in = DoubleValue(value=3.14)
    kafka_data = serializer(msg_in, context)
    assert isinstance(kafka_data, bytes)
    proto_data = msg_in.SerializeToString()

    assert kafka_data[-len(proto_data) :] == proto_data
    assert len(kafka_data) - len(proto_data) == 6
    deserializer._msg_class().ParseFromString(proto_data)

    with pytest.raises(
        AttributeError, match="'NoneType' object has no attribute 'topic'"
    ):
        deserializer(kafka_data, ctx=None)

    msg_out = deserializer(
        kafka_data, ctx=SerializationContext("test-topic-1", MessageField.VALUE)
    )

    deserializer_no_subject_name = ProtobufDeserializer(
        DoubleValue,
        {"use.deprecated.format": False, "subject.name.strategy": lambda *_: None},
    )
    msg_out2 = deserializer_no_subject_name(kafka_data, ctx=None)
    assert msg_out == msg_out2

    assert isinstance(msg_out, DoubleValue)
    assert msg_out == msg_in

    kafka_data_back = serializer(msg_out, context)

    assert kafka_data_back == kafka_data


if __name__ == "__main__":
    test_end_to_end_kafka()

```

### LINKED ISSUE #1973: Handling of null ctx in AvroDeserializer
Similar issue reported in #1939 but for AvroDeserializer...

So when avro deserializer is called without (optional) parameter named `ctx` following Attribute Error is raised:

```txt
AttributeError: 'NoneType' object has no attribute 'topic'
```

in here:

```py
    def topic_subject_name_strategy(ctx, record_name: Optional[str]) -> Optional[str]:
        """
        Constructs a subject name in the form of {topic}-key|value.
    
        Args:
            ctx (SerializationContext): Metadata pertaining to the serialization
                operation.
    
            record_name (Optional[str]): Record name.
    
        """
>       return ctx.topic + "-" + ctx.field
E       AttributeError: 'NoneType' object has no attribute 'topic'
```

Steps to reproduce:

```py
schema_registry_client = SchemaRegistryClient(...)
local_schema = '''
{
  "type": "record",
  "name": "FooRecord",
  "namespace": "com.confluent_kafka.issue",
  "fields": [
    {
      "name": "foo",
      "type": ["string"]
    }
  ]
}
'''

deserializer = AvroDeserializer(schema_registry_client, local_schema)
msg = deserializer(b'\x00\x00\x00\x10\x00\bar')
```

and-ratajski:
Handled `None` value for optional parameter called `ctx` in both AvroDeserializer and ProtobufDeserializer.

closes #1939
closes #1973

----

The "optional" `ctx` parameter in `__call__` method in AvroDeserializer and ProtobufDeserializer isn't optional at all... Despite method annotation that specifies type of this parameter as `ctx: Optional[SerializationContext] = None` is was required in the method logic and `None` value wasn't handled. This misbehavior (compered to previous releases) was already noticed in the issues mentioned above.

This PR contains fix that handles None value and hence gives backward compatibility.

Checklist
------------------

References
----------
GH:
- #1939 
- #1973 

Open questions / Follow-ups
--------------------------

If this behavior is intended for some reason I would suggest:
- marking it as breaking change (e.g. version 3.0.0)
- changing type annotation for `ctx` parameter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Confluent's Python client for Apache Kafka
 
+## v2.10.1
+
+v2.10.0 is a fix release with the following fixes
+
+- Handled `None` value for optional `ctx` parameter in  `ProtobufDeserializer` (#1939)
+- Handled `None` value for optional `ctx` parameter in  `AvroDeserializer` (#1973)
+
 ## v2.10.0
 
 v2.10.0 is a feature release with the following fixes and enhancements:
--- a/src/confluent_kafka/schema_registry/avro.py
+++ b/src/confluent_kafka/schema_registry/avro.py
@@ -557,7 +557,7 @@ class AvroDeserializer(BaseDeserializer):
                                      "message was not produced with a Confluent "
                                      "Schema Registry serializer".format(len(data)))
 
-        subject = self._subject_name_func(ctx, None)
+        subject = self._subject_name_func(ctx, None) if ctx else None
         latest_schema = None
         if subject is not None:
             latest_schema = self._get_reader_schema(subject)
@@ -573,7 +573,7 @@ class AvroDeserializer(BaseDeserializer):
             writer_schema = self._get_parsed_schema(writer_schema_raw)
 
             if subject is None:
-                subject = self._subject_name_func(ctx, writer_schema.get("name"))
+                subject = self._subject_name_func(ctx, writer_schema.get("name")) if ctx else None
                 if subject is not None:
                     latest_schema = self._get_reader_schema(subject)
 
--- a/src/confluent_kafka/schema_registry/protobuf.py
+++ b/src/confluent_kafka/schema_registry/protobuf.py
@@ -565,14 +565,15 @@ class ProtobufSerializer(BaseSerializer):
             raise ValueError("message must be of type {} not {}"
                              .format(self._msg_class, type(message)))
 
-        subject = self._subject_name_func(ctx,
-                                          message.DESCRIPTOR.full_name)
-        latest_schema = self._get_reader_schema(subject, fmt='serialized')
+        subject = self._subject_name_func(ctx, message.DESCRIPTOR.full_name) if ctx else None
+        latest_schema = None
+        if subject is not None:
+            latest_schema = self._get_reader_schema(subject, fmt='serialized')
+
         if latest_schema is not None:
             self._schema_id = latest_schema.schema_id
-        elif subject not in self._known_subjects:
-            references = self._resolve_dependencies(
-                ctx, message.DESCRIPTOR.file)
+        elif subject not in self._known_subjects and ctx is not None:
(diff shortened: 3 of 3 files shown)


rayokota (approved):
Thanks @and-ratajski , LGTM

PR #2000, opened 2025-06-24
Fixed non-tool path prefix file names for python conventions

MSeal:
A minor fix to project files to not have multiple conventions.
----

A few file renames.

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

I originally changed the files in `tools`, but reverted that as they are treated like bash scripts in that folder and all the command line instructions use the same `-` separators. It felt better to keep it consistent with the `.sh` files in that folder.


emasab (approved):
Thanks for the change

Issue #1857, opened 2024-11-21
License metadata issues


Kyle-Neale:
Description
===========
We have license validations we perform on dependencies we're shipping and `confluent-kafka-python` has changed. The license metadata looks bungled going from 2.6.0 to 2.6.1.
- https://pypi.org/project/confluent-kafka/2.6.0/
- https://pypi.org/project/confluent-kafka/2.6.1/

![image](https://github.com/user-attachments/assets/a5e55bb6-7ff2-42da-bb30-279f9861e748)

How to reproduce
================
N/A

Checklist
=========
Please provide the following information:

 - [x] confluent-kafka-python and librdkafka version (`confluent_kafka.version()` and `confluent_kafka.libversion()`):

pranavrth:
I have fixed this issue with [v2.8.0](https://pypi.org/project/confluent-kafka/2.8.0/). Please check.

PR #2001, opened 2025-06-25
Manifest tweak

MSeal:
Moved license to pyproject.toml, and removed redundant README from Manifest as it's already in the toml definition.
----

See above

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "confluent-kafka"
+license = {file = "LICENSE"}
 version = "2.10.1"
 description = "Confluent's Python client for Apache Kafka"
 classifiers = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -4,6 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "confluent-kafka"
+license = {file = "LICENSE"}
```
pranavrth, line 7: There was an [issue](https://github.com/confluentinc/confluent-kafka-python/issues/1857) with LICENSE here. I fixed it by moving to MANIFEST.in file [earlier](https://github.com/confluentinc/confluent-kafka-python/pull/1891/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L17).

MSeal: Hmm that's odd, since that's where it _should_ be according to [docs](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files)

PR #2004, opened 2025-07-01
remove explicit calls to pyenv

cchristous:
What
----

This changes the Semaphore pipeline to no longer contain explicit calls to `pyenv`. 

This approach will enable an upcoming change to Semaphore agents to replace `pyenv` usage with `mise`, which is an analogous tool that supports many different languages.

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


emasab (approved):
Approving for @confluentinc/clients after @MSeal review

PR #2019, opened 2025-08-12
Fix error propagation rule for Python's C API

k-raina:
What
----

This PR fixes issue : https://github.com/confluentinc/confluent-kafka-python/issues/865

Couple of Python C voilation rule we want to understand: 
```
The Python C API Exception Rule, When a C function returns to Python:
1. If the function returns NULL (indicating an error), there MUST be a Python exception set
2. If the function returns a non-NULL value (success), there MUST NOT be a Python exception set
Violating this rule causes SystemError.
```

```
 C function should own the exception it sets when returning NULL.
```

**Previous Workflow (Broken)**
1. Application calls consumer.consume() → Python → C extension (Consumer_consume) → librdkafka
2. librdkafka triggers callback → error_cb() creates new Python execution context
4. Python callback raises exception → Exception set in Python global state within callback context
5. Return to librdkafka → librdkafka doesn't understand Python exceptions, continues normally
6. librdkafka returns to Consumer_consume() → Exception still in global state from different context ❌
7. Context mismatch detected → Consumer_consume() must return to Python but doesn't "own" this exception
8. User gets SystemError → Python detects inconsistent exception ownership between contexts 

**New Workflow (Fixed)**

1. Application calls consumer.consume() → Python → C extension (Consumer_consume) → librdkafka
2. librdkafka triggers callback → error_cb() creates new Python execution context, callback raises exception
4. Exception context transfer → PyErr_Fetch() extracts exception from callback context + clears global state ✅
5. librdkafka continues cleanly → No exception in global state, librdkafka completes normally
6. CallState_end() restores context → PyErr_Restore() transfers exception ownership to Consumer_consume() context
7. Consumer_consume() returns to Python → Now properly "owns" the exception, returns NULL with exception set
8. User gets single exception → Only RuntimeError, proper exception ownership maintained 

Checklist
------------------

References
----------
https://docs.python.org/3/c-api/exceptions.html
> As a general principle, a function that calls another function to perform some task should check whether the called function raised an exception, and if so, pass the exception state on to its caller.

> When a function must fail, it is important to set an exception condition and return an error value (usually a NULL pointer).

Test & Review
------------

```
#!/usr/bin/env python3
"""
Minimal reproducer for Issue #865: SystemError when error_cb raises

This script is based on the exact reproducer from the GitHub issue.
It will try to connect to a non-existent broker, which will trigger
the error callback, which will raise an exception, which will cause
the SystemError.
"""

import sys
import os

# Use our locally built version with the fix
sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'src'))

import confluent_kafka


def error_cb(error):
    """Error callback that raises an exception - this causes the SystemError"""
    # print(f"Error callback triggered: {error}")
    raise confluent_kafka.KafkaException(error)


def demo():
    """Main demo function that reproduces the SystemError"""
    consumer = confluent_kafka.Consumer({
        "bootstrap.servers": "nonexistent:9092",  # This will fail to resolve
        "group.id": "demo",
        "error_cb": error_cb,
    })
    
    consumer.subscribe(["test"])
    
    # This will trigger the error callback and cause SystemError
    return consumer.consume(1, timeout=1)


if __name__ == "__main__":
    print("Reproducing SystemError from Issue #865")
    print("=" * 50)
    print()
    
    print("This will attempt to connect to 'nonexistent:9092' which will fail.")
    print("The error_cb will be triggered and will raise a KafkaException.")
    print("You should see TWO exceptions:")
    print("1. Your KafkaException (the intended one)")
    print("2. A SystemError (the bug)")
    print()
    
    try:
        result = demo()
        print(f"Unexpected success: {result}")
    except Exception as e:
        print(f"Caught exception: {type(e).__name__}: {e}")
        print()
        
        # Print the full traceback to show both exceptions
        import traceback
        print("Full traceback showing the double exception:")
        print("-" * 50)
        traceback.print_exc()

```

### DIFF
--- a/tests/test_Consumer.py
+++ b/tests/test_Consumer.py
@@ -321,7 +321,7 @@ def test_callback_exception_no_system_error():
 
     # Test error_cb callback
     error_called = []
-    
+
     def error_cb_that_raises(error):
         """Error callback that raises an exception"""
         error_called.append(error)
@@ -348,7 +348,7 @@ def test_callback_exception_no_system_error():
 
     # Test stats_cb callback
     stats_called = []
-    
+
     def stats_cb_that_raises(stats_json):
         """Stats callback that raises an exception"""
         stats_called.append(stats_json)
@@ -376,7 +376,7 @@ def test_callback_exception_no_system_error():
 
     # Test throttle_cb callback (may not be triggered in this scenario)
     throttle_called = []
-    
+
     def throttle_cb_that_raises(throttle_event):
         """Throttle callback that raises an exception"""
         throttle_called.append(throttle_event)
@@ -402,25 +402,25 @@ def test_callback_exception_no_system_error():
         if "Test exception from throttle_cb" in str(exc_info.value):
             assert len(throttle_called) > 0
             print("Throttle callback was triggered and raised exception")
-    
+
     consumer3.close()
 
 
 def test_error_callback_exception_different_error_types():
     """Test error callback with different exception types"""
-    
+
     def error_cb_kafka_exception(error):
         """Error callback that raises KafkaException"""
         raise KafkaException(error)
-    
+
     def error_cb_value_error(error):
         """Error callback that raises ValueError"""
         raise ValueError(f"Custom error: {error}")
-    
+
     def error_cb_runtime_error(error):
         """Error callback that raises RuntimeError"""
         raise RuntimeError(f"Runtime error: {error}")
-    
+
     # Test with KafkaException
     consumer1 = TestConsumer({
         'group.id': 'test-kafka-exception',
@@ -430,11 +430,11 @@ def test_error_callback_exception_different_error_types():
         'error_cb': error_cb_kafka_exception
     })
     consumer1.subscribe(['test-topic'])
-    
+
     with pytest.raises(KafkaException):
         consumer1.consume(timeout=0.1)
     consumer1.close()
-    
+
     # Test with ValueError
     consumer2 = TestConsumer({
         'group.id': 'test-value-error',
@@ -444,12 +444,12 @@ def test_error_callback_exception_different_error_types():
         'error_cb': error_cb_value_error
     })
     consumer2.subscribe(['test-topic'])
-    
+
     with pytest.raises(ValueError) as exc_info:
         consumer2.consume(timeout=0.1)
     assert "Custom error:" in str(exc_info.value)
     consumer2.close()
-    
+
     # Test with RuntimeError
     consumer3 = TestConsumer({
         'group.id': 'test-runtime-error',
@@ -459,14 +459,8 @@ def test_error_callback_exception_different_error_types():
         'error_cb': error_cb_runtime_error
     })
     consumer3.subscribe(['test-topic'])
-    
+
     with pytest.raises(RuntimeError) as exc_info:
         consumer3.consume(timeout=0.1)
     assert "Runtime error:" in str(exc_info.value)
     consumer3.close()
-
-
-
-
-
-
--- a/tests/test_Producer.py
+++ b/tests/test_Producer.py
@@ -974,6 +974,7 @@ def test_produce_batch_api_compatibility():
         # AvroProducer not available - skip this part
         pytest.skip("AvroProducer not available")
 
+
 def test_callback_exception_no_system_error():
     delivery_reports = []
 
@@ -1007,48 +1008,48 @@ def test_callback_exception_no_system_error():
 
 def test_core_callbacks_exception_different_types():
     """Test error_cb, throttle_cb, and stats_cb exception handling with different exception types"""
-    
+
     # Test error_cb with different exception types
     def error_cb_kafka_exception(error):
         raise KafkaException(KafkaError._FAIL, "KafkaException from error_cb")
-    
+
     def error_cb_value_error(error):
         raise ValueError("ValueError from error_cb")
-    
+
     def error_cb_runtime_error(error):
         raise RuntimeError("RuntimeError from error_cb")
-    
+
     # Test error_cb exceptions - these should be triggered by connection failure
     producer = Producer({
         'bootstrap.servers': 'nonexistent-broker:9092',
         'socket.timeout.ms': 100,
         'error_cb': error_cb_kafka_exception
     })
-    
+
     with pytest.raises(KafkaException) as exc_info:
         producer.produce('test-topic', value='test-message')
         producer.flush(timeout=2.0)
     assert "KafkaException from error_cb" in str(exc_info.value)
-    
+
     # Test error_cb with ValueError
     producer = Producer({
         'bootstrap.servers': 'nonexistent-broker:9092',
         'socket.timeout.ms': 100,
         'error_cb': error_cb_value_error
     })
-    
+
     with pytest.raises(ValueError) as exc_info:
         producer.produce('test-topic', value='test-message')
         producer.flush(timeout=2.0)
     assert "ValueError from error_cb" in str(exc_info.value)
-    
+
     # Test error_cb with RuntimeError
     producer = Producer({
         'bootstrap.servers': 'nonexistent-broker:9092',
         'socket.timeout.ms': 100,
         'error_cb': error_cb_runtime_error
     })
-    
+
     with pytest.raises(RuntimeError) as exc_info:
         producer.produce('test-topic', value='test-message')
         producer.flush(timeout=2.0)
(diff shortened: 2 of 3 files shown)


MSeal:
Looks like some whitespace issue are making the linter fail

src/confluent_kafka/src/Consumer.c
```diff
@@ -1575,6 +1575,8 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
 		if (result)
 			Py_DECREF(result);
 		else {
+			
+			PyErr_Fetch(&cs->exception_type, &cs->exception_value, &cs->exception_traceback);
```
pranavrth: [PyErr_Fetch](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch) is deprecated since 3.12. Its suggested to use [PyErr_GetRaisedException()](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_GetRaisedException) instead.

tests/test_Consumer.py
```diff
@@ -324,3 +324,38 @@ def test_consumer_without_groupid():
...
+        'socket.timeout.ms': 100,
+        'session.timeout.ms': 1000,
+        'error_cb': error_cb_that_raises
+    })
+
+    consumer.subscribe(['test-topic'])
+
+    # This should trigger the error callback due to connection failure
+    # Before fix: Would get RuntimeError + SystemError
+    # After fix: Should only get RuntimeError (no SystemError)
```
pranavrth: Add issue number here.

tests/test_Producer.py
```diff
@@ -283,3 +283,34 @@ def test_producer_bool_value():
...
+        'message.timeout.ms': 10,  # Very short timeout to trigger delivery failure quickly
+        'on_delivery': delivery_cb_that_raises
+    })
+
+    # Produce a message - this will trigger delivery report callback when it fails
+    producer.produce('test-topic', value='test-message')
+
+    # Flush to ensure delivery reports are processed
+    # Before fix: Would get RuntimeError + SystemError
+    # After fix: Should only get RuntimeError (no SystemError)
```
pranavrth, line 998: Same add issue number here.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1810,6 +1812,8 @@ static void throttle_cb (rd_kafka_t *rk, const char *broker_name, int32_t broker
                 goto done;
         }
 
+        PyErr_Fetch(&cs->exception_type, &cs->exception_value, &cs->exception_traceback);
+
```
pranavrth: Put this into an else condition just to maintain the consistency across the callbacks.

src/confluent_kafka/src/Consumer.c
MSeal: You'll need to check if `PyErr_SetRaisedException(exc);` is necessary as well to set the context temporarily rather than passing it into GetRaisedException

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2592,8 +2601,19 @@ int CallState_end (Handle *h, CallState *cs) {
...
 	PyEval_RestoreThread(cs->thread_state);
 
-	if (PyErr_CheckSignals() == -1 || cs->crashed)
+	if (PyErr_CheckSignals() == -1)
+		return 0;
+
+	if (cs->crashed) {
+		/* Restore the saved exception if we have one */
+		if (cs->exception_type) {
+			PyErr_Restore(cs->exception_type, cs->exception_value, cs->exception_traceback);
```
MSeal: Similarly you'll need to use https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetRaisedException here instead as `PyErr_Restore` is also deprecated

k-raina:
@MSeal @pranavrth 

Thank you for the feedback regarding the deprecated PyErr_Fetch/PyErr_Restore APIs. 

After reviewing the codebase, I'd like to discuss the implementation approach given our current Python version support requirements. The confluent-kafka-python library currently supports [Python >=3.7](https://github.com/confluentinc/confluent-kafka-python/blob/cc4c8d80d2a777b9277be833e2d896ca188e90e1/pyproject.toml#L17), while the new exception APIs (PyErr_GetRaisedException/PyErr_SetRaisedException) were introduced in `Python 3.12`.

I see two main approaches to address the deprecated API concerns:

**Option 1: Defer API Migration**
- **Approach**: Continue using PyErr_Fetch/PyErr_Restore until we update minimum Python version to 3.12+
- **Pros**: 
  - Simple, proven implementation
  - No version-conditional code complexity
  - Current fix works reliably across all supported versions
- **Cons**: 
  - Uses deprecated APIs (though still functional)
  - Will need update in future releases once minimum python version support is changed to 3.12+

**Option 2: Compatibility Layer**
- **Approach**: Add version-conditional wrapper functions in `confluent_kafka.h` (check [this implementation](https://github.com/confluentinc/confluent-kafka-python/pull/2019/files#diff-bf55d1bd9c3eacae8dc0083ad7e18a66d074c892159cd14bbeb2cbbd817b6003R283-R331))
- **Pros**: 
  - Uses modern APIs where available
  - Future-proofs the codebase
- **Cons**: 
  - Increases code complexity
  - Higher risk of introducing bugs in critical error handling
  
IMO, i am fine with both options. Please let me know your preference.

pranavrth:
Option 2 is the way to go in this case as we should aim for correctness and future compatibility. I don't think the code is too complex as well. 
Also, we will have to eventually reach Option 2 when the deprecated APIs are removed as we will still support lower python version.

src/confluent_kafka/src/Admin.c
```diff
@@ -5082,7 +5082,7 @@ static void Admin_background_event_cb (rd_kafka_t *rk, rd_kafka_event_t *rkev,
                         PyObject *trace = NULL;
 
                         /* Fetch (and clear) currently raised exception */
-                        PyErr_Fetch(&exctype, &error, &trace);
+                        cfl_exception_fetch(&exctype, &exc, &trace);
```
MSeal: Was the Fetch call against `error` instead of `exc` incorrect before?

MSeal (approved):
One question and kicked off the consumer test against as it looked unrelated failure to initialize.

src/confluent_kafka/src/Admin.c
k-raina: Thanks @MSeal for catch
I went through code and it was functionally correct, although semantically it looked wrong. I have reverted change.

src/confluent_kafka/src/Consumer.c
```diff
@@ -1575,6 +1575,8 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
 		if (result)
 			Py_DECREF(result);
 		else {
+
```
pranavrth: Remove extra line. From other places as well.

MSeal (approved):

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -275,8 +275,66 @@ int  Handle_traverse (Handle *h, visitproc visit, void *arg);
...
+static inline void
+cfl_exception_fetch(PyObject **exc_type, PyObject **exc_value, PyObject **exc_traceback) {
+#if PY_VERSION_HEX >= 0x030c0000
+    /* Python 3.12+ - use new API */
+    PyObject *exc = PyErr_GetRaisedException();
+    if (exc) {
+        *exc_type = (PyObject *)Py_TYPE(exc);
+        Py_INCREF(*exc_type);
+        *exc_value = exc;
+        *exc_traceback = PyException_GetTraceback(exc);
```
pranavrth: We don't need `exc_type` and `exc_traceback` in the new C APIs. Check if we can remove these safely. This removal will simplify ref count handling as well.

Check `CallState_end` as well. `cs->exception_type` usage is added there in the PR.

```
if (cs->crashed) {
		/* Restore the saved exception if we have one */
		if (cs->exception_type) {
			CallState_restore_exception(cs);
		}
		return 0;
	}
```

src/confluent_kafka/src/Consumer.c
```diff
@@ -1575,6 +1575,8 @@ static void Consumer_rebalance_cb (rd_kafka_t *rk, rd_kafka_resp_err_t err,
 		if (result)
 			Py_DECREF(result);
 		else {
+			
+			CallState_fetch_exception(cs);
```
pranavrth, line 1590: There are other places as well where we need to handle this. For example `Consumer_offset_commit_cb`. Please check other places where this is needed.

tests/test_Consumer.py
```diff
@@ -324,3 +324,38 @@ def test_consumer_without_groupid():
     with pytest.raises(ValueError) as ex:
         TestConsumer({'bootstrap.servers': "mybroker:9092"})
     assert ex.match('group.id must be set')
+
+
+def test_callback_exception_no_system_error():
```
pranavrth, line 319: I would suggest adding more test cases related to librdkafka layer, C binding layer and python layer.

pranavrth (dismissed):
Provided a few comments as discussed over call.

k-raina:
Thanks @MSeal 
Tests are passing, merging PR!

PR #2020, opened 2025-08-13
Set registered schema subject correctly

jekel:
What
----
Schema Registry server on new schema registration with specifiied subject could return [no subject](https://docs.confluent.io/platform/current/schema-registry/develop/api.html#post--subjects-(string-%20subject)-versions) name in response,
like Confluent Schema Registry, so, register_schema_full_response method does not set it correctly, and returned registered schema is without subject.
This pull request fixs this issue.


rayokota (approved):
Thanks @jekel , LGTM

PR #2024, opened 2025-08-18
Add missing endpoints (/compatibility, /mode) to SchemaRegistryClient

fangnx:
What
----

Add the missing endpoints and update the existing ones to SR client, for /compatibility and /mode endpoints.

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/DGS-21591

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/confluent_kafka/schema_registry/common/json_schema.py
+++ b/src/confluent_kafka/schema_registry/common/json_schema.py
@@ -73,6 +73,14 @@ def transform(
     field_ctx = ctx.current_field()
     if field_ctx is not None:
         field_ctx.field_type = get_type(schema)
+    original_type = schema.get("type")
+    if isinstance(original_type, list) and len(original_type) > 0:
+        subschema = _validate_subtypes(schema, message, ref_registry)
+        try:
+            if subschema is not None:
+                return transform(ctx, subschema, ref_registry, ref_resolver, path, message, field_transform)
+        finally:
+            schema["type"] = original_type  # restore original type
     all_of = schema.get("allOf")
     if all_of is not None:
         subschema = _validate_subschemas(all_of, message, ref_registry)
@@ -134,6 +142,22 @@ def _transform_field(
             message[prop_name] = new_value
     finally:
         ctx.exit_field()
+
+
+def _validate_subtypes(
+    schema: JsonSchema, message: JsonMessage, registry: Registry
+) -> Optional[JsonSchema]:
+    schema_type = schema.get("type")
+    if not isinstance(schema_type, list) or len(schema_type) == 0:
+        return None
+    for typ in schema_type:
+        schema["type"] = typ
+        try:
+            validate(instance=message, schema=schema, registry=registry)
+            return schema
+        except ValidationError:
+            pass
+    return None
 
 
 def _validate_subschemas(
--- a/tests/schema_registry/_async/test_json_serdes.py
+++ b/tests/schema_registry/_async/test_json_serdes.py
@@ -573,6 +573,72 @@ async def test_json_cel_field_transform():
     assert obj2 == newobj
 
 
+async def test_json_cel_field_transform_with_nullable():
+    conf = {'url': _BASE_URL}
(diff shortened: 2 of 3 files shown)


src/confluent_kafka/schema_registry/_async/schema_registry_client.py
```diff
@@ -1222,13 +1223,36 @@ async def test_compatibility(
...
+        self, subject_name: str, schema: 'Schema', normalize: bool = False, verbose: bool = False
+    ) -> bool:
+        """
+        Test the input schema against all of the subject's schemas for compatibility.
+
+        Args:
+            subject_name (str): Name of the schema against which compatibility is to be tested.
+            schema (Schema): Schema instance.
+            normalize (bool): Whether to normalize the input schema. # TODO: missing in cp + cc docs
+            verbose (bool): Wehther to return detailed error messages.
```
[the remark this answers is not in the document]
MSeal: Nice to have copilot find typos to save reading for them :D

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
```diff
@@ -1291,12 +1294,40 @@ async def test_compatibility(
...
 
         request = schema.to_dict()
-
         response = await self._rest_client.post(
-            'compatibility/subjects/{}/versions/{}'.format(_urlencode(subject_name), version), body=request
+            'compatibility/subjects/{}/versions/{}?normalize={}&verbose={}'.format(
+                _urlencode(subject_name), version, normalize, verbose),
+            body=request
         )
+        return response['is_compatible']  # TODO: should it return entire response (including error messages)?
```
MSeal: Maybe return a tuple [bool, Response] so it's easy to have a simple response vs a parsing the complex object? But I think this should probably return the full response by itself with a reference in the docstring to the `is_compatible` as a likely attribute for consumption.

src/confluent_kafka/schema_registry/_sync/schema_registry_client.py
```diff
@@ -1360,6 +1391,104 @@ def get_config(self, subject_name: Optional[str] = None) -> 'ServerConfig':
...
+            str: Schema Registry mode. Must be one of IMPORT, READONLY, READWRITE (default).
+
+        Raises:
+            SchemaRegistryError: if the request was unsuccessful.
+
+        See Also:
+            `GET Global Mode API Reference <https://docs.confluent.io/current/schema-registry/develop/api.html#get--mode>`_
+        """  # noqa: E501
+        result = self._rest_client.get('mode')
+        return result['mode']
```
MSeal, line 1429: Consider a debug log here with the result if you're not returning the full payload so the contents aren't lost (or maybe generally have a logging debug level option for all rest requests made to SR service)

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
naxinf: My concern is this would be a breaking change,  although in this case I think it might be necessary to just make the change (or add a separate method and mark this one as @deprecated): not sure how customers usually interact with test_compatibility, but intuitively I think the "messages" field for test_compatibility is important to include
cc @rayokota if you have any thoughts on this :)

src/confluent_kafka/schema_registry/_sync/schema_registry_client.py
naxinf: In this case, `mode` is the only field in the JSON returned from this API

rayokota (approved):
Thanks @fangnx , LGTM

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
rayokota: We shouldn't make any breaking changes, let's keep this as-is for now

MSeal (approved):
Might need a merge/rebase from master

PR #2027, opened 2025-08-19
Integrate Schema Registry with ducktape load tests

fangnx:
What
----

Add basic ducktape test cases for producer w/ schema registry integrated, covering 3 supported serialization formats.

As a follow-up item, we can explore testing producing with with different sizes of messages to be serialized.

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/DGS-21924

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/tests/ducktape/test_producer_with_schema_registry.py
+++ b/tests/ducktape/test_producer_with_schema_registry.py
@@ -29,8 +29,7 @@ class SimpleProducerTestWithSchemaRegistry(Test):
         self.logger.info("Verifying connection to external Kafka at localhost:9092")
 
         if not self.kafka.verify_connection():
-            raise ConnectionError("Cannot connect to Kafka at localhost:9092. "
-                            "Please ensure Kafka is running.")
+            raise ConnectionError("Cannot connect to Kafka at localhost:9092. Please ensure Kafka is running.")
 
         self.logger.info("Successfully connected to Kafka")
 
@@ -56,7 +55,8 @@ class SimpleProducerTestWithSchemaRegistry(Test):
 
         self.logger.info("Successfully completed %s test with comprehensive metrics", serialization_type)
 
-    def produce_messages_with_serialization(self, producer, topic_name, serializer, string_serializer, test_duration, message_value_func, serialization_type):
+    def produce_messages_with_serialization(self, producer, topic_name, serializer, string_serializer,
+                                            test_duration, message_value_func, serialization_type):
         """Produce messages using the given serializer with comprehensive metrics collection"""
         # Initialize metrics collection and bounds
         metrics = MetricsCollector()
@@ -97,7 +97,9 @@ class SimpleProducerTestWithSchemaRegistry(Test):
             try:
                 # Calculate message size for metrics
                 serialized_key = string_serializer(message_key)
-                serialized_value = serializer(message_value, SerializationContext(topic_name, MessageField.VALUE))
+                serialized_value = serializer(
+                    message_value, SerializationContext(topic_name, MessageField.VALUE)
+                )
                 message_size = len(serialized_key) + len(serialized_value)
 
                 # Record message being sent with metrics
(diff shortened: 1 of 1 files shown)


tests/ducktape/test_producer_with_schema_registry.py
```diff
@@ -0,0 +1,127 @@
...
+        # Set up Kafka client and Schema Registry client
+        self.kafka = KafkaClient(test_context, bootstrap_servers="localhost:9092")
+        self.schema_registry_client = SchemaRegistryClient({'url': 'http://localhost:8081'})
+
+    def setUp(self):
+        """Set up test environment"""
+        self.logger.info("Verifying connection to external Kafka at localhost:9092")
+
+        if not self.kafka.verify_connection():
+            raise Exception("Cannot connect to Kafka at localhost:9092. "
```
MSeal: As SonarQube suggests never throw an `Exception` or `BaseException`. I'd maybe raise a `ConnectionError` here.

tests/ducktape/test_producer_with_schema_registry.py
```diff
@@ -0,0 +1,286 @@
...
+from confluent_kafka.schema_registry._sync.protobuf import ProtobufSerializer
+from confluent_kafka.schema_registry.avro import AvroSerializer
+from confluent_kafka.serialization import MessageField, SerializationContext, StringSerializer
+from tests.ducktape.services.kafka import KafkaClient
+from tests.ducktape.benchmark_metrics import MetricsCollector, MetricsBounds, validate_metrics, print_metrics_report
+from tests.integration.schema_registry.data.proto import PublicTestProto_pb2
+
+try:
+    from confluent_kafka import Producer
+except ImportError:
```
MSeal: Why would we ever be in this state? We should just import the Producer no?

tests/ducktape/test_producer_with_schema_registry.py
```diff
@@ -0,0 +1,286 @@
...
+        # Finalize metrics collection
+        metrics.finalize()
+
+        # Get comprehensive metrics summary and validate
+        metrics_summary = metrics.get_summary()
+        self.calculate_and_verify_results(metrics_summary, bounds, serialization_type)
+
+    def test_basic_produce_with_avro_serialization(self):
+        """Test producing messages with Avro serialization using Schema Registry"""
+        if Producer is None:
```
MSeal: I don't think we should support this, see comment above

tests/ducktape/test_producer_with_schema_registry.py
```diff
@@ -0,0 +1,286 @@
...
+            self.logger.error("confluent_kafka not available, skipping test")
+            return
```
MSeal: BTW pytest has decorators for skipping tests conditionally which is better than returning early if you need this pattern in the future

tests/ducktape/test_producer_with_schema_registry.py
```diff
@@ -0,0 +1,286 @@
...
+            json_serializer,
+            string_serializer,
+            test_duration,
+            lambda messages_sent: {'name': f"User{messages_sent}", 'age': messages_sent},
+            "JSON"
+        )
+
+    def test_basic_produce_with_protobuf_serialization(self):
+        """Test producing messages with Protobuf serialization using Schema Registry"""
+        if Producer is None:
```
MSeal: Same as above

MSeal (changes requested):

tests/ducktape/test_producer_with_schema_registry.py
naxinf: I think it's a mistake (code not cleaned up by me properly) during my testing. Just updated the PR to get rid of them. Thanks for the feedback 👍

MSeal (approved):

PR #2033, opened 2025-08-22
Support key password for schema registry client.

insightfuls:
Support key password for schema registry client
----
When a client certificate is used, it is common for the private key to be encrypted. The schema registry client currently does not support this. This PR adds the support, at the same time migrating away from features that have been deprecated in current versions of `httpx`.

Checklist
------------------
- [X] Contains customer facing changes? Including API/behavior changes
- [X] Did you add sufficient unit test and/or integration test coverage for this PR?
  - I have updated the existing test covering this code, and added a new one. The existing coverage is slim, and I have not increased it, however all the heavy lifting is done by `httpx` and Python's `ssl` module.

References
----------
- `httpx` code that I have mimicked in part of this implementation (which however has a bug preventing usage of custom ca certs with client certificates which I have _not_ mimicked): https://github.com/encode/httpx/blob/26d48e0634e6ee9cdc0533996db289ce4b430177/httpx/_config.py#L23-L69
- `httpx` dependencies, indicating that `certifi` is a transitive dependency, so the direct dependency I have added is merely formal, not introducing any new dependency in practice: https://github.com/encode/httpx/blob/26d48e0634e6ee9cdc0533996db289ce4b430177/pyproject.toml#L31

Test & Review
------------
In addition to the updated unit tests, I am working on testing this in a real-world environment.

### DIFF
--- a/requirements/requirements-schemaregistry.txt
+++ b/requirements/requirements-schemaregistry.txt
@@ -1,5 +1,6 @@
 attrs>=21.2.0
 cachetools>=5.5.0
+certifi
 httpx>=0.26
 authlib>=1.0.0
 orjson >= 3.10
--- a/src/confluent_kafka/schema_registry/_async/schema_registry_client.py
+++ b/src/confluent_kafka/schema_registry/_async/schema_registry_client.py
@@ -16,8 +16,11 @@
 # limitations under the License.
 #
 import asyncio
+import certifi
 import orjson
 import logging
+import os
+import ssl
 import time
 import urllib
 from urllib.parse import unquote, urlparse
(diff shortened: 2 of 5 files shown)


cchristous:
Hi, it looks like the `cla-status-check` is failing because there is a commit by `Ben Schmidt` included and that isn't a valid GitHub account. I'd suggest squashing your commits so they are all owned by `insightfuls`, and retrying the `cla-status-check`.

insightfuls:
> Hi, it looks like the `cla-status-check` is failing because there is a commit by `Ben Schmidt` included and that isn't a valid GitHub account. I'd suggest squashing your commits so they are all owned by `insightfuls`, and retrying the `cla-status-check`.

Thanks for that. Fixed.

insightfuls:
Fixed the linting issue from [Semaphore](https://confluentinc.semaphoreci.com/workflows/1bdb795b-d7b4-4fa1-8cfa-874dca71e3e5?pipeline_id=0f143a38-8f88-48b0-8c91-5354f70cffe9) but there's another issue; seems to be more related to the build env than the code change to me, but please let me know if I need to fix something.

insightfuls:
Thanks for accepting and merging this @rayokota . FWIW, I've successfully tested this in a real world environment now.

PR #2036, opened 2025-08-26
Add unit tests for asyncio producer + consumer

fangnx:
What
----

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/DGS-21925

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/tests/test_AIOConsumer.py
+++ b/tests/test_AIOConsumer.py
@@ -3,7 +3,7 @@
 import pytest
 import asyncio
 import concurrent.futures
-from unittest.mock import Mock, patch, AsyncMock
+from unittest.mock import Mock, patch
 
 from confluent_kafka import TopicPartition
 from confluent_kafka.aio._AIOConsumer import AIOConsumer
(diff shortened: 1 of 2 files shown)


tests/test_AIOConsumer.py
```diff
@@ -0,0 +1,129 @@
...
+        """Basic consumer configuration."""
+        return {
+            'bootstrap.servers': 'localhost:9092',
+            'group.id': 'test-group',
+            'auto.offset.reset': 'earliest'
+        }
+
+    @pytest.mark.asyncio
+    async def test_constructor_behavior(self, mock_consumer, mock_common, basic_config):
+        """Test constructor creates consumer with correct configuration."""
```
MSeal: Not quite sure what we're testing for -- that _max_workers on executor isn't overwritten by consumer max_workers?

tests/test_AIOProducer.py
```diff
@@ -0,0 +1,274 @@
...
+        """Test successful message production."""
+        producer = AIOProducer(basic_config, auto_poll=False)
+
+        mock_msg = Mock()
+        loop = asyncio.get_event_loop()
+
+        def sync_produce(*args, **kwargs):
+            callback = kwargs.get('on_delivery')
+            if callback:
+                # Simulate successful delivery - callback would normally come from poll()
```
MSeal, line 111: Nice setup

MSeal (approved):
Over looks good, I would change the file naming convention to test_aioconsumer.py and test_aioproducer.py OR we should do a follow-up PR to change the test files to not have capital letters for classes past the `test_` prefix to follow standard Python convention.

tests/test_AIOConsumer.py
naxinf: Yes, it's to cover the logic that we only set max_workers when using the default executor: https://github.com/confluentinc/confluent-kafka-python/blob/async/src/confluent_kafka/aio/_AIOConsumer.py#L24-L33. It's a pretty trivial test but it's nice to have it covered

naxinf:
> Over looks good, I would change the file naming convention to test_aioconsumer.py and test_aioproducer.py OR we should do a follow-up PR to change the test files to not have capital letters for classes past the `test_` prefix to follow standard Python convention.

Updated the naming 👍 The repo has many other files not following the convention. I think we can raise a separate PR to clean them up

PR #2041, opened 2025-09-05
Add type hinting

fangnx:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


src/confluent_kafka/_model/__init__.py
```diff
@@ -60,7 +61,7 @@ class ConsumerGroupTopicPartitions:
         List of topic partitions information.
     """
 
-    def __init__(self, group_id, topic_partitions=None):
+    def __init__(self, group_id: str, topic_partitions: Optional[List['cimpl.TopicPartition']] = None) -> None:
```
MSeal: Why'd this need the cimpl path prefix to work here? Curious to chat about how the cbinding type errors were worked out in development

src/confluent_kafka/_util/conversion_util.py
```diff
@@ -12,12 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Union, Type, TypeVar
 from enum import Enum
 
+# Generic type for enum conversion
+E = TypeVar('E', bound=Enum)
```
MSeal: Why do we need this? Can't we just use `type[Enum]` below?

src/confluent_kafka/admin/_cluster.py
```diff
@@ -34,11 +36,12 @@ class DescribeClusterResult:
...
     """
 
-    def __init__(self, controller, nodes, cluster_id=None, authorized_operations=None):
+    def __init__(self, controller: Node, nodes: List[Node], cluster_id: Optional[str] = None,
+                 authorized_operations: Optional[List[Any]] = None) -> None:
         self.cluster_id = cluster_id
         self.controller = controller
         self.nodes = nodes
-        self.authorized_operations = None
+        self.authorized_operations: Optional[List[AclOperation]] = None
```
MSeal: Why the hint here and why does it differ from the hint in the def?

src/confluent_kafka/admin/_scram.py
```diff
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .. import cimpl
-
+from typing import List, Optional, Any
 from enum import Enum
 
+from .. import cimpl
```
MSeal, line 18: Why the move?

src/confluent_kafka/cimpl.pyi
```diff
@@ -0,0 +1,443 @@
+"""
```
MSeal, line 15: I do dislike how we'll have two locations for type/actual definitions we need to maintain now. However it may be the ideal solution for now. This really makes me want to try a pass on converting the c files to Cython and then getting the type hinting for free while avoiding manually defining so many Python things in C binding calls. Not in scope for this PR though it would eliminate the above dual definition problem. For now let's add a warning comment to the C code files to update this file when/if editing interfaces therein

MSeal (commented):
Would like to spend some time over zoom talking through the work on this PR but overall looks reasonable as a solution PR

src/confluent_kafka/_model/__init__.py
naxinf: We are only importing the module `from .. import cimpl` so the type checker would not find the directly referenced TopicPartition. We can switch to `from ..cimpl import TopicPartition` to make the code a bit cleaner

src/confluent_kafka/_util/conversion_util.py
naxinf: Yes I think we can just leave it `type[Enum]`. I think `Type[E]` creates a stricter type relationship with the specific enum class, but it overcomplicates the code a bit and I don't think this function (meant to handle types itself) can benefit from it

src/confluent_kafka/admin/_scram.py
naxinf: Honestly I think cursor moved it during the first pass and I didn't notice it. But after some investigation, I think the move makes sense as it adheres to the pep8 import rule: https://peps.python.org/pep-0008/#imports

src/confluent_kafka/admin/_cluster.py
naxinf: Fixed. Also update authorized_operations to be more specific: `Optional[List[Union[str, int, AclOperation]]]`, as those 3 types are handled in ConversionUtil.convert_to_enum

src/confluent_kafka/_model/__init__.py
```diff
@@ -165,7 +167,7 @@ class IsolationLevel(Enum):
     READ_UNCOMMITTED = cimpl.ISOLATION_LEVEL_READ_UNCOMMITTED  #: Receive all the offsets.
     READ_COMMITTED = cimpl.ISOLATION_LEVEL_READ_COMMITTED  #: Skip offsets belonging to an aborted transaction.
 
-    def __lt__(self, other):
+    def __lt__(self, other) -> Any:
```
MSeal: should be  `-> bool`

src/confluent_kafka/_model/__init__.py
```diff
@@ -184,7 +186,7 @@ class ElectionType(Enum):
     #: Unclean election
     UNCLEAN = cimpl.ELECTION_TYPE_UNCLEAN
 
-    def __lt__(self, other):
+    def __lt__(self, other) -> Any:
```
MSeal: should be  `-> bool`

src/confluent_kafka/admin/__init__.py
```diff
@@ -634,15 +648,17 @@ def delete_topics(self, topics, **kwargs):
...
 
         return super(AdminClient, self).list_topics(*args, **kwargs)
 
-    def list_groups(self, *args, **kwargs):
+    def list_groups(self, *args: Any, **kwargs: Any) -> GroupMetadata:
 
         return super(AdminClient, self).list_groups(*args, **kwargs)
 
-    def create_partitions(self, new_partitions, **kwargs):
+    def create_partitions(  # type: ignore[override]
```
MSeal, line 659: Why do we need the type ignore here?

src/confluent_kafka/admin/__init__.py
```diff
@@ -634,15 +648,17 @@ def delete_topics(self, topics, **kwargs):
...
+        self, new_partitions: List[NewPartitions], **kwargs: Any
```
MSeal, line 660: TIL **kwargs you only set the value type on the hint

src/confluent_kafka/admin/_config.py
```diff
@@ -172,31 +176,33 @@ def __init__(self, restype, name,
...
             return "ConfigResource(%s,%s,%r)" % (self.restype, self.name, self.error)
         else:
             return "ConfigResource(%s,%s)" % (self.restype, self.name)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash((self.restype, self.name))
 
-    def __lt__(self, other):
+    def __lt__(self, other: 'ConfigResource') -> bool:
```
MSeal: Enforcement here might be awkward, since technically `__lt__` should be handling non-type match argumentsand giving a default answer or raising is NotImplemented -- which it appears to not implement

src/confluent_kafka/admin/_group.py
```diff
@@ -73,10 +80,8 @@ class MemberAssignment:
         The topic partitions assigned to a group member.
     """
 
-    def __init__(self, topic_partitions=[]):
-        self.topic_partitions = topic_partitions
-        if self.topic_partitions is None:
-            self.topic_partitions = []
+    def __init__(self, topic_partitions: List[TopicPartition] = []) -> None:
```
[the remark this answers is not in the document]
MSeal: This is catching a real bug here -- we should fix this to be None as it's currently sharing the instance value incorrectly across instances without provided inputs

src/confluent_kafka/admin/_listoffsets.py
```diff
@@ -52,18 +57,18 @@ def max_timestamp(cls):
...
+    def __new__(cls, index: int):
         # Trying to instantiate returns one of the subclasses.
         # Subclasses can be instantiated but aren't accessible externally.
         if index < 0:
             return cls._values[index]
         else:
             return cls.for_timestamp(index)
 
-    def __lt__(self, other):
+    def __lt__(self, other) -> Any:
```
MSeal: `other` missing type

src/confluent_kafka/admin/_metadata.py
```diff
@@ -73,25 +76,25 @@ class TopicMetadata (object):
     # Sphinx issue where it tries to reference the same instance variable
     # on other classes which raises a warning/error.
 
-    def __init__(self):
+    def __init__(self) -> None:
         self.topic = None
```
MSeal: Missing string type if we're also typing partitions

src/confluent_kafka/admin/_metadata.py
```diff
@@ -73,25 +76,25 @@ class TopicMetadata (object):
...
         """Topic name"""
-        self.partitions = {}
+        self.partitions: Dict[int, 'PartitionMetadata'] = {}
         """Map of partitions indexed by partition id. Value is a PartitionMetadata object."""
         self.error = None
```
MSeal: Missing KafkaError type if we're also typing partitions

src/confluent_kafka/admin/_scram.py
```diff
@@ -25,7 +26,7 @@ class ScramMechanism(Enum):
     SCRAM_SHA_256 = cimpl.SCRAM_MECHANISM_SHA_256  #: SCRAM-SHA-256 mechanism
     SCRAM_SHA_512 = cimpl.SCRAM_MECHANISM_SHA_512  #: SCRAM-SHA-512 mechanism
 
-    def __lt__(self, other):
+    def __lt__(self, other) -> Any:
```
MSeal: bool return type

src/confluent_kafka/experimental/aio/producer/_AIOProducer.py
```diff
@@ -199,10 +224,10 @@ async def flush(self, *args, **kwargs):
             # Update buffer activity since we just flushed
             self._buffer_timeout_manager.mark_activity()
 
-        # Then flush the underlying producer and wait for delivery confirmation
```
MSeal, line 202: I think the old text was grammatically correct

src/confluent_kafka/admin/__init__.py
fangnx: You are right - no need for this

src/confluent_kafka/admin/__init__.py
naxinf: Yep, I assume the reason is that the type of key is always `str`

src/confluent_kafka/admin/_config.py
naxinf: Good point. I think maybe I should add the type check here and in other places to follow the pattern below:
```
def __lt__(self, other: object) -> bool:
    if not isinstance(other, MyClass):
        return NotImplemented
    return self.attribute < other.attribute
```

src/confluent_kafka/experimental/aio/producer/_AIOProducer.py
naxinf: Cursor was trying to smart and made a bunch of creative text edits :(

src/confluent_kafka/admin/__init__.py
naxinf: I think it's unfortunately required to work around the current setup:
1. In C extension layer, Admin_create_partitions returns None
2. In the cimpl.pyi stub file where we define the C-extension class `_AdminClientImpl`, we set return type of create_partitions to None
3. In the python wrapper layer, `AdminClient` inherits `_AdminClientImpl` but create_partitions returns a future instead of None

src/confluent_kafka/admin/_config.py
MSeal: Yes I think it should be that

src/confluent_kafka/experimental/aio/_common.py
```diff
@@ -14,42 +14,69 @@
...
+        loop: asyncio.AbstractEventLoop,
+        logger: logging.Logger
+    ) -> None:
         self.loop = loop
         self.logger = logger
 
-    def log(self, *args, **kwargs):
-        self.loop.call_soon_threadsafe(self.logger.log, *args, **kwargs)
+    def log(self, *args: Any, **kwargs: Any) -> None:
+        self.loop.call_soon_threadsafe(lambda: self.logger.log(*args, **kwargs))
```
naxinf, line 35: Another fix to pass the function in correctly
```
def call_soon_threadsafe(self, callback, *args, context=None)
```

src/confluent_kafka/experimental/aio/producer/_kafka_batch_executor.py
```diff
@@ -106,10 +108,24 @@ def _produce_batch_and_poll() -> int:
         loop = asyncio.get_running_loop()
         return await loop.run_in_executor(self._executor, _produce_batch_and_poll)
 
-    def _handle_partial_failures(
-        self,
-        batch_messages: List[Dict[str, Any]]
-    ) -> None:
+    async def flush_librdkafka_queue(self, timeout=-1):
```
MSeal, line 111: I think since you picked up master changes you now have some more methods to type -- I'll approve PR and we can do a second pass for these separately as part of putting code checks on typing.

PR #2044, opened 2025-09-09
Add AsyncIo producer and consumer test

k-raina:


tests/ducktape/producer_strategy.py
```diff
@@ -0,0 +1,194 @@
...
+        super().__init__(*args, **kwargs)
+        self._producer_instance = None
+    
+    def create_producer(self):
+        from confluent_kafka.aio import AIOProducer
+        # Enable logging for AIOProducer
+        import logging
+        logging.basicConfig(level=logging.INFO)
+        
+        self._producer_instance = AIOProducer({'bootstrap.servers': self.bootstrap_servers}, max_workers=1, auto_poll=True)
```
naxinf: Is max_workers=1 intended? I thought the benchmark was based on 20 workers

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -72,8 +72,9 @@ def on_delivery(err, msg):
                 loop.call_soon_threadsafe(result.set_result, msg)
 
         kwargs['on_delivery'] = on_delivery
+        # Wait for message to be queued, but don't wait for delivery
```
fangnx, line 75: Nice catch

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+            *args, **kwargs: Additional parameters like partition, timestamp, headers
+        """
+        # Get current running event loop
+        result = asyncio.get_running_loop().create_future()
+        
+        # Track send time for latency calculation if metrics are enabled
+        send_time = None
+        message_id = None
+        if self._metrics_callback:
+            import time
```
MSeal: Just put the import at the top, no need for JIT for this one

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+                # Handle delivery success
+                if self._metrics_callback and message_id:
+                    latency_ms = 0.0
+                    if message_id in self._send_times:
+                        latency_ms = (time.time() - self._send_times[message_id]) * 1000
+                        del self._send_times[message_id]
+                    self._metrics_callback(latency_ms, msg.topic(), msg.partition(), True)  # success=True
+                result.set_result(msg)
+        
+        kwargs['on_delivery'] = on_delivery
```
MSeal: Did we overwrite the user on_delivery argument here? I think the intermediate solution checked if the callback was async or not and called it alongside the result setting. It'd probably be good to still do that even though it's not a common pattern for async with a future returned but since we're maintaining api compatibility.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+            import time
+            send_time = time.time()
+            message_id = f"{topic}:{key}:{id(result)}"  # Unique message identifier
+            self._send_times[message_id] = send_time
+
+        # Pre-bind variables to avoid closure overhead
+        def on_delivery(err, msg):
+            if err:
+                # Handle delivery failure
+                if self._metrics_callback and message_id:
```
MSeal: If we're intending the capture these client metrics for retrieval by end users this should stay in here. But if we're only using it for the benchmark test we should move it there instead of polluting the function.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+        send_time = None
+        message_id = None
+        if self._metrics_callback:
+            import time
+            send_time = time.time()
+            message_id = f"{topic}:{key}:{id(result)}"  # Unique message identifier
+            self._send_times[message_id] = send_time
+
+        # Pre-bind variables to avoid closure overhead
+        def on_delivery(err, msg):
```
MSeal: My knee jerk reaction for a tight loop call was to ask to avoid doing a function generation within that loop. But that code path has been heavily optimized since Python 2 days (and I checked that `lambda` has a very similar low cost profile without any perf differences).

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+        # Internal metrics tracking
+        self._metrics_callback = None
+        self._send_times = {}  # Track send times for latency calculation
+
+    def set_metrics_callback(self, callback):
+        """Set callback for delivery metrics: callback(latency_ms, topic, partition, success)"""
+        self._metrics_callback = callback
+
+    # ========================================================================
+    # HYBRID OPERATIONS - Blocking behavior depends on parameters
```
MSeal: I liked these section indicators

tests/ducktape/test_producer.py
```diff
@@ -30,10 +30,18 @@ def setUp(self):
 
         self.logger.info("Successfully connected to Kafka")
 
-    def test_basic_produce(self):
+    def createProducer(self, producer_type):
```
MSeal:
```suggestion
    def create_producer(self, producer_type):
```

MSeal (changes requested):
I still think the future resolving contention was a root cause of perf, but moving the producer to direct call being viable is a better solution so long as we don't hit any periodic latency issues with the queue insertion to librdkafka locking it's thread.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+    # ========================================================================
+
+    async def poll(self, timeout=0, *args, **kwargs):
+        """Processes callbacks - blocking behavior depends on timeout
+        
+        Args:
+            timeout: 0 = non-blocking, >0 = block up to timeout seconds, -1 = block indefinitely
```
naxinf: Should we elaborate more what timeout means in this context? Without context it might be not be super clear :)

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,102 @@
...
+        """
+        if timeout > 0:
+            # Blocking call - use ThreadPool to avoid blocking event loop
+            return await self._call(self._producer.poll, timeout, *args, **kwargs)
+        else:
```
naxinf: Does the else branch also handle the -1 case

tests/ducktape/test_producer.py
```diff
@@ -260,11 +188,11 @@ def delivery_callback(err, msg):
 
         self.logger.info("Successfully completed %ds batch production test with comprehensive metrics", test_duration)
 
-    @matrix(compression_type=['none', 'gzip', 'snappy'])
-    def test_produce_with_compression(self, compression_type):
+    @matrix(producer_type=["sync", "async"], compression_type=['none', 'gzip', 'snappy'])
```
naxinf, line 199: At the moment I don't think compression_type is being used, unless I miss something from the code

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -74,36 +75,26 @@ async def produce(self, topic, value=None, key=None, *args, **kwargs):
...
-                    if message_id in self._send_times:
-                        latency_ms = (time.time() - self._send_times[message_id]) * 1000
-                        del self._send_times[message_id]
-                    self._metrics_callback(latency_ms, msg.topic(), msg.partition(), True)  # success=True
                 result.set_result(msg)
+                
+                # Call user's callback on successful delivery
+                if user_callback:
+                    try:
+                        user_callback(err, msg)  # err is None here
```
MSeal: I'd do the check here for if it's a sync or async function with `asyncio.iscoroutinefunction(func)` and await if it's true. Since the parent context is async it's likely the user may pass a async callback without thinking much about it.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -74,36 +75,26 @@ async def produce(self, topic, value=None, key=None, *args, **kwargs):
...
+                    except Exception:
+                        # Log but don't propagate user callback errors to avoid breaking delivery confirmation
```
MSeal: Don't swallow / catch general exceptions. This leads to silent failures and can catch system signals that should terminate. If the user really wants to each eat exceptions they can within the function they pass through.

tests/ducktape/producer_strategy.py
```diff
@@ -24,12 +24,28 @@ def produce_messages(self, topic_name, test_duration, start_time, message_format
 
 
 class SyncProducerStrategy(ProducerStrategy):
-    def create_producer(self):
-        producer = Producer({'bootstrap.servers': self.bootstrap_servers})
+    def create_producer(self, config_overrides=None):
+        config = {
```
MSeal: Fun trick, you can do ** expansion to inline update dicts
```
{'bootstrap.servers': self.bootstrap_servers, **(config_overrides or {})}
```

tests/ducktape/producer_strategy.py
```diff
@@ -112,39 +130,52 @@ def delivery_callback(err, msg):
         print("=" * 45)
         
         return messages_sent
+        
+    def get_final_metrics(self):
+        """Return final metrics summary for the sync producer"""
+        if self.metrics:
+            return self.metrics
+        return None
```
MSeal, line 182: Technically not needed since no return defaults to None return but ok to be explicit

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,93 @@
...
+            timeout: Timeout in seconds for waiting for callbacks:
+                    - 0 = non-blocking, return immediately after processing available callbacks
+                    - >0 = block up to timeout seconds waiting for new callbacks to arrive
+                    - -1 = block indefinitely until callbacks are available
+        
+        Returns:
+            Number of callbacks processed during this call
+        """
+        if timeout > 0 or timeout == -1:
+            # Blocking call - use ThreadPool to avoid blocking event loop
```
emasab: Why do you say the auto poll was blocking the event loop before this change? the wait is executed in one of the threads of the `ThreadPoolExecutor`

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,93 @@
...
+                # Call user's callback on successful delivery
+                if user_callback:
+                    try:
+                        user_callback(err, msg)  # err is None here
+                    except Exception:
+                        # Log but don't propagate user callback errors to avoid breaking delivery confirmation
+                        pass
+        
+        kwargs['on_delivery'] = on_delivery
+        self._producer.produce(topic, value, key, *args, **kwargs)
```
emasab: The `producer.produce` isn't exactly non-blocking, if everything goes well it is, but in `rd_kafka_producev` there are locks, `rd_kafka_wrlock(rk);`, `rd_kafka_topic_rdlock(rkt);`, `mtx_lock(&rk->rk_curr_msgs.lock);` a custom partitioner (that at the moment cannot be used in Python because of a deadlock to solve in librdkafka), some interceptors `rd_kafka_interceptors_on_send`, `rd_kafka_interceptors_on_acknowledgement` that` could have any code and be used later for tracing.

We can see if it's possible to implement `rd_kafka_produce_batch` in Python so if we have N futures still not completed we accumulate to a list and then when one of them finishes send the list of messages (through the ThreadPoolExecutor)

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -21,36 +21,93 @@
...
+        
+        Returns:
+            Number of callbacks processed during this call
+        """
+        if timeout > 0 or timeout == -1:
+            # Blocking call - use ThreadPool to avoid blocking event loop
+            return await self._call(self._producer.poll, timeout, *args, **kwargs)
+        else:
+            # Non-blocking call (timeout=0) - direct call is safe
+            return self._producer.poll(timeout, *args, **kwargs)
```
emasab: There are some issues with calling it directly, the `poll` could call a wrapped `stats_cb` that has to spawn an async function in the event loop and wait it synchronously. So needs to be called from a dedicated thread in the `ThreadPoolExecutor`.

MSeal: Good catch, I missed this in the last set of edits. I believe poll has to always be thread isolated with the current state of librdkafka

src/confluent_kafka/aio/_AIOProducer.py
MSeal: He is referencing that the future assignment was getting blocked by poll because it was awaiting on that future instead of returning it. I had suggested trying returning the future without awaiting but I think he was still hitting contention. We should talk through that more

src/confluent_kafka/aio/_AIOProducer.py
naxinf: Do you think introducing batch-produce at the Python client layer (my understanding is that librdkafka already batches messages internally) can drastically improve the async produce() performance, similar to async consume() w/ batching, as the ThreadPool overhead will be amortized?

https://github.com/confluentinc/confluent-kafka-python/blob/55b55550acabc51cb75c7ac78190d6db71706690/src/confluent_kafka/src/Consumer.c#L1076

emasab: Right, as happened in the `consume` benchmark. Just in the consumer we cannot implement a single message poll with a larger consume call because when we call `consume` the full batch can be auto stored, we would need to use manual offset management and we don't want to do complex things here, just follow existing API, so we can just tell users to use the `consume` call for high throughput.
For the `produce` call instead we can accumulate the calls coming from produce and when finished sending a previous batch we send the next one. This will change a bit the ordering of messages but in the context of concurrent production we aren't ensuring it.

src/confluent_kafka/aio/_AIOProducer.py
emasab: Yes it has to be outside the event loop, but when an async callback needs to be called it calls it in the event loop and awaits its result synchronously in the callback wrapper.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -17,40 +17,122 @@
...
+        Args:
+            timeout: Timeout in seconds for waiting for callbacks:
+                    - 0 = non-blocking, return immediately after processing available callbacks
+                    - >0 = block up to timeout seconds waiting for new callbacks to arrive
+                    - -1 = block indefinitely until callbacks are available
+        
+        Returns:
+            Number of callbacks processed during this call
+        """
+        if timeout > 0 or timeout == -1:
```
MSeal: I think we're missing the else statement as timeout=0 is a no-op. But that being said does it makes sense to have anything but non-blocking as an option here? Maybe it should be forced as timeout=0 mode always for an async caller.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -17,40 +17,122 @@
...
+            key: Message key (optional)
+            *args, **kwargs: Additional parameters like partition, timestamp, headers
+        """
+        # Start Time
+        start_time = time.perf_counter()
+        
+        result = asyncio.get_running_loop().create_future()
+        user_callback = kwargs.get('on_delivery')
+        
+        await self._buffer_lock.acquire()
```
MSeal: This is not necessary. Only one thread will be accessing this area at once as an async function. Even if multiple threads are using producers they would have separate internal state.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
-        def on_delivery(err, msg):
-            if err:
-                loop.call_soon_threadsafe(result.set_exception,
-                                          _KafkaException(err))
-            else:
-                loop.call_soon_threadsafe(result.set_result, msg)
+    async def flush(self, *args, **kwargs):
+        """Waits until all messages are delivered or timeout"""
+        # First, flush any remaining messages in the buffer for all topics
+        async with self._buffer_lock:
```
MSeal: Again don't need with an async setup

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                # Flush all topics - the _flush_buffer method now properly handles multiple topics
+                await self._flush_buffer()
+        
+        # Then flush the producer
+        return await self._call(self._producer.flush, *args, **kwargs)
 
-        kwargs['on_delivery'] = on_delivery
-        await self._call(self._producer.produce, *args, **kwargs)
-        return await result
+    async def purge(self, *args, **kwargs):
```
MSeal: This needs to also clear the local `_message_buffer`

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+        """
+        # Call produce_batch first
+        self._producer.produce_batch(target_topic, batch_messages, on_delivery=batch_delivery_callback)
+        
+        # Immediately poll to process delivery callbacks in the same worker
+        poll_result = self._producer.poll(0)
+        
+        return poll_result
+
+    async def _flush_buffer(self, target_topic=None):
```
MSeal: This is quite a complex function :/

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -17,40 +17,122 @@
...
+            if 'timestamp' in kwargs:
+                msg_data['timestamp'] = kwargs['timestamp']
+            if 'headers' in kwargs:
+                msg_data['headers'] = kwargs['headers']
+            
+            # Store user callback in message data for later execution
+            if user_callback:
+                msg_data['user_callback'] = user_callback
+            
+            self._message_buffer.append(msg_data)
```
MSeal: I don't see it in the code yet in review but I think we're missing the edge where the we write to `_message_buffer` but never hit the `_batch_size` or manually `_flush_buffer`. In that case we'd lose message written or hold them indefinitely if throughput slows to a crawl. This could happen in a bursty operation mode or by an exception happening later and cleanup not being aware to clear local buffer. The `__del__` method can sometimes be used to catch this, though an async context wrapper pattern is better here to ensure buffer cleanup/flush on exit / exception as the entry point instead of direct calls to the method. But that changes the interface we offer. Another option that I think is easier and cleaner is to have a check for an async task that sleeps for 1-5 configurable seconds and flushes the message buffer if it's not been flushed in that time period. Keeps large buffer settings from holding messages indefinitely. Basically create the async task when init occurs and have it only self-repeat if the object still exists via a https://docs.python.org/3/library/weakref.html.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                
+                # Add optional fields if present
+                if 'partition' in msg_data:
+                    batch_msg['partition'] = msg_data['partition']
+                if 'timestamp' in msg_data:
+                    batch_msg['timestamp'] = msg_data['timestamp']
+                if 'headers' in msg_data:
+                    batch_msg['headers'] = msg_data['headers']
+                
+                batch_messages.append(batch_msg)
```
MSeal: Why can't we just append the original `msg_data` or pass `group_data['messages']` directly? If we really need to avoid reference mutation why not just `copy.copy` or `copy.deepcopy`?

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                    user_callback = callback_state['callbacks'][idx]
+                    
+                    if err:
+                        future.set_exception(_KafkaException(err))
+                    else:
+                        future.set_result(msg)
+                        if user_callback:
+                            try:
+                                user_callback(err, msg)
+                            except Exception:
```
MSeal: I would not try to catch all exceptions here, let them file and debug themselves. If they want exception swallowing let the user_callback define it inside their function.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+        """Flush the current message buffer using produce_batch via thread pool
+        
+        This method now properly handles messages for different topics by grouping
+        them and calling produce_batch separately for each topic.
+        """
+        if not self._message_buffer:
+            return
+        
+        # Group messages by topic
+        topic_groups = {}
```
MSeal: I would break this into a separate function for cyclomatic complexity reduction

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+        for i, msg_data in enumerate(self._message_buffer):
+            topic = msg_data['topic']
+            if topic not in topic_groups:
+                topic_groups[topic] = {
```
MSeal: Uhg, we're reliant on index ordering because batch produce librdkafka argument pattern aren't we

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+        - Reduces thread pool overhead (1 call instead of 2)
+        - Eliminates context switching between operations
+        - Ensures immediate callback processing
+        - Lower latency for delivery confirmation
+        """
+        # Call produce_batch first
+        self._producer.produce_batch(target_topic, batch_messages, on_delivery=batch_delivery_callback)
+        
+        # Immediately poll to process delivery callbacks in the same worker
+        poll_result = self._producer.poll(0)
```
MSeal: Can just return here to save a line

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                            except Exception:
+                                pass
+                    
+                    callback_state['index'] += 1
+            
+            try:
+                # Call produce_batch and poll for this topic
+                await self._call(self._produce_batch_and_poll, topic, batch_messages, batch_delivery_callback)
+                        
+            except Exception as e:
```
MSeal: Don't forget to re-raise the exception after marking as failed (and add a test for this case)

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                # If batch fails, fail all remaining futures for this topic
+                start_idx = callback_state['index']
+                for i in range(start_idx, len(batch_futures)):
+                    if not batch_futures[i].done():
+                        batch_futures[i].set_exception(e)
+
+    async def flush_buffer(self):
+        """Manually flush the current message buffer for all topics"""
+        async with self._buffer_lock:
```
MSeal: Don't need

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                await self._call(self._produce_batch_and_poll, topic, batch_messages, batch_delivery_callback)
+                        
+            except Exception as e:
+                # If batch fails, fail all remaining futures for this topic
+                start_idx = callback_state['index']
+                for i in range(start_idx, len(batch_futures)):
+                    if not batch_futures[i].done():
+                        batch_futures[i].set_exception(e)
+
+    async def flush_buffer(self):
```
MSeal: I think you can either delete this method or delete the private variant `_flush_buffer` and keep 1 function as this wrapper no longer servers a purpose with the lock removed.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+                if 'partition' in msg_data:
+                    batch_msg['partition'] = msg_data['partition']
+                if 'timestamp' in msg_data:
+                    batch_msg['timestamp'] = msg_data['timestamp']
+                if 'headers' in msg_data:
+                    batch_msg['headers'] = msg_data['headers']
+                
+                batch_messages.append(batch_msg)
+            
+            # Setup callback with pre-allocated state
```
MSeal: I would add to the comment what the difference between future and user callback is here.

src/confluent_kafka/aio/_AIOProducer.py
```diff
@@ -60,47 +142,254 @@ async def _call(self, blocking_task, *args, **kwargs):
...
+            try:
+                # Call produce_batch and poll for this topic
+                await self._call(self._produce_batch_and_poll, topic, batch_messages, batch_delivery_callback)
+                        
+            except Exception as e:
+                # If batch fails, fail all remaining futures for this topic
+                start_idx = callback_state['index']
+                for i in range(start_idx, len(batch_futures)):
+                    if not batch_futures[i].done():
+                        batch_futures[i].set_exception(e)
```
MSeal: Probably also need to trigger user_callback for the index for error handling

src/confluent_kafka/aio/_AIOProducer.py
MSeal: The time you'd need an asyncio.Lock would be if you awaited in the middle of working on memory objects that might be manipulated by the other async boundary which we don't need as we're done with manipulation by the time we might do `_flush_buffer`. This lock isn't threadsafe btw for preserving any threaded operations if that was the intent (which I don't think we should guard against as per the prior comment)

tests/ducktape/producer_strategy.py
```diff
@@ -0,0 +1,299 @@
...
+        
+        # Log the configuration for validation
+        self.logger.info("=== SYNC PRODUCER CONFIGURATION ===")
+        for key, value in config.items():
+            self.logger.info(f"{key}: {value}")
+        self.logger.info("=" * 40)
+        
+        return producer
+    
+    def produce_messages(self, topic_name, test_duration, start_time, message_formatter, delivered_container, failed_container=None):
```
MSeal: A lot of shared code with Async variant -- we'll want to squash that smaller at some point

tests/ducktape/producer_strategy.py
```diff
@@ -0,0 +1,299 @@
...
+            # Print producer latency stats
+            producer.get_latency_stats()
+
+            # Flush producer
+            flush_start = time.time()
+            await producer.flush(timeout=30)
+            flush_time = time.time() - flush_start
+
+            # Wait for all pending futures to complete (for delivery confirmation only)
+            for delivery_future, message_key in pending_futures:
```
MSeal, line 281: Maybe we should track how long we took to put messages into librdkafka in between total e2e delivery times? Ensures we know if it's on the delivery or generation end now and in the future.

MSeal (changes requested):
If we make an async `batch_produce` that just sends the batch with a single future to resolve does perf catch up to sync produce times? If so we can narrow down if it's future generation or something else in the logic that's slowing it down. If not then we need to look at why batch context swapping isn't working. All of my comments where on correctness or improvements but I don't think any will dramatically affect perf -- maybe the asyncio lock acquisition is affecting things but I doubt it.

src/confluent_kafka/aio/_common.py
```diff
@@ -48,6 +49,26 @@ def wrap_conf_logger(loop, conf):
         conf['logger'] = AsyncLogger(loop, conf['logger'])
 
 
+async def async_call(executor, blocking_task, *args, **kwargs):
```
MSeal, line 52: Good idea, I did this in another project that had similar wrapper needs

aio_producer_simple_diagram.md
```diff
@@ -0,0 +1,47 @@
+# AIOProducer Simple Class Diagram
```
MSeal, line 1: Good addition, AI is nice for making docs for future dev in these complex components

src/confluent_kafka/aio/producer/_buffer_timeout_manager.py
```diff
@@ -0,0 +1,146 @@
...
+        
+        Creates an async task that runs in the background and periodically checks
+        if messages have been sitting in the buffer for too long without being flushed.
+        
+        Key design decisions:
+        1. **Weak Reference**: Uses weakref.ref(self) to prevent circular references
+        2. **Self-Canceling**: The task stops itself if the manager is garbage collected
+        3. **Adaptive Check Interval**: Uses timeout to determine check frequency
+        """
+        if self._timeout <= 0:
```
MSeal: Common to use None for disabling these things in Python. This will catch that case cleanly
```suggestion
        if not self._timeout or self._timeout <= 0:
```

src/confluent_kafka/aio/producer/_buffer_timeout_manager.py
```diff
@@ -0,0 +1,146 @@
...
+                not manager._batch_processor.is_buffer_empty()):
+                
+                try:
+                    # Flush the buffer due to timeout
+                    await manager._flush_buffer_due_to_timeout()
+                    # Update activity since we just flushed
+                    manager.mark_activity()
+                except Exception:
+                    logger.error("Error flushing buffer due to timeout", exc_info=True)
+                    # Don't let buffer flush errors crash the timeout task
```
MSeal: Disagree -- don't swallow errors here. It could be system errors or user code errors that need to stop execution.

src/confluent_kafka/aio/producer/_buffer_timeout_manager.py
```diff
@@ -0,0 +1,146 @@
...
+        batches = self._create_batches_for_timeout_flush()
+        
+        # Execute all batches
+        for batch in batches:
+            await self._kafka_executor.execute_batch(batch.topic, batch.messages)
+        
+        # Clear the buffer since all messages were processed
+        self._batch_processor.clear_buffer()
+    
+    def _create_batches_for_timeout_flush(self):
```
MSeal: I'd remove this method as it's a 1 line call that's only called from one place

src/confluent_kafka/aio/producer/_callback_manager.py
```diff
@@ -0,0 +1,191 @@
...
+            else:
+                # Message delivered successfully  
+                if not self.future.done():  # Prevent double-setting
+                    self.future.set_result(msg)
+                if self.user_callback and self.pool:
+                    self.pool.handle_user_callback(self.user_callback, err, msg)
+                        
+        except Exception as e:
+            logger.error(f"Error in reusable message delivery callback: {e}", exc_info=True)
+            # Ensure future gets resolved even if there's an error in callback processing
```
MSeal: Do we do this in other languages? I believe in we raise the exception if the callback has an error in those, which I believe is the correct thing to do instead of setting a future error. Here we're masking that the message was delivered but the user callback had a problem. This might result in double sending messages from the client handling the future error on the message. Better to just let the user_callback throw and put responsibility on the library user to send a safe message or catch how they'd like within their callback.

src/confluent_kafka/aio/producer/_callback_manager.py
```diff
@@ -0,0 +1,191 @@
...
+
+import asyncio
+import logging
+
+from confluent_kafka import KafkaException as _KafkaException
+
+logger = logging.getLogger(__name__)
+
+
+class ReusableMessageCallback:
```
MSeal: Is this manager really necessary? Do you know what perf gains we get from this reference manager? If we are keeping this it needs a better block comment. To understand if it's resusing a future or not you have to jump 5-6 calls up from use. It seems like we always still need to call the loop to create a future, so this must be for avoiding making the closure over the user callback rather than for future juggling. If that's the case do we really get a good return? Seems like maybe we should just get rid of the callback for async class and tell the user to await the future return to run any callback operations. Then we could deleted several classes and a lot of complexity. Let's discuss this one asap before we merge

src/confluent_kafka/aio/producer/_kafka_batch_executor.py
```diff
@@ -0,0 +1,109 @@
...
+            - Handle partial batch failures for messages that fail immediately
+            - Poll for delivery reports to trigger callbacks for successful messages
+            """
+            # Call produce_batch with individual callbacks (no batch callback)
+            # Convert tuple to list since produce_batch expects a list
+            messages_list = list(batch_messages) if isinstance(batch_messages, tuple) else batch_messages
+            self._producer.produce_batch(topic, messages_list)
+            
+            # Handle partial batch failures: Check for messages that failed during produce_batch
+            # These messages have their msgstates destroyed in Producer.c and won't get callbacks
```
MSeal, line 73: Really?! That's not great

src/confluent_kafka/aio/producer/_kafka_batch_executor.py
```diff
@@ -0,0 +1,109 @@
...
+                callback = msg_dict.get('callback')
+                if callback:
+                    # Extract the error from the message dict (set by Producer.c)
+                    error = msg_dict['_error']
+                    # Manually invoke the callback with the error
+                    # Note: msg is None since the message failed before being queued
+                    try:
+                        callback(error, None)
+                    except Exception as e:
+                        logger.error(f"Error in callback for failed message: {e}", exc_info=True)
```
MSeal: Repeat of don't swallow exceptions

MSeal: In this case you could ensure each callback is called in the partial list before raising the first exception hit... but that's a bit awkward too. I think it's probably better to just raise and tell callers to ensure exceptions aren't thrown by their code

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -0,0 +1,61 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import namedtuple
+from typing import List, Any, Optional
+
+
+# Create immutable MessageBatch value object
+MessageBatch = namedtuple('MessageBatch', [
```
MSeal: If you're going to type this file anyway, can just use 

```
    from typing import NamedTuple, Sequence, Callable, Any

    # Define a named tuple type with type hints
    class MessageBatch(NamedTuple):
        topic: str
        messages: Sequence[Any] # Or just str?!
        futures: Sequence[asyncio.Future]
        callbacks: Sequence[Callable]
```

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -0,0 +1,61 @@
...
+    'topic',        # str: Target topic for this batch
+    'messages',     # tuple: Prepared message dictionaries with callbacks
+    'futures',      # tuple: asyncio.Future objects to resolve
+    'callbacks'     # tuple: User callback functions (can be None)
+])
+
+
+def create_message_batch(topic: str, messages: List[dict], futures: List[Any], callbacks: List[Optional[callable]]) -> MessageBatch:
```
MSeal: Can set the types to `Sequence` instead of `List` to allow for tuples

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -0,0 +1,61 @@
...
+        messages: List of prepared message dictionaries
+        futures: List of asyncio.Future objects
+        callbacks: List of user callback functions
+        
+    Returns:
+        MessageBatch: Immutable batch object
+    """
+    return MessageBatch(
+        topic=topic,
+        messages=tuple(messages),
```
MSeal: Is the only reason to copy the list / tuple here? Do we need to ensure an immutable list passed? If so you could do a `messages = tuple(messages) if not isinstance(messages, tuple) else messages` to avoid copy if already immutable.

src/confluent_kafka/aio/producer/_message_batch.py
MSeal: Not entirely sure this method is needed if it's just call the class constructor, which could do the argument cleanup / copies in an __init__ if it was a NamedTuple Class

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -0,0 +1,61 @@
...
+        futures=tuple(futures), 
+        callbacks=tuple(callbacks)
+    )
+
+
+# Add convenience properties to MessageBatch
+def _get_batch_size(self) -> int:
```
MSeal: Oh yeah just use the NamedTuple -- doing a faux self at the top level function to fake beinh a method isn't needed in Python 3

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -0,0 +1,61 @@
...
+    """Get the number of messages in this batch"""
+    return len(self.messages)
+
+def _get_batch_info(self) -> str:
```
MSeal: See above

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -0,0 +1,61 @@
...
+    """Get a string representation of batch info"""
+    return f"MessageBatch(topic='{self.topic}', size={len(self.messages)})"
+
+# Monkey-patch methods onto the namedtuple
```
MSeal: See above

src/confluent_kafka/aio/producer/_buffer_timeout_manager.py
MSeal: Comment above still applied

MSeal (changes requested):
The test functions are a bit long and hard to review, unsure if it's worth cleaning up. I wanted to talk about the callback reuse manager some when you're online.

src/confluent_kafka/aio/producer/_AIOProducer.py
```diff
@@ -181,12 +173,11 @@ async def produce(self, topic, value=None, key=None, *args, **kwargs):
         if 'timestamp' in kwargs:
             msg_data['timestamp'] = kwargs['timestamp']
         if 'headers' in kwargs:
+            # NOTE: Headers are currently ignored in batch mode due to librdkafka API limitations.
```
MSeal: We should forcibly raise a `NotImplemented` exception in this case

PR #2052, opened 2025-09-16
WIP: Add more async consumer integration tests

fangnx:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/tests/ducktape/test_consumer.py
+++ b/tests/ducktape/test_consumer.py
@@ -12,6 +12,8 @@ from tests.ducktape.consumer_benchmark_metrics import (ConsumerMetricsCollector,
                                                      validate_consumer_metrics, print_consumer_metrics_report)
 from tests.ducktape.consumer_strategy import SyncConsumerStrategy, AsyncConsumerStrategy
 from confluent_kafka import Producer
+from confluent_kafka.aio import AIOConsumer
+import asyncio
 
 
 class SimpleConsumerTest(Test):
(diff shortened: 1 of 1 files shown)


tests/ducktape/services/kafka.py
```diff
@@ -124,3 +124,30 @@ def wait_for_topic(self, topic_name, max_wait_time=30, initial_wait=0.1):
 
         self.logger.error("Timeout waiting for topic '%s' after %ds", topic_name, max_wait_time)
         return False
+
+    def add_partitions(self, topic_name, new_partition_count):
+        """Add partitions to an existing topic"""
+        try:
+            from confluent_kafka.admin import AdminClient, NewPartitions
```
MSeal: move imports to top, no need for JIT here

tests/ducktape/test_consumer.py
```diff
@@ -147,73 +185,245 @@ def test_basic_consume(self, consumer_type, batch_size):
...
+
+                # Verify functionality
+                self.produce_test_messages(topic_name, num_messages=1)
+                msg = await consumer1.poll(timeout=5.0)
+                assert msg is not None, "Consumer should receive fresh message"
+
+            finally:
+                await consumer1.close()
+
+        asyncio.run(async_rebalance_test())
```
MSeal, line 150: A quick look at ducktape looked like it doesn't support async test defs to avoid this wrapper. Maybe we could make an async supporting wrapper class that does this in the Metaclass? Could contribute it back to enable async tests in ducktape later on. Can be a follow-up task -- don't block this PR on it

tests/ducktape/test_consumer.py
```diff
@@ -147,73 +185,245 @@ def test_basic_consume(self, consumer_type, batch_size):
...
+            # Produce messages to new partitions to trigger metadata refresh
+            self.produce_test_messages(topic_name, num_messages=5)
+
+            # Force rebalance by creating a new consumer that joins the group
+            # This will trigger metadata refresh and rebalancing for all consumers
+            consumer3 = self.create_consumer("async", group_id)
+            await consumer3.subscribe([topic_name], on_assign=track_rebalance)
+
+            # Poll all consumers until they detect new partitions and rebalance
+            for _ in range(15):  # Max 15 seconds for partition discovery
```
MSeal: Might be longer given the consumers each poll as well for 1 second consecutively correct?

tests/ducktape/test_consumer.py
```diff
@@ -147,73 +185,245 @@ def test_basic_consume(self, consumer_type, batch_size):
...
+            assert messages_received > 0, "Consumers should receive messages from new partitions"
+
+            # Clean up
+            await consumer1.close()
+            await consumer2.close()
+            await consumer3.close()
+
+        asyncio.run(async_topic_change_test())
+
+    # TODO: verify if the current behavior is correct/intended
```
MSeal: I would be surprised if it swallowed a callback error without reraising tbh

MSeal (changes requested):
Minor change request, glad to see wider edge case coverage

tests/ducktape/test_consumer.py
naxinf: Yes ducktape doesn't support it right now. Agreed - it's kinda verbose, and I can add a backlog ticket to work on after the main tasks are done

tests/test_AIOConsumer.py
```diff
@@ -124,3 +124,44 @@ async def test_multiple_concurrent_operations(self, mock_consumer, mock_common,
         results = await asyncio.gather(*tasks)
         assert len(results) == 3
         assert all(result is not None for result in results)
+
+    @pytest.mark.asyncio
+    async def test_concurrent_operations_error_handling(self, mock_consumer, mock_common, basic_config):
+        """Test concurrent async operations handle errors gracefully."""
+        # Mock: 2 poll calls fail, assignment succeeds
```
k-raina: Does this test verifies that consumer get KafkaError when assignment is empty?

tests/ducktape/services/kafka.py
```diff
@@ -124,3 +124,30 @@ def wait_for_topic(self, topic_name, max_wait_time=30, initial_wait=0.1):
...
+
+            admin_client = AdminClient({'bootstrap.servers': self.bootstrap_servers_str})
+            metadata = admin_client.list_topics(timeout=10)
```
k-raina: i see multiple instances of `timeout=10`. Maybe define it static ?

tests/ducktape/test_consumer.py
```diff
@@ -74,13 +81,36 @@ def produce_test_messages(self, topic_name, num_messages):
...
+    @matrix(consumer_type=["sync", "async"])
+    def test_basic_poll(self, consumer_type):
+        """Test basic message polling (single message) with comprehensive metrics and bounds validation"""
+        self._run_consumer_performance_test(
+            consumer_type=consumer_type,
+            operation_type="poll",
+        )
 
-        topic_name = f"test-{consumer_type}-consumer-topic"
+    def _run_consumer_performance_test(self, consumer_type, operation_type, batch_size=None):
```
k-raina: Maybe move these private helper functions to bottom of file?

naxinf: Will do

tests/test_AIOConsumer.py
naxinf: I don't think assign() and empty list (empty partitions) should cause consumer to return errors, unless I am mistaken @k-raina 
I removed the assignment reference (since it's not doing much) in this test case to avoid confusion

PR #2056, opened 2025-09-17
Add integration tests for transactions

fangnx:
What
----

Add integration tests for transactions:
  - `test_commit_transaction`: committed messages visible to `read_committed`.
  - `test_abort_transaction_then_retry_commit`: abort invisible; retry+commit visible.
  - `test_send_offsets_to_transaction`: offsets committed correctly via `send_offsets_to_transaction`.
  - `test_commit_multiple_topics_partitions`: atomic commit across topics/partitions.

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/NONJAVACLI-3991

Test & Review
------------

Open questions / Follow-ups
--------------------------


tests/ducktape/run_ducktape_test.py
```diff
@@ -34,7 +38,7 @@ def get_test_info(test_type):
 def main():
     """Run the ducktape test based on specified type"""
     parser = argparse.ArgumentParser(description="Confluent Kafka Python - Ducktape Test Runner")
-    parser.add_argument('test_type', choices=['producer', 'consumer', 'producer_sr'],
+    parser.add_argument('test_type', choices=['producer', 'consumer', 'producer_sr', 'transactions'],
```
k-raina: Will it be possible to run specific tests if we use "producer" string parameters . Eg. 
`ducktape tests/ducktape/test_producer.py::SimpleProducerTest.test_basic_produce`

tests/ducktape/test_transactions.py
```diff
@@ -0,0 +1,238 @@
...
+import asyncio
+from ducktape.tests.test import Test
+
+from tests.ducktape.producer_strategy import AsyncProducerStrategy
+from tests.ducktape.consumer_strategy import AsyncConsumerStrategy
+from tests.ducktape.services.kafka import KafkaClient
+from confluent_kafka.aio import AIOConsumer
+
+
+class TransactionsTest(Test):
```
k-raina, line 14: I think we need overall design on how to perf test transaction. One of options can be : Run n number of transaction end to end for async vs sync and compare performance

k-raina (changes requested):
I am wondering if we should have seperate folders/test_file for performance and integration/e2e tests? 

I see currently test_transaction does e2e testing, as there are no performance bounds defined

tests/ducktape/run_ducktape_test.py
naxinf: Currently I extracted the transaction tests into a standalone file as I think those involve both producer + consumer

tests/ducktape/test_transactions.py
naxinf: Good point - I was mostly doing integration tests to verify behavior, but maybe we should benchmark test transactions as well. Let me think about this

tests/ducktape/producer_strategy.py
```diff
@@ -0,0 +1,77 @@
...
+        test_duration,
+        start_time,
+        message_formatter,
+        delivered_container,
+        failed_container=None,
+    ):
+        raise NotImplementedError()
+
+
+class AsyncProducerStrategy(ProducerStrategy):
```
naxinf: Added it temporarily to unblock this PR. Will remove once the PR to add is merged and will rebase

tests/ducktape/test_transactions.py
MSeal: Making transactions an optional configuration of the perf tests seems best as a follow-up. Make the perf tests essentially a config driven execution engine for parameters of a performance scenario we can pick and choose for a given context.

MSeal (approved):

PR #2069, opened 2025-09-26
Group messages by topic partition before passing to produce_batch API

k-raina:
**Issue**
- librdkafka's rd_kafka_produce_batch API only supports a single partition for the entire batch
- Individual message partitions are not supported by the underlying librdkafka batch API

**Fix**
- Group messages by topic and partition before calling batch_produce from Async Producer.-

**Additional changes**
- In purge API, all futures should be moved to cancelled state. ` self._batch_processor.cancel_pending_futures()`
- Reduce default batch timeout to 1s from 5s for low throughput


src/confluent_kafka/aio/producer/_kafka_batch_executor.py
```diff
@@ -107,4 +112,4 @@ def _handle_partial_failures(self, batch_messages):
                         callback(error, None)
                     except Exception:
                         logger.warning("Exception in callback during partial failure handling", exc_info=True)
-                        raise
+                        raise
```
MSeal: Missing eof newline

src/confluent_kafka/aio/producer/_message_batch.py
```diff
@@ -52,12 +54,14 @@ def create_message_batch(topic: str,
...
         MessageBatch: Immutable batch object
     """
     return MessageBatch(
         topic=topic,
         messages=tuple(messages) if not isinstance(messages, tuple) else messages,
-        futures=tuple(futures) if not isinstance(futures, tuple) else futures
-    )
+        futures=tuple(futures) if not isinstance(futures, tuple) else futures,
+        partition=partition
+    )
```
MSeal: same here

PR #2071, opened 2025-09-29
AsyncIO: Only clear messages from buffer if executor passed

k-raina:
### Problem:

- AIOProducer was losing messages during broker disconnections
- Messages were cleared from buffer when executor failed, causing data loss when network failures occurred

 ### Solution:

- Modified ProducerBatchManager.flush_buffer() to clear buffer after successful execution. This atomic batch processing to ensure (at-least-once delivery 

 ###  Key Changes:
- `_producer_batch_processor.py`:Fixed buffer clearing logic in flush_buffer() method
- `_buffer_timeout_manager.py`: Updated timeout flush to use consistent error handling
- Refactored duplicate code into common buffer proccessor file's _execute_batches() function

### DIFF
--- a/src/confluent_kafka/aio/producer/_producer_batch_processor.py
+++ b/src/confluent_kafka/aio/producer/_producer_batch_processor.py
@@ -123,6 +123,7 @@ class ProducerBatchManager:
         self._message_buffer = messages_to_keep
         self._buffer_futures = futures_to_keep
 
+
     async def flush_buffer(self, target_topic=None):
         """Flush the current message buffer using produce_batch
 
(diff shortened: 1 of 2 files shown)


src/confluent_kafka/aio/producer/_producer_batch_processor.py
```diff
@@ -137,27 +138,53 @@ async def flush_buffer(self, target_topic=None):
...
             try:
-                # Execute batch using the Kafka executor with partition info
+                # Execute batch using the Kafka executor
                 await self._kafka_executor.execute_batch(batch.topic, batch.messages, batch.partition)
-
+                
+                # Only clear messages from buffer after successful execution
+                # Collect the messages and futures that were successfully processed
+                messages_to_clear.extend(batch.messages)
+                futures_to_clear.extend(batch.futures)
```
MSeal: You don't use this -- was that an accident?

src/confluent_kafka/aio/producer/_producer_batch_processor.py
```diff
@@ -137,27 +138,53 @@ async def flush_buffer(self, target_topic=None):
...
+                
             except Exception as e:
                 # Handle batch failure by failing all unresolved futures for this batch
                 self._handle_batch_failure(e, batch.futures)
                 # Re-raise the exception so caller knows the batch operation failed
                 raise
 
+        # Clear successfully processed messages from buffer
+        if messages_to_clear:
```
MSeal: I don't think this can ever be False with any non-empty `batches` since you either populate `messages_to_clear` or you have an exception and interrupt. So I'd just do the clear logic. Need to rerun the perf test to make sure this didn't impact anything

k-raina: Thanks for review. 
```
=== Performance Metrics Report ===
Duration: 5.78s
Messages sent: 876000
Messages delivered: 876000
Send throughput: 151439.79 msg/s
Data throughput: 4.15 MB/s

Latency Metrics:
  Average: 2920.81ms
  P50: 2971.81ms
  P95: 5037.35ms
  P99: 5181.73ms
  Max: 5222.79ms

Reliability:
  Success rate: 1.0000
  Error rate: 0.0000

Message Size Analysis:
  Average: 29 bytes
  Median: 29 bytes
  Range: 19 - 29 bytes
```

There is no performance degradation

src/confluent_kafka/aio/producer/_producer_batch_processor.py
k-raina: Missed this, fixed in latest commit

MSeal (approved):

PR #2077, opened 2025-10-01
Fix broken tests 

k-raina:
Minor Tests updated

### DIFF
--- a/tests/test_AIOProducer.py
+++ b/tests/test_AIOProducer.py
@@ -139,7 +139,7 @@ class TestAIOProducer:
         batch_called = asyncio.Event()
         captured_messages = None
 
-        def mock_produce_batch(topic, messages):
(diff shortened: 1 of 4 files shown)


tests/test_AIOProducer.py
```diff
@@ -209,10 +209,10 @@ async def test_multiple_concurrent_produce(self, mock_producer, mock_common, bas
         """Test multiple concurrent produce operations with batching."""
         producer = AIOProducer(basic_config, max_workers=3, batch_size=1)  # Force immediate flush
 
-        completed_produces = []
+        completed_produces = set()
```
naxinf: Curious: did the race condition issue exist, or just the test wasn't set up correctly? Would the same message (in python buffer) be flushed to librdkafka several times, if multiple threads are invoking it concurrently

PR #2097, opened 2025-10-09
Treat warnings as error in doc generation

pranavrth:


docs/conf.py
```diff
@@ -133,7 +133,7 @@ def _read_version_from_pyproject(pyproject_path=None):
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+# html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
```
MSeal: I'm assuming the theme still loads from line 131 and this plus import becomes unnecessary. Did you double check the rendering is preserving the theme? If so I'd just delete rather than comment these lines

pranavrth: I built the documentation before and after the removal and compared both versions visually and using vimdiff. Only a minor difference was observed: the character used to represent links after section headers had changed. This is not a significant issue, so we can proceed with removing that part.

MSeal (approved):

PR #2103, opened 2025-10-15
Enhanced the BufferTimeoutManager to flush the librdkafka queue

k-raina:
What
----
Enhanced the `BufferTimeoutManager` to flush the librdkafka queue during timeout-based flushes, ensuring messages are not only batched and sent to librdkafka but also delivered to the broker.

**Changes:**
1. **Added `flush_librdkafka_queue()` method** to `ProducerBatchExecutor` 
   - Wraps `producer.flush()` call in thread pool to avoid blocking event loop
   
2. **Enhanced `_flush_buffer_due_to_timeout()`** in `BufferTimeoutManager`
   - Now performs complete flush workflow:
     1. Flush local buffer to librdkafka (`batch_processor.flush_buffer()`)
     2. Flush librdkafka queue to broker (`kafka_executor.flush_librdkafka_queue()`)
   
3. **Added comprehensive test coverage** (`test_buffer_timeout_manager.py`)

**Why:**
Previously, timeout-based flushes only moved messages from the local buffer to librdkafka's queue but didn't ensure delivery to the broker.  Applications needed to call poll() or flush() before waiting delivery future. This PR removes this dependency from applications. 

Test & Review
------------
**Testing:**
```bash
# Run all buffer timeout manager tests
pytest tests/test_buffer_timeout_manager.py -v

# Run specific timeout flush test
pytest tests/test_buffer_timeout_manager.py::TestBufferTimeoutManager::test_flush_buffer_due_to_timeout -v

```

Sample reproduction code example : 
```
async def start_producer():
    print("Starting producer...")
    topic = "testtopic"
    producer = AIOProducer(configure_common(
        {
            'transactional.id': 'producer1'
        }), max_workers=5)
    print("Producer created")

    await producer.init_transactions()
    transaction_active = False

    try:
        while True:
            await producer.begin_transaction()
            transaction_active = True

            produce_futures = [await producer.produce(
                                 topic=topic,
                                 key=f'testkey{i}',
                                 value=f'testvalue{i}')
                               for i in range(10)]
            print(f"Produced {len(produce_futures)} messages")

            # If you put the message reading after the transaction commit you don't need this flush call
            # await producer.flush()
            print("Waiting for messages to be produced")
            for msg in await asyncio.gather(*produce_futures):
                print(
                    'Produced to: {} [{}] @ {}'.format(msg.topic(),
                                                       msg.partition(),
                                                       msg.offset()))

            await producer.commit_transaction()

            transaction_active = False
            await asyncio.sleep(1)

    except Exception as e:
        logger.error(e)
    finally:
        if transaction_active:
            await producer.abort_transaction()
        await producer.close()
        logger.info('Closed producer')

async def main():
    print("Main function started")
    await start_producer()

print("Script starting...")
try:
    asyncio.run(main())
except asyncio.exceptions.CancelledError as e:
    logger.warning(f'Asyncio task was cancelled: {e}')
```

### DIFF
--- a/src/confluent_kafka/experimental/aio/producer/_kafka_batch_executor.py
+++ b/src/confluent_kafka/experimental/aio/producer/_kafka_batch_executor.py
@@ -93,8 +93,8 @@ class ProducerBatchExecutor:
     async def flush_librdkafka_queue(self, timeout=-1):
         """Flush the librdkafka queue and wait for all messages to be delivered
 
-        This method blocks until all outstanding produce requests are completed
-        or the timeout is reached.
+        This method awaits until all outstanding produce requests are completed
+        or the timeout is reached, unless the timeout is set to 0 (non-blocking).
 
         Args:
             timeout: Maximum time to wait in seconds:


tests/test_buffer_timeout_manager.py
```diff
@@ -0,0 +1,296 @@
...
+
+            # Task should stop itself when manager is garbage collected
+            # In practice, the task checks if manager_ref() is None
+            # and breaks out of the loop
+            # This test mainly verifies the pattern is in place
+
+        self.loop.run_until_complete(async_test())
+
+
+if __name__ == '__main__':
```
MSeal: Don't think you want a main entry point here

tests/test_buffer_timeout_manager.py
```diff
@@ -0,0 +1,297 @@
...
+            
+            # Wait for task to notice manager is gone
+            await asyncio.sleep(0.6)
+            
+            # Task should stop itself when manager is garbage collected
+            # In practice, the task checks if manager_ref() is None
+            # and breaks out of the loop
+            # This test mainly verifies the pattern is in place
+
+        self.loop.run_until_complete(async_test())
```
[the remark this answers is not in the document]
MSeal: This is reasonable, we should have a check that the task successfully noticed the absence of the manager rather than assume it did without an exception in that await period.

src/confluent_kafka/experimental/aio/producer/_kafka_batch_executor.py
```diff
@@ -88,6 +90,23 @@ def _produce_batch_and_poll():
         loop = asyncio.get_running_loop()
         return await loop.run_in_executor(self._executor, _produce_batch_and_poll)
 
+    async def flush_librdkafka_queue(self, timeout=-1):
+        """Flush the librdkafka queue and wait for all messages to be delivered
+
+        This method blocks until all outstanding produce requests are completed
```
MSeal: Nit: should read that it "awaits until all outstanding..." since it's immediately returning the await callable instead of actually blocking

MSeal: Should also mention in the text unless the timeout is set to 0

PR #2104, opened 2025-10-16
Fix use.schema.id, broken by signature change to sr.lookup_schema().

insightfuls:
What
----
The `use.schema.id` functionality for publishing broke between v2.11.1 and v2.12.0, because the signature of the `sr.lookup_schema()` function changed to add the `fmt` argument before the `deleted` argument ([diff showing `schema_registry_client.py` showing the change at target line 828](https://github.com/confluentinc/confluent-kafka-python/commit/aa4c836410915291d4ab79412820620c1ac359da#diff-f6de176cdb0830da1d023413d442f7d25e956fcd575082d5333291a27db38582R826-R828)) so when [called passing arguments positionally as part of looking up a provided schema ID](https://github.com/confluentinc/confluent-kafka-python/blob/v2.12.0/src/confluent_kafka/schema_registry/_async/serde.py#L50), the `True` intended for `deleted` is passed as `fmt`, and a `POST` is made with a query string including `&deleted=False&format=True` where it should have had simply `&deleted=True`. This is causing our schema registry, at least, to throw back a `500 Server Error`. I've fixed it by passing some arguments by name, including in some similar methods.

Checklist
------------------
  - I'm not familiar enough with the codebase to easily add a unit test for this. Clearly the breakage slipped through the test suite, though, so it would be preferable to add one.

Test & Review
------------
~I'll report back once I'm able to test the changes in a real environment.~ I have successfully tested this fixes the issue I was experiencing in a real environment.

Open questions / Follow-ups
--------------------------
- Was the breaking signature change even intentional?

Review stakeholders
------------------
- @fangnx committed the code which broke it, so may have a stake in the fix.
- @rayokota assisted reviewing my previous PR to related code, so may be able to assist with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ p = Producer({'bootstrap.servers': 'mybroker1,mybroker2'})
 
 def delivery_report(err, msg):
     """ Called once for each message produced to indicate delivery result.
-        Triggered by poll() or flush(). """
+        Triggered by poll() or flush()."""
     if err is not None:
         print('Message delivery failed: {}'.format(err))
     else:


rayokota:
@insightfuls , thanks for the PR.  There are some `flake8` lint errors that are causing the build to fail.

MSeal:
We're about to get a hotfix prepared for early next week and this can be included.

insightfuls:
- Thanks, @rayokota. I've pushed a commit to address the PEP-8 issues.
- @MSeal, that would be fantastic if this could make a hotfix.
- @ Everyone, I've now successfully tested this in a real environment.

MSeal (approved):

PR #2107, opened 2025-10-17
Set up type hinting: add correct types and apply type fixes in schema-registry module

fangnx:
What
----

Follow-up PR after https://github.com/confluentinc/confluent-kafka-python/pull/2041: adding missing types + correcting existing types, according to mypy static checker, in schema-registry module 

What's left are functions that might require refactoring and more thorough investigation to get the types right:
- common/avro.py, json_schema.py, protobuf.py: schemas are defined as unions, and we need to add guards to verify types accordingly during transformations
- schema_registry_client: handling None return types for several functions
- rules/cel: adding types for CEL operations 
- rules/encryption: handling None values propagated through functions. Need to understand the logic better

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/DGS-22076

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/requirements/requirements-rules.txt
+++ b/requirements/requirements-rules.txt
@@ -2,6 +2,8 @@ azure-identity
 azure-keyvault-keys
 boto3>=1.35
 cel-python>=0.4.0
+# Pin google-re2 to last version with Python 3.9 wheels (see https://pypi.org/project/google-re2/1.1.20251105/#files)
+google-re2<1.1.20251105
 google-auth
 google-api-core
 google-cloud-kms


src/confluent_kafka/schema_registry/rules/encryption/hcvault/hcvault_client.py
```diff
@@ -29,29 +29,27 @@ class HcVaultKmsClient(tink.KmsClient):
     """Basic HashiCorp Vault client for AEAD."""
 
     def __init__(
-        self, key_uri: Optional[str], token: Optional[str], ns: Optional[str] = None,
+        self, key_uri: str, token: Optional[str], ns: Optional[str] = None,
```
naxinf, line 32: I removed the optional part as we are already checking emptiness for key_url in https://github.com/confluentinc/confluent-kafka-python/blob/master/src/confluent_kafka/schema_registry/rules/encryption/hcvault/hcvault_driver.py#L53, and I think HcVaultKmsClient is supposed to be only created with a specified key_uri, according to the function doc

Not sure if this is considered a breaking change (if customers initializes HcVaultKmsClient directly in their code). @rayokota would love to hear your thoughts on this

src/confluent_kafka/schema_registry/rules/encryption/azurekms/azure_client.py
```diff
@@ -30,27 +30,25 @@ class AzureKmsClient(tink.KmsClient):
     """Basic Azure client for AEAD."""
 
     def __init__(
-        self, key_uri: Optional[str], credentials: TokenCredential
+        self, key_uri: str, credentials: TokenCredential
```
naxinf, line 31: Similar change as the one in hcvault_client.py

src/confluent_kafka/schema_registry/_sync/avro.py
```diff
@@ -365,7 +375,11 @@ def _get_parsed_schema(self, schema: Schema) -> AvroSchema:
             return parsed_schema
 
         named_schemas = _resolve_named_schema(schema, self._registry)
+        if schema.schema_str is None:
```
naxinf, line 400: In practice the schema_str field should never be empty, and even it is I think it makes sense to raise the error to fail early here

src/confluent_kafka/schema_registry/_async/avro.py
```diff
@@ -57,12 +57,13 @@ async def _resolve_named_schema(
     named_schemas = {}
     if schema.references is not None:
         for ref in schema.references:
-            referenced_schema = await schema_registry_client.get_version(ref.subject, ref.version, True)
-            ref_named_schemas = await _resolve_named_schema(referenced_schema.schema, schema_registry_client)
+            # References in registered schemas are validated by server to be complete
+            referenced_schema = await schema_registry_client.get_version(ref.subject, ref.version, True)  # type: ignore[arg-type]
```
MSeal: Why is there a need for type ignoring here? Can we set the ref/referenced_schema type to avoid this?

src/confluent_kafka/schema_registry/_async/avro.py
```diff
@@ -270,8 +277,11 @@ async def __init_impl(
                 # i.e. {"type": "string"} has a name of string.
                 # This function does not comply.
                 # https://github.com/fastavro/fastavro/issues/415
-                schema_dict = json.loads(schema.schema_str)
-                schema_name = parsed_schema.get("name", schema_dict.get("type"))
+                if schema.schema_str is not None:
+                    schema_dict = json.loads(schema.schema_str)
+                    schema_name = parsed_schema.get("name", schema_dict.get("type"))  # type: ignore[union-attr]
```
MSeal: Maybe add a comment that `'type'` in the dict is for the schema type and not any language type hinting? Seeing it might be confusing without context right next to a type ignore call.

src/confluent_kafka/schema_registry/_async/avro.py
```diff
@@ -333,26 +343,26 @@ async def __serialize(self, obj: object, ctx: Optional[SerializationContext] = N
...
-            value = obj
+            value = obj  # type: ignore[assignment]
 
         if latest_schema is not None:
-            parsed_schema = await self._get_parsed_schema(latest_schema.schema)
+            parsed_schema = await self._get_parsed_schema(latest_schema.schema)  # type: ignore[arg-type]
             def field_transformer(rule_ctx, field_transform, msg): return (  # noqa: E731
                 transform(rule_ctx, parsed_schema, msg, field_transform))
-            value = self._execute_rules(ctx, subject, RuleMode.WRITE, None,
+            value = self._execute_rules(ctx, subject, RuleMode.WRITE, None,  # type: ignore[arg-type]
```
MSeal: I think we should spread this function args out per line if we're type commenting one. It read weird like this and I think is error prone to refactor bugs

src/confluent_kafka/schema_registry/_async/avro.py
```diff
@@ -333,26 +343,26 @@ async def __serialize(self, obj: object, ctx: Optional[SerializationContext] = N
...
-            parsed_schema = self._parsed_schema
+            parsed_schema = self._parsed_schema  # type: ignore[assignment]
 
         with _ContextStringIO() as fo:
             # write the record to the rest of the buffer
             schemaless_writer(fo, parsed_schema, value)
             buffer = fo.getvalue()
 
-            if latest_schema is not None:
+            if latest_schema is not None and ctx is not None and subject is not None:
```
MSeal, line 387: This changed the logic here. Are we certain it's a correct / tested change?

MSeal (changes requested):
A couple questions on the PR. We are introducing a lot of ignore comments in places like avro that feel off but maybe not worth tackling in this pass anyway.

src/confluent_kafka/schema_registry/_async/avro.py
naxinf: In practice, subject and version of SchemaReference never be None, but probably for historical reason (or tech debt) they have been typed as optional. I think updating the types will be a breaking change

Here, get_version() requires non-empty subject and version, so that's why I added the type ignore there. Alternatively we can do:
```
            if ref.subject is None or ref.version is None:
               # maybe log something
                continue
            referenced_schema = await schema_registry_client.get_version(ref.subject, ref.version, True)
```

src/confluent_kafka/schema_registry/__init__.py
```diff
@@ -72,58 +72,82 @@
...
-            operation.
+            operation. **Required** - will raise ValueError if None.
 
         record_name (Optional[str]): Record name.
 
+    Raises:
+        ValueError: If ctx is None.
+
     """
+    if ctx is None:
```
naxinf, line 89: I refactored the name_strategy functions a bit: they have to follow the same function signature but SerializationContext is only required for some

src/confluent_kafka/schema_registry/_async/avro.py
naxinf: _execute_rules_with_phase() already requires SerializationContext and subject to build RuleContext (https://github.com/confluentinc/confluent-kafka-python/blob/f026e726c1dd1351e04d9b10679103322bf2d6b7/src/confluent_kafka/schema_registry/_async/serde.py#L70)

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
```diff
@@ -281,43 +285,43 @@ def __init__(self, conf: dict):
...
-                    self.max_retries, self.retries_wait_ms,
-                    self.retries_max_wait_ms)
-            elif self.bearer_auth_credentials_source == 'STATIC_TOKEN':
-                if 'bearer.auth.token' not in conf_copy:
-                    raise ValueError("Missing bearer.auth.token")
-                static_token = conf_copy.pop('bearer.auth.token')
-                self.bearer_field_provider = _StaticFieldProvider(static_token, logical_cluster, identity_pool)
-                if not isinstance(static_token, string_type):
-                    raise TypeError("bearer.auth.token must be a str, not " + str(type(static_token)))
+                if self.bearer_auth_credentials_source == 'OAUTHBEARER':
```
naxinf, line 288: Moving the indentation is intended to address the type issue of logical_cluster and identity_pool for building _AsyncOAuthClient: they must be non-empty, which we already check in line 280 and 284

This doesn't affect code logic: we check self.bearer_auth_credentials_source in {'OAUTHBEARER', 'STATIC_TOKEN'} for the outer block, and those if-else branches are for OAUTHBEARER and STATIC_TOKEN respectively

src/confluent_kafka/schema_registry/common/json_schema.py
```diff
@@ -166,29 +184,35 @@ def _validate_subschemas(
...
+                    subschema = resolver.lookup(ref).contents
+                validate(instance=message, schema=subschema, registry=registry, resolver=resolver)
+                return subschema
+            except ValidationError:
+                pass
     return None
 
 
 def get_type(schema: JsonSchema) -> FieldType:
-    if isinstance(schema, list):
```
naxinf, line 184: JsonSchema union type is either bool or dict. This was likely coped from avro.py

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
```diff
@@ -281,43 +285,43 @@ def __init__(self, conf: dict):
...
+                    if not isinstance(self.token_endpoint, string_type):
+                        raise TypeError("bearer.auth.issuer.endpoint.url must be a str, not "
+                                        + str(type(self.token_endpoint)))
+
+                    self.bearer_field_provider = _AsyncOAuthClient(
+                        self.client_id, self.client_secret, self.scope,
+                        self.token_endpoint, logical_cluster, identity_pool,
+                        self.max_retries, self.retries_wait_ms,
+                        self.retries_max_wait_ms)
+                else: # STATIC_TOKEN
```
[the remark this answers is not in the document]
naxinf: It's already the outer block `if self.bearer_auth_credentials_source in {'OAUTHBEARER', 'STATIC_TOKEN'}:` (line 273)

src/confluent_kafka/schema_registry/_sync/schema_registry_client.py
```diff
@@ -281,43 +285,43 @@ def __init__(self, conf: dict):
...
+                    if not isinstance(self.token_endpoint, string_type):
+                        raise TypeError("bearer.auth.issuer.endpoint.url must be a str, not "
+                                        + str(type(self.token_endpoint)))
+
+                    self.bearer_field_provider = _OAuthClient(
+                        self.client_id, self.client_secret, self.scope,
+                        self.token_endpoint, logical_cluster, identity_pool,
+                        self.max_retries, self.retries_wait_ms,
+                        self.retries_max_wait_ms)
+                else: # STATIC_TOKEN
```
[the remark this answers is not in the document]
naxinf: It's already the outer block `if self.bearer_auth_credentials_source in {'OAUTHBEARER', 'STATIC_TOKEN'}:`

src/confluent_kafka/schema_registry/__init__.py
```diff
@@ -72,58 +72,82 @@
...
+        raise ValueError(
+            "SerializationContext is required for topic_record_subject_name_strategy. "
```
rayokota:
```suggestion
            "SerializationContext is required for topic_name_strategy and topic_record_subject_name_strategy. "
```

src/confluent_kafka/admin/__init__.py
```diff
@@ -653,7 +653,7 @@ def list_topics(self, *args: Any, **kwargs: Any) -> ClusterMetadata:
         return super(AdminClient, self).list_topics(*args, **kwargs)
 
     def list_groups(self, *args: Any, **kwargs: Any) -> List[GroupMetadata]:
-
+
```
MSeal: Extra newline

src/confluent_kafka/schema_registry/_async/json_schema.py
```diff
@@ -341,27 +358,33 @@ async def __serialize(self, obj: object, ctx: Optional[SerializationContext] = N
 
             self._known_subjects.add(subject)
 
+        value: Any
```
MSeal, line 364: this line feels wrong, but alternatives seem just as bad since value = obj might happen

src/confluent_kafka/schema_registry/_async/json_schema.py
```diff
@@ -623,29 +657,33 @@ async def __deserialize(self, data: bytes, ctx: Optional[SerializationContext] =
...
+            def field_transformer(rule_ctx, field_transform, message): return (  # noqa: E731
+                transform(rule_ctx, reader_schema, reader_ref_registry, reader_ref_resolver,
+                          "$", message, field_transform))
+            if ctx is not None and subject is not None:
+                obj_dict = self._execute_rules(ctx, subject, RuleMode.READ, None,
+                                               reader_schema_raw, obj_dict, None,
+                                               field_transformer)
 
-        if self._validate:
+        if self._validate and reader_schema_raw is not None and reader_schema is not None and reader_ref_registry is not None:
```
MSeal: I wonder if we should log if one is None before moving to the self._from_dict check? If these should just always be set nbd, but then why check here?

src/confluent_kafka/schema_registry/_async/mock_schema_registry_client.py
```diff
@@ -212,7 +212,10 @@ async def lookup_schema(
 
         raise SchemaRegistryError(404, 40400, "Schema Not Found")
 
-    async def get_subjects(self) -> List[str]:
+    async def get_subjects(
```
MSeal, line 215: This seems wrong, we don't use any of the arguments in the function

src/confluent_kafka/schema_registry/_async/mock_schema_registry_client.py
```diff
@@ -236,30 +239,36 @@ async def get_latest_with_metadata(
         raise SchemaRegistryError(404, 40400, "Schema Not Found")
 
     async def get_version(
-        self, subject_name: str, version: int,
+        self, subject_name: str, version: Union[int, Literal["latest"]] = "latest",
```
MSeal, line 258: Well that type hint reveals an interesting choice for defaults in get_version

src/confluent_kafka/schema_registry/_async/mock_schema_registry_client.py
naxinf: I think we should maintain the consistency of signatures in actual functions and mocked ones, so I updated the mock functions to use them + add comment for unsupported

src/confluent_kafka/schema_registry/_async/json_schema.py
naxinf: Added logs

src/confluent_kafka/schema_registry/_async/mock_schema_registry_client.py
naxinf: Yes, it's a special api design :)

src/confluent_kafka/schema_registry/common/json_schema.py
```diff
@@ -166,29 +184,35 @@ def _validate_subschemas(
...
+    if isinstance(schema, bool):
         return FieldType.COMBINED
-    elif isinstance(schema, dict):
-        schema_type = schema.get("type")
-    else:
-        # string schemas; this could be either a named schema or a primitive type
-        schema_type = schema
 
+    schema_type = schema.get("type")
```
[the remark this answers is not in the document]
naxinf: Hallucinating - it's literally returning in line 213

MSeal (approved):

Issue #807, opened 2020-03-22
Wakeable poll


royreznik:
Description
===========
When i poll a message, if there is no message, i can not terminate the process.
Would it be intersting to wrap the polling with wakeable poll?

like wraping this function:
```c
rkm = rd_kafka_consumer_poll(self->rk, cfl_timeout_ms(tmout));
```
inside a `while start_time + timeout > current_time` statement, and put a little timout inside,
so it will be wakeable?

Interesting for a PR?

edenhill:
We want to address this problem in librdkafka, allowing all blocking calls to be cancelled, something we'll look into later this year.

Issue #209, opened 2017-06-21
Ctrl + C can not terminate the program 


15663671003:
Ctrl + C can not terminate the program.
The program may block the signal set inside
may be blocked in the consumer.poll () call, this method to solve

DXist:
+1

15663671003:
Set consumer.poll (timeout = 1) can be resolved

DXist:
Yes, it solves the problem. This issue is more about improving the documentation.

edenhill:
There is currently no way to cancel a blocking call into librdkafka, when this is added to librdkafka we'll make the proper changes to the Python module to fix it.

fpietka:
@edenhill is there by any chance any related issue in edenhill/librdkafka ?

edenhill:
The closest thing would be https://github.com/edenhill/librdkafka/issues/614

royreznik:
Is there anythin that going to be done about this?

edenhill:
Yes, but it is quite involved and requires changes to librdkafka.

PR #2126, opened 2025-11-07
Make long waiting consumer and producer API wakeable

k-raina:
## Summary

This PR fixes a critical usability issue where `Consumer.poll()`, `Consumer.consume()`, `Producer.poll()`, and `Producer.flush()` would block indefinitely and not respond to Ctrl+C (KeyboardInterrupt) signals.

**Fixes:** [#209](https://github.com/confluentinc/confluent-kafka-python/issues/209) and [#807](https://github.com/confluentinc/confluent-kafka-python/issues/807)

## Problem

When calling blocking operations with infinite or very long timeouts:
- `Consumer.poll()` / `Consumer.consume()` - would block indefinitely waiting for messages
- `Producer.poll()` / `Producer.flush()` - would block indefinitely waiting for delivery callbacks or queue flushing

Because these operations release the Python Global Interpreter Lock (GIL) during blocking calls to the underlying librdkafka C library, Python's signal handling mechanism couldn't detect Ctrl+C signals, making it impossible to gracefully interrupt these operations.

## Solution

The fix implements a "wakeable poll" pattern that:

1. **Chunks long timeouts**: Instead of making a single blocking call with the full timeout, the implementation breaks it into smaller 200ms chunks

2. **Periodic signal checking**: Between chunks, the code re-acquires the Python GIL and calls `PyErr_CheckSignals()` to detect pending KeyboardInterrupt signals

3. **Immediate interruption**: If a signal is detected, the operation returns immediately, allowing the KeyboardInterrupt to propagate to the Python code

## Impact

This fix significantly improves the developer and operational experience for applications using  Python client.

**Before this fix:**
- Developers had to forcefully kill processes during development and testing
- No way to gracefully stop consumer/producer loops during debugging
- Producer `flush()` operations couldn't be interrupted, causing issues during shutdown

**After this fix:**
- Developers can use standard Ctrl+C to interrupt all blocking operations
- Clean shutdown during development and testing

### Consumer Tests (`tests/test_Consumer.py`)

1. **Utility function tests**:
   - `test_calculate_chunk_timeout_utility_function()`: Tests chunk timeout calculation logic
   - `test_check_signals_between_chunks_utility_function()`: Tests signal detection between chunks
   - `test_wakeable_poll_utility_functions_interaction()`: Tests interaction between both utilities

2. **Poll interruptibility tests**:
   - `test_wakeable_poll_interruptibility_and_messages()`: Tests poll() can be interrupted and still handles messages correctly
   - `test_wakeable_poll_edge_cases()`: Tests edge cases (zero timeout, closed consumer, short timeouts)

3. **Consume interruptibility tests**:
   - `test_wakeable_consume_interruptibility_and_messages()`: Tests consume() can be interrupted and still handles messages correctly
   - `test_wakeable_consume_edge_cases()`: Tests edge cases (zero timeout, invalid parameters, short timeouts)

### Producer Tests (`tests/test_Producer.py`)

1. **Utility function tests**:
   - `test_wakeable_poll_utility_functions_interaction()`: Tests chunk calculation and signal checking work together

2. **Poll interruptibility tests**:
   - `test_wakeable_poll_interruptibility_and_messages()`: Tests poll() can be interrupted and still handles delivery callbacks correctly
   - `test_wakeable_poll_edge_cases()`: Tests edge cases (zero timeout, closed producer, short timeouts)

3. **Flush interruptibility tests**:
   - `test_wakeable_flush_interruptibility_and_messages()`: Tests flush() can be interrupted and still delivers messages correctly
   - `test_wakeable_flush_edge_cases()`: Tests edge cases (zero timeout, closed producer, short timeouts, empty queue)

### Shared Utility Tests (`tests/test_wakeable_utilities.py`)
- Tests for newly added utility functions 

### Integration Tests

- `tests/integration/consumer/test_consumer_wakeable_poll_consume.py`: Verifies wakeable pattern doesn't interfere with normal message consumption
- `tests/integration/producer/test_producer_wakeable_poll_flush.py`: Verifies wakeable pattern doesn't interfere with normal message production and delivery callbacks

All tests use a helper function `TestUtils.send_sigint_after_delay()` to simulate Ctrl+C in automated tests.

## Performance Impact
- **Minimal overhead**: Only adds signal checks between 200ms chunks
- **No impact on short timeouts**: Timeouts < 200ms bypass chunking entirely

### Replicable code
[test_wakeable_consume_interrupt.py](https://github.com/user-attachments/files/23774411/test_wakeable_consume_interrupt.py)
[test_wakeable_poll_interrupt.py](https://github.com/user-attachments/files/23774413/test_wakeable_poll_interrupt.py)
[test_wakeable_producer_flush_interrupt.py](https://github.com/user-attachments/files/23774414/test_wakeable_producer_flush_interrupt.py)
[test_wakeable_producer_poll_interrupt.py](https://github.com/user-attachments/files/23774415/test_wakeable_producer_poll_interrupt.py)

### Example Run
```
 Queue before flush: 101000 messages
  (Many messages are in queue and in-flight, waiting for acknowledgment)

[20:45:54] Calling flush() with infinite timeout...
    (This will block until all messages are flushed or Ctrl+C)
    Background thread is continuously adding messages to the queue
    Current queue length: 101000 messages
    Press Ctrl+C to test interruptibility...

^C
======================================================================
✓ KeyboardInterrupt caught!
  Interrupted after 1.07 seconds
  Background thread produced 1167722 messages during this time
  (Had 53 production errors)
  With wakeable pattern, interruption should occur within ~200ms of Ctrl+C
  ⚠ Interruption took 1.07s (may indicate wakeable pattern issue)
======================================================================

Stopping background producer thread...
  Final stats: 1167722 messages produced
  (53 production errors)

Closing producer...
Producer closed.
  File "/Users/kaushikraina/projects/njc/confluent-kafka-python/test_wakeable_producer_flush_interrupt.py", line 292, in <module>
    main()
  File "/Users/kaushikraina/projects/njc/confluent-kafka-python/test_wakeable_producer_flush_interrupt.py", line 223, in main
    remaining = producer.flush(timeout=-1.0)  # Infinite timeout - will block
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt

```


tests/test_Consumer.py
```diff
@@ -1,14 +1,35 @@
...
+    Utility function for testing interruptible poll/consume operations.
+    Used to simulate Ctrl+C in automated tests.
+    
+    Args:
+        delay_seconds: Delay in seconds before sending SIGINT
+    """
+    time.sleep(delay_seconds)
+    try:
+        os.kill(os.getpid(), signal.SIGINT)
+    except Exception:
```
MSeal: Should probably catch the KeyboardInterrupt here instead so you don't mask other errors being raised instead

MSeal (approved):
One minor comment, tested out things locally with a 30s timeout with this vs master... much better experience of always being able to interrupt.

tests/test_Consumer.py
k-raina: Thanks
`os.kill()` raises OSError, not KeyboardInterrupt, i have removed try catch block as any exception raised in this function should not be masked

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -536,6 +536,88 @@ static CFL_UNUSED CFL_INLINE int cfl_timeout_ms(double tmout) {
...
+                /* Note: GIL is already held, but CallState_end expects to
+                 * restore it */
+                /* Save thread state again so CallState_end can restore it
+                 * properly */
+                cs->thread_state = PyEval_SaveThread();
+                if (!CallState_end(self, cs)) {
+                        /* CallState_end detected signal and cleaned up */
+                        return 1; /* Signal detected */
+                }
+                return 1;
```
MSeal, line 608: We simply always return 1 -- was this intentional? Seems like we'd either have a return 0 in one call path or cleanup the code to just call CallState_end and then return no matter what

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -536,6 +536,88 @@ static CFL_UNUSED CFL_INLINE int cfl_timeout_ms(double tmout) {
...
+        }
+
+        /* Re-release GIL for next iteration */
+        cs->thread_state = PyEval_SaveThread();
```
MSeal: Maybe put this in an else block so you have clear control paths for calling `PyEval_SaveThread` appropriately if things get refactored. I've found those explicit conditional blocks help a lot down the line for clarity of intent after additional changes

src/confluent_kafka/src/Consumer.c
```diff
@@ -105,6 +105,16 @@ static int Consumer_traverse(Handle *self, visitproc visit, void *arg) {
 }
 
 
+/****************************************************************************
+ *
+ *
+ * Consumer Methods
```
MSeal: If we're going to have a giant comment block it maybe should be more descriptive. Though I'm not sure it's highly valuable here to do this anyway

src/confluent_kafka/src/Consumer.c
```diff
@@ -974,16 +1007,52 @@ static PyObject *Consumer_poll(Handle *self, PyObject *args, PyObject *kwargs) {
...
+                        if (check_signals_between_chunks(self, &cs)) {
+                                return NULL;
+                        }
+                }
+        }
+
+        /* Final GIL restore and signal check */
         if (!CallState_end(self, &cs)) {
                 if (rkm)
                         rd_kafka_message_destroy(rkm);
```
MSeal, line 1046: I always found C code less risky to refactor errors with single line if's include wrapping brackets, but I don't believe we follow that convention here as of now. If we're doing that convention within a function let's be consistent to the rest of the calls in the same scope imo

src/confluent_kafka/src/Consumer.c
```diff
@@ -1024,7 +1093,27 @@ Consumer_memberid(Handle *self, PyObject *args, PyObject *kwargs) {
         return memberidobj;
 }
 
-
+/**
+ * @brief Consume a batch of messages from the subscribed topics.
+ *
+ * Instead of a single blocking call to rd_kafka_consume_batch_queue() with the
```
MSeal, line 1095: Nothing to change but commenting that just enough code differences that it's probably best left separate even though it looks similar enough to share more code.

MSeal (dismissed):
Minor changes requested but up to you if you feel they are worth it. Test blocks were a bit dense/difficult to read after a long weekend losing mental context. I'm not sure how to make them easier to maintain but the next person coming along who has to fix those for future change will need to reread them a few times I think.

src/confluent_kafka/src/Consumer.c
k-raina: Agree, i tried share more code into a common function, but it looked awful, so decided to keep it separate.

k-raina:
Thanks @MSeal for the review.

I've addressed the minor comments.

Regarding testing for the Wakeable feature:
- Consolidated unit tests into a single file `tests/test_Wakeable.py` with detailed comments explaining the testing strategy for both Producer and Consumer operations using signal injection (SIGINT) to simulate Ctrl+C behavior.
- Added testing strategy comments to integration test files (`test_consumer_wakeable_poll_consume.py` and `test_producer_wakeable_poll_flush.py`) that document how wakeability is tested with real Kafka clusters and message delivery scenarios.

Issue #1535, opened 2023-03-21
feature request: make Message type user-instantiable


chairmank:
Description
===========
This is an enhancement request to make it easier to write tests for user code that uses this client library. In unit tests, it would useful to be create synthetic `Message` instances without having to read them from a live Kafka cluster with a `Consumer`.

The current implementation of the `Message` type does not make it possible to instantiate a `Message` because the `PyTypeObject MessageType` definition does not specify the `tp_new` and `tp_init` fields.

In contrast, the `TopicPartitionType` definition does specify these fields, and we can instantiate a `TopicPartition`.

How to reproduce
================
Using confluent-kafka-python v2.0.2, I observe the following when I try to call the `__init__` method of `Message`:

```
>>> import confluent_kafka
>>> confluent_kafka.Message()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: cannot create 'cimpl.Message' instances
```

keithks:
Agree it will be more convenient, we are working around as the below:

```
    mock_message_class = mocker.patch('confluent_kafka.Message')
    mock_message = mock_message_class.return_value
    mock_message.key.return_value = serialized_key
    mock_message.value.return_value = serialized_value
   .....
```

joaoe:
Hi.
I just came here to report the same. 
I had to work around by creating my own mock Message class that reimplements the whole interface for tests, but that is unproductive if every one has to do the same.

ouphi:
+1 I agree I have the same problem

mesner:
+1 I have the same problem

armarik:
+1 I also have the same problem

Floodnut:
+1 I have the same problem.

keysersoza:
+1 same issue

AlessioPeluso:
+1

westover:
+1

nhelsherif:
+1

tmanninger:
+1

MSeal:
Agreed this is a large pain point. I'm marking this for short list of fixes that should be worked on

ojasvajain:
This issue has been fixed in #2128. Closing this issue.

PR #2128, opened 2025-11-10
Make Message Object Instantiable By Users

ojasvajain:
Overview
----
1. Fixes issue #1535 by allowing `Message` objects to be instantiable by users.
2. Modify type hints of `topic()`, `partition()` and `offset()` getters to make them consistent with their implementations ([ref](https://github.com/confluentinc/confluent-kafka-python/blob/36a9e6c8aaaeb1b7d395c68e84049b11c0599d9f/src/confluent_kafka/src/confluent_kafka.c#L463))

Problem
-----
Users were unable to create synthetic `Message` objects in their unit tests and had to resort to reading Message objects from a live Kafka cluster.

```
>>> import confluent_kafka
>>> confluent_kafka.Message()
TypeError: cannot create 'cimpl.Message' instances
```

Solution
----

A constructor has been defined in the Message class. All params have been intentionally kept optional to support easy instantiation without providing any params. This way of creating new messages is expected to be used only while writing unit tests. Producers and consumers will continue to use the internal factory method (`Message_new0`) for creating message objects.

```
>>> import confluent_kafka
>>> confluent_kafka.Message()
<cimpl.Message object at 0x101301640>
>>> 
```

Checklist
------------------

References
----------
https://github.com/confluentinc/confluent-kafka-python/issues/1535

### DIFF
--- a/tests/test_message.py
+++ b/tests/test_message.py
@@ -16,8 +16,10 @@ def test_init_no_params():
     assert m.latency() is None
     assert m.leader_epoch() is None
 
+
 def test_init_all_params():
-    m = Message(topic="test", partition=1, offset=2, key=b"key", value=b"value", headers=[("h1", "v1")], error=KafkaError(0),
+    m = Message(topic="test", partition=1, offset=2, key=b"key", value=b"value", headers=[("h1", "v1")],
+                error=KafkaError(0),
                 timestamp=(1, 1762499956), latency=0.05, leader_epoch=1762499956)
     assert m.topic() == "test"
     assert m.partition() == 1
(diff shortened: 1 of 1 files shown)


pranavrth:
Message class was intentionally like this because we didn't want user to instantiate a message. There are some APIs which take message as input like `commit` and `store_offsets`.

We can choose from the below:
- We provide a MockMessage class instead of making it instantiable.
- We add make this instantiable in other clients as well and test all the APIs that uses messages as input.

MSeal:
@pranavrth I advised against the Mock providing here as it won't solve the root issues and this is a non-default sharp edge that the user can be responsible for if they misuse. I suggested we either do checks on the init or checks in method uses to ensure bad inputs raises a clean exception if we want to prevent deep failures from bad inputs. But I also imagine 90% of the cases here will be user knowingly instantiating an object they know isn't real for e2e patterns.

Issue #1713, opened 2024-02-21
AdminClient is stuck at `CONNECT` when using `OAUTHBEARER` with `SASL_SSL` protocol


kumarappan-arumugam:
Description
===========
Admin client cannot connect to the bootstrap servers when using `SASL_SSL` with `OAUTHBEARER`.
I'm trying to connect to AWS MSK cluster bootstrap servers. Admin client can connect when using an unauthenticated endpoint. Producer and Consumer config work fine with `OAUTHBEARER`.

How to reproduce
================

```
>>> from confluent_kafka.admin import AdminClient, NewTopic
>>> from aws_msk_iam_sasl_signer import MSKAuthTokenProvider
>>> import socket

>>> def oauth_cb(oauth_config):
...     auth_token, expiry_ms = MSKAuthTokenProvider.generate_auth_token("region")
...     return auth_token, expiry_ms/1000

>>> config = {
...     'bootstrap.servers': 'xxx-1.kafka.us-west-2.amazonaws.com:9098,xxx-2.kafka.us-west-2.amazonaws.com:9098',
...     'security.protocol': 'SASL_SSL',
...     'sasl.mechanism': 'OAUTHBEARER',
...     'oauth_cb': oauth_cb,
... }

>>> admin = AdminClient(config)
>>> admin.list_topics(timeout=2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/venv3/lib/python3.8/site-packages/confluent_kafka/admin/__init__.py", line 603, in list_topics
    return super(AdminClient, self).list_topics(*args, **kwargs)
cimpl.KafkaException: KafkaError{code=_TRANSPORT,val=-195,str="Failed to get metadata: Local: Broker transport failure"}
```

Debug logs:
```
>>> config = {
...     'bootstrap.servers': 'xxx-1.kafka.us-west-2.amazonaws.com:9098,xxx-2.kafka.us-west-2.amazonaws.com:9098',
...     'security.protocol': 'SASL_SSL',
...     'sasl.mechanism': 'OAUTHBEARER',
...     'oauth_cb': oauth_cb,
...     'debug': 'all'
... }
>>> admin = AdminClient(config)
%7|1708548520.827|SASL|rdkafka#producer-2| [thrd:app]: Selected provider OAUTHBEARER (builtin) for SASL mechanism OAUTHBEARER
[...]

PR #2130, opened 2025-11-13
Handle OAuth Token Refreshes Using Background Thread For Admin, Producer and Consumer Clients

ojasvajain:
Overview
----
1. Fixes issue #1713 by delegating OAuth token refresh events to be handled by the background thread.
2. This has been implemented for Admin, Producer and Consumer clients to make behaviour consistent with the Java client.

Note that, `poll()` will not trigger the OAuth refresh callbacks anymore. 

Problem
-----
When providing `sasl.mechanisms` as `OAUTHBEARER` and a `oauth_cb` in the conf, users were unable to make API calls like `list_topics()`, `create_topics()` unless `poll` was called.
```
>>> from confluent_kafka.admin import AdminClient, NewTopic
>>> from aws_msk_iam_sasl_signer import MSKAuthTokenProvider
>>> import socket

>>> def oauth_cb(oauth_config):
...     auth_token, expiry_ms = MSKAuthTokenProvider.generate_auth_token("region")
...     return auth_token, expiry_ms/1000

>>> config = {
...     'bootstrap.servers': 'xxx-1.kafka.us-west-2.amazonaws.com:9098,xxx-2.kafka.us-west-2.amazonaws.com:9098',
...     'security.protocol': 'SASL_SSL',
...     'sasl.mechanism': 'OAUTHBEARER',
...     'oauth_cb': oauth_cb,
... }

>>> admin = AdminClient(config)
>>> admin.list_topics(timeout=2)
  File "/venv3/lib/python3.8/site-packages/confluent_kafka/admin/__init__.py", line 603, in list_topics
cimpl.KafkaException: KafkaError{code=_TRANSPORT,val=-195,str="Failed to get metadata: Local: Broker transport failure"}
```
(Snippet Source: https://github.com/confluentinc/confluent-kafka-python/issues/1713#issue-2147735306)

This happens because OAuth token refresh events get pushed to the main queue which is listened by poll(). These events are responsible for triggering the configured OAuth callback function and setting the token in the request. If `poll()` is not called before `list_topics()`, the token would not be set in the request, causing the request to fail.

Solution
----
Changes have been made to handle these refresh events in the background thread using the following librdkafka method calls:

1. `rd_kafka_conf_enable_sasl_queue(conf, 1)` - ensures all refresh events get pushed to the SASL queue instead of main queue

2. `rd_kafka_sasl_background_callbacks_enable(self->rk)` - ensures all events in SASL queue are forwarded to the background queue.

By using these 2 method calls, we have been able to ensure that OAuth callbacks get triggered from the background thread instead of `poll()`

Checklist
------------------

### DIFF
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -99,12 +99,12 @@ struct Admin_options {
  * Make sure this is kept up to date with Admin_options above. */
 #define Admin_options_INITIALIZER                                              \
         {                                                                      \
-            Admin_options_def_int,   Admin_options_def_float,                  \
-            Admin_options_def_float, Admin_options_def_int,                    \
-            Admin_options_def_int,   Admin_options_def_int,                    \
-            Admin_options_def_int,   Admin_options_def_ptr,                    \
-            Admin_options_def_cnt,   Admin_options_def_ptr,                    \
-            Admin_options_def_cnt,                                             \
+                Admin_options_def_int, Admin_options_def_float,                \
+                    Admin_options_def_float, Admin_options_def_int,            \
+                    Admin_options_def_int, Admin_options_def_int,              \
+                    Admin_options_def_int, Admin_options_def_ptr,              \
+                    Admin_options_def_cnt, Admin_options_def_ptr,              \
+                    Admin_options_def_cnt,                                     \
         }
 
 #define Admin_options_is_set_int(v)   ((v) != Admin_options_def_int)


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2053,11 +2053,56 @@ static int py_extensions_to_c (char **extensions, Py_ssize_t idx,
...
+ * Useful during client init as we want to ensure we have the token before we return back
+ *
+ * Returns 0 if token was set within the timeout period, -1 otherwise.
+ */
+int wait_for_oauth_token_set(Handle *h) {
+
+    if (!h->oauth_cb)
+        return 0;
+
+    int max_wait_sec = 5;
```
MSeal: Where did we pick this timeout? I'm not sure what the longest oauth wait time will be. Maybe this should be a bit longer in case the loopback is slow?

MSeal (commented):
Wanted to check on expected timeout here and where we picked the value first, but otherwise lgtm

src/confluent_kafka/src/confluent_kafka.c
ojasvajain: No particular reason to keep it 5 sec. It should be fine to increase it (say, to 10s) as this init is a one time thing and it will help reduce flakiness for clients with slower callback functions. For clients with faster callbacks, this shouldn't be a problem as we are breaking out of the loop if callback succeeds early.

ojasvajain: Increased to 10s

PR #1586, opened 2023-06-15
make Message type user-instantiable

joaoe:
Closes #1535 and #1585

PR #2153, opened 2025-12-08
Enhance message class to include serialisation support and rich comparison

k-raina:
### Summary
* Added `__reduce__` methods to both `Message` and `KafkaError` classes, enabling reliable serialization and deserialization using Python's pickle protocol. 
* Introduced new setter methods `set_topic` and `set_error` for the `Message` class, allowing direct modification of these fields after object creation.
* Implemented a custom rich comparison method (`__eq__`/`__ne__`) for the `Message` class, enabling meaningful equality checks based on all relevant fields.

* Updated the type names for both `Message` and `KafkaError` to use the fully qualified `confluent_kafka.cimpl` namespace, ensuring consistency and clarity in type identification.
* Refactored `tests/test_message.py` suite to cover object creation, property setters, pickling, comparison, and error handling. This ensures robust validation of all new and existing features.

Cherrypicked From : https://github.com/confluentinc/confluent-kafka-python/pull/1586

### DIFF
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -149,7 +149,6 @@ static PyObject *KafkaError_reduce(KafkaError *self,
         }
         if (!args) {
                 Py_DECREF(KafkaError_type);
-                Py_DECREF(reason);
                 return NULL;
         }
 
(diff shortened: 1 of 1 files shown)


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -540,6 +579,61 @@ static PyObject *Message_set_key(Message *self, PyObject *new_key) {
...
+                 * default */
+                latency_obj = PyFloat_FromDouble(-1.0);
+        }
+        result = Py_BuildValue(
+            "O(NiLNNNNOOi)", Message_type, Message_topic(self, NULL),
+            self->partition, self->offset, Message_key(self, NULL),
+            Message_value(self, NULL), Message_headers(self, NULL),
+            Message_error(self, NULL), Message_timestamp(self, NULL),
+            latency_obj, self->leader_epoch);
+        Py_DECREF(latency_obj);
```
MSeal, line 630: Technically you want try-finally the deref so you don't leak on error but memory leaking during value build isn't really important to protect against.

k-raina: Thanks! The code already decrements `latency_obj` unconditionally, so there's no leak even if `Py_BuildValue` fails.

joaoe:
@k-raina Hi! Thank you so much for merging this. I've move to other projects not using confluent kafka so my availability is none :)

k-raina:
Thanks for update @joaoe and appreciate your initial PR!

Issue #2155, opened 2025-12-16
v2.13.0 beta feedback


Tenzer:
I have upgraded our use of this library to v2.13.0b1, mainly to get the background freshing of OAuth tokens - which appear to work well.

The type annotations, specifically for the admin client, however don't work as well. It seems like mypy v1.19.1 uses the type annotations in `src/confluent_kafka/cimpl.pyi` rather than what are exposed in `src/confluent_kafka/admin/__init__.py`, which creates a mismatch between what types it thinks should be used and which ones the Python code actually has.

See for instance this example:
```py
from confluent_kafka.admin import AdminClient

with AdminClient({}) as admin:
    futures = admin.delete_topics(["foo", "bar"])

    for future in futures.values():
        future.result()
```

Running mypy on this file brings up the following errors:
```
delete_topics.py:4: error: Missing positional argument "future" in call to "delete_topics" of "_AdminClientImpl"  [call-arg]
delete_topics.py:4: error: "delete_topics" of "_AdminClientImpl" does not return a value (it only ever returns None)  [func-returns-value]
```

This is because the type annotations are used from here:
https://github.com/confluentinc/confluent-kafka-python/blob/b6637b3772200b1fb1133819677f6e68bea990f1/src/confluent_kafka/cimpl.pyi#L232-L234

Rather than here:
https://github.com/confluentinc/confluent-kafka-python/blob/b6637b3772200b1fb1133819677f6e68bea990f1/src/confluent_kafka/admin/__init__.py#L627-L629

Given that the C implementation of the admin client is marked as being private (it has a leading underscore), I wonder if the easiest way to fix this would be to rename the functions in the C implementation to, for instance, have a leading underscore as well, and call those methods from `AdminClient`. This means the method names between `_AdminClientImpl` and `AdminClient` won't clash, so there's no confusion as to which type annotations should be used.

ojasvajain:
Hi @Tenzer, thanks for sharing your feedback. I was able to trace the root cause, and it turns out this issue occurs only when we instantiate the AdminClient using a with clause. Looking further, I found that the __enter__() method in cimpl.pyi is type-hinted to return an _AdminClientImpl object. As a result, mypy applies the type annotations from _AdminClientImpl to delete_topics, rather than those from AdminClient. 

Let me get a fix out for this.

Tenzer:
Ah, good catch. I hadn't considered that switching to use the context managers as part of the upgrade would cause this.

[...]

PR #2157, opened 2025-12-17
Type hint Admin Client's __enter__() to return the same object type that called it

ojasvajain:
### Overview
Addresses issue #2155 by fixing type hint for _AdminClientImpl.__enter__() to return Self

### Problem
When using AdminClient with a context manager (the with statement), mypy incorrectly narrows the type from AdminClient to _AdminClientImpl, causing type checking errors for methods that are overridden in the AdminClient subclass.

```
  with AdminClient(config) as admin:
      # mypy thinks 'admin' is _AdminClientImpl, not AdminClient
      admin.delete_topics(['topic'])  # Type error: wrong signature
```

This happens because _AdminClientImpl.__enter__() was annotated to return "_AdminClientImpl" instead of the actual type of the instance that called it.

### Solution
 Changed the return type annotation to use Self:
```
  class _AdminClientImpl:
      def __enter__(self) -> Self: ...
```
This tells mypy: "return the same type that called __enter__"


k-raina (commented):
Do we need same for [Producer](https://github.com/confluentinc/confluent-kafka-python/blob/9809c1acb82cfe7082e5c0d243c0a30b1152cc01/src/confluent_kafka/cimpl.pyi#L209) and [Consumer](https://github.com/confluentinc/confluent-kafka-python/blob/9809c1acb82cfe7082e5c0d243c0a30b1152cc01/src/confluent_kafka/cimpl.pyi#L209)

ojasvajain:
> Do we need same for [Producer](https://github.com/confluentinc/confluent-kafka-python/blob/9809c1acb82cfe7082e5c0d243c0a30b1152cc01/src/confluent_kafka/cimpl.pyi#L209) and [Consumer](https://github.com/confluentinc/confluent-kafka-python/blob/9809c1acb82cfe7082e5c0d243c0a30b1152cc01/src/confluent_kafka/cimpl.pyi#L209)


Not required for them as their instantiation directly calls the `_init` and `_new` methods defined in Producer.c and Consumer.c. For Admin clients, there is a wrapper class defined in admin/__init__.py which then calls the respective `_init_` and `_new` methods of `_AdminClientImpl`.

Issue #2168, opened 2026-01-08
Missing error constants in the version 2.13.0


jjuris:
Hello

It looks like the most recent version of **confluent-kafka 2.13.0** is missing several error constants in the generated [cimpl.pyi](https://github.com/confluentinc/confluent-kafka-python/blob/v2.13.0/src/confluent_kafka/cimpl.pyi) file.

```
running ['mypy', '--ignore-missing-imports', '--show-error-codes', 'src/kafkawrapper'] 
src/kafkawrapper/retrying.py:5: error: "type[KafkaError]" has no attribute "REQUEST_TIMED_OUT"  [attr-defined]
src/kafkawrapper/retrying.py:6: error: "type[KafkaError]" has no attribute "NETWORK_EXCEPTION"  [attr-defined]
src/kafkawrapper/retrying.py:7: error: "type[KafkaError]" has no attribute "LEADER_NOT_AVAILABLE"  [attr-defined]
src/kafkawrapper/retrying.py:8: error: "type[KafkaError]" has no attribute "BROKER_NOT_AVAILABLE"  [attr-defined]
src/kafkawrapper/retrying.py:9: error: "type[KafkaError]" has no attribute "INVALID_SESSION_TIMEOUT"  [attr-defined]
src/kafkawrapper/consumer.py:166: error: Argument 1 to "subscribe" of "Consumer" has incompatible type "tuple[Any, ...]"; expected "list[str]"  [arg-type]
src/kafkawrapper/consumer.py:239: error: "type[KafkaError]" has no attribute "_PARTITION_EOF"  [attr-defined]
src/kafkawrapper/consumer.py:240: error: "type[KafkaError]" has no attribute "_TRANSPORT"  [attr-defined]
src/kafkawrapper/consumer.py:243: error: "type[KafkaError]" has no attribute "UNKNOWN_TOPIC_OR_PART"  [attr-defined]
src/kafkawrapper/consumer.py:252: error: Argument 1 to "append" of "list" has incompatible type "Message"; expected "bytes | None"  [arg-type]
src/kafkawrapper/consumer.py:294: error: Exception type must be derived from BaseException (or be a tuple of exception classes)  [misc]
Found 11 errors in 2 files (checked 3 source files)
```

version **2.12.2** works correctly.

Could you please take a look at this problem?

Best regards,
Jozef

howamith:
[...]

PR #2172, opened 2026-01-12
Type Hint Fixes

ojasvajain:
### Overview

Fixes issue #2168 

#### Part - 1
1. Updates KafkaError stub class with all the error codes available at runtime. 
2. A script has been added in tools directory that can look at the complied C extension and generate the error codes supported. 
3. Added check in `source-package-verification.sh` that compares the error codes in stub class and the ones available at runtime to catch such issues later when upgrading `librdkafka`

#### Part - 2
Fixed type hints of some methods to align with their C implementation.

### Problem
There are 150+ error codes available in librdkafka but only 4 error codes are added in the KafkaError stub class. This class is used by mypy for static checking and to support type hints so the absence of majority of the error codes leads to mypy showing errors. For example:

```
if err.code() == KafkaError.SSL_HELPER_FAILED:
    # mypy error: "type[KafkaError]" has no attribute "SSL_HELPER_FAILED"
    pass
```

It was also found that type hints of a few methods was incorrect and did not align with their C implementation. For example: https://github.com/confluentinc/confluent-kafka-python/issues/2168#issuecomment-3725752736

### Solution
1. KafkaError stub class has been updated with all the up to date error codes. Newer versions of librdkafka can introduce changes to the error codes so care has to be taken to update the KafkaError stub class with the updated error codes. To facilitate this, a script has been added in tools directory that can generate the list of error codes available at runtime. The KafkaError stub class can then be updated with the generated error codes. A check has also been added in `source-package-verification.sh` so that the inconsistency gets caught when updating librdkafka.
2. Type hints of methods like `Consumer.__init()__, Producer.__init()__, Producer.produce(), Consumer.commit()` has been fixed.


docs/conf.py
```diff
@@ -74,9 +74,6 @@ def _read_version_from_pyproject(pyproject_path=None):
 # The suffix of source filenames.
 source_suffix = '.rst'
 
-# The encoding of source files.
```
ojasvajain, line 77: This is an intended change. If we don't remove this then Sphinx throws a warning while building the docs.

` WARNING: Support for source encodings other than UTF-8 is deprecated and will be removed in Sphinx 10. Please comment at 
https://github.com/sphinx-doc/sphinx/issues/13665 if this causes a problem`


`utf-8` is the only encoding supported in newer versions of Sphnix so this config has been deprecated.

k-raina (changes requested):
Couple of feedbacks: 
1. Please verify if docs are getting generated correctly after docs/conf.py change
2. Update CHANGELOG

ojasvajain:
@k-raina Thanks for your review. 

> Please verify if docs are getting generated correctly after docs/conf.py change

Compared the HTML output with and without the change. The docs seem identical.
image


> Update CHANGELOG

Done

PR #2174, opened 2026-01-13
Type hint __enter__ to return the same object type that called it

leifwar:
What
----
Have change the return type hint for  both `Producer.__enter__()` and `Consumer.__enter__()` to `Self`. This will make it possible to extend this class through inheritance. Related to #2157.

Checklist
------------------
- [x ] Did you add sufficient unit test and/or integration test coverage for this PR?

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


ojasvajain:
Hi, thanks for raising this PR. But we don't foresee the need to extend Producer and Consumer classes. The reason AdminClient class extends _AdminClientImpl is because AdminClient class takes care of managing futures. In case of Producer and Consumer clients, that will not be required as we now have separate Async IO Producer and Consumer clients. Closing this issue.

davidkleiven:
For testing purposes one could foresee this being useful. Say one would like to just have something that counted the number of produce calls (or whatever would be useful in the context of the unittest)

```python
# Pseudo-code
class CountingProducer(Producer):
     def __init__(...) -> None:
          self.num_calls = 0
     
     def produce(self, ...) -> cimpl.Message:
          self.num_calls += 1
```

Another example can be found in this project itself

https://github.com/confluentinc/confluent-kafka-python/blob/612349e0bb30188d767dfa05ed14d6bf5093cd91/src/confluent_kafka/deserializing_consumer.py#L28


There is also an example for a producer that is extended:

https://github.com/confluentinc/confluent-kafka-python/blob/612349e0bb30188d767dfa05ed14d6bf5093cd91/src/confluent_kafka/serializing_producer.py#L28

@ojasvajain with this PR the type-hints for your extended producer/consumers would be correct and it would give users to extend it (if they have would like).

PR #2182, opened 2026-02-02
DGS-23437: Enable client omit Confluent-Identity-Pool-Id 

tobiogunbi:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/examples/oauth_schema_registry.py
+++ b/examples/oauth_schema_registry.py
@@ -71,6 +71,8 @@ def main():
     custom_sr_client = SchemaRegistryClient(custom_sr_config)
     print(custom_sr_client.get_subjects())
 
+
+def identity_pool_examples():
     # Example: Using union-of-pools with comma-separated pool IDs
     union_of_pools_config = {
         'url': 'https://psrc-123456.us-east-1.aws.confluent.cloud',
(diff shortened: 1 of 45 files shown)


examples/oauth_schema_registry.py
```diff
@@ -71,6 +71,33 @@ def custom_oauth_function(config):
     custom_sr_client = SchemaRegistryClient(custom_sr_config)
     print(custom_sr_client.get_subjects())
 
+    # Example: Using union-of-pools with comma-separated pool IDs
```
naxinf, line 87: nit: I think this should be extracted into a separate function in example. Combining this with the existing one might be confusing, since it's currently building two separate SR clients + calling get_subjects() twice in one call

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
```diff
@@ -412,7 +409,8 @@ async def handle_bearer_auth(self, headers: dict) -> None:
         if self.bearer_field_provider is None:
             raise ValueError("Bearer field provider is not set")
         bearer_fields = await self.bearer_field_provider.get_bearer_fields()
-        required_fields = ['bearer.auth.token', 'bearer.auth.identity.pool.id', 'bearer.auth.logical.cluster']
+        # Note: bearer.auth.identity.pool.id is optional; only token and logical.cluster are required
```
naxinf, line 444: Should logical.cluster be optional as well, since it's only required for CC not but CP? I know we discussed this for the OAuth CPP implementation (https://github.com/confluentinc/libschemaregistry/pull/13/changes) and decided and it should be optional (cc @Claimundefine). Wondering if we should fix it in python client as well

src/confluent_kafka/schema_registry/common/schema_registry_client.py
```diff
@@ -113,6 +118,35 @@ def full_jitter(base_delay_ms: int, max_delay_ms: int, retries_attempted: int) -
     return random.random() * min(no_jitter_delay, max_delay_ms)
 
 
+def normalize_identity_pool(identity_pool_raw: Any) -> Optional[str]:
```
naxinf: Let's avoid `Any` and use a union type
```
def normalize_identity_pool(identity_pool_raw: str | list[str] | None)
```

Also, would be great to have a simple unit test for this function, to cover the 3 different input types + expected outputs

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
tobiogunbi: logical cluster is always required for SDS, maybe I can just include a comment instructing users?

Claimundefine (approved):

PR #2217, opened 2026-03-12
[KIP-932] : Implement Share consumer interface with poll API

k-raina:
**Summary**
- Implements ShareConsumer, a new Kafka client type for KIP-932 share group consumption
- Added New file: `src/confluent_kafka/src/ShareConsumer.c`
- ShareConsumerHandle struct inheriting from Handle via first-member embedding — enables safe (Handle *)self casts and reuses all common callback/TLS infrastructure
- `subscribe(topics) / unsubscribe() / subscription()` — topic subscription management via `rd_kafka_share_*` APIs
- `consume_batch(timeout=-1)` — batch-only consumption with chunked polling for Ctrl+C interruptibility; uses - CallState_begin/end and check_signals_between_chunks for correct GIL and TLS lifecycle management
- `close()` — graceful shutdown: attempts broker close then always destroys handle

**Known limitations (TODOs in code)**
| Item | Blocked On |
| :--- | :--- |
| **rd_kafka_set_log_queue()** | Needs a specific `rd_kafka_share_set_log_queue()` wrapper to handle the share consumer handle. |
| **OAuth Background Callbacks** | Missing `rd_kafka_share_sasl_background_callbacks_enable()` export in the C API. |
| **max_poll_records** | Currently hardcoded to 10005; requires updating to the `librdkafka` double-pointer API for dynamic config reads. |

**Additional Changes**
- Additional commit bc6101ab4618bc48c1aeaf92f807bbdc77573f81 to build python binding with kip-932 branch for librdkafka. We can remove this commit in future before merging to master.


src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,567 @@
...
+ ****************************************************************************/
+
+typedef struct {
+        Handle base;  /* first member — (Handle *)self cast is always safe */
+                      /* base.rk, base.u unused: share uses rkshare, no rebalance */
+
+        rd_kafka_share_t *rkshare;
+
+        /* TODO: Remove after interface of librdkafka is updated to return double pointer */
+        size_t max_poll_records;
```
PratRanj07: Lets name it batch_size so as to not confuse with the actual configuration property max.poll.records

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,567 @@
...
+    0,                                                    /*tp_as_mapping*/
+    0,                                                    /*tp_hash */
+    0,                                                    /*tp_call*/
+    0,                                                    /*tp_str*/
+    0,                                                    /*tp_getattro*/
+    0,                                                    /*tp_setattro*/
+    0,                                                    /*tp_as_buffer*/
+    Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE |
+        Py_TPFLAGS_HAVE_GC, /*tp_flags*/
+    "A high-level Apache Kafka share consumer (KIP-932)\n"
```
PratRanj07:
```suggestion
    "A high-level Apache Kafka share consumer\n"
```

src/confluent_kafka/cimpl.pyi
```diff
@@ -535,6 +552,15 @@ class Consumer:
     def memberid(self) -> str: ...
     def set_sasl_credentials(self, username: str, password: str) -> None: ...
 
+class ShareConsumer:
+    """Share Consumer for queue-like message consumption (KIP-932)."""
+    def __init__(self, config: Dict[str, Union[str, int, float, bool]]) -> None: ...
+    def subscribe(self, topics: List[str]) -> None: ...
+    def unsubscribe(self) -> None: ...
+    def subscription(self) -> List[str]: ...
+    def consume_batch(self, timeout: float = -1) -> List[Message]: ...
```
PratRanj07: I think we should name this Poll only to match the signature of the Java Client atleast in Higher level clients

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,567 @@
...
+        return topics;
+}
+
+
+/**
+ * @brief Consume a batch of messages from the share consumer.
+ *
+ */
+static PyObject *
+ShareConsumer_consume_batch(ShareConsumerHandle *self,
```
PratRanj07: Reduce the cyclomatic complexity of this function

PratRanj07 (changes requested):
First pass comments

src/confluent_kafka/cimpl.pyi
k-raina: In past, It was discussed that we would align naming with librdkafka APis.

src/confluent_kafka/src/ShareConsumer.c
k-raina: This function definition is aligned with Consumer.consume_batch for simplicity

src/confluent_kafka/cimpl.pyi
pranavrth: I think we still need to finalize this name. Mostly we will go with `poll` to adhere to Java. It was mentioned in one of the recent comments in the librdkafka public interface docs.

pranavrth:
I feel there will be issues related to the common callbacks in Python client. We are populating Handle but rk in that handle is not present. I can see that we are using this `rk` in the callbacks like we are using `rd_kafka_yield` or `rd_kafka_fatal_error` in `error_cb`  which is using `rk`. The main decision in Python client is going to be related to reusing existing callback or Handle or devise a new code flow for the same. Let's go deeper in each callback atleast to figure out:

- How are we planning to use Handle?
- How are we panning to use internal rk handle? Some places are even destroying this handle with `rd_kafka_destroy`? Do we need to expose `rk` from librdkafka for Binding use?
- Do we need to expose some extra APIs to make Python client work? Some of the APIs might not even work as those were working in normal Consumer. For example, `rd_kafka_yield`.
- Verify each aspect of "ShareConsumerHandle struct inheriting from Handle via first-member embedding — enables safe (Handle *)self casts and reuses all common callback/TLS infrastructure".
- Let's have a thorough understanding of how all the callbacks will work

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,558 @@
...
+                if (error)
+                        rd_kafka_error_destroy(error);
+                return NULL;
+        }
+
+        /* Handle error from rd_kafka_share_consume_batch() */
+        if (error) {
+                const char *error_str = rd_kafka_error_string(error);
+                int is_fatal          = rd_kafka_error_is_fatal(error);
+                int is_retriable      = rd_kafka_error_is_retriable(error);
```
pranavrth: We are not planning to fill this information judiciously. Let's not use this.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,558 @@
...
+ */
+static PyObject *ShareConsumer_close(ShareConsumerHandle *self) {
+        rd_kafka_resp_err_t err;
+        CallState cs;
+
+        if (!self->rkshare)
+                Py_RETURN_NONE;
+
+        CallState_begin((Handle *)self, &cs);
+        err = rd_kafka_share_consumer_close(self->rkshare);
```
pranavrth, line 342: return type of close is going to change to error_t *. Add a TODO to change this when interface in librdkafka is changed. 

```
rd_kafka_error_t *rd_kafka_share_consumer_close(rd_kafka_share_t *rkshare);
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,558 @@
...
+     "\n"
+     "  Close the share consumer.\n"
+     "\n"
+     "  This method should be called to properly clean up the share consumer\n"
+     "  and leave the share group.\n"
+     "\n"
+     "  :raises KafkaException: on error\n"
+     "\n"},
+
+    {NULL}};
```
pranavrth, line 466: We will need to add `set_sasl_credentials` and Context manager functions.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,558 @@
...
+        return 0;
+}
+
+static void ShareConsumer_dealloc(ShareConsumerHandle *self) {
+        PyObject_GC_UnTrack(self);
+
+        if (self->rkshare) {
+                CallState cs;
+                CallState_begin((Handle *)self, &cs);
+                rd_kafka_share_destroy(self->rkshare);
```
pranavrth, line 66: Use `_flags` variant for `_destroy()` here. If you cant add right now, add a TODO with `TODO KIP-932`.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,558 @@
...
+
+                if (is_fatal) {
+                        PyErr_Format(PyExc_RuntimeError, "Fatal error: %s",
+                                     error_str);
+                } else {
+                        PyErr_Format(KafkaException,
+                                     "Error: %s (retriable: %s)", error_str,
+                                     is_retriable ? "yes" : "no");
+                }
+
```
pranavrth: Check how Consumer is raising the exception. Use that way instead. `cfl_PyErr_Format`

pranavrth (changes requested):
Some high level comments. Still reviewing deeper.

k-raina:
Thanks @pranavrth for review

1. `ShareConsumerHandle` embeds `Handle base` as its first member, so `(Handle *)self `casts are valid per the C standard. 
      - The fields actually used by ShareConsumer are:` tlskey, error_cb, stats_cb, throttle_cb`. 
      - Fields not used: `base.rk (left NULL), base.u.Consumer.*`.

2. When `rd_kafka_share_consumer_new(conf)` is called, it internally creates an` rd_kafka_t*` stored inside the opaque `rd_kafka_share_t`. The conf goes into this internal rk. So when callbacks fire, librdkafka passes this valid internal `rk` as the callback's first parameter.
      - `error_cb,  stats_cb , throttle_cb, log_cb ` - uses the callback's rk param (valid), not h->rk. So these are safe.
      - `oauth_cb ` - will SEGFAULT. There are multiple h->rk usages that pass NULL to APIs that dereference it .eg. rk->rk_sasl.handle inside librdkafka

3. Regarding yield and destry functions:
      - `rd_kafka_yield(rk)` — This function [never dereferences its rk parameter](https://github.com/confluentinc/librdkafka/blob/dev_kip-932_queues-for-kafka/src/rdkafka_queue.c#L37-L39) — it only sets `rd_kafka_yield_thread = 1` . So rd_kafka_yield(NULL) is safe.
      - `rd_kafka_destroy(rk)` — Would segfault with NULL. Not currently reachable for ShareConsumer but would be a problem if OAuth support is added naively with wait_for_oauth_token_set.

4. The cleanest fix is a public accessor in librdkafka: 
    - `rd_kafka_t *rd_kafka_share_get_rk(rd_kafka_share_t *rkshare);` : Then ShareConsumer_init can do self->base.rk = rd_kafka_share_get_rk(self->rkshare), and all existing callback code, rd_kafka_set_log_queue(), and rd_kafka_sasl_background_callbacks_enable() work unmodified.

src/confluent_kafka/src/ShareConsumer.c
k-raina: Added context manager. Will add ssl changes in callback handling PR

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+/****************************************************************************
+ *
+ *
+ * ShareConsumer
+ *
+ *
+ ****************************************************************************/
+
+typedef struct {
+        Handle base; /* first member — (Handle *)self cast is always safe */
```
pranavrth: Why do we need cast here?

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+        /* base.rk, base.u unused: share uses rkshare, no rebalance */
+
+        rd_kafka_share_t *rkshare;
+
+        /* TODO: Remove after interface of librdkafka is updated to return
```
pranavrth: "TODO KIP-932:" instead of "TODO"

src/confluent_kafka/cimpl.pyi
```diff
@@ -535,6 +552,17 @@ class Consumer:
     def memberid(self) -> str: ...
     def set_sasl_credentials(self, username: str, password: str) -> None: ...
 
+class ShareConsumer:
+    """Share Consumer for queue-like message consumption (KIP-932)."""
+    def __init__(self, config: Dict[str, Union[str, int, float, bool]]) -> None: ...
+    def subscribe(self, topics: List[str]) -> None: ...
+    def unsubscribe(self) -> None: ...
+    def subscription(self) -> List[str]: ...
+    def poll(self, timeout: float = -1) -> List[Message]: ...
```
pranavrth, line 566: it should be `consume_batch`

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+     "\n"
+     "  Get current topic subscription.\n"
+     "\n"
+     "  :returns: List of subscribed topics\n"
+     "  :rtype: list(str)\n"
+     "  :raises KafkaException: on error\n"
+     "  :raises RuntimeError: if called on a closed share consumer\n"
+     "\n"},
+
+    {"poll", (PyCFunction)ShareConsumer_poll, METH_VARARGS | METH_KEYWORDS,
```
pranavrth, line 409: it should be `consume_batch`

.semaphore/semaphore.yml
```diff
@@ -9,6 +9,8 @@ global_job_config:
   env_vars:
     - name: LIBRDKAFKA_VERSION
       value: v2.13.2
+    - name: LIBRDKAFKA_BRANCH
```
pranavrth, line 13: Add a TODO to remove this `env_vars` when the `LIBRDKAFKA_VERSION` has QfK changes.

.semaphore/semaphore.yml
```diff
@@ -332,7 +334,13 @@ blocks:
 
             # Build and install confluent-kafka from source
             - lib_dir=dest/runtimes/$OS_NAME-$ARCH/native
-            - tools/wheels/install-librdkafka.sh "${LIBRDKAFKA_VERSION#v}" dest
+            - |
```
pranavrth, line 335: We need these changes in other jobs as well. Either add them or add a TODO for those.

src/confluent_kafka/__init__.py
```diff
@@ -39,6 +39,7 @@
     Consumer,
     Message,
     Producer,
+    ShareConsumer,
```
pranavrth: nit: Have it after `Consumer`.

src/confluent_kafka/cimpl.pyi
```diff
@@ -535,6 +552,17 @@ class Consumer:
     def memberid(self) -> str: ...
     def set_sasl_credentials(self, username: str, password: str) -> None: ...
 
+class ShareConsumer:
+    """Share Consumer for queue-like message consumption (KIP-932)."""
+    def __init__(self, config: Dict[str, Union[str, int, float, bool]]) -> None: ...
```
pranavrth: The interface should be like the normal consumer. It should accept `dict` as well as `kwargs`. Check Consumer.

src/confluent_kafka/cimpl.pyi
```diff
@@ -63,12 +63,14 @@ class KafkaError:
     DELEGATION_TOKEN_REQUEST_NOT_ALLOWED: int
     DUPLICATE_RESOURCE: int
     DUPLICATE_SEQUENCE_NUMBER: int
+    DUPLICATE_VOTER: int
```
pranavrth, line 66: Some of these error codes are intentional right?

src/confluent_kafka/src/Admin.c
```diff
@@ -99,12 +99,12 @@ struct Admin_options {
  * Make sure this is kept up to date with Admin_options above. */
 #define Admin_options_INITIALIZER                                              \
         {                                                                      \
-            Admin_options_def_int,   Admin_options_def_float,                  \
```
pranavrth, line 102: We can skip this file change.

tests/test_Admin.py
```diff
@@ -153,7 +153,9 @@ def test_create_topics_api():
     try:
         a.create_topics([NewTopic("mytopic")])
     except Exception as err:
-        assert False, f"When none of the partitions, \
```
pranavrth, line 156: Let's not change this file.

tests/test_unasync.py
```diff
@@ -46,14 +46,18 @@ def test_unasync_file_check(temp_dirs):
     os.makedirs(os.path.dirname(sync_file), exist_ok=True)
```
pranavrth, line 46: Let's not change this file.

.semaphore/semaphore.yml
```diff
@@ -332,7 +334,13 @@ blocks:
...
+              if [[ -n $LIBRDKAFKA_BRANCH ]]; then
+                sudo apt-get install -y -qq libssl-dev libsasl2-dev liblz4-dev libzstd-dev
```
pranavrth: Do we need this `sudo apt-get install`? If needed then move them to the install script instead.

tools/wheels/build-librdkafka-branch.sh
```diff
@@ -0,0 +1,76 @@
...
+
+if [[ -f $DEST/build/native/include/librdkafka/rdkafka.h ]]; then
+    echo "$0: librdkafka already built in $DEST"
+    exit 0
+fi
+
+echo "$0: Building librdkafka branch '$BRANCH' into '$DEST'"
+
+ARCH=${ARCH:-x64}
+SRC=/tmp/librdkafka-branch-src
```
pranavrth: We should have branch name in the folder name.

tools/wheels/build-librdkafka-branch.sh
```diff
@@ -0,0 +1,76 @@
...
+INSTALL=$SRC/install
+
+[[ -d "$DEST" ]] || mkdir -p "$DEST"
+rm -rf "$SRC"
+
+git clone --depth 1 --branch "$BRANCH" \
+    https://github.com/confluentinc/librdkafka.git "$SRC"
+
+pushd "$SRC"
+./configure --prefix="$INSTALL" --disable-debug-symbols
```
pranavrth: We should configure with all the required features.

tools/wheels/build-librdkafka-branch.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
pranavrth, line 1: Add a TODO that this script is temporary.

tools/source-package-verification.sh
```diff
@@ -23,7 +23,11 @@ for version in 3.9.0 4.0.0; do
 done
 
 lib_dir=dest/runtimes/$OS_NAME-$ARCH/native
-tools/wheels/install-librdkafka.sh "${LIBRDKAFKA_VERSION#v}" dest
+if [[ -n $LIBRDKAFKA_BRANCH ]]; then
```
pranavrth, line 27: Add TODO for removal here as well.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+        if (total_timeout_ms >= 0 && total_timeout_ms < CHUNK_TIMEOUT_MS) {
+                /* Short timeout: single call */
+                error = rd_kafka_share_consume_batch(
+                    self->rkshare, total_timeout_ms, rkmessages,
+                    &rkmessages_size);
+        } else {
+                /* Long timeout: chunked with signal checking */
+                while (1) {
+                        /* Calculate timeout for this chunk */
+                        if (total_timeout_ms < 0) {
```
pranavrth: Use `calculate_chunk_timeout` function.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+ * When changing method signatures, parameters, or defaults, you MUST
+ * also update the corresponding type definitions in:
+ *   src/confluent_kafka/cimpl.pyi
+ *
+ * Failure to keep both in sync will result in incorrect type hints.
+ */
+
+#include "confluent_kafka.h"
+
+#define ERR_MSG_SHARE_CONSUMER_CLOSED "Share consumer closed"
```
pranavrth: Move this to `confluent_kafka.h`

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        Py_RETURN_NONE;
+}
+
+
+/**
+ * @brief Unsubscribe from current subscription.
+ */
+static PyObject *ShareConsumer_unsubscribe(ShareConsumerHandle *self) {
```
pranavrth: We should use `PyObject *ignore` as 2nd argument in `METH_NOARGS` methods of the ShareConsumer. Check Consumer for the same.

Check all the APIs for ShareConsumer. Multiple APIs have this issue.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+                return -1;
+        }
+
+        self->base.type = RD_KAFKA_CONSUMER;
+
+        if (!(conf = common_conf_setup(RD_KAFKA_CONSUMER, (Handle *)self, args,
+                                       kwargs)))
+                return -1; /* Exception raised by common_conf_setup() */
+
+        /* TODO: Remove after interface of librdkafka is updated to return
```
pranavrth: I prefer to keep `TODO KIP-932` instead of just `TODO` to identify our TODOs

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+    0,                                                    /*tp_as_buffer*/
+    Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE | Py_TPFLAGS_HAVE_GC, /*tp_flags*/
+    "A high-level Apache Kafka share consumer\n"
+    "\n"
+    ".. py:function:: ShareConsumer(config)\n"
+    "\n"
+    "Create a new ShareConsumer instance using the provided configuration "
+    "*dict*.\n"
+    "Share consumers enable queue-like consumption where each partition can be "
+    "\n"
```
pranavrth: Remove this wrong \n

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+                return NULL;
+        }
+
+        /* Build Python list from all returned messages. */
+        msglist = PyList_New(rkmessages_size);
+
+        for (i = 0; i < rkmessages_size; i++) {
+                PyObject *msgobj = Message_new0((Handle *)self, rkmessages[i]);
+
+#ifdef RD_KAFKA_V_HEADERS
```
pranavrth, line 297: Add the same explanatory comment that Consumer.c uses at line 1226-1227:

```
#ifdef RD_KAFKA_V_HEADERS
        /** Have to detach headers outside Message_new0 because it
         * declares the rk message as a const */
        rd_kafka_message_detach_headers(
            rkmessages[i], &((Message *)msgobj)->c_headers);
#endif
```

Without this comment, the second header call after Message_new0 looks redundant.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+     "  Poll for a batch of messages from the share consumer.\n"
+     "\n"
+     "  The application must check each Message object's error() method\n"
+     "  to distinguish between proper messages (error() returns None)\n"
+     "  and errors.\n"
+     "\n"
+     "  Batch size is controlled by the ``max.poll.records`` configuration\n"
+     "  property, not a runtime argument. Records are locked by the broker\n"
+     "  per fetch cycle and must not be discarded mid-batch.\n"
+     "\n"
```
pranavrth: I am not able to understand what it means or why is it needed in the doc?

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+ * ShareConsumer
+ *
+ *
+ ****************************************************************************/
+
+typedef struct {
+        Handle base; /* first member — (Handle *)self cast is always safe */
+        /* base.rk, base.u unused: share uses rkshare, no rebalance */
+
+        rd_kafka_share_t *rkshare;
```
pranavrth, line 42: Add `TODO KIP-932: Remove this handle once the callbacks issue is fixed.`

src/confluent_kafka/src/ShareConsumer.c
pranavrth: This is available now as part of [this PR](https://github.com/confluentinc/librdkafka/pull/5406). Use this branch.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+                CallState cs;
+                CallState_begin((Handle *)self, &cs);
+                /* TODO KIP-932: Use rd_kafka_share_destroy_flags() once
+                 * available in the librdkafka public API. */
+                rd_kafka_share_destroy(self->rkshare);
+                self->rkshare = NULL;
+                CallState_end((Handle *)self, &cs);
+        }
+
+        Handle_clear((Handle *)self);
```
pranavrth: We shouldn't type cast. Use the field instead. Same for every place.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+        rkmessages = malloc(self->batch_size * sizeof(*rkmessages));
+        if (!rkmessages) {
+                PyErr_SetString(PyExc_MemoryError,
+                                "Failed to allocate message array");
+                return NULL;
+        }
+
+        total_timeout_ms = cfl_timeout_ms(tmout);
+
+        CallState_begin((Handle *)self, &cs);
```
pranavrth: Don't type cast

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+
+        /* Chunked polling pattern for signal interruptibility */
+        if (total_timeout_ms >= 0 && total_timeout_ms < CHUNK_TIMEOUT_MS) {
+                /* Short timeout: single call */
+                error = rd_kafka_share_consume_batch(
+                    self->rkshare, total_timeout_ms, rkmessages,
+                    &rkmessages_size);
+        } else {
```
pranavrth: These if else can be merged. Please do that.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+                        }
+
+                        /* Exit if messages received */
+                        if (rkmessages_size > 0) {
+                                break;
+                        }
+
+                        /* No messages yet — check for Ctrl+C before next chunk.
+                         */
+                        chunk_count++;
```
pranavrth: For long waiting application, `chunk_count * CHUNK_TIMEOUT_MS` can be limited by the `int` max value. Check if using long is needed.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        /* Build Python list from all returned messages. */
+        msglist = PyList_New(rkmessages_size);
+
+        for (i = 0; i < rkmessages_size; i++) {
+                PyObject *msgobj = Message_new0((Handle *)self, rkmessages[i]);
+
+#ifdef RD_KAFKA_V_HEADERS
+                rd_kafka_message_detach_headers(
```
pranavrth, line 300: Add note here as well.

pranavrth (changes requested):
Comments for the implementation part. Please fix the build first and then I will review the testing part.

src/confluent_kafka/cimpl.pyi
k-raina: They weren't manually added,  they appeared automatically when we ran generate_kafka_error_stub_codes.py against the newer librdkafka build. 
CI also runs this with --check ([source-package-verification.sh:54](vscode-webview://0f7rcdsh088ai1u0e0q8e2pnf0goduajofg8cj18pf5djreh04tt/tools/source-package-verification.sh#L54)), so the stubs must match the compiled module.

src/confluent_kafka/src/ShareConsumer.c
k-raina: **With a finite timeout**, 
- The maximum number of loop iterations is bounded by the timeout itself:
`max iterations = total_timeout_ms / CHUNK_TIMEOUT_MS`

- The largest finite timeout you can pass is INT_MAX = 2,147,483,647 ms (~24.8 days). With 200ms chunks:
`max chunk_count = 2,147,483,647 / 200 = 10,737,418`

- The multiplication chunk_count * CHUNK_TIMEOUT_MS at its peak:
`10,737,418 * 200 = 2,147,483,600`
That's still under INT_MAX (2,147,483,647). No overflow.

**With an infinite timeout (total_timeout_ms = -1),** 
- calculate_chunk_timeout hits this branch:
```
if (total_timeout_ms < 0) {
        /* Infinite timeout - use chunk size */
        return chunk_timeout_ms;  // returns 200, ignores chunk_count entirely
}
```


However `chunk_count` being long can be insurance . It requires changes in `calculate_chunk_timeout` which  takes `int` — changing just the local variable would silently truncate on the call. 
The proper fix would also need to update the shared function signature, which affects Consumer.c and Producer.c too. 
Maybe this can be done in some latter PRs?

k-raina:
@pranavrth Thanks for review!

- Previously test was failing due to infra issue, which i had fixed in main branch : https://github.com/confluentinc/confluent-kafka-python/pull/2233
- Currently all tests are passing except `test_stats_cb_exception_propagates ` - which is failing due to close() internally calls something that drains pending callbacks, hitting the same stats_cb exception a second time. Once implementation of close on librdkafka side in merged in qfk branch , we will check again if this fails.

tests/test_Admin.py
k-raina: These are style fixes only , done to fix CI

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
...
+    assert ex.match('expected configuration dict')
+
+
+def test_constructor_with_valid_config():
+    """ShareConsumer can be created with valid configuration."""
+    sc = ShareConsumer(
+        {
+            'group.id': 'test-share-group',
+            'bootstrap.servers': 'localhost:9092',
+            'socket.timeout.ms': 100,
```
pranavrth: `socket.timeout.ms` shouldn't be a valid configuration?

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
...
+        }
+    )
+
+    sc.subscribe(['test-topic'])
+
+    # Should timeout and return empty list
+    messages = sc.poll(timeout=0.1)
+    assert isinstance(messages, list)
+    # May be empty or contain error messages
```
pranavrth: Add correct handling here instead of comment.

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
...
+                'bootstrap.servers': 'localhost:9092',
+            }
+        )
+    assert ex.match('group.id must be set')
+
+
+@pytest.mark.integration
+def test_concurrent_consumers():
+    """Test multiple consumers in same share group."""
+    pytest.skip("Requires running Kafka broker with Share Consumer support")
```
pranavrth: Why are we skipping this test?

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
...
+    def my_stats_cb(stats_json):
+        stats_called.append(stats_json)
+
+    sc = ShareConsumer(
+        {
+            'group.id': 'test-share-stats-cb',
+            'bootstrap.servers': 'localhost:19999',
+            'socket.timeout.ms': 100,
+            'statistics.interval.ms': 100,
+            'stats_cb': my_stats_cb,
```
pranavrth: We discussed earlier with Andrew as well that we don't want to add stats_cb support for share consumer as KIP-714 is supported by librdkafka now. Either correct the test or add a TODO to correct this test

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
...
+        sc.poll(timeout=0.5)
+
+    assert "Test exception from error_cb" in str(exc_info.value)
+    assert len(error_called) > 0
+    # close() may also drain pending callbacks and re-raise, so ignore
+    try:
+        sc.close()
+    except RuntimeError:
+        pass
+
```
pranavrth: Why are we doing this? We shouldn't pass if close had an error.

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
...
+    sc = ShareConsumer(
+        {
+            'group.id': 'test-share-group',
+            'bootstrap.servers': 'localhost:9092',
+            'socket.timeout.ms': 100,
+        }
+    )
+
+    sc.close()
+    sc.close()  # Should not raise
```
pranavrth: This should raise exception that the share consumer is already closed. Add a TODO to help future developer on this. Currently, there is no error raised from here.

tests/test_ShareConsumer.py
```diff
@@ -0,0 +1,332 @@
+#!/usr/bin/env python
```
pranavrth, line 1: Why have we not added any test to get records? It should be part of integration tests. We can merge this PR and raise another one for that or add those tests as well in the next PR.

tests/test_ShareConsumer.py
pranavrth, line 1: The config dict `{'group.id': 'test-share-group', 'bootstrap.servers': 'localhost:9092', 'socket.timeout.ms': 100}` is duplicated across 8 tests. Two issues:

1. **Brittle to config evolution**: `test_Consumer.py:31` includes `'session.timeout.ms': 1000` with comment "Avoid close() blocking too long". If ShareConsumer needs an analogous knob, all 8 copies must be updated.
2. **Resource leaks on test failure**: every test ends with `sc.close()` — if an assertion fails earlier, close is skipped. Combined with `rd_kafka_share_destroy` being blocking, a leaked consumer could hang pytest teardown.

Replace with a fixture:
```python
@pytest.fixture
def share_consumer():
    sc = ShareConsumer({
        'group.id': 'test-share-group',
        'bootstrap.servers': 'localhost:9092',
        'socket.timeout.ms': 100,
    })
    yield sc
    sc.close()
```
Tests that use non-default configs (callback tests, `test_required_group_id`) can keep their inline dicts.

pranavrth, line 1: **Nit:** `test_Consumer.py` uses a `TestConsumer` wrapper from `tests/common/__init__.py` that handles protocol-specific config setup. Worth considering an analogous `TestShareConsumer` wrapper before the test suite grows — easier to add now than to retrofit later.

pranavrth, line 1: Test verifies the callback fired but not what it received. A regression where `error_cb` is invoked with a string or None would slip through. Suggest:
```python
assert isinstance(error_called[0], KafkaError)
assert error_called[0].code() in (KafkaError._TRANSPORT, KafkaError._ALL_BROKERS_DOWN)
```

pranavrth, line 1: No tests exercise the argument validation paths in ShareConsumer.c (e.g., `PyList_Check` at line 319, type checks in subscribe). Either add negative tests covering invalid argument types — `sc.subscribe("not_a_list")`, `sc.subscribe([123, 456])`, `sc.subscribe([])`, `sc.poll(timeout="bad")` — or add a TODO for a follow-up PR.

pranavrth, line 1: No test for signal interruption. `ShareConsumer.poll` reuses the chunked-polling pattern specifically to support `KeyboardInterrupt` (docstring at `ShareConsumer.c:643` says so). `test_Wakeable.py` has the pattern for Consumer — worth adapting for ShareConsumer to verify the chunked logic actually interrupts.

pranavrth, line 1: **Nit:** Move `import json` to the top of the file with other imports.

src/confluent_kafka/src/ShareConsumer.c
pranavrth, line 1: `PyList_New(c_topics->cnt)` (line 189) can return NULL on OOM. The next line `PyList_SET_ITEM(topics, i, ...)` would dereference NULL → segfault. Even if we survived, `c_topics` would leak.

Defensive form:
```c
topics = PyList_New(c_topics->cnt);
if (!topics) {
    rd_kafka_topic_partition_list_destroy(c_topics);
    return NULL;
}
```

Note: Consumer.c has the same pattern unguarded in similar places, so this is a project-wide robustness issue not specific to ShareConsumer. Worth flagging here as a small fix; broader cleanup could be a follow-up.

pranavrth, line 1: Suggest hardening these two NULL-deref paths — improving on the existing Consumer.c pattern (lines 1221-1232) which has the same risk:

1. `PyList_New(rkmessages_size)` (line 292) can return NULL on OOM → `PyList_SET_ITEM(NULL, ...)` segfaults.
2. `Message_new0()` (line 295) can return NULL → `&((Message *)msgobj)->c_headers` dereferences NULL.

Since this is new code, worth adding the NULL checks here even though the older Consumer code has the same gap. Broader cleanup of the older pattern can be a follow-up.

pranavrth, line 1: Brief comment here would help — `RD_KAFKA_CONSUMER` is set to get `group.id` validation from `common_conf_setup`, but ShareConsumer isn't a regular consumer. Without the comment, this looks like a copy-paste mistake.

pranavrth, line 1: Add a TODO comment near init noting that callback dispatch for ShareConsumer (`error_cb`/`stats_cb`/`throttle_cb`) needs verification against librdkafka — these trampolines assert via `CallState_get(handle)` that the calling thread has an active CallState. Without `rd_kafka_poll_set_consumer` equivalent, callbacks firing from librdkafka's background thread would abort the process. Tracked separately.

pranavrth, line 1: FYI / future-maintainer note: `consumer_conf_set_special` (confluent_kafka.c:2477) INCREFs `on_commit` into `self->base.u.Consumer.on_commit` if a user passes it. Consumer cleans this up in `Consumer_clear0` but ShareConsumer doesn't have an equivalent — so passing `on_commit` to ShareConsumer would leak the callback reference. `on_commit` is meaningless for share consumers anyway (no offset commits), so worth either ignoring this or adding a TODO for cleanup.

pranavrth (changes requested):
- I am providing few more comments on the implementation part as well though those are nits or for future callback implementation. I would suggest going through them and add a TODO for each on of them so that you can tackle them in the next PR.
- Tests comments should be fixed in this PR or maybe in next immediate PR if we require to merge this quickly.
- Please don't use MERGE to get the latest changes from the master. It affects commit history as well. Better to use rebase or cherry pick your comments on top of the base branch.

pranavrth:
Fix the build as well.

tests/integration/share_consumer/test_share_consumer.py
pranavrth, line 1: Add a lock timeout / redelivery test. `test_implicit_ack_no_redelivery` verifies the happy path ("when ack happens, no redelivery"). The inverse ("when ack doesn't happen, broker redelivers to another consumer after the lock timeout") isn't covered. This is the defining at-least-once promise of share groups — worth locking down.

pranavrth, line 1: Practical concern for the test above: the default share lock timeout is ~30s, making this a slow test. To run reliably in CI, would need either (a) the test cluster configured with a shorter `share.record.lock.duration.ms`, or (b) tracking as a follow-up integration test once cluster knobs are available. Either is fine — flagging because it affects when this lands.

tests/test_ShareConsumer.py
pranavrth, line 1: The 3 `__init__` overloads in `cimpl.pyi:559-563` (positional dict, dict + kwargs, kwargs only) — only the positional dict form is tested. The other two are unverified type-stub claims. If the C extension rejects either form, type checkers would happily accept user code that crashes at runtime. Worth adding two small tests covering the kwargs-only and dict+kwargs forms.

pranavrth, line 1: Minor: `subscribe()` called twice in succession isn't directly tested in unit tests. The integration test `test_resubscribe_to_different_topic` likely covers it implicitly (real broker), but a unit-level lock-down — `subscribe(['a']); subscribe(['b']); assert subscription() == ['b']` — would catch any regression where librdkafka changes the merge-vs-replace semantics. Worth adding as a quick sanity check.

pranavrth, line 1: Minor: `subscription()` on a fresh consumer (before any `subscribe()` call) isn't tested. Worth a quick `assert sc.subscription() == []` to lock down the no-subscription state. Catches any regression where librdkafka changes how it represents the empty subscription case.

tests/integration/share_consumer/test_share_consumer.py
pranavrth, line 1: Add a `poll(timeout=0)` non-blocking test. The interesting case isn't just "returns fast on first call" but "tight-loop polling with timeout=0 eventually receives produced records." Real-world use cases (event-loop integration, asyncio bridges, custom polling loops) depend on this contract. Produce N messages, then `while collected < N: msgs = sc.poll(timeout=0); collected.extend(msgs)` — asserts both that poll(0) doesn't block AND that records flow through the non-blocking path correctly.

src/confluent_kafka/__init__.py
pranavrth, line 1: Add an example file `examples/share_consumer.py`.

tests/integration/share_consumer/test_share_consumer.py
pranavrth, line 1: Created topics aren't deleted after tests. Add cleanup so topics don't accumulate across runs on long-lived test clusters.

tests/integration/share_consumer/test_share_consumer.py
```diff
@@ -0,0 +1,382 @@
...
+    drain_share_consumers,
+    unique_id,
+    warmup_share_consumers,
+)
+
+# KIP-932 share groups require Kafka 4.1+ with the share rebalance protocol
+# enabled. Skip unless either:
+#   - TEST_CONSUMER_GROUP_PROTOCOL=share is set.
+#   - or BROKERS is set.
+if not (TestUtils.use_group_protocol_share() or os.environ.get('BROKERS')):
```
pranavrth: There is no protocol like share. Please update this path. We need to test this always.

.semaphore/semaphore.yml
```diff
@@ -228,6 +228,16 @@ blocks:
...
             - tools/source-package-verification.sh
+        - name: Build and Tests with 'share' group protocol
+          commands:
+            - sem-version python 3.11
+            - sem-version java 17
+            - pip install uv
+            # use a virtualenv
+            - uv venv _venv --python "$(command -v python)" && source _venv/bin/activate
+            - chmod u+r+x tools/source-package-verification.sh
+            - export TEST_CONSUMER_GROUP_PROTOCOL=share
```
pranavrth: This is incorrect as this env variable is used to define group protocol in normal consumer. There is no concept of group protocol in Share Consumer. Let's remove this.

.semaphore/semaphore.yml
```diff
@@ -228,6 +228,16 @@ blocks:
             - chmod u+r+x tools/source-package-verification.sh
             - export TEST_CONSUMER_GROUP_PROTOCOL=consumer
             - tools/source-package-verification.sh
+        - name: Build and Tests with 'share' group protocol
+          commands:
+            - sem-version python 3.11
```
pranavrth: Let's add a TODO to test with multiple versions of python.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -48,14 +48,29 @@ typedef struct {
...
+        /* consumer_conf_set_special INCREFs on_commit into
+         * base.u.Consumer.on_commit for any consumer type, even though
+         * ShareConsumer never registers the offset commit trampoline so the
+         * callback can't fire. Match the INCREF with a DECREF here to avoid
+         * leaking the user's callback. Mirrors Consumer_clear0. */
+        if (self->base.u.Consumer.on_commit) {
+                Py_DECREF(self->base.u.Consumer.on_commit);
+                self->base.u.Consumer.on_commit = NULL;
+        }
+}
```
pranavrth, line 66: We will have to remove this assignment which setting up the conf when we fix callback related handling. Let's add a TODO to fix this when callbacks are implemented.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -73,6 +88,10 @@ static void ShareConsumer_dealloc(ShareConsumerHandle *self) {
 
 static int
 ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
+        /* See ShareConsumer_clear: pair the DECREF with a Py_VISIT so cyclic
+         * GC can find on_commit if a user accidentally passed one. */
+        if (self->base.u.Consumer.on_commit)
+                Py_VISIT(self->base.u.Consumer.on_commit);
```
pranavrth, line 104: I think instead of this, we shouldn't let user to pass this.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -465,6 +501,9 @@ ShareConsumer_init(PyObject *selfobj, PyObject *args, PyObject *kwargs) {
 
         self->base.type = RD_KAFKA_CONSUMER;
 
+        /* RD_KAFKA_CONSUMER is intentional, not a copy-paste from Consumer.c:
+         * it makes common_conf_setup enforce "group.id must be set", which
+         * share consumers also need. */
         if (!(conf = common_conf_setup(RD_KAFKA_CONSUMER, &self->base, args,
```
pranavrth: We will be looking again at this handling for Share Consumer. Mainly config setup. Let's add a TODO.

tests/common/__init__.py
```diff
@@ -46,12 +49,24 @@ def send_sigint_after_delay(delay_seconds):
 
     @staticmethod
     def broker_version():
+        if TestUtils.use_group_protocol_share():
+            # KIP-932 share groups require Kafka 4.1+ with the share rebalance
+            # protocol.
+            return '4.2.0'
         return '4.0.0' if TestUtils.use_group_protocol_consumer() else '3.9.0'
```
pranavrth: We should move to common `4.2.0` broker version.

tests/common/__init__.py
```diff
@@ -46,12 +49,24 @@ def send_sigint_after_delay(delay_seconds):
...
 
     @staticmethod
     def broker_conf():
         broker_conf = ['transaction.state.log.replication.factor=1', 'transaction.state.log.min.isr=1']
-        if TestUtils.use_group_protocol_consumer():
+        if TestUtils.use_group_protocol_share():
+            broker_conf.extend(
+                [
+                    'group.coordinator.rebalance.protocols=classic,consumer,share',
```
pranavrth: This is not correct. `share` is not an acceptable property. Even this property is going to get deprecated in future. Check [KIP-1237](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1237%3A+Deprecate+group.coordinator.rebalance.protocols+config). Share consumer is enabled by default in 4.2.0. No need for this property.

tests/common/__init__.py
```diff
@@ -46,12 +49,24 @@ def send_sigint_after_delay(delay_seconds):
...
+                    'share.coordinator.state.topic.replication.factor=1',
+                    'share.coordinator.state.topic.min.isr=1',
```
pranavrth: Add `group.share.min.record.lock.duration.ms=1000` property as well to make acquisition lock time out faster so that test suite can run faster as well.

tests/common/__init__.py
```diff
@@ -61,12 +76,20 @@ def _broker_major_version():
 
     @staticmethod
     def use_kraft():
-        return TestUtils.use_group_protocol_consumer() or _trivup_cluster_type_kraft()
+        return (
+            TestUtils.use_group_protocol_consumer()
+            or TestUtils.use_group_protocol_share()
```
pranavrth: Remove

tests/common/__init__.py
```diff
@@ -61,12 +76,20 @@ def _broker_major_version():
...
+            or _trivup_cluster_type_kraft()
+        )
 
     @staticmethod
     def use_group_protocol_consumer():
         return _GROUP_PROTOCOL_ENV in os.environ and os.environ[_GROUP_PROTOCOL_ENV] == 'consumer'
 
+    @staticmethod
+    def use_group_protocol_share():
```
pranavrth: Remove

tests/common/__init__.py
```diff
@@ -118,3 +141,53 @@ def unassign(self, partitions):
...
+def warmup_share_consumers(consumers, duration_s=8.0):
+    """Drive heartbeats so share consumers register with the share coordinator.
+
+    KIP-932 share groups only deliver records produced after the consumer has
+    joined. Call after subscribing and before producing test records.
+    """
+    deadline = time.time() + duration_s
+    while time.time() < deadline:
+        for sc in consumers:
+            sc.poll(timeout=0.5)
```
pranavrth: As discussed, remove this warmup code. We should use auto offset reset to reset the offset to earliest to make sure that the tests are predictable.

tests/common/__init__.py
```diff
@@ -118,3 +141,53 @@ def unassign(self, partitions):
...
+    return received
+
+
+class TestShareConsumer(ShareConsumer):
+    """Test wrapper around ShareConsumer"""
+
+    __test__ = False  # not a pytest collection target despite the Test* prefix
+
+    def __init__(self, conf=None, **kwargs):
+        merged = {'bootstrap.servers': DEFAULT_BOOTSTRAP_SERVERS}
```
pranavrth: Improve name for `merged`

tests/test_ShareConsumer.py
```diff
@@ -4,9 +4,29 @@
...
+    unique_id,
+)
+
+
+@pytest.fixture
+def share_consumer():
+    """Default-configured ShareConsumer with teardown."""
+    sc = TestShareConsumer(
+        {
+            'group.id': 'test-share-group',
```
pranavrth: We should isolate the group for each test. Have same group creates transient issues.

tests/test_ShareConsumer.py
```diff
@@ -241,67 +282,8 @@ def error_cb_that_raises(error):
...
-    )
-
-    sc.subscribe(['test-topic'])
-
-    with pytest.raises(RuntimeError) as exc_info:
-        sc.poll(timeout=0.5)
 
-    assert "Test exception from stats_cb" in str(exc_info.value)
-    assert len(stats_called) > 0
+    raising[0] = False
```
pranavrth, line 369: Close might also be raising. Let's understand what is expected from the close and update the test.

src/confluent_kafka/__init__.py
pranavrth: I think you missed this comment.

tests/common/__init__.py
```diff
@@ -46,12 +49,24 @@ def send_sigint_after_delay(delay_seconds):
...
+        if TestUtils.use_group_protocol_share():
+            broker_conf.extend(
+                [
+                    'group.coordinator.rebalance.protocols=classic,consumer,share',
+                    'share.coordinator.state.topic.replication.factor=1',
+                    'share.coordinator.state.topic.min.isr=1',
+                ]
+            )
+        elif TestUtils.use_group_protocol_consumer():
             broker_conf.append('group.coordinator.rebalance.protocols=classic,consumer')
```
pranavrth: We can merge other configs here itself.

tests/common/__init__.py
```diff
@@ -118,3 +141,53 @@ def unassign(self, partitions):
...
         else:
             super(TestConsumer, self).unassign()
+
+
+def unique_id(prefix):
+    """Generate a topic/group id unique to this test run.
+
+    Avoids cross-test interference when running against a shared broker.
+    """
+    return f'{prefix}-{uuid.uuid4().hex[:10]}'
```
pranavrth, line 146: Isn't there anything same in the suite?

tests/common/__init__.py
```diff
@@ -118,3 +141,53 @@ def unassign(self, partitions):
...
+    KIP-932 share groups only deliver records produced after the consumer has
+    joined. Call after subscribing and before producing test records.
+    """
+    deadline = time.time() + duration_s
+    while time.time() < deadline:
+        for sc in consumers:
+            sc.poll(timeout=0.5)
+
+
+def drain_share_consumers(consumers, n_expected, timeout_s=20.0):
```
pranavrth: This will only work for the implicit ack mode. The issue is that if records come in multiple polls then we will get error in explicit mode that we have not acknowledged previously inflight records.

I think we have same issue in librdkafka tests as well. We should fix there as well.

pranavrth (changes requested):
I still have to go deeper into `test_share_consumer.py`. I am going through this in parallel. Posting the comments so that you are unblocked.

tests/common/__init__.py
k-raina: Currently python codebase dont have clean alternative for same.

k-raina:
@pranavrth
i have updated pipeline to cleanly support AK 4.2 broker for tests.
Added some TODOs to cleanup pipeline which can be handled in later PR before merging to main branch.

examples/share_consumer.py
```diff
@@ -0,0 +1,76 @@
...
+    broker = sys.argv[1]
+    group = sys.argv[2]
+    topics = sys.argv[3:]
+
+    # ShareConsumer configuration.
+    # See https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md
+    conf = {
+        'bootstrap.servers': broker,
+        'group.id': group,
+        'auto.offset.reset': 'earliest',
```
pranavrth: There is no `auto.offset.reset` property in share consumer.

examples/share_consumer.py
```diff
@@ -0,0 +1,76 @@
...
+    try:
+        while True:
+            messages = sc.poll(timeout=1.0)  # returns a list (possibly empty)
+            for msg in messages:
+                # This example is fail-fast on per-message
+                # errors — the first error terminates the consumer. For a long-running queue worker, prefer
+                # logging msg.error() and `continue` so transient delivery
+                # errors don't kill the process.
+                if msg.error():
+                    raise KafkaException(msg.error())
```
pranavrth: We shouldn't throw exception in this case. We should log and move on. These errors are mostly informational. We only fail in FATAL errors.

src/confluent_kafka/__init__.py
pranavrth: Still not moved.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,631 @@
...
+                CallState cs;
+                CallState_begin(&self->base, &cs);
+                /* TODO KIP-932: Use rd_kafka_share_destroy_flags() once
+                 * available in the librdkafka public API. */
+                rd_kafka_share_destroy(self->rkshare);
+                self->rkshare = NULL;
+                CallState_end(&self->base, &cs);
+        }
+
+        Handle_clear(&self->base);
```
pranavrth, line 91: Add a TODO to call `ShareHandle_clear` directly here when `ShareHandle_clear0` is removed.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,631 @@
...
+
+        Py_TYPE(self)->tp_free((PyObject *)self);
+}
+
+static int
+ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
+        /* See ShareConsumer_clear: pair the DECREF with a Py_VISIT so cyclic
+         * GC can find on_commit if a user accidentally passed one. */
+        if (self->base.u.Consumer.on_commit)
```
pranavrth, line 103: TODO

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,631 @@
...
+                free(rkmessages);
+                if (error)
+                        rd_kafka_error_destroy(error);
+                return NULL;
+        }
+
+        /* Handle error from rd_kafka_share_consume_batch() */
+        if (error) {
+                cfl_PyErr_from_error_destroy(error);
+                free(rkmessages);
```
pranavrth, line 344: Let's add a defensive handling for destroying messages as well.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,631 @@
...
+                        /* Cleanup on Message_new0 failure:
+                         * - msglist DECREF releases successfully-built msgobjs
+                         *   for indices [0, i) (PyList_SET_ITEM stole their
+                         *   refs).
+                         * - rkmessages[0..i-1] were already destroyed at the
+                         *   end of their respective loop iterations.
+                         * - rkmessages[i..N-1] (current failure plus any
+                         *   unprocessed) are destroyed by the loop below. */
+                        Py_DECREF(msglist);
+                        for (; i < rkmessages_size; i++)
```
pranavrth, line 374: This is correct handling. Check that we are destroying all the received messages in all the flows before exiting this function.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,631 @@
...
+                return NULL;
+        }
+
+        /* Build Python list from all returned messages. */
+        msglist = PyList_New(rkmessages_size);
+        if (!msglist) {
+                for (i = 0; i < rkmessages_size; i++)
+                        rd_kafka_message_destroy(rkmessages[i]);
+                free(rkmessages);
+                return NULL;
```
pranavrth, line 359: TODO to move this to goto.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,631 @@
...
+
+        return topics;
+}
+
+
+/**
+ * @brief Poll for a batch of messages from the share consumer.
+ *
+ */
+static PyObject *ShareConsumer_poll(ShareConsumerHandle *self,
```
pranavrth, line 247:
```
if (self->u.Consumer.rkqu) {
                rd_kafka_queue_destroy(self->u.Consumer.rkqu);
                self->u.Consumer.rkqu = NULL;
        }

```

We are creating a consumer queue specifically from Python client to fetch records. Check if that is needed in ShareConsumer as well.

tests/common/__init__.py
```diff
@@ -118,3 +136,66 @@ def unassign(self, partitions):
...
+    consumer-group-join timing: records produced before subscribe() are still
+    delivered. Tests that need 'latest' semantics must override explicitly.
+    """
+
+    __test__ = False  # not a pytest collection target despite the Test* prefix
+
+    def __init__(self, conf=None, **kwargs):
+        effective_conf = {
+            'bootstrap.servers': DEFAULT_BOOTSTRAP_SERVERS,
+            'auto.offset.reset': 'earliest',
```
pranavrth: Remove this

tools/source-package-verification.sh
```diff
@@ -11,7 +11,7 @@ uv pip install -U build
 # Cache trivup Apache Kafka versions
 
 BASE=$PWD
-for version in 3.9.0 4.0.0; do
+for version in 3.9.0 4.0.0 4.2.0; do
```
pranavrth: Remove other versions if not required. Add TODO maybe.

pranavrth (commented):
Minor comments.

src/confluent_kafka/src/ShareConsumer.c
k-raina: 1. We need this queue in notmal consumer to read messages in batches

```
  rd_kafka_queue_t *rkqu = self->u.Consumer.rkqu;
  n = rd_kafka_consume_batch_queue(rkqu, timeout_ms, rkmessages, num_messages);
```

2. Another case where consumer uses queue is : when the commit finishes, drop the callback into this queue and let the   user pick it up on their next poll

```
if (async) {
      rkqu = self->u.Consumer.rkqu;                          // L525
  }
  err = rd_kafka_commit_queue(self->rk, c_offsets, rkqu,    // L535
                              async ? Consumer_offset_commit_cb
                                    : Consumer_offset_commit_return_cb,
                              ...);
```
We might need this when we handle callbacks for share consumer. Added TODO for now

tests/integration/cluster_fixture.py
```diff
@@ -134,6 +141,41 @@ def cimpl_consumer(self, conf=None):
...
+            incremental_configs=[
+                ConfigEntry(
+                    'share.auto.offset.reset',
+                    reset,
+                    incremental_operation=AlterConfigOpType.SET,
+                ),
+            ],
+        )
+        for f in self.admin().incremental_alter_configs([res]).values():
+            f.result()
```
pranavrth, line 179: Everytime you try to create a new share consumer with same group id, it will try to update auto.offset.reset which is not necessary. This shouldn't be part of share consumer fixture. It is a property of group.

tests/integration/share_consumer/test_share_consumer.py
```diff
@@ -0,0 +1,508 @@
...
+
+    producer = kafka_cluster.cimpl_producer()
+    for i in range(n):
+        producer.produce(topic, value=f'pre-{i}'.encode())
+    producer.flush(timeout=10.0)
+
+    # Override the suite-wide 'earliest' default: this test asserts that
+    # pre-join records are NOT delivered, which is only the contract under
+    # 'latest'.
+    sc = kafka_cluster.share_consumer({'auto.offset.reset': 'latest'})
```
pranavrth, line 171: This gives incorrect impression that offset.reset is property of consumer. Let's fix this.

tests/integration/share_consumer/test_share_consumer.py
```diff
@@ -0,0 +1,508 @@
...
+        sc.close()
+
+
+def test_resubscribe_to_different_topic(kafka_cluster):
+    """subscribe() replaces (does not extend) the prior subscription."""
+    topic_a = kafka_cluster.create_topic_and_wait_propogation('test-share-consumer-resub-a')
+    topic_b = kafka_cluster.create_topic_and_wait_propogation('test-share-consumer-resub-b')
+
+    sc = kafka_cluster.share_consumer()
+    producer = kafka_cluster.cimpl_producer()
```
pranavrth, line 446: Producer created outside try block — inconsistent with other tests. If `cimpl_producer()` raised, sc would leak. Move inside try for consistency.

tests/common/__init__.py
```diff
@@ -118,3 +136,60 @@ def unassign(self, partitions):
...
+    TODO KIP-932: when explicit-ack mode lands in the Python wrapper
+    (ShareConsumer.acknowledge()), update this helper to ack each message as
+    it's drained, otherwise the broker will return INFLIGHT-records errors
+    on the second poll. Same caveat exists in the librdkafka tests.
+    """
+    received = [[] for _ in consumers]
+    deadline = time.time() + timeout_s
+    while time.time() < deadline:
+        for sc, bucket in zip(consumers, received):
+            for m in sc.poll(timeout=poll_timeout_s):
```
pranavrth, line 181: After last poll, we are not acknowledging last set of records. Some of the tests assume that we do. Let's fix this with explicit ack part.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,654 @@
...
+                PyObject *uo, *uo8 = NULL;
+                if (!(uo = cfl_PyObject_Unistr(o))) {
+                        PyErr_SetString(PyExc_TypeError,
+                                        "expected list of unicode strings");
+                        rd_kafka_topic_partition_list_destroy(c_topics);
+                        return NULL;
+                }
+                rd_kafka_topic_partition_list_add(c_topics,
+                                                  cfl_PyUnistr_AsUTF8(uo, &uo8),
+                                                  RD_KAFKA_PARTITION_UA);
```
pranavrth, line 153: `cfl_PyUnistr_AsUTF8` can return NULL. Passing NULL to rd_kafka_topic_partition_list_add would crash. Worth checking the return value here.

`Consumer.c` has the same gap — pre-existing project pattern. Maybe have a TODO to note this so that we fix this atleast for ShareConsumer.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -0,0 +1,659 @@
...
+        }
+
+        Py_RETURN_NONE;
+}
+
+
+/**
+ * @brief ShareConsumer methods.
+ */
+static PyMethodDef ShareConsumer_methods[] = {
```
pranavrth, line 466: We need to mention that Share Consumer is not thread safe and also if user tries to use it from different threads, then we should maybe throw an error or atleast verify that wrapper code in python is not causing any issue. Even though this is being done in the librdkafka but Python has wrapper code as well. Add a TODO to verify this.

pranavrth (changes requested):
Some minor fixes which can be taken as part of next PR. Just a TODOs for these and we are good to go.

pranavrth (approved):
LGTM! Thanks @k-raina !.

PratRanj07 (approved):
LGTM! Thanks @k-raina

PR #2219, opened 2026-03-16
Remove callback wrapping for oauth_cb in AIO clients

### LINKED ISSUE #2171: AIOProducer doesn't work with OAUTHBEARER/oauth_cb
Hi,

We are trying to get the new `AIOProducer` to work with the `oauth_cb` mechanism. However, we are failing to get it to work. It seems the generated token is never set on the underlying producer.

This is a minimal example that fails:

```python
import asyncio
import os

import confluent_kafka
import confluent_kafka.aio
from tokenprovider import TokenProvider

token_provider = TokenProvider()

def cb(error):
    print(error)

class Prod:
    async def produce(self):
        producer_config = {
            "bootstrap.servers": os.getenv("BOOTSTRAP_URL"),
            "security.protocol": "SASL_SSL",
            "sasl.mechanisms": "OAUTHBEARER",
            "oauth_cb": token_provider.get_token,
            "error_cb": cb,
            "debug": "broker,protocol,msg,topic,metadata,queue,all",
        }
        self.producer = confluent_kafka.aio.AIOProducer(producer_config)
        print("Producer created")

        await self.producer.begin_transaction()

        future = await self.producer.produce(
            "test_topic",
           value='{"a": 1}'.encode('utf-8'),
        )
        await self.producer.flush()
        await self.producer.commit_transaction()
        return future

async def run():
    prod = Prod()
    results = await prod.produce()
    print(results)

asyncio.run(run())
```

The tokenprovider is pretty much exactly the one from Google's example on Managed Kafka [here](https://docs.cloud.google.com/managed-service-for-apache-kafka/docs/quickstart-python), but suffice to say that token_provider.get_token simply returns a (token, expiry) tuple as required. With the non-async Producer this works perfectly fine, and we are able to connect and produce to Kafka without problems.

What we see during execution is that:
- `print("Producer created")` is never reached
- `token_provider.get_token` is executed and completes succesfully, returning the correct token and expiry

Instead, we get stuck with output like this until the new 10 second timeout is triggered:

```
%7|1767953740.400|CONNECT|rdkafka#producer-1| [thrd:main]: Cluster connection already in progress: no cluster connection
%7|1767953741.400|CONNECT|rdkafka#producer-1| [thrd:main]: Cluster connection already in progress: no cluster connection
```

After which it fails with `cimpl.KafkaException: KafkaError{code=SASL_AUTHENTICATION_FAILED,val=58,str="OAuth token not set within 10 seconds timeout"}`

Are we doing something wrong with the `oauth_cb` or is this a bug? Should we do something different?

ojasvajain:
Fixes #2171 

- Fixed AIOConsumer/AIOProducer initialization failures by removing oauth_cb wrapping in `_common.wrap_common_callbacks()`. The OAuth callback is invoked by librdkafka's
  background thread and must be called directly, not scheduled on the event loop.
  
 - Created dedicated test file `test_oauth_cb.py`, moved oauth_cb related test cases from `test_misc.py` for better organization.

### DIFF
--- a/tests/test_oauth_cb.py
+++ b/tests/test_oauth_cb.py
@@ -1,4 +1,18 @@
 #!/usr/bin/env python
+#
+# Copyright 2026 Confluent Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
 import asyncio
 import time


src/confluent_kafka/aio/_common.py
```diff
@@ -83,5 +83,6 @@ def wrap_common_callbacks(loop: asyncio.AbstractEventLoop, conf: Dict[str, Any])
     wrap_conf_callback(loop, conf, 'error_cb')
     wrap_conf_callback(loop, conf, 'throttle_cb')
     wrap_conf_callback(loop, conf, 'stats_cb')
-    wrap_conf_callback(loop, conf, 'oauth_cb')
+    # oauth_cb must NOT be wrapped because it is invoked by librdkafka's background thread
```
k-raina: Should we remove this code instead of comment?

tests/test_oauth_cb.py
```diff
@@ -0,0 +1,246 @@
+#!/usr/bin/env python
```
k-raina, line 1: All new files needs to include license aggrements on top.

ojasvajain: Do we follow this for test files as well? Asking because I didn't see license agreements in any of the test files. Files like test_AIOProducer.py, test_AIOConsumer.py and test_unasync.py were created last year.

src/confluent_kafka/aio/_common.py
ojasvajain: Fine either ways. Removed.

tests/test_oauth_cb.py
k-raina: Yes ([example](https://github.com/confluentinc/confluent-kafka-python/blob/master/tests/avro/test_mock_client.py)), i think it got missed in other files in /tests repo.

ojasvajain: Sure, added

k-raina (approved):
Thanks for PR. Approved!

PR #2228, opened 2026-04-02
DGS-23862 Fix URL joining in Python client 

rayokota:
What
----
Fix URL joining in Python client to avoid double slashes

Checklist
------------------
- [Y] Contains customer facing changes? Including API/behavior changes 
- [Y] Did you add sufficient unit test and/or integration test coverage for this PR?

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/confluent_kafka/schema_registry/_async/schema_registry_client.py
+++ b/src/confluent_kafka/schema_registry/_async/schema_registry_client.py
@@ -571,7 +571,11 @@ class _AsyncRestClient(_AsyncBaseRestClient):
         response = None
         for i in range(self.max_retries + 1):
             response = await self.session.request(
-                method, url="/".join([base_url, url]), headers=headers, content=body, params=query
+                method,
+                url="/".join([base_url.rstrip("/"), url.lstrip("/")]),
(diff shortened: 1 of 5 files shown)


DhirajSuri27 (approved):

src/confluent_kafka/schema_registry/_async/schema_registry_client.py
```diff
@@ -571,7 +571,11 @@ async def send_http_request(
         response = None
         for i in range(self.max_retries + 1):
             response = await self.session.request(
-                method, url="/".join([base_url, url]), headers=headers, content=body, params=query
+                method,
+                url="/".join([base_url.rstrip("/"), url.lstrip("/")]),
```
DhirajSuri27, line 575: Nit: Are there other clients that are prone to this issue? If so, we could make the legacy URI compliance FF default to true in CP SR.

rayokota: I checked the other clients and they don't seem to have the same issue

PR #2230, opened 2026-04-14
KIP-932: Add Share Consumer soak tests + high-throughput perf variant

Ankith-Confluent:
What

- Add `--share` flag to soakclient.py that runs Producer + ShareConsumer
- Implicit and Explicit
- Adds ShareConsumer (KIP-932) coverage to the CKPy soak client and introduces a
higher-throughput `soakclient_perf.py` variant for the HT soak

----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


k-raina:
@Ankith-Confluent 
Could you please rebase this branch?

tests/soak/soakclient.py
```diff
@@ -446,50 +637,108 @@ def filter_config(conf, filter_out, strip_prefix):
...
-        self.consumer_thread = threading.Thread(target=self.consumer_thread_main)
-        self.consumer_thread.start()
+        self.consumer = None
+        self.consumer_thread = None
+        self.share_consumer = None
+        self.share_thread = None
+
+        if enable_share:
+            if not HAS_SHARE_CONSUMER:
+                raise RuntimeError("ShareConsumer requested but not available in this " "confluent_kafka build.")
```
k-raina: In case share consumer fails, should producer also shut down?

tests/soak/soakclient.py
```diff
@@ -329,6 +337,188 @@ def producer_error_cb(self, err):
...
+                        if err_details:
+                            self.logger.warning(
+                                "share: commit_sync had {} partition error(s): {}".format(
+                                    len(err_details), "; ".join(err_details)
+                                )
+                            )
+                            self.share_err_cnt += 1
+                            self.incr_counter("consumer.error", 1)
+                    else:
+                        self.share_consumer.commit_async()
```
k-raina, line 547: Should we plan to add ack callback to catch errors?

tests/soak/soakclient.py
```diff
@@ -329,6 +337,188 @@ def producer_error_cb(self, err):
...
+                # consumer: in explicit mode, released records get redelivered
+                # non-monotonically, which would drive false dup/gap counters.
+                if self.share_mode == 'implicit':
+                    hwkey = "{}-{}".format(msg.topic(), msg.partition())
+                    hw = hwmarks[hwkey]
+
+                    if hw > 0:
+                        if msg.offset() <= hw:
+                            self.logger.warning(
+                                "share: Old or duplicate message {} "
```
k-raina, line 504: Does this logic prints  "Old or duplicate message" for delivery retries?

tests/soak/soakclient.py
k-raina, line 298: Do we want to count corrupt messages also?

tests/soak/soakclient.py
```diff
@@ -329,6 +337,188 @@ def producer_error_cb(self, err):
...
+                            self.share_err_cnt += 1
+                            self.incr_counter("consumer.error", 1)
+                    else:
+                        self.share_consumer.commit_async()
+                except KafkaException as ex:
+                    self.logger.error("share: commit_{} exception: {}".format("sync" if use_sync else "async", ex))
+                    self.share_err_cnt += 1
+                    self.incr_counter("consumer.error", 1)
+
+        self.share_consumer.close()
```
k-raina: nit: should close be moved into finally?

tests/soak/soakclient.py
```diff
@@ -329,6 +337,188 @@ def producer_error_cb(self, err):
...
+
+    def share_thread_main(self):
+        """Share consumer thread main function"""
+        try:
+            self.share_run()
+        except KeyboardInterrupt:
+            self.logger.info("share: aborted by user")
+            self.run = False
+        except Exception as ex:
+            self.logger.fatal("share: fatal exception: {}\n{}".format(ex, traceback.print_exc()))
```
k-raina:
```suggestion
            self.logger.fatal("share: fatal exception: {}\n{}".format(ex, traceback.format_exc()))
```
Check if format_exec is more relavant? Same for rest of files

tests/soak/soakclient.py
```diff
@@ -446,50 +637,108 @@ def filter_config(conf, filter_out, strip_prefix):
             return out
 
         # Create topic (might already exist)
-        aconf = filter_config(conf, ["consumer.", "producer."], "admin.")
+        aconf = filter_config(conf, ["consumer.", "producer.", "share."], "admin.")
```
k-raina, line 690: Adding ".share" in filter config changes share.acknowledgement.mode -> acknowledgement.mode, which is an invalid config. i think this will cause crash for explicit mode. Is my understanding correct?

k-raina (commented):
Is there scope to common out code between soakclient.py and soakclient_perf.py?

tests/soak/soakclient.py
Ankith-Confluent: The second arg to filter_config is the drop list, not a rename. So share.acknowledgement.mode is dropped from aconf, not renamed to acknowledgement.mode.
Explicit mode still works because `sconf['share.acknowledgement.mode'] = 'explicit'` is set directly on the share conf after filter_config
I verified it end-to-end with --share --explicit.

tests/soak/soakclient.py
Ankith-Confluent: I think we should not count them because we already have another metric counting the erros.
So it would pollute the readings.

tests/soak/soakclient.py
Ankith-Confluent: yes, I will change the the log text.

tests/soak/soakclient.py
Ankith-Confluent: Yes we can do that, when I implemented this before, that mechanism was not implemented yet in Librdkafka side.
Will add.

Ankith-Confluent:
Thanks! @k-raina 
I have addressed the comments.
Regarding the point to common out the code, I think it's better to keep it separated and simple, in the future if we want to modify the perf script, we will be required to change the soak client as well.
What are your thoughts about this?

k-raina:
In the current codebase, the two files (soakclient.py and soakclient_perf.py) overlap heavily:
- Out of the 31 function definitions, **28 are byte-for-byte identical.**
- **Only 3 functions differ** in actual behavior: SoakRecord.__init__, SoakRecord.serialize, producer_run

Because the two files are largely copies, any fix applied to one must be manually ported to the other. Over time this leads to silent, unintended drift (for example, divergence in logging detail, error handing etc.).

One way to address this is to make soakclient.py the base class and soakclient_perf.py a subclass.

tests/soak/soakclient.py
k-raina: Does this need to be?
```suggestion
        self.last_committed = None
```

k-raina: Nit: Does this crash for msg with no header?

tests/soak/soakclient.py
```diff
@@ -134,36 +151,66 @@ def producer_status(self):
...
-        while self.run:
+        if self.perf:
+            # perf: produce in batches and pace per-batch with time.sleep
+            # instead of one produce()+blocking poll() per message. At high
+            # rates the per-message poll() floor (~hundreds of us) caps
+            # throughput well below -r; pacing a batch lets us sleep a
+            # duration the OS can actually honor. -r remains the target rate
+            # (an upper bound: if a batch takes longer than its time budget,
+            # we run as fast as we can, below -r).
+            batch = max(1, int(self.rate / 100))  # ~100 batches/sec
```
k-raina, line 169: Note : By default , rate is hardcoded to 80 in run.sh:

`time opentelemetry-instrument $testdir/soakclient.py -i $TESTID -t $topic -r 80 -f $1 $share_flag $explicit_flag $perf_flag`

So each batch is one message and batching only kicks in at -r ≥ 200.

k-raina (commented):
Overall LGTM! 
Since delta btw files is tiny, runtime branching works fine for this PR. In future if there are many hooks or variants, we might need to revisit and adopt  Inheritance.

k-raina (approved):
LGTM! Thanks for PR

PR #2233, opened 2026-04-17
Fix uv venv picking up Python 3.14 on new Semaphore AMI

k-raina:
## Root Cause
New Semaphore AMIs ship with Python 3.14 as the system default. Even though our pipeline calls `sem-version python 3.11` before `uv venv`, `uv` ignores this and auto-discovers 3.14. Since `google-re2` has no pre-built wheel for 3.14, pip/uv falls back to building from source, which fails because `absl/strings/string_view.h` is not installed on the agent.

## Fix
Pass `--python "$(command -v python)"` to all `uv venv` calls in `semaphore.yml` to ensure the pinned Python version is used

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -205,7 +205,7 @@ blocks:
             - sem-version python 3.11
             - pip install uv
             # use a virtualenv
-            - uv venv _venv && source _venv/bin/activate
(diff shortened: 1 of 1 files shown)


rayokota (approved):
Thanks @k-raina , LGTM

pranavrth (approved):
LGTM!. Thanks @k-raina

PR #2235, opened 2026-04-20
Fix consume() breaking out of chunked loop on first result

k-raina:
**Summary**
* Refactored the `Consumer.consume()` implementation in `Consumer.c` to accumulate messages across multiple 200ms chunks, ensuring that the requested number of messages is collected before returning, unless a timeout or error occurs. Also ensures proper cleanup of accumulated messages on error or signal interruption.


emasab (approved):
Looks good now that is reverted to previous version, thanks. The polls have side effects that can affect at least once semantics, when implemented in future by default. Polled messages could be autocommitted before being received by the application. They also affect the `poll.idle.ratio.avg` calculation.

Reverting this method only is fine as a user can set a shorter interval of 200ms anyway.

A different approach, allowing to receive signals immediately during the call, would require a redesign of Python/C interactions only through Futures, similarly to async calls, but it's too much just for this fix.

PR #2239, opened 2026-04-27
[KIP-932] Implement acknowledgement RPC

k-raina:
##### Summary
- Add `ShareConsumer.acknowledge(`) and `acknowledge_offset()`, explicit-mode acknowledgement APIs in the C extension. . Internally delegates to `rd_kafka_share_acknowledge_offset()` to work around Message not retaining the underlying rd_kafka_message_t.
- 12 integration tests in tests/test_ShareConsumer_ack.py covering the full ack workflow against a real KIP-932 broker:.
- Type stubs updated in cimpl.pyi to keep mypy and IDE autocomplete in sync with the new C surface.


k-raina:
Unrelated test failure as per base branch https://github.com/confluentinc/confluent-kafka-python/pull/2217#issuecomment-4293972433

src/confluent_kafka/cimpl.pyi
```diff
@@ -564,6 +564,10 @@ class ShareConsumer:
     def unsubscribe(self) -> None: ...
     def subscription(self) -> List[str]: ...
     def poll(self, timeout: float = -1) -> List[Message]: ...
+    def acknowledge(self, message: Message, ack_type: int = 1) -> None: ...
+    def acknowledge_offset(
+        self, topic: str, partition: int, offset: int, ack_type: int = 1
+    ) -> None: ...
```
ojasvajain:
```suggestion
    def acknowledge(self, message: Message, ack_type: int = 1) -> None: ...
    def acknowledge_offset(
        self, topic: str, partition: int, offset: int, ack_type: Union[AcknowledgeType, int] = ...
    ) -> None: ...
```

Seems to me that we are accepting both enum and raw integer. Also regarding type hint for default values -https://typing.python.org/en/latest/guides/writing_stubs.html#functions-and-methods

src/confluent_kafka/cimpl.pyi
```diff
@@ -564,6 +564,10 @@ class ShareConsumer:
     def unsubscribe(self) -> None: ...
     def subscription(self) -> List[str]: ...
     def poll(self, timeout: float = -1) -> List[Message]: ...
+    def acknowledge(self, message: Message, ack_type: int = 1) -> None: ...
```
ojasvajain:
```suggestion
    def acknowledge(self, message: Message, ack_type: Union[AcknowledgeType, int] = ...) -> None: ...
```

Seems to me that we are accepting both enum and raw integer. Also regarding type hint for default values -https://typing.python.org/en/latest/guides/writing_stubs.html#functions-and-methods

src/confluent_kafka/_model/__init__.py
```diff
@@ -193,3 +193,25 @@ def __lt__(self, other: object) -> bool:
         if not isinstance(other, ElectionType):
             return NotImplemented
         return self.value < other.value
+
+
+class AcknowledgeType(Enum):
```
ojasvajain: Can we use `IntEnum` instead? It will have the `__index__` method defined by default but I am not sure what side-effects it can cause.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -310,6 +310,115 @@ static PyObject *ShareConsumer_poll(ShareConsumerHandle *self,
...
+                return NULL;
+
+        if (ack_type < (int)RD_KAFKA_SHARE_ACKNOWLEDGE_TYPE_ACCEPT ||
+            ack_type > (int)RD_KAFKA_SHARE_ACKNOWLEDGE_TYPE_REJECT) {
+                PyErr_Format(PyExc_ValueError,
+                             "Invalid ack_type %d (expected 1=ACCEPT, "
+                             "2=RELEASE, or 3=REJECT)",
+                             ack_type);
+                return NULL;
+        }
```
ojasvajain: This check is already present in `rd_kafka_share_acknowledge_offset`. Shall we skip it from here? We are already relying on validations inside `rd_kafka_share_acknowledge_offset` for partition and offset validation


```
if (!rkshare || !topic || partition < 0 || offset < 0)
                return RD_KAFKA_RESP_ERR__INVALID_ARG;
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -310,6 +310,115 @@ static PyObject *ShareConsumer_poll(ShareConsumerHandle *self,
...
```
ojasvajain: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2239/changes#r3167763083)

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -425,6 +534,44 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "  Acknowledge a previously polled message in explicit acknowledgement\n"
+     "  mode. Tells the broker how to handle the record.\n"
+     "\n"
+     "  :param Message message: A message returned by poll().\n"
+     "  :param AcknowledgeType ack_type: ACCEPT (default), RELEASE, or "
+     "REJECT.\n"
+     "  :raises KafkaException: if the consumer is not in explicit\n"
+     "                          acknowledgement mode, the message is no "
+     "longer\n"
+     "                          in-flight, or the offset is a GAP record.\n"
```
ojasvajain: I am not sure if this applicable to the acknowledge API. If it is a GAP record, does the user even get the message object for it?

src/confluent_kafka/_model/__init__.py
```diff
@@ -193,3 +193,25 @@ def __lt__(self, other: object) -> bool:
...
+
+    Values:
+    -------
+    """
+
+    #: Record was processed successfully — broker will not redeliver it.
+    ACCEPT = 1
+    #: Could not process — return to the share group for another delivery attempt.
+    RELEASE = 2
+    #: Permanently failed — do NOT redeliver (poison pill).
```
PratRanj07:
```suggestion
    #: Could not process - Do not release for another delivery attempt
```

src/confluent_kafka/_model/__init__.py
```diff
@@ -193,3 +193,25 @@ def __lt__(self, other: object) -> bool:
...
+    Share Consumer acknowledgement type used to tell the broker how to
+    handle a polled message in explicit acknowledgement mode.
+
+    Values:
+    -------
+    """
+
+    #: Record was processed successfully — broker will not redeliver it.
+    ACCEPT = 1
+    #: Could not process — return to the share group for another delivery attempt.
```
PratRanj07:
```suggestion
    #: Could not process — Release it for another delivery attempt
```

PratRanj07 (changes requested):
Some nits

k-raina:
@pranavrth Addressed feedback. 
i have updated pipeline to cleanly support  AK 4.2 broker for tests. 
Added some TODOs to cleanup pipeline which can be handled in later PR before merging to main branch.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -310,6 +310,104 @@ static PyObject *ShareConsumer_poll(ShareConsumerHandle *self,
 }
 
 
+/**
+ * @brief Acknowledge previously polled message.
+ *
+ * Internally delegates to rd_kafka_share_acknowledge_offset() because the
+ * Python Message object does not retain the underlying rd_kafka_message_t
+ * pointer (Message_new0 copies fields out and destroys the rkm)
+ */
```
PratRanj07, line 408: Maybe add a TODO to check whether we should match the Java functionality that we should only acknowledge the successful messages with this API and the error messages with the offset API?
Like for the case of Java, they throw an error if the successful messages are processed with the offset API. I think that decision will be finalised once we have the discussion on the differences between the Java clients and the NJC clients

tests/test_ShareConsumer_ack.py
```diff
@@ -0,0 +1,591 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
ojasvajain: We should add license details

ojasvajain (commented):
I have resolved my earlier comments. Thanks.
Just one minor comment.

src/confluent_kafka/_model/__init__.py
```diff
@@ -193,3 +193,19 @@ def __lt__(self, other: object) -> bool:
...
+class AcknowledgeType(IntEnum):
+    """
+    Share Consumer acknowledgement type.
+
+    Used to tell the broker how to
+    handle a polled message in explicit acknowledgement mode.
+    """
+
+    #: Record was processed successfully — broker will not redeliver it.
+    ACCEPT = 1
```
PratRanj07: Should we match the values here to the librdkafka values as done for the other APIs [here](https://github.com/confluentinc/confluent-kafka-python/pull/2239/changes/fff458458c8c46e9916d2ae90d043d9f69f2fa83#diff-0317ccd97e65b2360105af3772dbfbf985d7dfe657a9d80f1495adc99e5d5eedR188)

**PREFERRED = cimpl.ELECTION_TYPE_PREFERRED**

tests/test_ShareConsumer_ack.py
```diff
@@ -0,0 +1,591 @@
...
+        sc.subscribe([topic])
+        _warmup(sc)
+        _produce(topic, 1)
+
+        msgs = _collect(sc, 1, POLL_TIMEOUT)
+        assert msgs, 'no messages received'
+
+        with pytest.raises(KafkaException) as ex:
+            sc.acknowledge(msgs[0], AcknowledgeType.ACCEPT)
+        assert ex.value.args[0].code() == -172  # __STATE
```
PratRanj07: Why do we have hard coded error number? KafkaError.__STATE?

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -393,29 +310,127 @@ static PyObject *ShareConsumer_poll(ShareConsumerHandle *self,
...
+                return NULL;
+
+        /* No need to range-check ack_type here —
+         * rd_kafka_share_acknowledge_offset() does it and returns INVALID_ARG,
+         * which we'll surface as KafkaException. */
+
+        if (!msg->topic || msg->topic == Py_None) {
+                PyErr_SetString(PyExc_ValueError, "Message topic is None");
+                return NULL;
+        }
```
PratRanj07: The topic check is also being done on the librdkafka side. Keeping this cause some inconsistency as we will have ValueError for message topic null whereas KafkaException for other errors

tests/test_ShareConsumer_ack.py
```diff
@@ -0,0 +1,591 @@
...
+    ShareConsumer,
+)
+from confluent_kafka.admin import (
+    AdminClient,
+    AlterConfigOpType,
+    ConfigEntry,
+    ConfigResource,
+)
+
+BOOTSTRAP_SERVERS = 'localhost:9092'
```
PratRanj07: Is this intended?

tests/test_ShareConsumer_ack.py
```diff
@@ -0,0 +1,591 @@
...
+        while not first_batch and time.monotonic() < deadline:
+            for m in sc.poll(timeout=2.0):
+                if m.error() is None:
+                    first_batch.append(m)
+        assert first_batch, 'never received any messages'
+
+        with pytest.raises(KafkaException) as ex:
+            sc.poll(timeout=2.0)
+        err = ex.value.args[0]
+        assert err.code() == -172
```
PratRanj07: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2239/changes#r3234901415)

tests/test_ShareConsumer_ack.py
```diff
@@ -0,0 +1,591 @@
...
+        # Sleep past the lock duration — broker reverts record to Available.
+        time.sleep(SHORT_LOCK_MS / 1000.0 + LOCK_EXPIRY_BUFFER_S)
+
+        # Acking the now-stale message: librdkafka may raise or silently
+        # accept. Either way, the record is the broker's authority.
+        try:
+            sc.acknowledge(first, AcknowledgeType.ACCEPT)
+        except KafkaException as e:
+            code = e.args[0].code()
+            assert code in (-172, 121), f'unexpected error code {code}'
```
PratRanj07: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2239/changes#r3234901415)

PratRanj07 (changes requested):
Some comments

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,742 @@
...
+            sc.acknowledge(m, AcknowledgeType.ACCEPT)
+        sc.poll(timeout=2.0)
+    finally:
+        sc.close()
+
+
+# --- queue-style distribution across multiple consumers --------------------
+
+
+def test_explicit_two_consumers_share_workload(kafka_cluster):
```
PratRanj07: We are producing only 30 messages, Most probably the 1st consumer will get all those messages. Maybe try 10000 or something?

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,742 @@
...
+                    seen.add((m.topic(), m.partition(), m.offset()))
+                    sc.acknowledge(m, AcknowledgeType.ACCEPT)
+        assert seen == {released_coords}, f'expected only {released_coords}, got {seen}'
+    finally:
+        sc.close()
+
+
+def test_double_ack_same_record_is_lenient(kafka_cluster):
+    """Double-ack is silent today (Java raises; NJC may tighten later).
+    Next poll must not raise _STATE."""
```
PratRanj07: We can ack any number of times we want before committing in java too

PratRanj07 (commented):
Some comments

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,742 @@
...
+        assert msgs, 'no messages received'
+
+        with pytest.raises(KafkaException) as ex:
+            sc.acknowledge(msgs[0], AcknowledgeType.ACCEPT)
+        assert ex.value.args[0].code() == KafkaError._STATE
+    finally:
+        sc.close()
+
+
+def test_implicit_mode_autocommits_on_next_poll(kafka_cluster):
```
PratRanj07, line 64: Maybe add a TODO to check if the acknowledgements are successful using acknowledgement callback when it has been implemented

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,742 @@
...
+        sc1.close()
+        sc2.close()
+
+
+# --- delivery limit, atomicity, transactions ------------------------------
+
+# TODO KIP-932: late-ack semantics differ between Java (raises
+# InvalidRecordStateException) and NJC (silent success). Add a test when the
+# Java-vs-NJC contract is settled. Basic lock-expiry redelivery is already
+# covered by test_share_consumer.py::test_records_redelivered_after_lock_timeout.
```
PratRanj07: This error should be returned by the broker which we would capture? I dont understand the context of this

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,742 @@
...
+        sc2.subscribe([topic])
+        leftovers = drain_share_consumers([sc2], 1, timeout_s=5.0, ack_type=AcknowledgeType.ACCEPT)[0]
+        assert leftovers == [], 'acknowledge_offset(ACCEPT) should prevent redelivery'
+    finally:
+        sc2.close()
+
+
+def test_acknowledge_offset_invalid_coords_raise(kafka_cluster):
+    """Bogus coords raise KafkaException; the consumer stays usable.
+    librdkafka collapses bogus-topic/partition/offset to _STATE rather than _INVALID_ARG.
```
PratRanj07: For bogus topic partition offset the librdkafka sends INVALID _ARG only

PratRanj07 (approved):
LGTM

ojasvajain (approved):

ojasvajain:
~~Approved, but~~ let's get the checks passing.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -240,45 +240,6 @@ def test_unacked_records_block_next_poll(kafka_cluster):
         sc.close()
 
 
-# --- queue-style distribution across multiple consumers --------------------
-
-
-def test_explicit_two_consumers_share_workload(kafka_cluster):
```
PratRanj07: Why are we removing this test?
We should run both the consumer one by one till we get all the records and just make sure we have unique records

PratRanj07 (changes requested):

tests/integration/share_consumer/test_share_consumer_ack.py
k-raina: 1. High counts (e.g., n=10000): Consumer A drains the full batch in its first poll. Before it can ack all records, the broker's 1s record lock expires and the same records are redelivered to consumer B. assert overlap == set() then fails — both consumers receive the same offsets.
2. Low counts (e.g., n<100): Consumer A fetches and acks  the entire batch before consumer B issues its first fetch. Consumer B receives nothing, and assert received_1 and received_2 fails.

There is no value of n that avoids both failure modes — tuning the record count just trades one failure for the other, which is why the test is fundamentally unreliable.

PratRanj07 (approved):
LGTM! Thanks @k-raina 
Add the above discussed test in the next PR

ojasvajain (approved):

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -514,6 +615,44 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "\n"
+     "  :param Message message: A message returned by poll().\n"
+     "  :param AcknowledgeType ack_type: ACCEPT (default), RELEASE, or "
+     "REJECT.\n"
+     "  :raises KafkaException: if the consumer is not in explicit\n"
+     "                          acknowledgement mode, the message is no "
+     "longer\n"
+     "                          in-flight, ack_type is invalid, or "
+     "message.topic() is None.\n"
+     "  :raises RuntimeError: if called on a closed share consumer.\n"
```
pranavrth, line 653: Add the below TODO. We might return Value and Type errors as well. In some cases, we still need to understand what to send like __INVALID_ARG is sent as KafkaError in normal producer and consumer. I think it should have been converted to ValueError etc.

```
/* TODO KIP-932: librdkafka error code → Python exception mapping is
 * provisional. Currently, we use following for the Share Consumer 
 *   _INVALID_ARG → ValueError  (matches Java IllegalArgumentException)
 *   _STATE       → RuntimeError (matches Java IllegalStateException)
 *   Need to understand if per-partition broker errors stay can as KafkaError
 *   in commit_sync's result dict (mirrors Java's Map<TopicIdPartition, Optional<...>>)
 *   or we need to change that part as well.
 * 
 * We should have proper translation logic for this.
 * 
 * Revisit this mapping holistically once:
 *   - librdkafka's share-consumer error surface is stable (some codes
 *     may be redefined as work progresses), and
 *   - the equivalent translation lands on commit_sync / commit_async /
 *     ack-callback paths (currently TODO'd separately).
 * The goal is for each librdkafka return code in the share consumer
 * surface to map to the Python exception a user porting from Java
 * would expect. */
```

src/confluent_kafka/cimpl.pyi
```diff
@@ -568,6 +569,10 @@ class ShareConsumer:
     def unsubscribe(self) -> None: ...
     def subscription(self) -> List[str]: ...
     def poll(self, timeout: float = -1) -> List[Message]: ...
```
pranavrth, line 577: Add a TODO to remember that we need to create a class like Messages as return type from poll. Java return a new class oject instead of List of records. This provides a container class to hold more information if needed.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -514,6 +615,44 @@ static PyMethodDef ShareConsumer_methods[] = {
      "  :raises KeyboardInterrupt: if Ctrl+C pressed during consumption\n"
      "\n"},
 
+    {"acknowledge", (PyCFunction)ShareConsumer_acknowledge,
```
pranavrth, line 635: It can raise TypeError as well.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -514,6 +615,44 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "REJECT.\n"
+     "  :raises KafkaException: if the consumer is not in explicit\n"
+     "                          acknowledgement mode, the message is no "
+     "longer\n"
+     "                          in-flight, ack_type is invalid, or "
+     "message.topic() is None.\n"
+     "  :raises RuntimeError: if called on a closed share consumer.\n"
+     "\n"},
+
+    {"acknowledge_offset", (PyCFunction)ShareConsumer_acknowledge_offset,
```
pranavrth, line 656: It can raise TypeError as well.

src/confluent_kafka/src/ShareConsumer.c
pranavrth, line 653: I am hoping that these are not final wordings and we will update this later.

src/confluent_kafka/cimpl.pyi
```diff
@@ -568,6 +569,10 @@ class ShareConsumer:
...
+    def acknowledge(self, message: Message, ack_type: Union[AcknowledgeType, int] = ...) -> None: ...
+    def acknowledge_offset(
+        self, topic: str, partition: int, offset: int, ack_type: Union[AcknowledgeType, int] = ...
```
pranavrth:
```suggestion
        self, topic: str, partition: int, offset: int, ack_type: AcknowledgeType = ...
```

src/confluent_kafka/cimpl.pyi
```diff
@@ -568,6 +569,10 @@ class ShareConsumer:
     def unsubscribe(self) -> None: ...
     def subscription(self) -> List[str]: ...
     def poll(self, timeout: float = -1) -> List[Message]: ...
+    def acknowledge(self, message: Message, ack_type: Union[AcknowledgeType, int] = ...) -> None: ...
```
pranavrth:
```suggestion
    def acknowledge(self, message: Message, ack_type: AcknowledgeType = ...) -> None: ...
```

tests/common/__init__.py
```diff
@@ -146,33 +146,19 @@ def unique_id(prefix):
...
-
-    TODO KIP-932: when explicit-ack mode lands in the Python wrapper
-    (ShareConsumer.acknowledge()), update this helper to ack each message as
-    it's drained, otherwise the broker will return INFLIGHT-records errors
-    on the second poll. Same caveat exists in the librdkafka tests.
-
-    TODO KIP-932: after the final poll() in the loop below, the last batch of
-    records is never implicitly acknowledged (no subsequent poll to piggyback
-    on). Some tests assume the tail batch is ack'd. Fix once explicit-ack is
-    exposed: emit an explicit ack for the last drained batch before returning.
```
pranavrth, line 175: This still holds true.

tests/common/__init__.py
```diff
@@ -181,6 +167,8 @@ def drain_share_consumers(consumers, n_expected, timeout_s=20.0, poll_timeout_s=
             for m in sc.poll(timeout=poll_timeout_s):
```
pranavrth: We will still need some commit calls here. Add a TODO maybe for that.

tests/integration/share_consumer/test_share_consumer_ack.py
pranavrth, line 1: I think we should write unit test as well to verify type and value of input parameters. Maybe add it in commit PR with those APIs. But add a TODO to remember.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Integration tests for ShareConsumer acknowledgement (KIP-932)."""
+
+import time
+
+import pytest
+
+from confluent_kafka import AcknowledgeType, KafkaError, KafkaException, Producer
```
pranavrth: We might need to think about the module for `ShareConsumer`. In Java, its `org.apache.kafka.clients.consumer.KafkaShareConsumer` and normal consumer is also at the same level `org.apache.kafka.clients.consumer.KafkaConsumer`. This is the same case with us as well. But confirming. Let's put this in our interface doc.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+    """Set one dynamic share-group config."""
+    res = ConfigResource(
+        ConfigResource.Type.GROUP,
+        group_id,
+        incremental_configs=[
+            ConfigEntry(name, str(value), incremental_operation=AlterConfigOpType.SET),
+        ],
+    )
+    for f in kafka_cluster.admin().incremental_alter_configs([res]).values():
+        f.result()
```
pranavrth: Better to move to test common so that it can be used anywhere.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+    try:
+        sc1.subscribe([topic])
+
+        producer = kafka_cluster.cimpl_producer()
+        produced = [f'msg-{i}'.encode() for i in range(n)]
+        for v in produced:
+            producer.produce(topic, value=v)
+        producer.flush(timeout=10.0)
+
+        first = drain_share_consumers([sc1], n)[0]
```
pranavrth: You might not even get any record after drain. Drain is doing exactly what this test is doing. Every time you fetch records, drain would just acknowledge previous batch. Only last batch is returned. I think we should have another function which returns first batch of received records. That can be used here.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+        assert ex.value.args[0].code() == KafkaError._STATE
+    finally:
+        sc.close()
+
+
+def test_implicit_mode_autocommits_on_next_poll(kafka_cluster):
+    """Next poll() in implicit mode commits the previous batch."""
+    topic = kafka_cluster.create_topic_and_wait_propogation('test-share-consumer-ack-implicit-autocommit')
+    group_id = unique_id('test-share-consumer-ack-implicit-autocommit')
+    n = 3
```
pranavrth: Use better name.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+
+    sc1 = kafka_cluster.share_consumer({'group.id': group_id, 'share.acknowledgement.mode': 'implicit'})
+    try:
+        sc1.subscribe([topic])
+
+        producer = kafka_cluster.cimpl_producer()
+        produced = [f'msg-{i}'.encode() for i in range(n)]
+        for v in produced:
```
pranavrth: better name instead of `v`

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+            producer.produce(topic, value=v)
+        producer.flush(timeout=10.0)
+
+        first = drain_share_consumers([sc1], n)[0]
+        assert len(first) == n
+
+        sc1.poll(timeout=2.0)  # implicit-ack the prior batch
```
pranavrth: We should check that this poll has not received more records. By with my above suggested way, what you do is first find the first batch of records, then acknowledge and then drain remaining records. Verify x records received with first call then n-x should be received after draining.

This is one way. You can think of something else if you wish but atleast check the number of records here.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+
+        producer = kafka_cluster.cimpl_producer()
+        for i in range(n):
+            producer.produce(topic, value=f'msg-{i}'.encode())
+        producer.flush(timeout=10.0)
+
+        msgs = drain_share_consumers([sc1], n, ack_type=AcknowledgeType.ACCEPT)[0]
+        assert len(msgs) == n
+
+        sc1.poll(timeout=2.0)  # flush ACCEPTs
```
pranavrth: Verify returned records here as well. Same everywhere.

tests/common/__init__.py
pranavrth: Better naming for `m`.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+
+        producer = kafka_cluster.cimpl_producer()
+        for i in range(n):
+            producer.produce(topic, value=f'msg-{i}'.encode())
+        producer.flush(timeout=10.0)
+
+        msgs = drain_share_consumers([sc1], n, ack_type=AcknowledgeType.REJECT)[0]
+        assert len(msgs) == n
+
+        sc1.poll(timeout=2.0)  # flush REJECTs
```
pranavrth: There is nothing like flush. Remove flush part.
poll() is not used to flush records. You can receive more records.

tests/common/__init__.py
```diff
@@ -146,33 +146,19 @@ def unique_id(prefix):
...
-    TODO KIP-932: when explicit-ack mode lands in the Python wrapper
-    (ShareConsumer.acknowledge()), update this helper to ack each message as
-    it's drained, otherwise the broker will return INFLIGHT-records errors
-    on the second poll. Same caveat exists in the librdkafka tests.
-
-    TODO KIP-932: after the final poll() in the loop below, the last batch of
-    records is never implicitly acknowledged (no subsequent poll to piggyback
-    on). Some tests assume the tail batch is ack'd. Fix once explicit-ack is
-    exposed: emit an explicit ack for the last drained batch before returning.
+def drain_share_consumers(consumers, n_expected, timeout_s=20.0, poll_timeout_s=0.5, *, ack_type=None):
```
pranavrth: Better name for `n_expected`.
Is it per consumer or overall?

tests/common/__init__.py
```diff
@@ -181,6 +167,8 @@ def drain_share_consumers(consumers, n_expected, timeout_s=20.0, poll_timeout_s=
             for m in sc.poll(timeout=poll_timeout_s):
                 if m.error() is None:
                     bucket.append(m)
+                    if ack_type is not None:
+                        sc.acknowledge(m, ack_type)
         if sum(len(b) for b in received) >= n_expected:
```
pranavrth: Better name for `b`

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+        sc.subscribe([topic])
+
+        producer = kafka_cluster.cimpl_producer()
+        producer.produce(topic, value=b'msg-0')
+        producer.flush(timeout=10.0)
+
+        first_batch = drain_share_consumers([sc], 1)[0]
+        assert first_batch
+        first = first_batch[0]
+        coords = (first.topic(), first.partition(), first.offset())
```
pranavrth: Better name for `coords`. So the idea is that whoever is reading the code should understand what is happening. They don't need to understand the logic to understand what is happening.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+            for m in sc.poll(timeout=2.0):
+                if m.error() is None:
+                    first_batch.append(m)
+        assert first_batch, 'never received any messages'
+
+        with pytest.raises(KafkaException) as ex:
+            sc.poll(timeout=2.0)
+        err = ex.value.args[0]
+        assert err.code() == KafkaError._STATE
+        assert 'not been acknowledged' in err.str()
```
pranavrth: I think we should remove this. Anything changing in librdkafka will break this test.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+        assert seen, 'consumer unusable after unsubscribe + stale ack'
+    finally:
+        sc.close()
+
+
+def test_double_close_is_idempotent(kafka_cluster):
+    """close() twice must be a no-op (__exit__ relies on this)."""
+    sc = kafka_cluster.share_consumer()
+    sc.close()
+    sc.close()
```
pranavrth: Move this to other share consumer test file. I think 2nd `sc.close()` will raise some exception.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+                    sc.acknowledge(m, AcknowledgeType.RELEASE)
+                    releases += 1
+                    if releases >= 5:
+                        break
+        assert releases >= 5, f'only saw {releases} releases, expected 5'
+        assert coords is not None
+
+        # Flush the final RELEASE.
+        try:
+            sc.poll(timeout=2.0)
```
pranavrth, line 302: Verify we don't receive the record here.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+        except KafkaException:
+            pass
+
+        # Watch for ~10s — must not redeliver.
+        deadline = time.time() + 10.0
```
pranavrth, line 315: WE need to wait for acquisition lock timeout and then try to poll for few attempts or seconds.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+        while time.time() < deadline:
+            for m in sc.poll(timeout=2.0):
+                if m.error() is None and (m.topic(), m.partition(), m.offset()) == coords:
+                    pytest.fail(f'record {coords} redelivered past delivery.attempt.limit')
+    finally:
+        sc.close()
+
+
+def test_open_transaction_stalls_share_group(kafka_cluster):
```
pranavrth: This should be in `test_share_consumer.py` file. There is nothing related to acks that we are testing here.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+        # the batch the len() assertion below trips.
+        batch = []
+        deadline = time.time() + 20.0
+        while time.time() < deadline and len(batch) < n:
+            for m in sc.poll(timeout=1.0):
+                if m.error() is None:
+                    batch.append(m)
+            if len(batch) >= n:
+                break
+        assert len(batch) == n, f'expected {n} in a single Acquired set, got {len(batch)}'
```
pranavrth: This might give you error if record comes in multiple batches.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,725 @@
...
+                    second_batch.append(m)
+            if second_batch:
+                break
+        assert second_batch, 'never received second-batch record'
+
+        with pytest.raises(KafkaException) as ex:
+            sc.acknowledge_offset(*stale_coords, AcknowledgeType.ACCEPT)
+        assert ex.value.args[0].code() == KafkaError._STATE
+
+        # Ack the live batch so close() doesn't trip.
```
pranavrth: What does trip mean here?

tests/integration/share_consumer/test_share_consumer_ack.py
pranavrth, line 1: Overall good tests.

pranavrth (changes requested):
Overall looks good. Nothing major. Mostly nits.

tests/integration/share_consumer/test_share_consumer_ack.py
```diff
@@ -0,0 +1,731 @@
...
+        # (num_messages - x records) so the total adds up.
+        remaining = drain_share_consumers([sc1], num_messages - len(first_batch), timeout_s=10.0)[0]
+
+        received_values = {m.value() for m in first_batch + remaining}
+        assert received_values == set(
+            produced
+        ), f'sc1 missed records: produced={set(produced)} received={received_values}'
+
+        # Final poll acks the tail batch returned by drain.
+        sc1.poll(timeout=2.0)
```
pranavrth, line 102: Use commit_async to flush final records. Using poll means that you are expecting records.

pranavrth (approved):
Minor nit to remember for next PR.

LGTM!. Thanks @k-raina !

PR #2241, opened 2026-04-29
[KIP-932] Add share consumer commit workflow

k-raina:
#### Summary
- Add ShareConsumer.commit(asynchronous=True, timeout=-1) to the C extension, wired to librdkafka's rd_kafka_share_commit_async() / rd_kafka_share_commit_sync().
- 28 integration tests in tests/test_ShareConsumer_commit.py against a real KIP-932 broker
- Type stub updated in cimpl.pyi.

####  Design decisions
- Single method with asynchronous= kwarg, not separate commit_sync() / commit_async(). Matches the existing Consumer.commit() precedent in this repo.
- Default asynchronous=True to match Consumer.commit(). 
- No chunked polling on sync commit. `rd_kafka_share_commit_sync()` drains pending acks into the request, so retrying it would silently lose the first call's results. Sync commit just releases the GIL and blocks.


src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -419,6 +419,73 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+ *
+ * Implicit mode auto-converts ACQUIRED → ACCEPT inside librdkafka before
+ * sending. Sync mode returns a per-partition result list; async mode returns
+ * None. By default, commit is asynchronous. In async mode, the application
+ * receives no immediate feedback. In sync mode, the application blocks until
+ * the broker responds or the timeout expires, and receives a list of
+ * TopicPartition where each entry's .error field carries the per-partition
+ * acknowledgement result.
+ */
+static PyObject *ShareConsumer_commit(ShareConsumerHandle *self,
```
PratRanj07: To match the Java signature we should have two separate commit Apis, commitSync and commitAsync

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -419,6 +419,73 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
 
+/**
+ * @brief Commit pending acknowledgements.
+ *
+ * Implicit mode auto-converts ACQUIRED → ACCEPT inside librdkafka before
+ * sending. Sync mode returns a per-partition result list; async mode returns
+ * None. By default, commit is asynchronous. In async mode, the application
+ * receives no immediate feedback. In sync mode, the application blocks until
+ * the broker responds or the timeout expires, and receives a list of
+ * TopicPartition where each entry's .error field carries the per-partition
```
PratRanj07: Java returns a Map<TopicPartition, Optional<KafkaException>>, we should match that signature

PratRanj07 (changes requested):

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: Agreed. Also, there are 2 variations of commitSync in java, one that doesn't accept a timeout and the one that does.

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: You might want to address this todo now, now that close changes are merged. Otherwise, an error is thrown when trying to link with the qfk branch of librdkafka.

ojasvajain (changes requested):
Left a couple of comments. Can review again after we modify the APIs according to Java client.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1084,6 +1104,13 @@ PyObject *Message_new0(const Handle *handle, const rd_kafka_message_t *rkm) {
         else
                 self->latency = -1;
 
+        /* Share consumer: 0 from librdkafka means unavailable; store as -1
+         * so the Python accessor returns None for non-share consumers. */
+        {
```
PratRanj07: Remove these line seperators

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,85 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+ *
+ * @returns dict mapping TopicPartition -> None on success or KafkaError on
+ *          per-partition failure (matching Java's
+ *          Map<TopicIdPartition, Optional<KafkaException>>). Empty dict when
+ *          no acknowledgements are pending.
+ */
+static PyObject *ShareConsumer_commit_sync(ShareConsumerHandle *self,
+                                           PyObject *args,
+                                           PyObject *kwargs) {
+        /* Default matches Java's default.api.timeout.ms (60s). */
```
PratRanj07:
```suggestion
        /* default timeout */
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -653,6 +732,48 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "(``share.acknowledgement.mode=implicit``,\n"
+     "  the default), all records returned by the previous :py:func:`poll` "
+     "call\n"
+     "  are auto-converted to ACCEPT before being sent. In explicit mode, "
+     "only\n"
+     "  records previously passed to :py:func:`acknowledge` /\n"
+     "  :py:func:`acknowledge_offset` are sent.\n"
+     "\n"
+     "  :param float timeout: Maximum time to block (seconds). Default: 60\n"
+     "                        (matches Java's ``default.api.timeout.ms``).\n"
```
PratRanj07: I dont think we should mention these Java related comments in the public docs

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -653,6 +732,48 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "                        Pass -1 for infinite.\n"
+     "  :returns: Dict mapping TopicPartition to None on success or "
+     "KafkaError\n"
+     "           on per-partition failure (matching Java's\n"
```
PratRanj07: Same

src/confluent_kafka/cimpl.pyi
ojasvajain, line 1: Lots of style related changes in this file that are unrelated to the changes done. Are they required?

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: Can we remove this `matches Java's ``default.api.timeout.ms``)`?

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -653,6 +732,48 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "  the default), all records returned by the previous :py:func:`poll` "
+     "call\n"
+     "  are auto-converted to ACCEPT before being sent. In explicit mode, "
+     "only\n"
+     "  records previously passed to :py:func:`acknowledge` /\n"
+     "  :py:func:`acknowledge_offset` are sent.\n"
+     "\n"
+     "  :param float timeout: Maximum time to block (seconds). Default: 60\n"
+     "                        (matches Java's ``default.api.timeout.ms``).\n"
+     "                        Pass -1 for infinite.\n"
```
ojasvajain, line 790: Is this a Python client convention we are trying to follow as i don't see -1 being supported in java?

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2241/changes#r3261182273)

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,85 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+/**
+ * @brief Synchronously commit pending acknowledgements.
+ *
+ * Implicit mode auto-converts the records returned by the latest poll() to
+ * ACCEPT inside librdkafka before sending. Blocks until all broker replies
+ * arrive or the timeout expires.
+ *
+ * @returns dict mapping TopicPartition -> None on success or KafkaError on
+ *          per-partition failure (matching Java's
+ *          Map<TopicIdPartition, Optional<KafkaException>>). Empty dict when
```
ojasvajain: Same

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1380 @@
...
+
+        msgs = drain_share_consumers([sc], n, timeout_s=30.0)[0]
+        assert len(msgs) >= n // 2, f'only drained {len(msgs)}/{n}'
+
+        result = sc.commit_sync(timeout=15.0)
+        assert isinstance(result, dict)
+        for tp, err in result.items():
+            assert err is None, f'{tp.topic}[{tp.partition}] -> {err}'
+    finally:
+        sc.close()
```
PratRanj07, line 206: 100 records wont be a large batch. Maybe if you want to check for a large batch go for 10k+ messages?

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1380 @@
...
+
+# --- state-machine edge cases --------------------------------------------
+
+
+def test_commit_after_lock_expiry(kafka_cluster):
+    """Acking + committing past the lock duration: librdkafka may raise
+    _STATE on the ack, or the commit may surface INVALID_RECORD_STATE
+    per partition. Either is acceptable.
+
+    Relies on the suite-wide 1s lock duration from broker_conf().
```
PratRanj07: Librdkafka doesnt send anything on its own in this case, it relies o broker to reject it

PratRanj07 (commented):
Some more comments

ojasvajain (commented):
Halfway through the PR

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,85 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+        PyThreadState *thread_state              = NULL;
+        PyObject *result;
+        static char *kws[] = {"timeout", NULL};
+
+        if (!self->rkshare) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                ERR_MSG_SHARE_CONSUMER_CLOSED);
+                return NULL;
+        }
+
```
ojasvajain, line 525: Check if we need to release GIL so that callbacks can we triggered.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,85 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+
+/**
+ * @brief Asynchronously commit pending acknowledgements.
+ *
+ * Returns immediately; broker results are delivered via the
+ * share_acknowledgement_commit_cb (when configured).
+ */
+static PyObject *ShareConsumer_commit_async(ShareConsumerHandle *self,
+                                            PyObject *ignore) {
```
ojasvajain, line 566: Check if we need to release GIL so that callbacks can we triggered.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1570,6 +1597,40 @@ PyObject *c_parts_to_py(const rd_kafka_topic_partition_list_t *c_parts) {
...
+        for (i = 0; i < (size_t)c_parts->cnt; i++) {
+                const rd_kafka_topic_partition_t *rktpar = &c_parts->elems[i];
+                PyObject *key = c_part_to_py(rktpar);
+                PyObject *val = KafkaError_new_or_None(rktpar->err, NULL);
+
+                if (!key || !val || PyDict_SetItem(result, key, val) == -1) {
+                        Py_XDECREF(key);
+                        Py_XDECREF(val);
+                        Py_DECREF(result);
+                        return NULL;
```
ojasvajain: Should we return an exception in this case?

src/confluent_kafka/src/ShareConsumer.c
k-raina: Code to release GIL is already at https://github.com/confluentinc/confluent-kafka-python/pull/2241/changes/BASE..b00a3c97a81926e215579e84ae993921f905e91f#diff-0125d4ace79a1ba13349437806f29c794cb9e577ef69e2e8857cd6f52c9aed06R529

src/confluent_kafka/src/ShareConsumer.c
k-raina: Commit Async is non-blocking on librdkafka side, so we should not need to release GIL

src/confluent_kafka/src/ShareConsumer.c
k-raina: This is aligned with Consumer.c

src/confluent_kafka/cimpl.pyi
k-raina: This is done by style fix, as we added change for delivery count, commit sync and asyc

src/confluent_kafka/src/confluent_kafka.c
k-raina: Yes, NULL will represent exception to upstream functions. Which will handle it gracefully

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: Keeping it open as this is something we want to finalize if we want to follow the Python way or the Java way.

src/confluent_kafka/src/confluent_kafka.c
k-raina: Check

k-raina: @ojasvajain 
Adding a CallState is not needed here as this helper runs with the GIL held and doesn't cross any librdkafka/callback boundary, so there's nothing to propagate across threads.

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: I think it will be required because in librdkafa, as part of commit_async, we drain the cb queue and invoke them. Executing the callbacks would require GIL.

src/confluent_kafka/src/confluent_kafka.c
ojasvajain: Didn't get you. I was talking about setting an exception before returning NULL.

k-raina: For each condition in check, error is already set by respective CPython function: 
- If key is NULL, then c_part_to_py failed — and inside c_part_to_py, CPython's function hit a wall and wrote: "MemoryError".
 - If val is NULL, then KafkaError_new_or_None failed — same function wrote "MemoryError".
 - If PyDict_SetItem returned -1, PyDict_SetItem function itself wrote "TypeError: unhashable key" or "MemoryError".

src/confluent_kafka/src/ShareConsumer.c
k-raina: Addressed

src/confluent_kafka/src/confluent_kafka.c
ojasvajain: Understood, thanks!

ojasvajain (approved):
Looks good. Assuming error mapping and public interface changes (for eg -1 as timeout) will be revisited in a separate PR.  

Also while implementing ack callback, please ensure to write test cases for callback getting triggered in commit sync, commit async so that callstate changes get tested.

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+    try:
+        sc.subscribe([topic])
+
+        producer = kafka_cluster.cimpl_producer()
+        produced = [f'msg-{i}'.encode() for i in range(num_messages)]
+        for value in produced:
+            producer.produce(topic, value=value)
+        producer.flush(timeout=10.0)
+
+        msgs = drain_share_consumers([sc], num_messages, ack_type=AcknowledgeType.ACCEPT)[0]
```
PratRanj07: I dont think we should use drain_share_consumers specially in explicit mode, we should call poll and for every batch that we get we should accept them and call commit for every batch. See the tests in 176 in librdkafka

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+        received_values = {m.value() for m in msgs}
```
PratRanj07: Better naming than m, maybe msg

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+    try:
+        sc.subscribe([topic])
+
+        producer = kafka_cluster.cimpl_producer()
+        produced = [f'msg-{i}'.encode() for i in range(num_messages)]
+        for value in produced:
+            producer.produce(topic, value=value)
+        producer.flush(timeout=10.0)
+
+        msgs = drain_share_consumers([sc], num_messages, ack_type=AcknowledgeType.ACCEPT)[0]
```
PratRanj07: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2241/changes#r3284611946)

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+        received_values = {m.value() for m in msgs}
```
PratRanj07: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2241/changes#r3284613996)

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+
+        msgs = drain_share_consumers([sc], num_messages, timeout_s=60.0)[0]
+        assert len(msgs) >= num_messages // 2, f'only drained {len(msgs)}/{num_messages}'
+
+        result = sc.commit_sync(timeout=30.0)
+        assert isinstance(result, dict)
+        for tp, err in result.items():
+            assert err is None, f'{tp.topic}[{tp.partition}] -> {err}'
+    finally:
+        sc.close()
```
PratRanj07, line 206: commit_sync here takes care of the last batch only here

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+        for tp, err in result.items():
+            assert err is None
+
+        deadline = time.time() + 15.0
+        seen_again = False
+        while not seen_again and time.time() < deadline:
+            for m in sc.poll(timeout=2.0):
+                if m.error() is None:
+                    if (m.topic(), m.partition(), m.offset()) == released_coords:
+                        seen_again = True
```
PratRanj07: Use delivery count here

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+
+        types = [
+            AcknowledgeType.ACCEPT,
+            AcknowledgeType.RELEASE,
+            AcknowledgeType.REJECT,
+        ]
+        gathered = 0
+        deadline = time.time() + 10.0
+        while gathered < num_messages and time.time() < deadline:
+            for m in sc.poll(timeout=2.0):
```
PratRanj07: Better naming than m. Pretty confusing

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1562 @@
...
+                        seen_again = True
+                    sc.acknowledge(m, AcknowledgeType.ACCEPT)
+                    if seen_again:
+                        break
+        assert seen_again, f'released record {released_coords} never came back'
+    finally:
+        sc.close()
+
+
+def test_commit_with_mixed_ack_types(kafka_cluster):
```
PratRanj07, line 287: We should also check we are recieving the Released messages again

PratRanj07 (commented):
Hey
I have provided some comments mostly regarding the use of drain_share_consumers. My thinking is that basically for where we are testing the commit, we shouldn't use this function and call poll in loop and acknowledge and then commit to check the results. 
The naming can also be improved
It applies to all the test scenarios.
If you want we can discuss this

tests/integration/share_consumer/test_share_consumer_commit.py
```diff
@@ -0,0 +1,1690 @@
...
+    try:
+        sc1.subscribe([topic])
+
+        producer = kafka_cluster.cimpl_producer()
+        produced = [f'msg-{i}'.encode() for i in range(num_messages)]
+        for value in produced:
+            producer.produce(topic, value=value)
+        producer.flush(timeout=10.0)
+
+        msgs = drain_share_consumers([sc1], num_messages, ack_type=AcknowledgeType.ACCEPT)[0]
```
PratRanj07: I think we can use poll_ack_commit_loop here too

PratRanj07 (commented):
One small comment

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -622,11 +628,11 @@ static PyObject *Message_reduce(Message *self, PyObject *Py_UNUSED(ignored)) {
                 latency_obj = PyFloat_FromDouble(-1.0);
         }
         result = Py_BuildValue(
-            "O(NiLNNNNOOi)", Message_type, Message_topic(self, NULL),
+            "O(NiLNNNNOOih)", Message_type, Message_topic(self, NULL),
```
pranavrth, line 631: Write a test for this.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -794,15 +810,17 @@ static int Message_init(PyObject *self0, PyObject *args, PyObject *kwargs) {
...
+                              "leader_epoch", "delivery_count", NULL};
 
-        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|OiLOOOOOdi", kws,
-                                         &topic, &partition, &offset, &key,
-                                         &value, &headers, &error, &timestamp,
-                                         &latency, &leader_epoch)) {
+        if (!PyArg_ParseTupleAndKeywords(
+                args, kwargs, "|OiLOOOOOdih", kws, &topic, &partition, &offset,
+                &key, &value, &headers, &error, &timestamp, &latency,
+                &leader_epoch, &delivery_count)) {
```
pranavrth, line 823: Change this to short for delivery_count.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -700,6 +706,16 @@ static PyMethodDef Message_methods[] = {
...
      "\n"},
+    {"delivery_count", (PyCFunction)Message_delivery_count, METH_NOARGS,
+     "Retrieve the share-consumer delivery count for this message.\n"
+     "The first time a record is acquired by a share group, delivery_count "
+     "is 1; it increments each time the record is re-acquired after a "
+     "RELEASE or acquisition-lock timeout.\n"
+     "\n"
+     "  :returns: delivery count (>= 1) for share-consumer messages, "
+     "or None if not available (e.g. standard consumer or producer).\n"
+     "  :rtype: int or None\n"
```
pranavrth, line 717: WE need to return optional of short type.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1084,6 +1103,11 @@ PyObject *Message_new0(const Handle *handle, const rd_kafka_message_t *rkm) {
         else
                 self->latency = -1;
 
+        /* Share consumer: 0 from librdkafka means unavailable; store as -1
+         * so the Python accessor returns None for non-share consumers. */
+        int16_t dc           = rd_kafka_message_delivery_count(rkm);
```
pranavrth:
```suggestion
        int16_t delivery_count           = rd_kafka_message_delivery_count(rkm);
```

src/confluent_kafka/cimpl.pyi
```diff
@@ -579,6 +584,8 @@ class ShareConsumer:
     def acknowledge_offset(
         self, topic: str, partition: int, offset: int, ack_type: AcknowledgeType = ...
     ) -> None: ...
+    def commit_sync(self, timeout: float = 60) -> Dict[TopicPartition, Optional[KafkaError]]: ...
```
pranavrth, line 591: We will need to create a `TopicIdPartition` class to keep it in parity with Java. Add a TODO and we should implement that once interfaces are finalized.

pranavrth, line 591: Understand if we need KafkaError or KafkaException here.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,98 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+ * Implicit mode auto-converts the records returned by the latest poll() to
+ * ACCEPT inside librdkafka before sending. Blocks until all broker replies
+ * arrive or the timeout expires.
+ *
+ * @returns dict mapping TopicPartition -> None on success or KafkaError on
+ *          per-partition failure. Empty dict when no acknowledgements are
+ *          pending.
+ */
+static PyObject *ShareConsumer_commit_sync(ShareConsumerHandle *self,
+                                           PyObject *args,
```
pranavrth, line 509: Can we use positional argument for timeout instead of kwargs? Though the pattern is same everywhere in librdkakfa, its worth thinking about.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,98 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+
+        CallState_begin(&self->base, &cs);
+        error = rd_kafka_share_commit_sync(self->rkshare, cfl_timeout_ms(tmout),
+                                           &c_parts);
+        if (!CallState_end(&self->base, &cs)) {
+                if (c_parts)
+                        rd_kafka_topic_partition_list_destroy(c_parts);
+                if (error)
+                        rd_kafka_error_destroy(error);
+                return NULL;
```
pranavrth: Use goto for this.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1570,6 +1594,39 @@ PyObject *c_parts_to_py(const rd_kafka_topic_partition_list_t *c_parts) {
...
+        result = PyDict_New();
+        if (!result)
+                return NULL;
+
+        for (i = 0; i < (size_t)c_parts->cnt; i++) {
+                const rd_kafka_topic_partition_t *rktpar = &c_parts->elems[i];
+                PyObject *key                            = c_part_to_py(rktpar);
+                PyObject *val = KafkaError_new_or_None(rktpar->err, NULL);
+
+                if (!key || !val || PyDict_SetItem(result, key, val) == -1) {
```
pranavrth: Goto

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -458,6 +458,8 @@ rd_kafka_conf_t *common_conf_setup(rd_kafka_type_t ktype,
                                    PyObject *kwargs);
 PyObject *c_part_to_py(const rd_kafka_topic_partition_t *c_part);
 PyObject *c_parts_to_py(const rd_kafka_topic_partition_list_t *c_parts);
+PyObject *
+c_parts_to_commit_result_dict(const rd_kafka_topic_partition_list_t *c_parts);
```
pranavrth: Change name to something like - `c_parts_to_dict_topic_partition_to_error`

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -674,6 +766,45 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+     "\n"
+     "  :param float timeout: Maximum time to block (seconds). Default: 60.\n"
+     "                        Pass -1 for infinite.\n"
+     "  :returns: Dict mapping TopicPartition to None on success or "
+     "KafkaError\n"
+     "           on per-partition failure. Empty dict when no\n"
+     "           acknowledgements are pending.\n"
+     "  :rtype: dict(TopicPartition, KafkaError | None)\n"
+     "  :raises KafkaException: on error\n"
+     "  :raises RuntimeError: if called on a closed share consumer\n"
```
pranavrth, line 797: TypeError as well.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -674,6 +766,45 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+    {"commit_async", (PyCFunction)ShareConsumer_commit_async, METH_NOARGS,
+     ".. py:function:: commit_async()\n"
+     "\n"
+     "  Asynchronously commit pending acknowledgements. Returns immediately;\n"
+     "  broker results are delivered via the configured\n"
+     "  ``share_acknowledgement_commit_cb`` callback.\n"
+     "\n"
+     "  :returns: None\n"
+     "  :raises KafkaException: on error\n"
+     "  :raises RuntimeError: if called on a closed share consumer\n"
```
pranavrth, line 810: Same

pranavrth (changes requested):
Overall implementation looks good. Few comments on implementation part.

src/confluent_kafka/src/ShareConsumer.c
k-raina: keeping it as kwargs for parity as Consumer.poll, Producer.poll, Producer.flush all take timeout as METH_VARARGS|METH_KEYWORDS with kws = {"timeout", NULL}

src/confluent_kafka/src/confluent_kafka.c
k-raina: Python doesn't have a short type

src/confluent_kafka/cimpl.pyi
k-raina: Internally, KafkaException is wrapper around KafkaError. KafkaException exists to carry a KafkaError when raising. 

If we use KafkaException, user would have to write result[tp].args[0].code() instead of result[tp].code(). 

Using KafkaError is consistant with rest of codebase, but differs from Java Interface. 

```
class Message:
      def __init__(self, ..., error: Optional[KafkaError] = ..., ...) -> None: ...
      def error(self) -> Optional[KafkaError]: ...
```

src/confluent_kafka/src/confluent_kafka.c
k-raina: @pranavrth 

From a Python application's point of view, Message.delivery_count() returns an int and the constructor accepts an int. Python ints are unbounded, so callers reasonably assume the value can be anything an integer could be.

Under the hood,  it's a 16-bit short returned by librdkafka.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1570,6 +1594,46 @@ PyObject *c_parts_to_py(const rd_kafka_topic_partition_list_t *c_parts) {
...
+                key = NULL;
+                val = NULL;
+        }
+
+        return result;
+
+err:
+        Py_XDECREF(key);
+        Py_XDECREF(val);
+        Py_DECREF(result);
```
pranavrth:
```suggestion
        Py_XDECREF(result);
```

result can still be NULL.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1570,6 +1594,46 @@ PyObject *c_parts_to_py(const rd_kafka_topic_partition_list_t *c_parts) {
...
+PyObject *c_parts_to_dict_topic_partition_to_error(
+    const rd_kafka_topic_partition_list_t *c_parts) {
+        PyObject *result = NULL;
+        PyObject *key    = NULL;
+        PyObject *val    = NULL;
+        size_t i;
+
+        result = PyDict_New();
+        if (!result)
+                return NULL;
```
pranavrth: goto err;

src/confluent_kafka/src/ShareConsumer.c
pranavrth: Let's add a TODO to check this part.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,99 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+ *
+ * @returns dict mapping TopicPartition -> None on success or KafkaError on
+ *          per-partition failure. Empty dict when no acknowledgements are
+ *          pending.
+ */
+static PyObject *ShareConsumer_commit_sync(ShareConsumerHandle *self,
+                                           PyObject *args,
+                                           PyObject *kwargs) {
+        /* default timeout */
+        double tmout                             = 60.0;
```
pranavrth, line 513: Add a TODO to confirm.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,99 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+                goto err;
+
+        if (error) {
+                cfl_PyErr_from_error_destroy(error);
+                error = NULL;
+                goto err;
+        }
+
+        if (!c_parts)
+                return PyDict_New();
```
pranavrth, line 544: Ideally this shouldn't happen but let's add a TODO to verify this part. We might need to improve something on librdkafka part.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,99 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+        rd_kafka_error_t *error                  = NULL;
+        rd_kafka_topic_partition_list_t *c_parts = NULL;
+        PyObject *result                         = NULL;
+        CallState cs;
+        static char *kws[] = {"timeout", NULL};
+
+        if (!self->rkshare) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                ERR_MSG_SHARE_CONSUMER_CLOSED);
+                return NULL;
```
pranavrth: goto err;

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -494,6 +494,99 @@ static PyObject *ShareConsumer_acknowledge_offset(ShareConsumerHandle *self,
...
+        }
+
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|d", kws, &tmout))
+                return NULL;
```
pranavrth: goto err

pranavrth (changes requested):
Overall looks fine from the Implementation's perspective. Few minor comments.

pranavrth (approved):
LGTM! from implementation's perspective.

Let's merge this once @PratRanj07 and @ojasvajain are convinced from implementation and testing perspective as well. Let me know if you are changing anything from the implementation perspective.

Thanks @k-raina !.

PratRanj07 (approved):
LGTM! Thanks @k-raina !

PR #2252, opened 2026-05-24
[KIP-932] Support callbacks for share consumer python binding

k-raina:
**Summary** 

 1. Enables `error_cb, throttle_cb, log_cb, and oauth_cb` on ShareConsumer, which dispatch through the share-consumer drain paths i.e `poll, commit_sync, commit_async`. 
  2. Temporarily Rejects stats_cb, statistics.interval.ms, and on_commit at construction time with ValueError. Pre-filter in ShareConsumer_init scans `args[0] + kwargs` before `common_conf_setup`.
  3. Wires the share-consumer log queue onto `rk_rep` via `rd_kafka_share_set_log_queue(rkshare, NULL)` so log records reach the Python logger through the same drain as poll/commit. 
  4. OAuth init contract: `rd_kafka_share_sasl_background_callbacks_enable` + `wait_for_oauth_token_set` so construction returns a ready-to-use client. `wait_for_oauth_token_set` refactored to no longer own destroy on  failure, and each _init now invokes the correct teardown API i.e `rd_kafka_destroy vs rd_kafka_share_destroy`
  5. All callback trampolines now use the librdkafka-supplied rk arg in `rd_kafka_yield` / `rd_kafka_oauthbearer_set_token` calls instead of `h->rk` . Also `error_cb`  additionally tolerates dispatch with no CallState and saves/restores any pending Python exception across the user callback so the constructor's original error isn't silently swallowed.
  6. New unit suite `tests/test_ShareConsumer_callbacks.py` covers dispatch, exception propagation, config rejection, and the CallState wrap on commit paths against an unreachable broker. Integration: throttle_cb under `tests/integration/share_consumer/` and full OAUTHBEARER handshake/refresh/auth-failure under `tests/integration/share_consumer_oauth/ `


src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2060,17 +2058,52 @@ static void
 error_cb(rd_kafka_t *rk, int err, const char *reason, void *opaque) {
```
pranavrth, line 2060: Create a separate path for share consumer internally if we want the fix to be available in share consumer.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2179,7 +2226,8 @@ static int stats_cb(rd_kafka_t *rk, char *json, size_t json_len, void *opaque) {
         else {
                 CallState_fetch_exception(cs);
                 CallState_crash(cs);
-                rd_kafka_yield(h->rk);
+                /* Use callback's rk — h->rk is NULL on share consumer. */
+                rd_kafka_yield(rk);
```
pranavrth, line 2190: Remove this as we are not planning to use stats_cb for share consumer.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2399,8 +2444,9 @@ oauth_cb(rd_kafka_t *rk, const char *oauthbearer_config, void *opaque) {
         goto done;
 
 fail:
+        /* h->rk is NULL on share consumer; use the callback's rk. */
         err_code = rd_kafka_oauthbearer_set_token_failure(
-            h->rk, "OAuth callback raised exception");
+            rk, "OAuth callback raised exception");
```
pranavrth, line 2414: Create mixed callback and mixed client type tests.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2060,17 +2058,52 @@ static void
...
+        PyGILState_STATE gstate;
+        int have_cs;
+
+        /* This callback can fire on threads that didn't run CallState_begin:
+         * librdkafka's background dispatch, and the drain inside
+         * rd_kafka_share_destroy() when a constructor is unwinding.
+         * CallState_get() returns NULL in those cases — the original code
+         * then deref'd it and segfaulted in release builds. Look it up
+         * directly and fall back to PyGILState_Ensure() if absent. */
+#ifdef WITH_PY_TSS
```
pranavrth: Revert or create this change in a new PR.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -834,6 +794,45 @@ static PyMethodDef ShareConsumer_methods[] = {
...
+ *        kwargs; common_conf_setup merges them later with kwargs winning,
+ *        so checking presence in either source matches the eventual
+ *        resolved view. Keeps confluent_kafka.c ignorant of share
+ *        consumers — no flag on Handle, no share-specific branches in
+ *        the shared config setup path.
+ *
+ * @returns 0 if no rejected keys are present, -1 on rejection (a Python
+ *          ValueError is set).
+ */
+static int ShareConsumer_reject_incompatible_config(PyObject *args,
```
pranavrth, line 828: Add a TODO that we would depend on librdkafka reject these. We need to remove these from here and when librdkafka has implemented this part.

pranavrth (changes requested):

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2410,7 +2414,7 @@ oauth_cb(rd_kafka_t *rk, const char *oauthbearer_config, void *opaque) {
         goto done;
 err:
         PyGILState_Release(gstate);
-        rd_kafka_yield(h->rk);
+        rd_kafka_yield(rk);
 done:
         PyGILState_Release(gstate);
```
pranavrth, line 2427: Why release is present twice? Is this expected?

src/confluent_kafka/src/Admin.c
```diff
@@ -5537,9 +5537,16 @@ static int Admin_init(PyObject *selfobj, PyObject *args, PyObject *kwargs) {
...
+         * destroy on failure — wait_for_oauth_token_set no longer touches
+         * self->rk (see refactor note in confluent_kafka.c). */
+        if (self->oauth_cb && wait_for_oauth_token_set(self) == -1) {
+                CallState cs;
+                CallState_begin(self, &cs);
+                rd_kafka_destroy(self->rk);
+                CallState_end(self, &cs);
+                self->rk = NULL;
+                return -1;
         }
```
pranavrth, line 5553: Similar change at other places as well.

```suggestion
        /* Wait for the background thread to set the token. Caller owns
         * destroy on failure — wait_for_oauth_token_set no longer touches
         * self->rk (see refactor note in confluent_kafka.c). */
         * 
        ret_wait_oauth = wait_for_oauth_token_set(self);
        if (ret_wait_oauth == -1) {
                CallState cs;
                CallState_begin(self, &cs);
                rd_kafka_destroy(self->rk);
                CallState_end(self, &cs);
                self->rk = NULL;
        }
        
        if(self->oauth_cb)
            return ret_wait_oauth;
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -887,14 +880,56 @@ ShareConsumer_init(PyObject *selfobj, PyObject *args, PyObject *kwargs) {
...
+        if (self->base.logger) {
+                rd_kafka_resp_err_t err =
+                    rd_kafka_share_set_log_queue(self->rkshare, NULL);
+                if (err) {
+                        cfl_PyErr_Format(err,
+                                         "Failed to set share log queue: %s",
+                                         rd_kafka_err2str(err));
+                        CallState cs;
+                        CallState_begin(&self->base, &cs);
+                        rd_kafka_share_destroy(self->rkshare);
```
pranavrth: We have introduced an atomic lock to make librdkafka API calls sequential. We might not be able to call core share APIs from another share API or callback. Check if we are not using destroy or any other share API in the callbacks.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2379,7 +2382,7 @@ oauth_cb(rd_kafka_t *rk, const char *oauthbearer_config, void *opaque) {
         }
 
         err_code = rd_kafka_oauthbearer_set_token(
```
pranavrth, line 2391: We should expose share equivalent of this API. Can you verify what more are needed on the oauthbearer side in librdkafka as well?

Its fine for now in python client side (as we have rk here) but we might need in librdkafka to make SASL work properly with Share Consumer.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2399,8 +2402,9 @@ oauth_cb(rd_kafka_t *rk, const char *oauthbearer_config, void *opaque) {
         goto done;
 
 fail:
+        /* h->rk is NULL on share consumer; use the callback's rk. */
         err_code = rd_kafka_oauthbearer_set_token_failure(
```
pranavrth, line 2413: We should expose share equivalent of this API. And even for `rd_kafka_fatal_error()` APIs.

Its fine for now in python client side but we might need in librdkafka to make SASL work properly with Share Consumer.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2410,7 +2414,7 @@ oauth_cb(rd_kafka_t *rk, const char *oauthbearer_config, void *opaque) {
         goto done;
 err:
         PyGILState_Release(gstate);
-        rd_kafka_yield(h->rk);
+        rd_kafka_yield(rk);
```
pranavrth, line 2424: The working of `rd_kafka_yield` is not tested with share consumer. It might or might not work in expected way. I think I will ask the team to prioritize wakeup for share consumer on the librdkafka side for this or we need to remove this yield handling here.

I can think of a problem here when yield is set from any callback, only that queue is yielded and not any other queue. If some API uses multiple queues then it won't work properly. Like consume_batch call uses q_serve 3 times.

pranavrth (changes requested):

src/confluent_kafka/src/ShareConsumer.c
pranavrth: share_destroy can return error now. We need to destroy the returned error. Check other places as well including older places where this is being used like in  `ShareConsumer_init`, `ShareConsumer_dealloc`, `ShareConsumer_close` etc.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2214,7 +2220,7 @@ log_cb(const rd_kafka_t *rk, int level, const char *fac, const char *buf) {
         else {
                 CallState_fetch_exception(cs);
                 CallState_crash(cs);
-                rd_kafka_yield(h->rk);
+                rd_kafka_yield((rd_kafka_t *)rk);
```
pranavrth, line 2227: Let's discuss and finalize if we need to keep the yield in callbacks for now or not?

pranavrth (commented):
Let's discuss the path forward with yield.

src/confluent_kafka/src/confluent_kafka.c
k-raina: Thanks @pranavrth 
**Update thread:** As discussed on call, we will keep `rd_kafka_yield` for preview. In GA, once wakeup is completed, we will use that.

src/confluent_kafka/src/confluent_kafka.c
k-raina: Thanks @pranavrth , as discussed adding new apis to librdkafka will be handled seperately

src/confluent_kafka/src/confluent_kafka.c
k-raina: [Same](https://github.com/confluentinc/confluent-kafka-python/pull/2252#discussion_r3369904542)

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -887,14 +880,56 @@ ShareConsumer_init(PyObject *selfobj, PyObject *args, PyObject *kwargs) {
...
-        /* TODO KIP-932: call rd_kafka_sasl_background_callbacks_enable() for
-         * OAuth once librdkafka adds a share-level wrapper for the same reason.
-         */
+        /* OAuth wiring: enable SASL background callbacks so the
+         * oauthbearer_token_refresh_cb fires on librdkafka's own thread
+         * during init (otherwise we'd have the chicken-and-egg problem
+         * — can't poll until connected, can't connect without a token).
+         * Then block until the initial token is set. Mirrors
+         * Consumer.c:1812-1833. */
+        if (self->base.oauth_cb) {
```
pranavrth, line 927: **`Handle_clear` / `Handle_traverse` are missing entries — this PR makes them reachable for ShareConsumer.**

`oauth_cb` is INCREF'd in `common_conf_setup` (`confluent_kafka.c:2962`) but never DECREF'd. `Handle_clear` covers `error_cb` / `throttle_cb` / `stats_cb` / `logger`; not `oauth_cb`. `Handle_traverse` covers `error_cb` / `throttle_cb` / `stats_cb`; not `oauth_cb` or `logger`.

Failure modes:
- **Bound-method `oauth_cb`** (e.g. `oauth_cb=self.refresh`) — full cycle, invisible to GC because traverse skips `oauth_cb`. Handle leaks forever.
- **Closure `oauth_cb`** capturing locals — clear never DECREFs, refs leak per Handle constructed.
- Free function — `+1` ref, cosmetic.

Pre-existing on Producer/Consumer/Admin, but this PR enables OAuth on ShareConsumer so the cycle mode becomes reachable here too.

```c
// Handle_clear
if (h->oauth_cb) { Py_DECREF(h->oauth_cb); h->oauth_cb = NULL; }

// Handle_traverse — oauth_cb AND logger are both missing
if (h->oauth_cb) Py_VISIT(h->oauth_cb);
if (h->logger)   Py_VISIT(h->logger);
```

Verify these changes and fix if required.

pranavrth (changes requested):

src/confluent_kafka/src/ShareConsumer.c
k-raina: I checked all the callbacks, and none of them call a share API,  so we're never calling into a share API from inside a callback. Every share API call happens in the regular ShareConsumer methods (init, subscribe, consume, commit, close, dealloc), which all run on the application's own thread. The callbacks only call core librdkafka APIs i.e `rd_kafka_yield`


We will need to check separately for the auth callback.  It calls `rd_kafka_oauthbearer_set_token`, set_token_failure, and `rd_kafka_yield`, all on the underlying rk.

pranavrth:
Build is failing. Please fix it.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -2214,7 +2220,7 @@ log_cb(const rd_kafka_t *rk, int level, const char *fac, const char *buf) {
         else {
                 CallState_fetch_exception(cs);
                 CallState_crash(cs);
-                rd_kafka_yield(h->rk);
+                rd_kafka_yield(rk);
```
pranavrth: This is giving warning during compile time.

```suggestion
                rd_kafka_yield((rk_kafka_t *) rk);
```

src/confluent_kafka/src/Consumer.c
```diff
@@ -1825,11 +1825,21 @@ static int Consumer_init(PyObject *selfobj, PyObject *args, PyObject *kwargs) {
...
+        if (ret_wait_oauth == -1) {
+                CallState cs;
+                CallState_begin(self, &cs);
+                rd_kafka_destroy(self->rk);
+                CallState_end(self, &cs);
+                self->rk = NULL;
         }
 
+        if (self->oauth_cb)
+                return ret_wait_oauth;
```
pranavrth: I didn't see the if (self->oauth_cb) { condition removed. I think we should follow below way in all older client types. ShareConsumer is already fine.


```suggestion
if (self->oauth_cb) {
    int ret_wait_oauth = wait_for_oauth_token_set(self);
    if (ret_wait_oauth == -1) {
        CallState cs;
        CallState_begin(self, &cs);
        rd_kafka_destroy(self->rk);
        CallState_end(self, &cs);
        self->rk = NULL;
    }
    return ret_wait_oauth;
}
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -603,14 +566,23 @@ static PyObject *ShareConsumer_close(ShareConsumerHandle *self,
 
         CallState_begin(&self->base, &cs);
         error = rd_kafka_share_consumer_close(self->rkshare);
-        rd_kafka_share_destroy(self->rkshare);
-        self->rkshare = NULL;
+        rd_kafka_error_t *destroy_error = rd_kafka_share_destroy(self->rkshare);
```
pranavrth: declaration at the top.

src/confluent_kafka/src/confluent_kafka.c
pranavrth, line 2302:
```c
/**
 * @brief Waits for OAuth callback to set the token during client init.
 *
 * @returns 0 if the token was set within the timeout (or if oauth_cb is
 *          not configured); -1 on timeout, with a Python exception set.
 *
 * @remark On failure (-1) the caller owns teardown. This function
 *         intentionally does NOT call rd_kafka_destroy() or
 *         rd_kafka_share_destroy() — each client type knows which
 *         destroy API to use for what it allocated.
 */
int wait_for_oauth_token_set(Handle *h) {
```

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -1595,10 +1595,8 @@ PyObject *c_parts_to_py(const rd_kafka_topic_partition_list_t *c_parts) {
 }
 
 /**
- * @brief Convert a per-partition commit result list into a Python
- * dict(TopicPartition -> KafkaError | None).
- *
- * @returns The new Python dict, or NULL on allocation failure.
```
pranavrth, line 1601: Why removed?

src/confluent_kafka/src/Producer.c
```diff
@@ -1468,11 +1468,21 @@ static int Producer_init(PyObject *selfobj, PyObject *args, PyObject *kwargs) {
         if (self->logger)
                 rd_kafka_set_log_queue(self->rk, NULL);
 
-        /* Wait for the background thread to set the token */
-        if (self->oauth_cb) {
-                return wait_for_oauth_token_set(self);
+        /* Wait for the background thread to set the token. Caller owns
+         * destroy on failure — wait_for_oauth_token_set no longer touches
+         * self->rk (see refactor note in confluent_kafka.c). */
+        int ret_wait_oauth = wait_for_oauth_token_set(self);
```
pranavrth: same

pranavrth (changes requested):

k-raina:
Seems like new changes are merged to librdkafka. I have Rebased librdkafka PR https://github.com/confluentinc/librdkafka/pull/5462/changes, this should pass the build.

.semaphore/semaphore.yml
```diff
@@ -11,7 +11,7 @@ global_job_config:
       value: v2.14.2
     # TODO KIP-932: Remove LIBRDKAFKA_BRANCH once LIBRDKAFKA_VERSION includes share consumer support
     - name: LIBRDKAFKA_BRANCH
-      value: dev_kip-932_queues-for-kafka
+      value: dev_kip-932_queues-for-kafka_additional_api_python
```
pranavrth, line 14: Change branch name.

pranavrth (approved):
LGTM!. Thanks @k-raina !

Issue #517, opened 2019-01-07
"json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)" when running example avro producer 


ADDale:
Description
===========
Getting below exception while running simple Avro Producer in python. Please suggest if I need to upgrade to any other version?
Traceback (most recent call last):
  File "C:\Users\daleamr\Desktop\python_kafka_avro_test\producer.py", line 39, in <module>
    main() 
  File "C:\Users\daleamr\Desktop\python_kafka_avro_test\producer.py", line 34, in main
    produce() 
  File "C:\Users\daleamr\Desktop\python_kafka_avro_test\producer.py", line 11, in produce
    avroProducer.produce(topic='avro_test_3', value=value)
  File "C:\Python3\lib\site-packages\confluent_kafka\avro\__init__.py", line 80, in produce
    value = self._serializer.encode_record_with_schema(topic, value_schema, value)
  File "C:\Python3\lib\site-packages\confluent_kafka\avro\serializer\message_serializer.py", line 98, in encode_record_with_schema
    schema_id = self.registry_client.register(subject, schema)
  File "C:\Python3\lib\site-packages\confluent_kafka\avro\cached_schema_registry_client.py", line 141, in register
    result, code = self._send_request(url, method='POST', body=body)
  File "C:\Python3\lib\site-packages\confluent_kafka\avro\cached_schema_registry_client.py", line 97, in _send_request
    return response.json(), response.status_code
  File "C:\Python3\lib\site-packages\requests\models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "C:\Python3\lib\json\__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "C:\Python3\lib\json\decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\Python3\lib\json\decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

How to reproduce
================
I am running below simple program and getting an exception
from confluent_kafka import avro 
from confluent_kafka.avro import AvroProducer

def produce():
    value_schema = avro.load('user.avsc')
    key_schema = avro.load('key.avsc')
    value ={ "name": u'abcd'
        }
    #key = {"name": "Key"}
[...]

PR #2258, opened 2026-05-29
Fix tag-release CI: Linux wheel verify + Windows wheel build

Devarsh010:
Linux x64/arm64 "Wheel Verification" failed because tools/test-manylinux.sh downloaded bootstrap.pypa.io/get-pip.py and ran it against Python 3.9. Upstream pypa dropped Py3.9 support from the unversioned endpoint; the error itself recommends the versioned URL https://bootstrap.pypa.io/pip/3.9/get-pip.py.

Windows "Wheels" + "Wheel Verification" failed because sem-version python 3.11.9 reported "No GitHub artifact attestations found". mise (used by Semaphore's Windows AMI) requires GitHub artifact attestations on python-build-standalone artifacts. Attestation support was added to python-build-standalone in PR #517 (2025-02-11); Py3.11.9 was packaged before that. Bump to 3.11.15, which is shipped by every recent attested release.

CI validation : https://semaphore.ci.confluent.io/workflows/fa36311e-c7e7-468c-9318-9d9f0ed23c0a?pipeline_id=63c64c70-738c-441a-ba1e-812770e01f77

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -179,7 +179,7 @@ blocks:
             - name: MSYSTEM
               value: UCRT64
           commands:
-            - sem-version python 3.11.15
+            - sem-version python 3.11
             - bash tools/mingw-w64/semaphore_commands.sh
             - bash tools/wheels/install-librdkafka.sh $env:LIBRDKAFKA_VERSION.TrimStart("v") dest
             - tools/wheels/build-wheels.bat x64 win_amd64 dest wheelhouse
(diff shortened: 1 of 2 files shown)


.semaphore/semaphore.yml
```diff
@@ -179,7 +179,7 @@ blocks:
             - name: MSYSTEM
               value: UCRT64
           commands:
-            - sem-version python 3.11.9
+            - sem-version python 3.11.15
```
emasab: Can we set `3.11` here? To download the latest patch release.

.semaphore/semaphore.yml
```diff
@@ -509,7 +509,7 @@ blocks:
       jobs:
         - name: Verify
           commands:
-            - sem-version python 3.11.9
+            - sem-version python 3.11.15
```
emasab: and here

emasab (approved):

PR #2264, opened 2026-06-02
Support protobuf 7 by replacing removed FieldDescriptor.label

emasab:
protobuf 7 removed the deprecated FieldDescriptor.label property in favor of the is_repeated boolean property, which broke the protobuf serde and CEL rule code that compared label against LABEL_REPEATED.

Add a PROTOBUF_MAJOR_VERSION-gated _is_repeated() helper in both common/protobuf.py and rules/cel/constraints.py: use fd.is_repeated on protobuf >= 7 and fall back to fd.label == LABEL_REPEATED on older runtimes, so the library keeps working across protobuf 4.x through 7.x.

What
----

Checklist
------------------
  - No
  - Tried the existing tests

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/confluent_kafka/schema_registry/common/protobuf.py
+++ b/src/confluent_kafka/schema_registry/common/protobuf.py
@@ -5,6 +5,7 @@ from collections import deque
 from decimal import MAX_PREC, Context, Decimal
 from typing import Any, Deque, List, Set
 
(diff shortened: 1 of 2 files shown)


rayokota (approved):
Thanks @emasab , LGTM

PR #2265, opened 2026-06-04
[KIP-932] Add deserializing share consumer

k-raina:
**Summary**

- Adds DeserializingShareConsumer, a high-level KIP-932 share consumer that deserializes record keys and values via the `key.deserializer` / `value.deserializer`  config (same callables as DeserializingConsumer).
- Overrides poll() to return a *list* of messages (mirroring ShareConsumer) and deserializes each one in place, so the returned messages stay valid arguments to acknowledge().
- A per-record deserialization failure no longer drops the rest of the batch: the raw bytes are kept and msg.error() is set to KEY_DESERIALIZATION / _VALUE_DESERIALIZATION, so apps can detect it with the usual `if msg.error():` check and acknowledge with REJECT.
-   Exports the class from `confluent_kafka` and adds the missing `Message.set_error` to the cimpl type stub.
- Adds unit tests , an integration test, and supporting share-consumer test fixtures


src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,158 @@
...
+        is not lost. The deserializer calls and the context/header setup they
+        use are guarded, so a failure marks only this record instead of
+        aborting the batch.
+        """
+
+        topic = msg.topic()
+        if topic is None:
+            # no topic to deserialize against; flag it rather than raise
+            msg.set_error(KafkaError(KafkaError._VALUE_DESERIALIZATION, "Message topic is None"))
+            return
```
pranavrth: This should be a TypeError. Check Deserializing Consumer and make sure we are following those conventions.

src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,158 @@
...
+
+        try:
+            key = msg.key()
+            if self._key_deserializer is not None:
+                ctx.field = MessageField.KEY
+                key = self._key_deserializer(key, ctx)
+        except Exception as se:
+            msg.set_error(KafkaError(KafkaError._KEY_DESERIALIZATION, f"{type(se).__name__}: {se}"))
+            return
```
pranavrth, line 159: Why are we not using "ValueDeserializationError" or "KeyDeserializationError" here like done in Deserializing Consumer?

tests/integration/share_consumer/test_share_consumer_deserialization.py
pranavrth, line 1: I think we should write integration tests with Avro, Protobuf and JSONSchema as well. A positive case and an error case with each is good enough to make sure that everything is fine.

src/confluent_kafka/deserializing_share_consumer.py
pranavrth, line 1: I would recommend that you check deserializing consumer and implement very similar logic here.

pranavrth (changes requested):
I have provided few comments. Please fix those and then I will review this PR again.

tests/integration/cluster_fixture.py
```diff
@@ -180,6 +180,53 @@ def share_consumer(self, conf=None):
...
+
+        Returns:
+            DeserializingShareConsumer: A new TestDeserializingShareConsumer instance
+        """
+        share_conf = self.client_conf({'group.id': str(uuid1())})
+
+        if conf is not None:
+            share_conf.update(conf)
+
+        reset = share_conf.pop('auto.offset.reset', 'earliest')
```
pranavrth: This is incorrect use of the share consumer. Let's fix this even in above share consumer scenario and extract a common private function to reuse the same code.

src/confluent_kafka/cimpl.pyi
```diff
@@ -300,6 +300,7 @@ class Message:
     def set_headers(self, headers: HeadersType) -> None: ...
     def set_key(self, key: Any) -> None: ...
     def set_value(self, value: Any) -> None: ...
+    def set_error(self, error: KafkaError) -> None: ...
```
pranavrth: Do we really want to expose this function to the user? If yes then we have to add error handling in the case when the user is sending not error in the error field.

pranavrth (changes requested):

src/confluent_kafka/deserializing_share_consumer.py
k-raina: Setting "ValueDeserializationError" or "KeyDeserializationError breaks mgs.error().code()

```
msg.error().code() 
    - On a KafkaError: .code() is a method → works.
    - On a ValueDeserializationError: .code is a property (an int) → .code() raises TypeError: 'int' object is not callable
```

k-raina:
Below are known test failures fixed in PR https://github.com/confluentinc/confluent-kafka-python/pull/2253/changes#top
```
1. test_subscribe_with_empty_list_raises
2. test_share_consumer_oauth_expired_token_surfaces_error_cb
```

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -595,10 +595,23 @@ static PyObject *Message_set_topic(Message *self, PyObject *new_topic) {
...
+                                "error must be a KafkaError or None");
+                return NULL;
+        }
+
+        Py_XDECREF(self->error);
+
+        if (new_error == Py_None) {
+                /* None clears the error; Message stores "no error" as NULL
+                 * (Message_error then returns None for it). */
+                self->error = NULL;
```
pranavrth: We never set error to NULL in python. It should be KafkaError or it should be None.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -595,10 +595,23 @@ static PyObject *Message_set_topic(Message *self, PyObject *new_topic) {
...
-        if (self->error)
-                Py_DECREF(self->error);
-        self->error = new_error;
-        Py_INCREF(self->error);
+        if (new_error != Py_None &&
+            !PyObject_TypeCheck(new_error, &KafkaErrorType)) {
+                PyErr_SetString(PyExc_TypeError,
+                                "error must be a KafkaError or None");
+                return NULL;
+        }
```
pranavrth: Add this breaking change in Changelog as well and add a TODO.

src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,160 @@
...
+        marks only this record instead of aborting the batch.
+
+        A message with no topic is a broken invariant rather than a per-record
+        data error, so it raises :py:exc:`TypeError` (matching
+        :py:class:`DeserializingConsumer`).
+        """
+
+        topic = msg.topic()
+        if topic is None:
+            raise TypeError("Message topic is None")
```
pranavrth, line 141: Add TODO to verify if this is correct or not for GA.

We have 3 options:
1) TypeError - but user didn't give the topic name so I think TypeError is not correct.
2) RuntimeError - I feel this is correct as this is internal issue not related to user.
3) KafkaError - set in the message field so that we don't throw exception.

src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,160 @@
...
+            deserialized objects.
+
+            Records that arrived with an error (``msg.error()`` is not None) are
+            returned unchanged. Records whose key or value fails to deserialize are
+            returned with their raw bytes preserved and ``msg.error()`` set to a
+            ``_KEY_DESERIALIZATION`` or ``_VALUE_DESERIALIZATION`` error.
+        """
+
+        messages = super(DeserializingShareConsumer, self).poll(timeout)
+        for msg in messages:
```
pranavrth, line 114: We need to check if we want to send error or exception in a message. We should use the pythonic way. Let's add a TODO to discuss it in the final interfaces part.

src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,160 @@
...
+        data error, so it raises :py:exc:`TypeError` (matching
+        :py:class:`DeserializingConsumer`).
+        """
+
+        topic = msg.topic()
+        if topic is None:
+            raise TypeError("Message topic is None")
+
+        try:
+            ctx = SerializationContext(topic, MessageField.VALUE, msg.headers())
```
pranavrth: Remove this to outside as it is being used below as well.

src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,160 @@
...
+        A message with no topic is a broken invariant rather than a per-record
+        data error, so it raises :py:exc:`TypeError` (matching
+        :py:class:`DeserializingConsumer`).
+        """
+
+        topic = msg.topic()
+        if topic is None:
+            raise TypeError("Message topic is None")
+
+        try:
```
pranavrth, line 144: Move try to after `if self._value_deserializer is not None:` part.

src/confluent_kafka/deserializing_share_consumer.py
```diff
@@ -0,0 +1,160 @@
...
+            return
+
+        try:
+            key = msg.key()
+            if self._key_deserializer is not None:
+                ctx.field = MessageField.KEY
+                key = self._key_deserializer(key, ctx)
+        except Exception as se:
+            msg.set_error(KafkaError(KafkaError._KEY_DESERIALIZATION, str(se)))
+            return
```
pranavrth, line 159: Same as Value part.

pranavrth (changes requested):

src/confluent_kafka/deserializing_share_consumer.py
pranavrth: This is not done.

src/confluent_kafka/deserializing_share_consumer.py
pranavrth: Moreover, ValueDeserializationError and KeyDeserializationError are really Exception and we want Error here and not Exceptions.

pranavrth (changes requested):
1 comment is missed.

pranavrth (approved):
Looks good to me. I am approving the PR but I would prefer a pass from @rayokota or somebody from the @confluentinc/data-governance team to review this as well.

rayokota (approved):
Thanks @k-raina , LGTM

PR #2271, opened 2026-06-10
validate protobuf message index bounds on deserialize

uwezkhan:
What
----
The protobuf deserializer picks the writer message type by walking the message-index array from the Confluent wire framing into the parsed `FileDescriptorProto`. `SchemaId._read_index_array` caps the array length but not the individual index values, and those values are zigzag varints, so a crafted message can carry a negative or out-of-range index. `_get_message_desc_proto` then indexes `desc.message_type` / `desc.nested_type` with the value directly.

Before: a negative index (for example uvarint `0x01`, which zigzag-decodes to `-1`) wraps around and resolves to a different message type in the same file, so the payload is decoded against the wrong descriptor. An out-of-range positive index raises a bare `IndexError`.

After: an index outside `0 <= index < len(...)` raises `SerializationError` with the index and the available count, matching the bound check the confluent-kafka-go reference (`toMessageDesc`) already applies. Valid indexes behave the same as today.

Tradeoff: a message that previously decoded against an unintended message type now fails fast instead of returning a mis-typed object. That is the point of the change, but it does convert a silent wrong-result into a surfaced error.

Checklist
------------------

References
----------
JIRA:

Test & Review
------------
Added `test_message_index_in_range` and `test_message_index_out_of_range` to `tests/schema_registry/_async/test_proto.py` (and the generated `_sync` counterpart). The out-of-range cases `[-1]`, `[2]`, `[0, -1]`, `[0, 5]` reproduce the issue on master and pass after the change. `_sync` files were regenerated with `tools/unasync.py`.

Open questions / Follow-ups
--------------------------
None.

### DIFF
--- a/src/confluent_kafka/schema_registry/_async/protobuf.py
+++ b/src/confluent_kafka/schema_registry/_async/protobuf.py
@@ -788,13 +788,23 @@ class AsyncProtobufDeserializer(AsyncBaseDeserializer):
     ) -> Tuple[str, descriptor_pb2.DescriptorProto]:
         index = msg_index[0]
         if isinstance(desc, descriptor_pb2.FileDescriptorProto):
-            msg = desc.message_type[index]
+            messages = desc.message_type
+            if index < 0 or index >= len(messages):
+                raise SerializationError(
+                    "message index {} out of range, schema has {} top-level message(s)".format(index, len(messages))
+                )
+            msg = messages[index]
             path = path + "." + msg.name if path else msg.name
             if len(msg_index) == 1:
                 return path, msg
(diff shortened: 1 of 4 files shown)


uwezkhan:
The build-test-release failure looks unrelated to this change. The diff is just the bounds check in `_get_message_desc_proto` plus the two regression tests, so it doesn't reach the integration paths that run in that job. Locally flake8, black, isort, and the new tests all pass on the changed files.

Pushed an empty commit to re-trigger the build. The new push resets the external CI gate, so it'll need another /sem-approve to run.

uwezkhan:
any update?

rayokota (approved):
Thanks @uwezkhan , LGTM

PR #2277, opened 2026-06-13
[KIP-932] - Address KIP-932 ToDos 1/N

k-raina:
**Summary**
* Added the share `set_sasl_credentials` method to the `ShareConsumer` API, enabling runtime updates of SASL username and password. This method can be called before or during consumption, and credentials will take effect on the next broker (re)authentication. 
* Modified the `ShareConsumer` deallocation logic to use `rd_kafka_share_destroy_flags`


src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -73,10 +73,8 @@ static void ShareConsumer_dealloc(ShareConsumerHandle *self) {
         if (self->rkshare) {
                 CallState cs;
                 CallState_begin(&self->base, &cs);
-                /* TODO KIP-932: Use rd_kafka_share_destroy_flags() once
-                 * available in the librdkafka public API. */
-                rd_kafka_error_t *destroy_error =
-                    rd_kafka_share_destroy(self->rkshare);
+                rd_kafka_error_t *destroy_error = rd_kafka_share_destroy_flags(
+                    self->rkshare, RD_KAFKA_DESTROY_F_NO_CONSUMER_CLOSE);
                 if (destroy_error)
```
ojasvajain: nit: If destroy fails, the rkshare will not be deallocated so it will eventually be leaked. Should we add a log indicating that destroy failed? This is a very unlikely scenario though.

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: I think we should not set self->rkshare as null when destroy fails because rkshare instance would still be alive so it will not be possible to call destroy_flags on it in dealloc if rkshare is null.

ojasvajain (commented):
Left comments

tests/integration/share_consumer/test_share_consumer.py
ojasvajain, line 1: In all the added test cases, lets use `sasl_cluster` instead of `kafka_cluster` so that test cases are more realistic.

tests/integration/share_consumer_sasl/conftest.py
ojasvajain, line 1: Please add license

src/confluent_kafka/src/ShareConsumer.c
ojasvajain: The fix seems to be in the right direction. But when I tried running the client, it gave me seg fault. 
To repro, you can add 
```
/* TEMP TEST: force a fake destroy error to exercise the
                 * error branch below. REMOVE before committing. */
                destroy_error = rd_kafka_error_new(RD_KAFKA_RESP_ERR__STATE,
                                                   "fake forced destroy error");`
```
                                                   
and see for yourself. This is the output I got:
```
(.venv) ➜  confluent-kafka-python git:(dev_kip-932_queues-for-kafka_address_todos_1_of_N) ✗ python
Python 3.10.0 (default, Nov  6 2025, 12:45:07) [Clang 17.0.0 (clang-1700.3.19.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from ShareConsumer imp
KeyboardInterrupt
>>> from confluent_kafka import ShareConsumer
>>> x = ShareConsumer({'group.id': 5})
>>> x = []
Destroying share consumer handleException ignored in: <cimpl.ShareConsumer object at 0x1043cd540>
RuntimeError: Failed to destroy share consumer; the handle has been leaked: fake forced destroy error
Destroying share consumer handle[1]    34680 segmentation fault  python
```                                        

Please see if you can also add a test case for this. Expected behaviour is what you intended - An exception should be set but it will be ignored, but there should no seg fault or program termination.

ojasvajain (approved):
LGTM, thanks @k-raina

ojasvajain (approved):

PR #2280, opened 2026-06-15
[KIP-932] Add tests for user callbacks in share consumer 2/N

k-raina:
**Summary**
  - Add 21 tests for ShareConsumer callbacks (error_cb, log_cb, oauth_cb, stats_cb, throttle_cb) across the unit and integration suites
  - Unit: error_cb dispatch + exception propagation through poll/commit_sync/commit_async/close; oauth_cb principal+extensions,  token-refresh success & failure-recovery, malformed-return; log_cb drain on
  commit/close, record format, and always-on (non-debug) delivery; stats_cb / statistics.interval.ms=0 rejection edges.
  - Integration: assert share-specific log facility tags (SHAREFETCH/SHAREACK/SHARESESSION/SHAREHEARTBEAT) and the cleared  throttle_time=0 throttle event.
  - Add a provision_client_quota fixture so throttle tests  self-provision/tear-down quotas and run unattended in CI — replacing the  manual SHARE_THROTTLE_QUOTA_CLIENT_ID env-var gating.
  - Document a real crash via 3 skipped tests: calling a ShareConsumer method from inside a generic callback SIGSEGVs (librdkafka's reentrancy guard covers  only the ack-commit cb); cross-thread correctly raises _CONFLICT — un-skip and assert _STATE once the guard is extended.

### DIFF
--- a/tests/integration/share_consumer/test_share_consumer_ack.py
+++ b/tests/integration/share_consumer/test_share_consumer_ack.py
@@ -1197,6 +1197,60 @@ def test_callback_reentrancy_guard(kafka_cluster):
         sc.close()
 
 
+def test_acknowledgement_commit_cb_reentrant_call_raises_state(kafka_cluster):
+    """poll() from inside the ack-commit callback raises _STATE, no crash.
+
+    The same reentrant poll() segfaults from error_cb / log_cb / throttle_cb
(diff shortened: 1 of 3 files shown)


PratRanj07 (approved):
LGTM! Thanks @k-raina

PR #2283, opened 2026-06-17
[KIP-932] Add tests for share consumer 3/N

k-raina:
**Summary**
- Add more tests for share consumer poll API


PratRanj07 (approved):
LGTM! Thanks @k-raina

PR #2284, opened 2026-06-17
[KIP-932] Add share consumer workflow tests 4/N

k-raina:
**Summary**
- Add tests for ack and commit workflows

### DIFF
--- a/tests/integration/share_consumer/test_share_consumer.py
+++ b/tests/integration/share_consumer/test_share_consumer.py
@@ -794,15 +794,9 @@ def test_open_transaction_stalls_share_group(kafka_cluster):
     group_id = unique_id('test-share-consumer-txn-stall')
 
     txn_producer = Producer(kafka_cluster.client_conf({'transactional.id': unique_id('txn')}))
-    try:
-        txn_producer.init_transactions(10)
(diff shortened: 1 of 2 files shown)


tests/integration/share_consumer/test_share_consumer.py
```diff
@@ -826,6 +826,177 @@ def test_open_transaction_stalls_share_group(kafka_cluster):
...
+    txn_producer = Producer(kafka_cluster.client_conf({'transactional.id': unique_id('txn')}))
+    try:
+        txn_producer.init_transactions(10)
+    except KafkaException as e:
+        pytest.skip(f'broker does not support transactions: {e}')
+
+    try:
+        set_group_config(kafka_cluster, group_id, 'share.isolation.level', 'read_committed')
+    except KafkaException as e:
+        pytest.skip(f'cannot set share.isolation.level on group: {e}')
```
PratRanj07: One thing the brokers above 4.2.0 allow this and I think all the share consumer tests should run after that only! So in that case I dont think we need these pytest.skips?

PratRanj07 (approved):
LGTM! Thanks @k-raina

PR #2287, opened 2026-06-19
[KIP-932] Add error translation in share consumer

k-raina:
**Why**
  - The KIP-932 share consumer wrapped every librdkafka error in KafkaException, blurring client misuse (illegal state, concurrent  access, bad args) and real broker errors. 
  - This PR translates each error code to the exception, which align with java share consumer,  i.e _STATE→IllegalStateException, _CONFLICT→ConcurrentModificationException, _INVALID_ARG→ValueError, everything else stays KafkaException — with the two new types added to cimpl and exported from confluent_kafka.

**Notes / caveats**
  - The new types subclass RuntimeError, not KafkaException, so except KafkaException no longer catches these for the share   consumer. args[0] is still a KafkaError in every translated case, so .code()/.str() keep working.
  - Scope is ShareConsumer only — Producer/Consumer/Admin are unchanged.
  - The ack-commit callback and commit_sync's per-partition map intentionally keep KafkaException (broker results, matching Java's  callback contract).


pranavrth:
`set_sasl_credentials` was missed by the translation — it's still raising a bare `RuntimeError` on a closed handle here, and plain `KafkaException` via `cfl_PyErr_from_error_destroy`. Please change that as well.

pranavrth:
I think we can create Pure python level exception instead of C Binding level and use it in C Binding. We should check the feasibility of that. Same we should do with `Messages` class.

k-raina:
Thanks @pranavrth for reviews!

Regarding 
> I think we can create Pure python level exception instead of C Binding level and use it in C Binding. We should check the feasibility of that. Same we should do with Messages class.

Our C module cimpl loads first and  error.py already does `from confluent_kafka.cimpl import ...`.
If we move the exception classes into error.py and have cimpl import them while it's starting up, we get a cycle: cimpl needs error.py, but error.py needs cimpl. Since  neither has finished loading, the import sees a half-built module and breaks. 

**The way around it** is to not grab the Python classes at startup, but lazily at the first time C
actually needs to raise one  and then cache the pointer so we only do it once. By then both modules are fully loaded, so there's no loop. It works fine, but **note that**: it adds a bit of C plumbing and a new edge case to get right.

pranavrth:
> Our C module cimpl loads first and error.py already does from confluent_kafka.cimpl import ....
If we move the exception classes into error.py and have cimpl import them while it's starting up, we get a cycle: cimpl needs error.py, but error.py needs cimpl. Since neither has finished loading, the import sees a half-built module and breaks.

> The way around it is to not grab the Python classes at startup, but lazily at the first time C
actually needs to raise one and then cache the pointer so we only do it once. By then both modules are fully loaded, so there's no loop. It works fine, but note that: it adds a bit of C plumbing and a new edge case to get right.

We are fine with the current approach. No need to complicate things.

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -3792,6 +3796,34 @@ static PyObject *_init_cimpl(void) {
...
+        /* Subclass RuntimeError, not KafkaException, but still carry a
+         * KafkaError as args[0]. Only the share consumer raises these today. */
+        IllegalStateException = PyErr_NewExceptionWithDoc(
+            "cimpl.IllegalStateException",
+            "Raised by ShareConsumer when an operation is attempted in an "
+            "invalid state, e.g. on a consumer that has already been "
+            "closed.\n"
+            "\n"
+            "Subclass of RuntimeError. ``exception.args[0]`` is a "
+            ":py:class:`KafkaError`.\n"
```
pranavrth: These are not subtype of `KafkaException` so why do we keep `KafkaError` in these exceptions? We can keep this if its a big change right now but we don't need to keep `KafkaError` in these exceptions.

pranavrth (changes requested):

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -116,6 +117,74 @@ ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
...
+
+        etype = ShareConsumer_translate_error(code);
+        if (etype == KafkaException) {
+                PyObject *eo = KafkaError_new0(code, "%s", buf);
+                if (!eo)
+                        return;
+                PyErr_SetObject(KafkaException, eo);
+                Py_DECREF(eo);
+        } else {
+                PyErr_SetString(etype, buf);
```
pranavrth:
```suggestion
                PyErr_SetString(etype, fmt ? buf : rd_kafka_err2str(code));
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -116,6 +117,74 @@ ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
...
+ * local errors have no retriable/fatal/abort flags worth preserving. */
+static void
+ShareConsumer_PyErr_Format(rd_kafka_resp_err_t code, const char *fmt, ...) {
+        char buf[512];
+        va_list ap;
+        PyObject *etype;
+
+        va_start(ap, fmt);
+        vsnprintf(buf, sizeof(buf), fmt, ap);
+        va_end(ap);
```
pranavrth:
```suggestion
if(rmt) {
        va_start(ap, fmt);
        vsnprintf(buf, sizeof(buf), fmt, ap);
        va_end(ap);
}
```

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -116,6 +117,74 @@ ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
...
+        }
+}
+
+/* Raise the translated exception type for a bare error code. KafkaException
+ * carries a KafkaError in args[0]; the non-KafkaException types
+ * (IllegalStateException / ConcurrentModificationException / ValueError) carry
+ * a plain message string — the Python type already conveys the code, and these
+ * local errors have no retriable/fatal/abort flags worth preserving. */
+static void
+ShareConsumer_PyErr_Format(rd_kafka_resp_err_t code, const char *fmt, ...) {
```
pranavrth, line 145: Check `KafkaError_new0`

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -116,6 +117,74 @@ ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
...
+        vsnprintf(buf, sizeof(buf), fmt, ap);
+        va_end(ap);
+
+        etype = ShareConsumer_translate_error(code);
+        if (etype == KafkaException) {
+                PyObject *eo = KafkaError_new0(code, "%s", buf);
+                if (!eo)
+                        return;
+                PyErr_SetObject(KafkaException, eo);
+                Py_DECREF(eo);
```
pranavrth: Use `cfl_PyErr_Format` instead.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -116,6 +117,74 @@ ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
...
+ * flags); the non-KafkaException types keep only librdkafka's error string as
+ * the message (the type conveys the code; these errors carry no flags). */
+static void ShareConsumer_PyErr_from_error_destroy(rd_kafka_error_t *error) {
+        rd_kafka_resp_err_t code = rd_kafka_error_code(error);
+        PyObject *etype          = ShareConsumer_translate_error(code);
+        if (etype == KafkaException) {
+                PyObject *eo = KafkaError_new_from_error_destroy(error);
+                if (!eo)
+                        return;
+                PyErr_SetObject(KafkaException, eo);
```
pranavrth, line 181: Use `cfl_PyErr_from_error_destroy` instead.

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -129,8 +198,8 @@ static PyObject *ShareConsumer_subscribe(ShareConsumerHandle *self,
         Py_ssize_t i;
 
         if (!self->rkshare) {
-                PyErr_SetString(PyExc_RuntimeError,
-                                ERR_MSG_SHARE_CONSUMER_CLOSED);
+                ShareConsumer_PyErr_Format(RD_KAFKA_RESP_ERR__STATE, "%s",
+                                           ERR_MSG_SHARE_CONSUMER_CLOSED);
```
pranavrth:
```suggestion
                PyErr_SetString(IllegalStateException,
                                ERR_MSG_SHARE_CONSUMER_CLOSED);
```

Same at other places

src/confluent_kafka/cimpl.pyi
```diff
@@ -275,6 +275,15 @@ class KafkaException(Exception):
...
     args: Tuple[Any, ...]
 
+# These subclass RuntimeError; str(exc) is the error message (not a KafkaError).
+class IllegalStateException(RuntimeError):
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
+    args: Tuple[Any, ...]
+
+class ConcurrentModificationException(RuntimeError):
+    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
+    args: Tuple[Any, ...]
```
pranavrth, line 285: Check if anything changed after we changed these errors to not take `KafkaError`

pranavrth (changes requested):

src/confluent_kafka/src/ShareConsumer.c
k-raina: Delegating to cfl_PyErr_from_error_destroy introduces a refcount leak i.e. the macros call PyErr_SetObject(KafkaException, _eo) but never Py_DECREF(_eo)

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -116,6 +117,78 @@ ShareConsumer_traverse(ShareConsumerHandle *self, visitproc visit, void *arg) {
...
+                vsnprintf(buf, sizeof(buf), fmt, ap);
+                va_end(ap);
+        }
+        msg   = fmt ? buf : rd_kafka_err2str(code);
+        etype = ShareConsumer_translate_error(code);
+
+        if (etype == KafkaException) {
+                PyObject *eo = KafkaError_new0(code, "%s", msg);
+                if (!eo)
+                        return;
```
pranavrth, line 162: We should remove this as it is not setting error and returning as it is.

src/confluent_kafka/src/ShareConsumer.c
pranavrth: Add a TODO for future then. When `cfl_PyErr_from_error_destroy` is fixed.

pranavrth (approved):
Build is failing. Check that.
Provided minor comment to be resolved in next PR.

Apart from that, LGTM!. Thanks @k-raina !.

pranavrth (approved):
LGTM!.

PR #2290, opened 2026-06-20
[KIP-932] - Implement new interface changes to share consumer

k-raina:
**Summary**
- Enabled stats_cb and configuration at python layer
- Change poll return from array to ConsumerRecords.
- Change return type of share consumer return type to KafkaError from Kafka Exception.


src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -368,7 +370,23 @@ static PyObject *ShareConsumer_poll(ShareConsumerHandle *self,
 
         rd_kafka_messages_destroy(rkmessages);
 
-        return msglist;
+        /* Wrap the plain list in ConsumerRecords so callers get records() /
+         * count() / is_empty(). */
+        ConsumerRecords_type =
+            cfl_PyObject_lookup("confluent_kafka", "ConsumerRecords");
```
pranavrth: Resolve it from the defining module `confluent_kafka._consumer_records` (maybe changed with `Messages` class) rather than the top-level package so it doesn't depend on `__init__` export ordering. Basically, we should import directly from the defining module instead.

pranavrth: `ConsumerRecords` is looked up on every `poll()`. We should cache it once in a function-static `PyObject *` object. The object should be initialized lazily and it will be kept for the lifecycle of the process.

src/confluent_kafka/_consumer_records.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+class ConsumerRecords(list):
```
pranavrth: `Messages` should be a composed container, not a `list` subclass. Here we should chose composition instead of inheritance.

Benefits of composition here:

- **Future-proof representation.** We can change the internal backing later (e.g. to `Map<TopicPartition, List[Message]>`) to add the Java-parity accessors `partitions()` / `records(tp)` / `nextOffsets()`, without changing the public type. A `list` subclass is permanently flat — adding those means either an O(n) re-scan on every call or a breaking change.
- **No leaking subclass identity.** Slicing, `+`, and `copy()` on a `list` subclass return a plain `list`, silently dropping `records()`/`count()`/`is_empty()`. A composed type stays itself.
- **Immutability.** A polled batch is an immutable snapshot of acquired records. Subclassing `list` exposes the whole mutating API (`append`/`extend`/`sort`/`pop`/`clear`/slice-assignment) — all meaningless on a batch and able to mask bugs. Composition exposes only what's meaningful.
- **Controlled, honest API surface.** We expose exactly the methods that make sense (sequence protocol + accessors) instead of inheriting ~30 list methods we never intend users to call. `repr`, `dir()`, and the type all read as "a ConsumerRecords," not "a list with extras."
- **No back-compat trap.** Once this ships as a `list` and callers rely on `isinstance(x, list)` / slicing / mutation, removing the list-ness is a breaking change. Composition avoids painting us into that corner — and pre-GA is the only cheap time to choose.
- **Still a drop-in for normal usage.** The sequence protocol (`__iter__`/`__len__`/`__getitem__`) keeps `for r in poll()`, `len(batch)`, `if batch:`, and `batch[i]` working exactly as today.

```python
class Messages:
    """Immutable batch returned by ShareConsumer.poll()."""
    def __init__(self, messages=()):
        self._records = list(messages)

    @classmethod
    def _from_list(cls, records):        # zero-copy path for the C layer
        obj = cls.__new__(cls)
        obj._records = records
        return obj

    def records(self):    return list(self._records)
    def count(self):      return len(self._records)
    def is_empty(self):   return not self._records

    def __iter__(self):       return iter(self._records)
    def __len__(self):        return len(self._records)
    def __getitem__(self, i): return self._records[i]
    def __repr__(self):       return f"ConsumerRecords({self._records!r})"

Comment (pranavrth) [inline src/confluent_kafka/_consumer_records.py]: We should change the name to `Messages` to match the existing naming convention.

File: src/confluent_kafka/_consumer_records.py [hunk 3]
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
```
pranavrth: I think this should be directly inside `__init__`

src/confluent_kafka/src/ShareConsumer.c
pranavrth, line 331: We should check the performance implication of this change as well to verify that everything is fine after the changes. We should relook at the overall approach if there is too much divergence.

pranavrth (changes requested):
Went through on the high level. I have posted few design level changes. You can start with those. I will also do a deep implementation level of review.

src/confluent_kafka/src/ShareConsumer.c
k-raina: Agree, we can rerun perf tests with this PR

src/confluent_kafka/src/ShareConsumer.c
```diff
@@ -880,30 +904,22 @@ static PyMethodDef ShareConsumer_methods[] = {
...
-     *   - the equivalent translation lands on commit_sync / commit_async /
-     *     ack-callback paths (currently TODO'd separately).
-     */
+    /* TODO: the error-code -> Python-exception mapping is still provisional.
+     * Every error code currently surfaces as KafkaException via
+     * cfl_PyErr_Format(), including the per-partition values in commit_sync's
+     * result dict. Eventually specific codes should map to the obvious
+     * exception type (_INVALID_ARG -> ValueError, _STATE -> RuntimeError).
+     * Revisit once the error surface settles and the same mapping lands on
+     * commit_sync / commit_async / the ack callback. */
```
pranavrth: Remove this todo.

src/confluent_kafka/_messages.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+#
+
+
+class Messages:
+    """Batch of messages returned by ShareConsumer.poll().
+
+    Read-only sequence (iterate, len, index, slice) plus count()/is_empty()/
+    records(). Not a list subclass on purpose -- a polled batch shouldn't be
+    mutated, and a private backing list stays swappable later.
```
pranavrth: We don't write the reason for choosing this. We need to explain what it is.

src/confluent_kafka/cimpl.pyi
```diff
@@ -45,6 +45,7 @@ except ImportError:
 
 from confluent_kafka.admin._metadata import ClusterMetadata, GroupMetadata
 
+from ._messages import Messages
```
pranavrth: Add entry for Messages in type hinting.

src/confluent_kafka/__init__.py
```diff
@@ -18,6 +18,7 @@
 
 import os
 
+from ._messages import Messages
```
pranavrth: Let's try to understand better place to keep this new class.

pranavrth (changes requested):

src/confluent_kafka/cimpl.pyi
pranavrth: This is not done yet.

k-raina: Change got lost in rebase, pushed in commit https://github.com/confluentinc/confluent-kafka-python/pull/2290/commits/c7c8851d577a3f65f50edc108e7212824f03ada0

pranavrth (approved):
LGTM!. Thanks @k-raina !

PR #2296, opened 2026-06-26
[DGS-24664] Harden HCVault KMS TLS verify against empty CA value; add tests

ap00rv:
## What

Follow-up hardening for #2294 (TLS verification for the HashiCorp Vault KMS client), plus the test coverage that PR did not include.

### Fix: empty CA value silently disabled verification

`HcVaultKmsDriver._get_verify` used `if ca_location is not None`. An empty `ssl.ca.location` config value or an empty `VAULT_CACERT` environment variable (an easy misconfiguration — an unset/templated env var resolves to `""`) returned `""`. `requests`/`hvac` treat a falsy `verify` as "do not verify the server certificate" (`CERT_NONE`), so this silently re-disabled TLS verification and reopened the certificate-validation hole #2294 closed.

The fix uses a truthiness check and falls back to the secure default `True`, so any empty/unset value keeps verification on, while a real CA path is still honored (config taking precedence over the `VAULT_CACERT` env var).

```python
ca_location = conf.get(_SSL_CA_LOCATION) or os.getenv("VAULT_CACERT")
if ca_location:
    return ca_location
return True
```

### Tests

Adds coverage that exercises the real wiring `new_kms_client -> HcVaultKmsClient -> hvac.Client` (the existing tests only covered the `_get_verify`/`_get_cert` helpers in isolation):

- default-configured driver builds an `hvac.Client` with `verify=True` / `cert=None`
- a configured CA bundle and client cert/key reach `hvac.Client` intact (guards against dropped/mis-ordered args)
- empty `ssl.ca.location` / `VAULT_CACERT` still verify (regression guard for the fix above)

## Test & Review

Validated locally against the fixed modules (full `tests/schema_registry/test_hcvault_driver.py`):

```
17 passed
```

Each new test was also confirmed to **fail** against the pre-fix code (negative control), e.g. the empty-`VAULT_CACERT` case produced `hvac.Client(..., verify="")` before the fix.

## References

JIRA: https://confluentinc.atlassian.net/browse/DGS-24664

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/confluent_kafka/schema_registry/rules/encryption/hcvault/hcvault_driver.py
+++ b/src/confluent_kafka/schema_registry/rules/encryption/hcvault/hcvault_driver.py
@@ -59,11 +59,14 @@ class HcVaultKmsDriver(KmsDriver):
     @staticmethod
     def _get_verify(conf: Dict[str, Any]) -> Union[bool, str]:
         # A CA bundle path enables verification against that bundle. The standard
-        # VAULT_CACERT environment variable is honored as a fallback.
-        ca_location = conf.get(_SSL_CA_LOCATION)
-        if ca_location is None:
-            ca_location = os.getenv("VAULT_CACERT")
-        if ca_location is not None:
+        # VAULT_CACERT environment variable is honored as a fallback. An empty or
+        # unset value must NOT disable verification: requests/hvac treat a falsy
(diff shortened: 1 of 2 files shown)


rayokota (approved):
Thanks @ap00rv , LGTM

PR #2304, opened 2026-07-06
[NOGIL] Add CI verification jobs for free-threaded Python 3.14t

ojasvajain:
First step toward free-threading (no-GIL) support: run source-package verification and integration tests against CPython 3.14t on every commit.

  - New Semaphore block with classic and consumer group-protocol jobs on 3.14t.
  - `requirements-tests-install-nogil.txt` — test deps without the rules/json-fast extras, whose compiled deps (tink, google-re2, grpcio; orjson) ship no free-threaded wheels; selected by the verification script on free-threaded builds.
  - Collection exclusions for the 11 schema_registry test modules importing those deps.
  - GIL-status fixture warning when the GIL is re-enabled; becomes an assertion once cimpl declares support.

  Note: cimpl doesn't yet declare Py_MOD_GIL_NOT_USED, so these jobs run with the GIL re-enabled — they validate the 3.14t toolchain and packaging, not lock-free execution. cp314t wheels and the thread-safety work will be added in a separate PR.

What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -233,6 +233,38 @@ blocks:
             - cp test-report.xml test-output
             - test-results publish test-output
             - artifact push workflow coverage.xml
+  - name: "Source package verification and Integration tests with Python 3.14t free-threaded (Linux x64)"
+    dependencies: []
+    task:
+      agent:
+        machine:
+          type: s1-prod-ubuntu24-04-amd64-3
+      env_vars:
+        - name: OS_NAME
+          value: linux
+      prologue:
+        commands:
+          - export ARCH=x64
+          - '[[ -z $DOCKERHUB_APIKEY ]] || docker login --username $DOCKERHUB_USER --password $DOCKERHUB_APIKEY'
+          - sudo apt-get update -qq && sudo apt-get install -y -qq clang-format
+      jobs:
+        - name: Build and Tests with 'classic' group protocol
+          commands:
+            - sem-version python 3.11
+            - sem-version java 17
+            - pip install uv
+            - uv venv _venv --python 3.14t && source _venv/bin/activate
+            - chmod u+r+x tools/source-package-verification.sh
+            - tools/source-package-verification.sh
+        - name: Build and Tests with 'consumer' group protocol
+          commands:
+            - sem-version python 3.11
+            - sem-version java 17
+            - pip install uv
+            - uv venv _venv --python 3.14t && source _venv/bin/activate
+            - chmod u+r+x tools/source-package-verification.sh
+            - export TEST_CONSUMER_GROUP_PROTOCOL=consumer
+            - tools/source-package-verification.sh
   - name: "Source package verification with Python 3 (Linux arm64)"
     dependencies: []
     task:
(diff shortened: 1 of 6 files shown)


rayokota (approved):
Thanks @ojasvajain , LGTM

.semaphore/semaphore.yml
```diff
@@ -233,6 +233,38 @@ blocks:
             - cp test-report.xml test-output
             - test-results publish test-output
             - artifact push workflow coverage.xml
+  - name: "Source package verification and Integration tests with Python 3.14t free-threaded (Linux x64)"
```
k-raina, line 236: Any reason why we are adding new jobs and not changing current ones? 
Adding new jobs and running old ones too means additional costs.

.semaphore/semaphore.yml
```diff
@@ -233,6 +233,38 @@ blocks:
...
+          value: linux
+      prologue:
+        commands:
+          - export ARCH=x64
+          - '[[ -z $DOCKERHUB_APIKEY ]] || docker login --username $DOCKERHUB_USER --password $DOCKERHUB_APIKEY'
+          - sudo apt-get update -qq && sudo apt-get install -y -qq clang-format
+      jobs:
+        - name: Build and Tests with 'classic' group protocol
+          commands:
+            - sem-version python 3.11
```
k-raina, line 253: Is using 3.11 version intentional?

requirements/requirements-tests-install-nogil.txt
k-raina, line 1: Considering that this PR merged into feature beanch, why are we not making changes to requirements/requirements-tests-install.txt ?

k-raina (changes requested):
Thanks for raising PR. Provided first pass of comments

.semaphore/semaphore.yml
ojasvajain: The existing job runs on 3.11 which is a GIL based Python. We need this new job because there are some differences in how the GIL free Python behaves:

1. There are some dependencies that do not provide free threaded wheels. If such deps are imported in the code, the job will fail. The code author should then take steps to exclude it and possibly document the behaviour saying that XYZ functionality is not supported on free threaded Python builds.

2. There is a good difference in the underlying CPython 3.14t vs any GIL based CPython. It is important we ensure all our existing test cases pass. Also I have added a fixture that will ensure GIL does not get enabled during a test run. That fixture will be applicable only when we run test cases on 3.14t. Currently the fixture is commented out, but I have a TODO to enable it once I make the core changes.

.semaphore/semaphore.yml
ojasvajain: Yes. `sem-version` does not provide 3.14t. So I am setting 3.11 as the host Python interpretor and then I am using `uv` and `venv` to set the version of the virtual environment to 3.14t
```uv venv _venv --python 3.14t && source _venv/bin/activate```.

requirements/requirements-tests-install-nogil.txt
ojasvajain: I have added a branch in source-package-verification.sh to use `requirements-tests-install` when running on GIL based Python. There are certain deps in `requirements-tests-install` that are not supported on GIL free python so I had to exclude them and create a new file called `requirements-tests-install-nogil`.

k-raina (approved):
Thanks for PR @ojasvajain . LGTM!

PR #2305, opened 2026-07-08
Bump authlib min version to 1.7.1

### LINKED ISSUE #2242: Use joserfc instead of authlib.jose
Confluent-kafka depends on authlib, which is deprecating authlib.jose. I get this warning with authlib==1.7.0:

`AuthlibDeprecationWarning: authlib.jose module is deprecated, please use joserfc instead.`

Can we get a confluent-kafka version that no longer uses authlib.jose?

The change in authlib is here: https://github.com/authlib/authlib/pull/852/changes#diff-b5d142ef6b84d7a63cc3a10fdfa6152539bd1b621829f319249e97eed1b6b1de

yvaishalirao:
Fixes #2242

authlib 1.7.0 raises an `AuthlibDeprecationWarning` for `authlib.jose` on import, even though this repo only uses `authlib.integrations.httpx_client.OAuth2Client` and never touches `authlib.jose` directly. It's an authlib-side import chain issue, fixed upstream in 1.7.1.

Bumped the `authlib` floor in `requirements-schemaregistry.txt` and `requirements-examples.txt` from `>=1.0.0` to `>=1.7.1`.

Verified `tests/schema_registry/_sync/test_bearer_field_provider.py` and the async counterpart still pass unchanged against 1.7.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Confluent Python Client for Apache Kafka - CHANGELOG
 
+### Fixes
+
+- Bump minimum `authlib` version to 1.7.1 to avoid a spurious `authlib.jose` deprecation
+  warning raised on import by `authlib` 1.7.0 (#2242)
+
 ## v2.15.0
 
 ### [KIP-932](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka) Queues for Kafka – Now in **Preview**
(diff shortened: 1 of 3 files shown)


yvaishalirao:
Hi @rayokota , it looks like the Semaphore job has failed multiple times due to what appears to be a CI issue rather than a code change. Would you mind rerunning the failed workflow when you have a chance? Thanks

PR #2310, opened 2026-07-20
Add DLQ rule action to Schema Registry serdes

Claimundefine:
Port the Java client's DlqAction to confluent-kafka-python. When a data-contract rule (CSFLE encryption, CEL condition, etc.) fails during serialize/deserialize and the rule's onFailure is set to "DLQ", the record is teed to a dead-letter-queue topic and a SerializationError is still raised (tee, not swallow).

- DlqAction + FieldRedactionExecutor under rules/dlq/
- Capture original key/value for the DLQ record via a contextvar (thread- and task-safe), mirroring the Java client's ThreadLocal
- Redact fields tagged by encryption rules before JSON-encoding so plaintext does not leak to the DLQ
- Skip a rule on replay when the record carries the matching __rule.name header
- Configure registered rule actions in each serde; close owns non-global registries

What
----

Checklist
------------------

References
----------
JIRA: 

DGS-14069

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@
 - Pass context when clients make KEK calls to DEK Registry (#2308)
 - Minor fix for subjectPrefix parameter in subjects API (#2311)
 
+### Fixes
+
+- `DeserializingConsumer` and `DeserializingShareConsumer` now deserialize the
+  message key before the value (previously the value was deserialized first),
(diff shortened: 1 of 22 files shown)


rayokota (approved):
Thanks @Claimundefine , LGTM

Issue #2092, opened 2025-10-07
Add Free-Threaded Python Support in 3.14


mkmoisen:
With the release of Python 3.14, free threaded support (i.e., GIL-less python) is now a sensible option. 

Third party packages that use C-Extensions like confluent_kafka however need to have code change to support free-threaded mode.

Here is the documentation on how to change C-Extensions to support free threaded mode:

https://docs.python.org/3.14/howto/free-threading-extensions.html#freethreading-extensions-howto

My understanding is that if a third party application does not support free-threaded python, then applications which import these applications will revert to using the GIL. This means that users are stuck with the GIL until all third party applications have ported to free-threaded compatibility. 

May we please put this on the road map for confluent_kafka?

Thank you.

MSeal:
Thanks for raising the issue @mkmoisen -- I've been a preliminary pass of the C code change requirements. It seems like the code changes will be minimal but testing this will require a bit of preparation work to ensure no deadlocks of segfaults are introduced against the librdkafka layer. I've added it to the roadmap, and will discuss with the team next week.

cclauss:
Is this fixed in the new v2.12.1 release?
* https://github.com/confluentinc/confluent-kafka-python/releases

dpdani:
As additional documentation, you can also take a look at https://py-free-threading.github.io/porting-extensions/

Apart from that page in particular, you may also find the rest of the site helpful.

I reckon the underlying librdkafka library is thread-safe, so most of the problems may arise from possible concurrent mutations on the [`Handle` object](https://github.com/confluentinc/confluent-kafka-python/blob/06c0744986b1796947b2c20629a2b3cac90e642f/src/confluent_kafka/src/confluent_kafka.h#L260), which should be solvable with [the new critical section API](https://py-free-threading.github.io/porting-extensions/#critical-sections).

I don't know a lot about the internals of this library but I do know free threading quite well, let me know if I can be of help in any way 🙏

MSeal:
@cclauss no, there is a wheel compiled for free-threading but the C-bindings aren't aware of this yet. @dpdani if you want to take a pass on updating the C-bindings, we're happy to have the contributions. The librdkafka is thread safe outside of some bug reports on specific call sequences that might need a patch, but testing here is the biggest concern as we don't have many multi-thread test examples at the Python level to feel confident we didn't miss anything.

dpdani:
[...]

PR #2312, opened 2026-07-21
[NOGIL] Declare cimpl free-threaded on CPython 3.14t

honglei:
Relates to #2092

## What

Declare the CPython 3.14t `cimpl` extension as free-threaded and protect the
lazy `ShareConsumer` `Messages` type cache when the GIL is disabled.

This is based on `dev_thread_free_support`, which already provides the 3.14t
build and test dependency preparation.

## Changes

- Call `PyUnstable_Module_SetGIL(m, Py_MOD_GIL_NOT_USED)` for free-threaded
  builds.
- Protect the process-lifetime `Messages` type initialization with `PyMutex`
  on free-threaded builds.
- Add isolated free-threading tests for independent and shared Producer use.
- Add an integration test using the existing `kafka_cluster` fixture to cover
  shared AdminClient, Producer, and Consumer operations.
- Run the strict free-threading tests in a separate process before the full
  regression suite. The full suite still reports warnings when optional
  third-party extensions, such as `fastavro`, re-enable the GIL.

## Validation

Validated on CPython 3.14.6 free-threading at `/opt/python314t` on 10.2:

- Formatting, shell syntax, and diff checks passed.
- Isolated no-GIL tests: `3 passed, 1 skipped` without a broker.
- Real-broker no-GIL tests: `4 passed` against `127.0.0.1:19092` before wiring
  the test into the project integration fixture.
- Final `kafka_cluster` integration test against the same broker:
  `1 passed in 1.42s` with `PYTHON_GIL=0`.
- Core regression tests: `108 passed`.
- `pip check`: no broken requirements.
- Earlier 60-second real-broker soak: 99 rounds, approximately 198,000
  messages, zero failures, and no leftover temporary topics.

The integration test is run in a separate `PYTHON_GIL=0` process because the
full test process imports optional extensions such as `fastavro` that can
re-enable the GIL. The existing source verification job already prepares the
Kafka artifact used by the `kafka_cluster` fixture.

## Scope

This change does not claim concurrent `close` while another thread is still
calling `poll` or `consume`. Callers must stop worker threads before closing a
shared client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ for a worked example.
 ### Enhancements
 
 - Add support for union-of-pools and auto pool mapping for Schema Registry (#2182)
+- Add free-threaded CPython 3.14 support to the core C extension.
 
 ### Fixes
 
--- a/src/confluent_kafka/src/ShareConsumer.c
+++ b/src/confluent_kafka/src/ShareConsumer.c
@@ -328,6 +328,26 @@ static PyObject *ShareConsumer_subscription(ShareConsumerHandle *self,
  * @brief Poll for a batch of messages from the share consumer.
  *
  */
+static PyObject *ShareConsumer_get_messages_type(void) {
+        static PyObject *messages_type = NULL;
+#ifdef Py_GIL_DISABLED
+        static PyMutex messages_type_mutex = {0};
+
+        PyMutex_Lock(&messages_type_mutex);
+#endif
+
+        if (!messages_type)
+                messages_type =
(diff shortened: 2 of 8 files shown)


ojasvajain:
Hi @honglei,

Thank you for putting together this PR and contributing!

Currently, our team is actively working on the free threading implementation internally, so we aren't accepting external PRs for this feature just yet. Additionally, there are a few edges that we want to address before declaring the module compatible without GIL.

Because of this, I’m going to close this PR for now. Thanks again for your time and effort.

honglei:
Any Updates？ @ojasvajain

PR #2316, opened 2026-07-25
feat(producer,consumer): expose cluster_id() via rd_kafka_clusterid()

shashank-reddy-nr:
What
----
Add a `cluster_id()` method to both `Producer` and `Consumer` that wraps
librdkafka's `rd_kafka_clusterid(rk, timeout_ms=0)`.

**Why**: OpenTelemetry's `messaging.kafka.cluster.id` semantic convention
attribute requires the Kafka cluster id on every span. Today there is no way
to retrieve this from a confluent-kafka-python producer or consumer without
creating a separate `AdminClient` and calling `list_topics()`.

**How it works**:
- `rd_kafka_clusterid(rk, 0)` returns the cached cluster id immediately
  (non-blocking, `timeout_ms=0`), or `NULL` if not yet fetched.
- An empty string result is also normalised to `None`.
- Memory is freed via `rd_kafka_mem_free()`, matching the existing
  `memberid()` implementation exactly.
- `cimpl.pyi` type stubs updated: `def cluster_id(self) -> Optional[str]: ...`

**Pattern followed**: `Consumer_memberid` in `Consumer.c` (the function is
structurally identical, just calling a different librdkafka function).

Checklist
------------------
- [x] Contains customer facing changes? Including API/behavior changes — yes, new public API method `cluster_id()` on Producer and Consumer
  - Integration tests added for both Consumer (`test_consumer_clusterid.py`) and Producer (`test_producer_clusterid.py`), covering: non-empty string after first produce/poll, and RuntimeError on closed handle.

References
----------
- librdkafka `rd_kafka_clusterid` signature: `src/rdkafka.h:3360`
- OpenTelemetry Kafka semantic conventions: `messaging.kafka.cluster.id` (incubating)
- Related cross-language work: [kafkajs#1797](https://github.com/tulios/kafkajs/pull/1797) (Node.js)

Test & Review
------------
Integration tests require a running Kafka cluster (standard test harness).

```python
from confluent_kafka import Producer, Consumer

p = Producer({'bootstrap.servers': 'localhost:9092'})
p.produce('my-topic', b'hello')
p.flush()
print(p.cluster_id())  # e.g. "abc123XYZ..."

c = Consumer({'bootstrap.servers': 'localhost:9092', 'group.id': 'g1'})
c.subscribe(['my-topic'])
c.poll(5)
print(c.cluster_id())  # e.g. "abc123XYZ..."
```

Before metadata is fetched, `cluster_id()` returns `None`; callers should
retry after a short delay if they need the value before the first poll.

Open questions / Follow-ups
--------------------------
- Should this also be surfaced on `AdminClient`? The same `Handle.rk` pattern
  applies but is out of scope for this PR.


ojasvajain:
Hi @shashank-reddy-nr, thanks for raising this PR. Wanted to point out that `list_topics` is available on Producer and Consumer instances as well, which can retrieve the cluster ID for you. It does trigger a metadata fetch call each time instead of reading from cache, but unless that's a worry for your use case, I'd suggest calling `list_topics` directly.

shashank-reddy-nr:
Thanks @ojasvajain. Yes, list_topics() is working fine, so I'm closing this PR. I can reopen it later if needed. My only concern is that fetching all other fields just to get the cluster ID isn't ideal, but it works for now. Good to go!

PR #2317, opened 2026-07-29
[NOGIL] Guard AdminClient against concurrent close() vs method-call races

ojasvajain:
**Problem**

An AdminClient instance can be shared across multiple threads — this is a common, previously-supported pattern. One long-standing race in this pattern: one thread destroys the Admin client (setting rk to NULL) while another thread is concurrently calling a librdkafka API on that same rk. This race has always existed, but on free-threaded builds the window widens significantly since there's no GIL to serialize the two threads.

**Change**
To close this race, every Admin API is now gated using active_calls/closing atomic variables. A CAS (compare-and-swap) construct has been added to __exit__() to correctly handle concurrent __exit__() calls. This mirrors the gating already implemented for Producer.

**Other changes**
1. APIs shared across clients (e.g. list_topics, set_sasl_credentials) were previously left unguarded. They are now gated for Producer, Consumer and Admin.  
2. Adds concurrency-related unit and integration test cases.

What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


tests/concurrency/test_admin_close_race.py
```diff
@@ -0,0 +1,217 @@
...
+    """__exit__() blocks until an in-flight poll() call finishes."""
+    admin = AdminClient(_ADMIN_CONF)
+    poll_started = threading.Event()
+    poll_duration = 10
+    poll_finished_at = None
+
+    def run_poll():
+        nonlocal poll_finished_at
+        poll_started.set()
+        admin.poll(poll_duration)
```
k-raina: If broker is unreachable `remaining_poll_duration is poll_finished_at − exit_start`, and since poll() already finished ~2s before exit_start, it's negative. So the assertion is (≥0) >= (negative) is always true.

tests/concurrency/test_admin_close_race.py
```diff
@@ -0,0 +1,217 @@
...
+                    except Exception:  # noqa: BLE001 - broker-less errors expected, not the bug we're after
+                        pass
+            except RuntimeError:
+                break
+
+    _race_exit_against(worker)
+
+
+@subprocess_isolated
+def test_exit_races_list_topics():
```
k-raina, line 102: Do we need similiar test for set_sasl_credentials ?

tests/concurrency/test_admin_close_race.py
```diff
@@ -0,0 +1,217 @@
...
+        while not stop_event.is_set():
+            try:
+                futmap = admin.create_topics([NewTopic('mytopic', num_partitions=1, replication_factor=1)])
+                for f in futmap.values():
+                    try:
+                        f.result(timeout=1)
+                    except Exception:  # noqa: BLE001 - broker-less errors expected, not the bug we're after
+                        pass
+            except RuntimeError:
+                break
```
k-raina, line 96: I am not sure if we should treat runtime error as sucess? Please check for rest of PR

tests/concurrency/test_admin_close_race.py
```diff
@@ -0,0 +1,217 @@
...
+
+@subprocess_isolated
+def test_exit_races_list_topics():
+    """Exiting the `with` block concurrent with list_topics() on another thread."""
+
+    def worker(admin, stop_event):
+        while not stop_event.is_set():
+            try:
+                admin.list_topics(timeout=0.05)
+            except RuntimeError:
```
k-raina: Same

k-raina (changes requested):
Couple of comments on tests

tests/concurrency/test_admin_close_race.py
ojasvajain: Runtime errors are expected when the call starts but close has already been initiated on another thread. To harden it, I can add an assertion inside the except to check if it is the "closing" error and nothing else.

tests/concurrency/test_admin_close_race.py
ojasvajain: Replied above

tests/concurrency/test_admin_close_race.py
ojasvajain: Added a test case

tests/concurrency/test_admin_close_race.py
ojasvajain: Nice catch. I have replaced poll() with list_topics() which actually blocks for the given timeout. Also added assertion that ensures list call finished after exit started.

tests/concurrency/test_admin_close_race.py
ojasvajain: Done

k-raina (approved):
LGTM!

PR #2322, opened 2026-08-06
[NOGIL] Restrict sharing of Consumer instances across threads

ojasvajain:
Introduces a reentrancy gate for the sync Consumer and AIOConsumer that serializes concurrent access to a single Consumer instance from a different caller — a colliding call waits for the current caller to finish rather than interleaving or racing directly against librdkafka's own consumer, which isn't thread-safe — while still allowing legitimate re-entrant calls to proceed immediately — e.g. a rebalance or commit callback calling back into the Consumer/AIOConsumer that triggered it (on_assign calling assign(), on_commit calling commit(), etc.).

A unified gate_owner/gate_depth mechanism (Handle_gate_enter()/Handle_gate_exit() in Consumer.c) is shared between both sync and async consumer types, and every gated method resolves its own caller identity internally. For the sync consumer, calling thread ID serves as the identity, while for the async consumer an ID is generated before dispatching each call and carried over through the task chain using context variables. For async consumers this is required, as callbacks may get scheduled on a different worker thread, so a thread ID can not be used as an identity.


tests/concurrency/test_producer_close_race.py
```diff
@@ -186,6 +190,10 @@ def worker(producer, stop_event):
 def test_close_races_abort_transaction():
     """close() concurrent with abort_transaction() on another thread."""
 
+    # TODO NOGIL: move to tests/integration -- abort_transaction() needs a
+    # real transaction coordinator to reach the race being tested; against
+    # the unreachable localhost:9092 used here it fails with a genuine
+    # _STATE KafkaException instead of the expected RuntimeError.
```
ojasvajain: Will revisit this in a separate PR that I plan to raise for Transactional Producer related audit.

src/confluent_kafka/aio/_common.py
```diff
@@ -14,13 +14,79 @@
...
+    The sync Consumer uses the calling thread's ID as its gate identity, but
+    that does not work for AIOConsumer: a call and the re-entrant calls its
+    callbacks make can run on different ThreadPoolExecutor worker threads (e.g.
+    a rebalance callback dispatched to one worker, then a re-entrant call from
+    within it scheduled on another). AIOConsumer therefore generates an
+    identity per top-level call and carries it in a ContextVar the gate reads
+    -- see Handle_gate_enter() in Consumer.c.
+    """
+
+    _var = cimpl._reentry_identity_var
```
k-raina, line 41: The gate identifies callers with a contextvar asyncio copies a task's context into every child task it spawns (gather, create_task), so any task launched from inside a callback inherits the same identity as the callback. 

- When an application writes callbacks like:
```
  async def on_assign(consumer, partitions):
      await asyncio.gather(*(consumer.seek(tp) for tp in partitions))

  def on_commit(err, partitions):
      asyncio.create_task(consumer.commit())
```
-  the spawned tasks inherit the callback's identity, so the owner == identity check in the gate is true for each of them and the gate admits them concurrently instead of serializing them.
-   when poll() is running and its rebalance callback fires. The callback does gather(consumer.seek(a), consumer.seek(b))  two child tasks, both carrying the callback's copied identity are dispatched onto two executor worker threads.
 -  Task A enters Handle_gate_enter. Gate is free, so A takes it. A is now inside the C code, using self->rk
 - Task B enters Handle_gate_enter. It sees gate_owner == identity (same copied badge), assumes it's a nested re-entrant call, does gate_depth++, and returns "admitted" immediately. 
 -  Now A and B are both inside the gated code at once, both touching self->rk.
 
Since running on two threads at once, those updates can collide with each other. When both threads increment and decrement the same object's reference count without synchronization, one of the updates can be lost, and the recorded count drops below the number of references that actually exist. The object is then freed while it is still in use, and the next thread to reach for it dereferences memory that has already been released . The use after free that can crash the process. 

Can we check if this can cause any issue?

src/confluent_kafka/src/Consumer.c
```diff
@@ -38,6 +44,83 @@
...
+                        return 1;
+                }
+
+                /* Someone else holds the gate: wait for it. */
+                CallState cs;
+                CallState_begin(h, &cs);
+#ifdef _WIN32
+                Sleep(1);
+#else
+                usleep(1000);
```
k-raina, line 107: Lets take an minimal example : 
```
 async def on_assign(c, partitions):
      await c.assign(partitions)

  await consumer.subscribe([topic], on_assign=on_assign)

  await asyncio.gather(
      consumer.poll(10),
      consumer.commit(),
  )

```

AIOConsumer runs every call on a fixed thread pool (default 2 workers). A poll() takes worker 1 and the gate, then blocks waiting for its rebalance callback to return. A concurrent commit() (different thread) takes worker 2 and busy-spins at the gate instead of releasing its worker. The callback now makes a re-entrant call (assign/pause) that needs a worker but both are taken. 

**Application impact** :  For applications doing manual offset management (seek/commit from a rebalance callback) while any background task also calls the consumer will be stuck in Deadlock. 
 
Could we check this?

k-raina, line 107: For code snippet: 
```
async def consume_loop():
    while running:
        msg = await consumer.poll(-1)     # block until the next message
        if msg:
            await process(msg)

async def shutdown_on_signal():
    await consumer.close()                # ← must wait behind poll(-1) at the gate

await asyncio.gather(consume_loop(), shutdown_on_signal())
```

1. consume_loop's poll(-1) calls Handle_gate_enter and takes the gate, then enters a chunk loop. With timeout = -1, poll  forever until a message arrives. The gate is held the entire time.
2. shutdown's close() calls Handle_gate_enter, sees the gate owned by poll (different identity), and spins in the wait loop (usleep) waiting for poll to release it.

**Application impact**
For  applications with Idle topic → no message → poll(-1) never returns → the gate is never released → close spins forever. The consumer is stuck.

Can we check this?

k-raina (changes requested):
Thanks for PR!
Provided couple of correctness comments on gating logic.

src/confluent_kafka/src/Consumer.c
ojasvajain: Yes this is possible if number of workers are less. But we can't have multiple tasks/threads touching the consumer handle at the same time and given AIO consumer schedules callbacks on different thread, we will need to accept this caveat. We will document this behaviour and recommend users to have enough workers in their thread pool. 

Since this change will be part of a minor version upgrade, we decided to serialize the calls rather than returning ConcurrentModificationException to the users, as this will not be backward compatible. In the next major version upgrade, we can replace the waiting behaviour with throwing the exception which will avoid such deadlocks and change them to exceptions.

src/confluent_kafka/aio/_common.py
ojasvajain: Yes I have documented it as a gap [here](https://github.com/confluentinc/confluent-kafka-python/blob/5e7c1138c6b5de6247755508a3bec9b9587be9a6/src/confluent_kafka/aio/_common.py#L120)(line 120). We can use python level lock to serialize the access. I will address this in a separate PR.

src/confluent_kafka/src/Consumer.c
ojasvajain: Yes. This will expected. And this is the cost of serializing the calls, rather then returning the exception but that will not be backward compatible. 

In the latest version of the python client, calling close() in parallel with poll(-1) is unsafe and causes hangs or segfault. So this is not a supported pattern anyway.

In Java client, close() will get a ConcurrentModificationException.

src/confluent_kafka/aio/_common.py
ojasvajain: Added a TODO at the same place

k-raina: Thanks @ojasvajain , 
As this has potential to crash, not sure if we should prioritise this fix. Feel free to ping me , i will review second PR ASAP.

src/confluent_kafka/src/Consumer.c
k-raina: Since we are documenting that "consumer should not be called from multiple user threads", should we mention that it would **deadlock** in document, so its not a surprise if some one faces it? 

Also as you mentioned above, AIOConsumer hits it because the concurrency is created by the library, not the user. The user writes ordinary single-event-loop async code, but AIOConsumer runs every call on its internal ThreadPoolExecutor (default 2 workers). Should we plan to update default threads?

k-raina: >In the latest version of the python client, calling close() in parallel with poll(-1) is unsafe and causes hangs or segfault. So this is not a supported pattern anyway.

This behavior might be working for current latest version of python client, as it doesn't support free threading. However with free threading, will above consumer stuck issue might be commonly encounted?

If we are documenting that consumer will not free threading, that will work for Sync Consumer, however will consumer stuck still be case for AIOConsumer running in NOGIL package?

tests/integration/consumer/test_aio_consumer_concurrent_serialization.py
```diff
@@ -0,0 +1,200 @@
...
+    consumer_conf = kafka_cluster.client_conf(
+        {
+            'group.id': str(uuid1()),
+            'auto.offset.reset': 'earliest',
+            'enable.auto.commit': False,
+        }
+    )
+    if conf:
+        consumer_conf.update(conf)
+    return TestAIOConsumer(consumer_conf, max_workers=3)
```
k-raina: Should we be using more workers to future proof tests, considering we have limitation regarding low count threads mentioned in comment https://github.com/confluentinc/confluent-kafka-python/pull/2322/changes#r3813055437?

tests/integration/consumer/test_aio_consumer_concurrent_serialization.py
```diff
@@ -0,0 +1,200 @@
...
+        len(loop_results) == n_iterations
+    ), f"expected all {n_iterations} re-entrant loop calls to succeed, got {len(loop_results)}"
+    assert not isinstance(poll_result, BaseException), f"poll_task unexpectedly raised: {poll_result}"
+    assert poll_result is not None
+    assert poll_result.value() == b'hello'
+    assert not isinstance(
+        commit_result, BaseException
+    ), f"expected the independent top-level commit() call to wait and then succeed, got: {commit_result}"
+
+    result = await consumer.assignment()
```
k-raina: Should we verify, if commit actually persisted to broker? Maybe Like

>  partitions = await consumer.assignment()
>  assert partitions
>  positions = await consumer.position(partitions)
>  committed = await consumer.committed(partitions, timeout=10)

k-raina (commented):
Comments on AIOConsumer mainly

src/confluent_kafka/src/Consumer.c
ojasvajain: > should we mention that it would deadlock in document, so its not a surprise if some one faces it?

We will mention - ensure there are adequate number of workers in the pool otherwise it can cause deadlocks. 


>  Should we plan to update default threads?

Yes that's what @emasab suggested today. Will think of what can be the new default value.

ojasvajain: > This behavior might be working for current latest version of python client, as it doesn't support free threading. 

As mentioned earlier, in current version, calling close() in parallel with any other conusmer call causes hangs or segfault because at librdkafka layer, these APIs are not thread safe with each other. By serializing the calls, we want to prioritize correctness of the consumer's behaviour. So with these changes, there will be no segfault or hang, but there will be indefinite wait because we are serializing the calls.

> However with free threading, will above consumer stuck issue might be commonly encountered?

Since it does not work in the current version, we can assume that nobody uses this pattern.

> If we are documenting that consumer will not free threading, that will work for Sync Consumer, however will consumer stuck still be case for AIOConsumer running in NOGIL package?

We will document that consumer is not thread safe and the statement will be applicable for both Sync and Async consumers. If worker thread 1 is calling poll() and after that, worker thread 2 calls close(), then worker thread 2 will wait for poll() to complete. During this time, worker 2 will be blocked and any new call will need a new worker.

tests/integration/consumer/test_aio_consumer_concurrent_serialization.py
ojasvajain: Yes, we will increase the max workers default value (I am still thinking of what should be the right value). Once we do that, we don't need to set max workers here.

tests/integration/consumer/test_aio_consumer_concurrent_serialization.py
ojasvajain: Ack, will check.

ojasvajain: Fixed

tests/integration/consumer/test_aio_consumer_concurrent_serialization.py
ojasvajain: Removed explicit setting of max_workers here as default value is increased

src/confluent_kafka/src/Consumer.c
ojasvajain: Bumped it to 100

src/confluent_kafka/aio/_AIOConsumer.py
```diff
@@ -31,7 +31,7 @@ class AIOConsumer:
     def __init__(
         self,
         consumer_conf: Dict[str, Any],
-        max_workers: int = 2,
+        max_workers: int = 100,
```
k-raina, line 53: Considering that we are spawning high number of threads, threads (and memory) pile up until the process is exhausted. Should we shut them down in close call?

src/confluent_kafka/src/Consumer.c
```diff
@@ -453,21 +590,30 @@ Consumer_commit(Handle *self, PyObject *args, PyObject *kwargs) {
         struct commit_return commit_return;
         PyThreadState *thread_state;
 
+        if (!Handle_gate_enter(self)) {
```
k-raina, line 594: Can we use goto syntax to make this function consistent with others?

src/confluent_kafka/src/Consumer.c
```diff
@@ -1114,38 +1367,47 @@ static PyObject *
 Consumer_consume(Handle *self, PyObject *args, PyObject *kwargs) {
```
k-raina, line 1368: Can we use goto syntax to make this function consistent with others?

src/confluent_kafka/src/Consumer.c
```diff
@@ -38,6 +44,83 @@
...
+        }
+}
+
+/**
+ * @brief Counterpart to Handle_gate_enter(): call once per successful
+ *        Handle_gate_enter(), on every return path.
+ */
+static void Handle_gate_exit(Handle *h) {
+        int depth = atomic_int_dec(&h->u.Consumer.gate_depth);
+        assert(depth >= 0);
```
k-raina, line 120: Will this asset be functional in production?

src/confluent_kafka/src/Consumer.c
k-raina, line 107: Does it makes sense to define these magic numbers as constants?

src/confluent_kafka/aio/_AIOConsumer.py
k-raina, line 30: Considering this class will be bit complex now, with NOGIL, should we add detailed  class docstring?

tests/test_AIOConsumer.py
```diff
@@ -60,21 +62,24 @@ async def test_constructor_invalid_max_workers(self, mock_consumer, mock_common,
 
     @pytest.mark.asyncio
     async def test_call_method_executor_usage(self, mock_consumer, mock_common, basic_config):
-        """Test that _call method properly uses ThreadPoolExecutor for async-to-sync bridging."""
+        """Test that _call method properly uses ThreadPoolExecutor for
+        async-to-sync bridging, setting the identity presented to the
+        Consumer gate on the reentry ContextVar before invoking it."""
+        mock_common.ReentryIdentity.get_or_generate.return_value = 42
         consumer = AIOConsumer(basic_config, max_workers=2)
```
k-raina: Any reason why we are using `max_workers=2` in tests? Can we use default i.e 100 ?

k-raina (changes requested):
Added few comments on code quality and tests

src/confluent_kafka/src/Consumer.c
ojasvajain: I modified 10-12 functions into having goto so that they only have one exit path. But for Consumer_commit() and Consumer_consume(), I found these function to be complex and modifying them to having goto may require good amount of testing to avoid introducing any regression. That's why I decided to skip them.

src/confluent_kafka/src/Consumer.c
ojasvajain: Same as above.

src/confluent_kafka/src/Consumer.c
ojasvajain: No. This is only applicable for debug builds, not production builds.

tests/test_AIOConsumer.py
ojasvajain: Yes, we can remove it.

src/confluent_kafka/aio/_AIOConsumer.py
ojasvajain: Checked this. Currently in AIO Consumer close, we simply call the Consumer_close from C bindings and do nothing else. The threads in the pool will be idle and alive until they get garbage collected so they will remain alive for an indeterministic time. As an improvement we can shutdown the executor (and its threads) immediately after close. We will do this only when the executor is not user-supplied. I am pushing a change for this.

src/confluent_kafka/src/Consumer.c
ojasvajain: I am thinking of creating a function which takes the duration as param and calls Sleep / usleep. I will do that refactoring as part of the admin PR because I need to add sleep for Admin as well. Once I move it in a function, we can store the duration in constant.

tests/test_AIOConsumer.py
ojasvajain: Removed from this and other UTs in this file where setting max workers is not relevant.

src/confluent_kafka/aio/_AIOConsumer.py
ojasvajain: I have added a docstring.

src/confluent_kafka/src/Consumer.c
ojasvajain: Added a TODO for this.

src/confluent_kafka/aio/_AIOConsumer.py
ojasvajain: Done

k-raina (approved):
Thanks for PR! LGTM. 
Please document findings/limitatons that were mentioned in review.

Issue #1696, opened 2023-12-29
please help me! I encountered difficulties using the "AdminClient. list_offsets" method


bcdezxam:
Description
===========
Hello！

I encountered difficulties using the "AdminClient. list_offsets" method, I want to obtain the timestamp corresponding to the offset, 
However, the timestamp obtained using the "result. timestamp" method is a negative timestamp(console: Timestamp: -1251536750), The timestamp of this message should be "1703850479000"

The code is as follows: 

`topic_partition_offsets = dict()`
`topic_partition = TopicPartition(topic, partition)`
`topic_partition_offsets[topic_partition] = OffsetSpec.for_timestamp(1703847600000)`
`future_map = AdminClient({xxx}).list_offsets(topic_partition_offsets, request_timeout=15)`
`for tp, future in future_map.items():`
`    result = future.result()`
`    print(result.timestamp)`

How to reproduce
================

Checklist
=========
Please provide the following information:
 - [x] confluent-kafka-python and librdkafka version: `2.3.0`
 - [x] Apache Kafka broker version: `2.8`
 - [x] Client configuration: `{'bootstrap.servers': xxx, 'security.protocol': 'SASL_PLAINTEXT', 'sasl.mechanism': 'SCRAM-SHA-256', 'sasl.username': xxx, 'sasl.password': xxx}`
 - [x] Operating system: `Windows 11`
 - [x] Provide client logs (with `'debug': '..'` as necessary)
 - [ ] Provide broker log excerpts
 - [x] Critical issue

mwdd146980:
This issue also affects how metric `kafka.broker_offset` appears in Datadog when using the `kafka_consumer` Datadog integration. Datadog has introduced a guardrail in the integration with this PR: https://github.com/DataDog/integrations-core/pull/24814. Negative offsets will be skipped starting in Datadog Agent v7.83.0.

PR #2325, opened 2026-08-10
Fix 64-bit offset/timestamp truncation on Windows in Admin/Uuid marshalling

piochelepiotr:
### What does this PR do?

Fixes a Windows-only bug where `AdminClient.list_offsets()` / `AdminClient.delete_records()` results, and `Uuid.most_significant_bits()` / `Uuid.least_significant_bits()`, silently truncate any value above `2^31 - 1` and can return a negative number.

Root cause: `cfl_PyDict_SetLong()` marshals a value into a Python dict via a parameter typed `long`. On Linux/macOS (LP64), `long` is 64 bits, so this is harmless there. On Windows (LLP64), `long` is always 32 bits, even in a 64-bit build. Every current call site passes a genuinely 64-bit `int64_t` value straight from librdkafka (`rd_kafka_topic_partition_t.offset`, `ListOffsets` timestamps, `rd_kafka_Uuid_t` halves) through this 32-bit-only helper, so on Windows the value wraps modulo `2**32` and gets reinterpreted as signed.

### Fix

Adds `cfl_PyDict_SetLongLong()` (backed by `PyLong_FromLongLong`, correctly 64-bit on every platform) and switches all five `cfl_PyDict_SetLong()` call sites that carry a 64-bit value, plus the two direct `cfl_PyLong_FromLong()` calls in the `Uuid` getters, to the new helper. `cfl_PyDict_SetLong()` itself is left as-is (doc comment added) in case a genuinely 32-bit-safe caller is ever needed; nothing currently calls it for a value that needs to be 64-bit.

### Motivation

A production incident: a Windows host on kafka_consumer (a Datadog Agent integration) reported `broker_offset = -1533701557` where the true high-watermark offset was `2761265739` — exactly `true - 2**32`. `TopicPartition.offset` (a separate, already 64-bit-safe code path using `PyLong_FromLongLong`, see `confluent_kafka.c:513`) reported the correct value in the same run, which is what localized the defect to this one marshalling helper.

This also affects `Uuid.most_significant_bits()`/`least_significant_bits()`, `AdminClient.list_offsets()` timestamps, and `delete_records()`'s `low_watermark` — any consumer of these on Windows can silently receive corrupted data above ~2.1 billion.

Relates to #1696, which reports the same symptom (a negative `ListOffsets` timestamp on Windows) but was never triaged to a root cause.

### Reproduction

The bug only manifests where `sizeof(long) == 4` (Windows LLP64, or any ILP32 platform). It doesn't reproduce on the Linux x64 / macOS CI this project runs functional tests on, which is presumably why it shipped unnoticed — the Windows CI job only builds wheels, it doesn't execute the test suite (`.semaphore/semaphore.yml`, "Wheels: Windows" job).

To reproduce without a Windows machine, the two files below extract the exact `cfl_PyDict_SetLong()` implementation from `confluent_kafka.c` (before) and the new `cfl_PyDict_SetLongLong()` (after) verbatim, feed in the real incident offset (`2761265739`), and run them in a 32-bit-`long` environment (i386 Debian container — same ABI class as Windows x64 for this purpose):

```c
// repro.c -- pre-fix, using the exact original cfl_PyDict_SetLong()
#include <Python.h>
#include <stdint.h>
#include <inttypes.h>

#define cfl_PyLong_FromLong(v) PyLong_FromLong(v)

void cfl_PyDict_SetLong(PyObject *dict, const char *name, long val) {
        PyObject *vo = cfl_PyLong_FromLong(val);
        PyDict_SetItemString(dict, name, vo);
        Py_DECREF(vo);
}

int main(void) {
        Py_Initialize();
        int64_t c_topic_partition_offset = 2761265739LL; /* AGENT-16736 */
        PyObject *kwargs = PyDict_New();
        cfl_PyDict_SetLong(kwargs, "offset", c_topic_partition_offset);
        long observed = PyLong_AsLong(PyDict_GetItemString(kwargs, "offset"));
        printf("sizeof(long)=%zu true=%" PRId64 " observed=%ld\n",
               sizeof(long), c_topic_partition_offset, observed);
        return 0;
}
```

```dockerfile
# Dockerfile.i386
FROM i386/debian:bookworm
RUN apt-get update && apt-get install -y --no-install-recommends gcc python3-dev python3
WORKDIR /repro
COPY repro.c .
RUN gcc -o repro repro.c $(python3-config --cflags --ldflags --embed) -lpython3.11
CMD ["./repro"]
```

```
$ docker build --platform linux/386 -f Dockerfile.i386 -t cfk-repro .
$ docker run --rm --platform linux/386 cfk-repro
sizeof(long)=4 true=2761265739 observed=-1533701557
```

`-1533701557` is an exact match for the incident report. Running the same offset through the patched `cfl_PyDict_SetLongLong()` (backed by `PyLong_FromLongLong`) in the identical i386 container:

```
sizeof(long)=4 sizeof(long long)=8 true=2761265739 observed=2761265739
FIX VERIFIED: offset survives the C->Python boundary intact.
```

I'm attaching a zip with both repro files (`repro.c`, `repro_fixed.c`) and their Dockerfiles so this can be re-run directly. Happy to also open a separate suggestion issue about adding an actual test-execution step to the Windows CI job, since that's the gap that let this ship — didn't want to bundle that into this fix.

### Checklist

- [x] Verified the exact truncation and the fix against the real incident offset, in an environment where `sizeof(long) == 4` (see above)
- [x] Verified no behavior change on `sizeof(long) == 8` (native macOS run, unaffected)
- [x] Updated CHANGELOG.md
- [ ] No new automated regression test — the defect is only observable when `sizeof(long) == 4`, which none of this project's CI targets exercise a test run on (see Reproduction section); a Python-level unit test on the existing Linux/macOS CI would pass identically before and after this fix and wouldn't guard against a regression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ### Fixes
 
+- Fixed `AdminClient.list_offsets()` and `AdminClient.delete_records()` returning a wrapped-around, negative offset/timestamp on Windows for values above 2^31. The 64-bit values from librdkafka were being marshalled into Python through a helper that took a C `long`, which is only 32 bits on Windows (LLP64), truncating the value; `Uuid.most_significant_bits()`/`least_significant_bits()` had the same issue. Fixes Issue [#1696](https://github.com/confluentinc/confluent-kafka-python/issues/1696).
 - Fixed memory leak in `Producer.produce()` when called with headers and raises `BufferError` (queue full) or `RuntimeError` (producer closed). The allocated `rd_headers` memory is now properly freed in error paths before returning. Fixes Issue [#2167](https://github.com/confluentinc/confluent-kafka-python/issues/2167).
 - Fixed type hinting of `KafkaError` class, `Consumer.__init()__`, `Producer.__init()__`, `Producer.produce()` and `Consumer.commit()` and introduced a script in tools directory to keep error codes up to date. Fixes Issue [#2168](https://github.com/confluentinc/confluent-kafka-python/issues/2168).
 
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -4948,9 +4948,10 @@ static PyObject *Admin_c_ListOffsetsResultInfos_to_py(
                         PyObject *args   = NULL;
                         PyObject *kwargs = NULL;
                         kwargs           = PyDict_New();
-                        cfl_PyDict_SetLong(kwargs, "offset",
-                                           c_topic_partition->offset);
-                        cfl_PyDict_SetLong(kwargs, "timestamp", c_timestamp);
+                        cfl_PyDict_SetLongLong(kwargs, "offset",
+                                               c_topic_partition->offset);
+                        cfl_PyDict_SetLongLong(kwargs, "timestamp",
+                                               c_timestamp);
                         cfl_PyDict_SetInt(
                             kwargs, "leader_epoch",
                             rd_kafka_topic_partition_get_leader_epoch(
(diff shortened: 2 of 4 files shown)


pranavrth:
Is this PR ready for review?

PR #2329, opened 2026-08-22
[NOGIL] Improve Test Coverage for Producer + Some Edge Case Fixes

ojasvajain:
1. Add integration test coverage for AIO Producer. This ensures AIO Producer works fine after all the changes done in Producer.c

2. Add more integration test cases related to Transactional Producer (using both sync and async Producer)

3. Addresses a TODO about calling abort_transaction after close has been initiated. A fix has been added to mimic Java client's behaviour by adding a abort_transaction call in close() itself. Ideally, the abort call should be made in the librdkafka layer but for the time being, it is added in the Python C bindings. With this fix in place, an abort_transaction call made after a close has been initiated will get an error (same as Java) but close will ensure it calls abort_transaction (same as Java) so calling abort explicitly will not be required.

4. Addresses a bug that allowed concurrent calls from inside a AIO consumer callback to be made. A lock has been introduced to serialize the concurrent calls. 

----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/confluent_kafka/aio/_common.py
+++ b/src/confluent_kafka/aio/_common.py
@@ -26,6 +26,8 @@ from confluent_kafka import cimpl
 T = TypeVar('T')
 
 
+# TODO NOGIL: Fix for fire and forget tasks created inside callbacks.
+# We don't want such tasks to escape serialization and run concurrently.
 class ReentryContext:
     """Internal use only. Per-call-chain context AIOConsumer carries through
     a ContextVar: the identity presented to the Consumer reentrancy gate, and
--- a/src/confluent_kafka/aio/_common.py
+++ b/src/confluent_kafka/aio/_common.py
@@ -26,10 +26,77 @@ from confluent_kafka import cimpl
 T = TypeVar('T')
 
 
+class Invocation:
+    """Internal use only. State for one callback invocation (e.g. one
+    on_assign firing): the identity it presents to the Consumer gate, the
+    lock serializing sibling calls dispatched from within it, whether the
+    callback has returned yet, and how many dispatches carrying its
+    identity are still in flight on executor workers.
+
+    The record is shared by reference into every context copied from the
+    callback's context (the tasks gather()/create_task() create), so
+    closing it here is visible to all of them -- including a task that
+    outlives the callback and only runs later (see AIOConsumer._call()).
+    """
+
+    __slots__ = ('identity', 'lock', '_alive', '_inflight', '_idle')
+
+    def __init__(self, identity: int) -> None:
+        self.identity = identity
+        self.lock = asyncio.Lock()
+        self._alive = True
+        self._inflight = 0
+        self._idle = asyncio.Event()
+        self._idle.set()
+
+    @property
+    def alive(self) -> bool:
+        """Whether the callback invocation is still open. Only a call made
+        while it is open may present its identity to the gate: that is
+        exactly when the gated call that fired the callback is guaranteed
+        to be parked, waiting for the callback to return."""
+        return self._alive
+
+    def dispatch_started(self) -> None:
+        """Record one gated dispatch carrying this invocation's identity.
+        Event-loop thread only, under self.lock, while alive."""
+        self._inflight += 1
+        self._idle.clear()
+
+    def dispatch_finished(self) -> None:
+        """Counterpart to dispatch_started(), called once the dispatch has
+        fully left the gate (or was cancelled before it ever started).
+        Event-loop thread only -- marshal via call_soon_threadsafe from
+        other threads."""
+        self._inflight -= 1
+        if self._inflight == 0:
+            self._idle.set()
+
+    async def close(self) -> None:
+        """Close the invocation and wait until no dispatch carrying its
+        identity is still inside the gate.
+
+        The gated call that fired the callback resumes as soon as this
+        returns, so it must not return while a same-identity dispatch could
+        still be executing -- even if the surrounding task is cancelled.
+        Cancellation is therefore re-raised only after the drain completes.
+        """
+        self._alive = False
+        cancelled = False
+        while self._inflight:
+            try:
+                await self._idle.wait()
+            except asyncio.CancelledError:
+                cancelled = True
+        if cancelled:
+            raise asyncio.CancelledError
+
+
 class ReentryContext:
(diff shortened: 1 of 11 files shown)


src/confluent_kafka/aio/_common.py
```diff
@@ -25,9 +26,10 @@
 T = TypeVar('T')
 
 
-class ReentryIdentity:
-    """Internal use only. The identity AIOConsumer presents to the Consumer
-    reentrancy gate.
+class ReentryContext:
```
k-raina, line 31: Two concurrent subscribe(), which double-frees callback references will be fixed by this fix but remains is the mutating call racing the application poll that is still running, still needs to be fixed?

eg. 
```
async def on_assign(consumer, partitions):
    asyncio.create_task(consumer.subscribe(...),)   
await consumer.subscribe(["topic"], on_assign=on_assign)
await consumer.poll(10)   
```

**Application workflow** 
1. poll() runs on executor worker W1,  holding the client handle rk.
2. librdkafka fires on_assign from inside that poll. W1 is now parked waiting for the callback to finish but its C-level poll is suspended mid-execution, still holding rk.
3. The callback starts consumer.subscribe(). Because asyncio copies the callback's context into the child task, the child also presents same identity.
4.. W2 runs Consumer_subscribe which runs/mutates on rk handle.
6. W1's suspended poll resumes and touches the mutated rk and crashs.

Can we check if this workflow should be fixed?

src/confluent_kafka/src/Producer.c
```diff
@@ -625,6 +628,22 @@ Producer_close(Handle *self, PyObject *args, PyObject *kwargs) {
...
         CallState_begin(self, &cs);
 
+        /* If a transaction is in progress, abort it now rather than leaving it
+         * dangling on the broker. Ideally, rd_kafka_destroy() should do it on
+         * its own.
+         * _ERR__STATE (no transaction in progress) and _ERR__NOT_CONFIGURED
+         * (non-transactional producer) are the expected outcome for most
+         * producers and are ignored; anything else is stashed to warn about
+         * below. */
+        txn_error = rd_kafka_abort_transaction(self->rk, -1);
```
k-raina, line 638: -1 means an infinite deadline. librdkafka's abort ends with an EndTxn round-trip to the coordinator, which waits on that infinite deadline which blocks forever. Should we add resonable timeout?

src/confluent_kafka/aio/producer/_AIOProducer.py
```diff
@@ -92,9 +92,9 @@ async def close(self) -> None:
         2. **Cancel Timeout Task**: Immediately cancels the buffer timeout monitoring task
         3. **Flush All Messages**: Flushes any buffered messages and waits for delivery confirmation
         4. **Shutdown ThreadPool**: Waits for all pending ThreadPool operations to complete
-        5. **Cleanup**: Ensures the underlying librdkafka producer is properly closed. The shutdown
-            is designed to be safe and non-blocking for the asyncio event loop
-            while ensuring all pending operations complete before the producer is closed.
+        5. **Cleanup** [Not implemented yet]: Ensures the underlying librdkafka producer is properly closed.
```
k-raina, line 95: Should we call  Producer.close() in AIOProducer.close()?

k-raina (changes requested):
Couple of comments on correctness

src/confluent_kafka/aio/producer/_AIOProducer.py
ojasvajain: Yes. It is a pre-existing bug that I am not planning to fix in this PR as it will require more testing.

src/confluent_kafka/src/Producer.c
ojasvajain: In Java, when we call producer.close() without timeout, it blocks indefinitely on aborting the tx (if broker is not reachable), which is the what above change also does. But Java does provide an option to set timeout for close (which librdkafka does not support). Do you think setting 60s of timeout would be fine?

src/confluent_kafka/aio/_common.py
ojasvajain: Yes this will break because task is outliving the callback. This can also happen in sync consumers if user starts some threads from within the callback. We should document that starting threads/tasks from within a callback is not recommended. I am not saying this can't be fixed but it can be taken as a future goal to fix maybe, if a requirement comes. In the current version of python client as well, the issue persists because there is no serialization at all so we are not breaking something that was supported. It's just that we are not supporting the serialization of consumer calls in all the cases, which should be fine.

src/confluent_kafka/src/Producer.c
ojasvajain: To check with @emasab

src/confluent_kafka/aio/_common.py
k-raina: Thanks. 
Lets verify this is issue with GIL mode also. If no, we might need to fix this in upcoming PRs.

src/confluent_kafka/aio/producer/_AIOProducer.py
k-raina: Since rk is leaking after application is closed, seems like it will be one time leak so severity is low. We should check in future PRs if this can be fixed and free threading is compatible for fix.

k-raina (approved):
Thanks for PR @ojasvajain

src/confluent_kafka/src/Producer.c
ojasvajain: On Emaneule's suggestion, I double checked the Java behaviour - the abort_tx call inside close waits indefinitely if broker is not reachable. So it should be fine to keep -1 as timeout here.

src/confluent_kafka/aio/_common.py
ojasvajain: I pushed a fix for this. I have now introduced a new field called Invocation that will be set in context and instead of its value, its reference will be passed on to the child tasks. So when a callback is complete, it can update this object and then while executing the child tasks (who will also have the value updated), we can check if we want to treat it as a re-entrant call or a separate top-level call.

ojasvajain: Have reverted the fix based on Emanuele's suggestion. It will be part of a separate PR later.

PR #2331, opened 2026-08-22
Fix KafkaError error strings raising/garbling on non-UTF-8 locales (#448)

### LINKED ISSUE #448: KafkaError error strings are unicode on Windows when language is non-english
Which messes up KafkaError__str__, assuming it has an ascii7 string.

iSparshP:
Fixes #448.

## Problem
`KafkaError.str()` decodes the underlying C error string with `PyUnicode_FromString()`, which assumes strict UTF-8. When librdkafka or the OS returns an error string in the system locale encoding rather than UTF-8 — e.g. on a non-English Windows install — this raises `UnicodeDecodeError` or produces garbled output, as reported in #448.

## Fix
Add a small helper `cfl_PyUnistr_FromStringSafe()` that decodes via `PyUnicode_DecodeUTF8(s, len, "replace")`. Invalid bytes degrade to U+FFFD (`�`) instead of raising, so error reporting never blows up on a locale-encoded string. Both the custom-message and `rd_kafka_err2str()` paths in `KafkaError_str()` now use it.

This is intentionally the minimal, non-raising fix: it guarantees a usable string in all locales rather than attempting full locale-encoding detection.

## Test
Added `test_kafka_error_non_ascii_str` covering a non-ASCII (Cyrillic) error message. Full `tests/test_error.py` passes locally (7 passed) against a fresh build.

### DIFF
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -96,9 +96,9 @@ static PyObject *KafkaError_code(KafkaError *self, PyObject *ignore) {
 
 static PyObject *KafkaError_str(KafkaError *self, PyObject *ignore) {
         if (self->str)
-                return cfl_PyUnistr(_FromString(self->str));
+                return cfl_PyUnistr_FromStringSafe(self->str);
         else
-                return cfl_PyUnistr(_FromString(rd_kafka_err2str(self->code)));
+                return cfl_PyUnistr_FromStringSafe(rd_kafka_err2str(self->code));
 }
 
 static PyObject *KafkaError_name(KafkaError *self, PyObject *ignore) {
@@ -136,7 +136,7 @@ static PyObject *KafkaError_reduce(KafkaError *self,
             cfl_PyObject_lookup("confluent_kafka.cimpl", "KafkaError");
 
         if (self->str) {
-                reason = cfl_PyUnistr(_FromString(self->str));
+                reason = cfl_PyUnistr_FromStringSafe(self->str);
         } else {
                 reason = NULL;
         }
--- a/src/confluent_kafka/src/confluent_kafka.h
+++ b/src/confluent_kafka/src/confluent_kafka.h
@@ -175,6 +175,20 @@ static __inline const char *cfl_PyUnistr_AsUTF8(PyObject *o, PyObject **uobjp) {
 #endif
 
 
+/**
+ * @returns a Unicode object from a NUL-terminated C string, decoding it as
+ *          UTF-8 with the "replace" error handler.
+ *
+ * Unlike PyUnicode_FromString(), which assumes strict UTF-8 and raises
+ * UnicodeDecodeError on invalid input, this never raises. Use it for strings
+ * that may come from librdkafka or the OS in the system locale encoding rather
+ * than UTF-8 (e.g. error strings on non-English Windows). See issue #448.
+ */
+static __inline PyObject *cfl_PyUnistr_FromStringSafe(const char *s) {
+        return PyUnicode_DecodeUTF8(s, (Py_ssize_t)strlen(s), "replace");
+}
+
+
 /****************************************************************************
  *
  *
--- a/tests/test_error.py
+++ b/tests/test_error.py
@@ -64,3 +64,16 @@ def test_new_produce_error_custom_message():
     assert pe.code == KafkaError._KEY_SERIALIZATION
     assert pe.name == u'_KEY_SERIALIZATION'
     assert pe.args[0].str() == "Unable to serialize key"
+
+
+def test_kafka_error_non_ascii_str():
+    # Covers the valid-UTF-8 non-ASCII path only: that the "replace" decode
+    # doesn't mangle legitimate non-English text. The public constructor can't
+    # inject invalid bytes, so the actual #448 invalid-byte case (reachable only
+    # from C or a genuinely non-UTF-8 locale) is not exercised here.
+    msg = u"Ошибка: недопустимое значение"
+    err = KafkaError(KafkaError._KEY_SERIALIZATION, msg)
+
+    assert err.str() == msg
+    assert msg in str(err)
+    assert repr(err)


prashah-confluent:
Hi @iSparshP,

Thanks for digging into this.

Nice diagnosis — the strict-vs-`replace` distinction is exactly the right root cause.

One gap: `KafkaError_reduce()` decodes the same `self->str` bytes through the strict
path, so it still raises. See `src/confluent_kafka/src/confluent_kafka.c:139`.

`__reduce__` is what `pickle` calls, and pickling happens implicitly whenever a
`KafkaError` crosses a process boundary or is `copy.deepcopy()`d — serialization
support was added deliberately in #2153. I verified this empirically against the
2.15.0 wheel by simulating the #448 condition (injecting a 0xE5 byte — CP1252 'å',
invalid UTF-8 — into a live error's `self->str` via ctypes):

- `repr(err)` / `print(err)` → work, render `v�rden` (already "replace" via
  `PyUnicode_FromFormat`)
- `err.str()` → `UnicodeDecodeError` (the #448 crash this PR fixes)
- `pickle.dumps(err)` / `copy.deepcopy(err)` → `SystemError: <built-in method
  __reduce__ ...> returned a result with an error set`, with the
  `UnicodeDecodeError` as `__cause__` (the unchecked NULL sends the function down
  the no-reason branch with an exception still pending, which CPython flags as a
  `SystemError`)
- a `KafkaError` returned from a `ProcessPoolExecutor` worker → the parent's
  `future.result()` raises that same `SystemError` (the result is pickled in the
  child on its way back), so the caller never receives the error object at all

Same one-line substitution as lines 99/101:

```c
        if (self->str) {
                reason = cfl_PyUnistr_FromStringSafe(self->str);
        } else {
                reason = NULL;
        }
```

This makes the decode non-raising, which also moots the unchecked-result issue for
the encoding case. Without it, #448 isn't fully closed — the same bytes still crash,
just from the pickling path instead of the display path.

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -175,6 +175,20 @@ static __inline const char *cfl_PyUnistr_AsUTF8(PyObject *o, PyObject **uobjp) {
...
+/**
+ * @returns a Unicode object from a NUL-terminated C string, decoding it as
+ *          UTF-8 with the "replace" error handler.
+ *
+ * Unlike PyUnicode_FromString(), which assumes strict UTF-8 and raises
+ * UnicodeDecodeError on invalid input, this never raises. Use it for strings
+ * that may come from librdkafka or the OS in the system locale encoding rather
+ * than UTF-8 (e.g. error strings on non-English Windows). See issue #448.
+ */
+static __inline PyObject *cfl_PyUnistr_FromStringSafe (const char *s) {
```
prashah-confluent: Nit: `cfl_PyUnistr_FromStringSafe (const char *s)` has a space before the paren - the repo's `.clang-format` has `SpaceBeforeParens: ControlStatements` (only
`if`/`while`/etc. get the space), and the neighboring `cfl_PyUnistr_AsUTF8(...)`
definitions follow that. clang-format would flag it.

prashah-confluent (changes requested):
Thanks @iSparshP 

Have few comments which could ensure that all the places are covered and the coding style guidelines are honoured.

tests/test_error.py
```diff
@@ -64,3 +64,12 @@ def test_new_produce_error_custom_message():
     assert pe.code == KafkaError._KEY_SERIALIZATION
     assert pe.name == u'_KEY_SERIALIZATION'
     assert pe.args[0].str() == "Unable to serialize key"
+
+
+def test_kafka_error_non_ascii_str():
```
prashah-confluent, line 69: The Cyrillic message is passed in from Python,
so `PyArg_ParseTupleAndKeywords` guarantees it arrives as valid UTF-8 - which means
it decodes fine under the *old* strict code too. I confirmed this by running the
PR's `tests/test_error.py` via pytest against the pre-fix 2.15.0 wheel from PyPI:
all 7 tests pass, including this one. So it doesn't exercise the #448 failure —
CI would go green with or without the fix.

That isn't really fixable from Python - the public constructor can't inject invalid
bytes, so the failing case is only reachable from C or on a genuinely non-UTF-8
locale. Two small asks rather than a code change:

- The commit message calls it a "regression test"; "non-ASCII round-trip test" would
  be more accurate.
- A one-line comment on the test noting it covers the valid-UTF-8 non-ASCII path
  only (i.e. that `"replace"` doesn't mangle legitimate non-English text), plus a
  note in the PR describing how the invalid-byte case was verified.

Mainly so nobody later assumes CI covers the invalid-byte path when it doesn't.

prashah-confluent (changes requested):

iSparshP:
Thanks for the thorough review @prashah-confluent, all three addressed in the amended commit:

1. **`__reduce__` gap** : good catch, that was the real hole. `KafkaError_reduce()` now decodes `self->str` through `cfl_PyUnistr_FromStringSafe()` too, so pickling / `copy.deepcopy()` (and anything crossing a process boundary, e.g. a `ProcessPoolExecutor` result) no longer raises on the same bytes. Verified locally: `pickle.loads(pickle.dumps(err))` and `copy.deepcopy(err)` both round-trip a non-ASCII message correctly, and routing the reduce path through the non-raising helper resolves the `SystemError` from the unchecked-NULL branch for the encoding case.

2. **clang-format nit**:  removed the space in `cfl_PyUnistr_FromStringSafe(const char *s)` to match `SpaceBeforeParens: ControlStatements` and the neighboring `cfl_PyUnistr_AsUTF8` definitions.

3. **Test scope**:  agreed the public constructor cannot inject invalid bytes, so this only exercises the valid-UTF-8 non-ASCII path. Reworded the commit from "regression test" to "non-ASCII round-trip test" and added a comment on the test spelling out that the invalid-byte case (reachable only from C or a genuinely non-UTF-8 locale) is not covered by CI. The invalid-byte crash itself was confirmed via your `ctypes` 0xE5 injection against the 2.15.0 wheel; both `str()` and now the reduce/pickle path route those bytes through the `"replace"` decode instead of raising.

`tests/test_error.py` passes (7 passed) against a fresh build. Also signing the CLA.

iSparshP:
Heads up on the CLA: the signing link from @confluent-cla-assistant (`https://ironcladapp.com/public-launch/66bcfa630dec9c9aeb2fc0da`) just renders an **"Error Occurred"** box and never loads the agreement.

I ruled out my setup, reproduced it in a clean Incognito window with all extensions disabled, and on a second browser, so it looks like an Ironclad-side / workflow issue rather than anything local. Re-running the check via an empty commit didn't help since the link is a fixed workflow ID.

Could a maintainer regenerate the CLA workflow or point me to an alternate way to sign? Happy to push an empty commit to re-run the check as soon as it's working.

prashah-confluent:
@iSparshP , can you try pushing an empty commit 

`git commit --allow-empty -m "retrigger cla check"`

prashah-confluent (dismissed):
Thanks @iSparshP for the fixes, we will be taking care of the `cla-status-check` issue and will merge the PR after one final review.

prashah-confluent:
@iSparshP , the team has fixed the cla link. Can you give it a try? If required do an empty commit as well?

src/confluent_kafka/src/confluent_kafka.c
```diff
@@ -136,7 +136,7 @@ static PyObject *KafkaError_reduce(KafkaError *self,
             cfl_PyObject_lookup("confluent_kafka.cimpl", "KafkaError");
 
         if (self->str) {
-                reason = cfl_PyUnistr(_FromString(self->str));
+                reason = cfl_PyUnistr_FromStringSafe(self->str);
```
k-raina, line 139: Can we add test case for pickle/__reduce__()?

k-raina (changes requested):
Thanks for PR! On comment

k-raina (approved):
Thanks @iSparshP for raising PR for fix. Thanks @prashah-confluent for addressing review comment

src/confluent_kafka/src/confluent_kafka.c
prashah-confluent: Added the unit test for this scenario

prashah-confluent (approved):

PR #2334, opened 2026-08-27
[NOGIL] Fix borrowed-reference/stale-count races

### LINKED ISSUE #2319: Free-threaded (no-GIL) use-after-free in Admin_create_topics: borrowed NewTopic from PyList_GET_ITEM dereferenced as a C struct against a stale count
Heads-up ahead of any free-threading support: `AdminClient.create_topics` (and the same-shaped `create_partitions` / topic-name paths) take a borrowed list item, cast it to a C struct, and dereference its fields while another thread can free it.

**This is latent today, not live.** The extension does not declare `Py_MOD_GIL_NOT_USED`, so a free-threaded interpreter **re-enables the GIL** on import and the code is safe as shipped. It becomes reachable only when the GIL is forced off (`PYTHON_GIL=0` / `-Xgil=0`), i.e. what will happen the day the module opts into free-threading. This is a "fix before advertising free-threading" report, not a "you are currently broken" one.

### Site

`src/confluent_kafka/src/Admin.c:609` on current `master`:

```c
tcnt = (int)PyList_Size(topics);                       // captured once (line 576)
c_objs = malloc(sizeof(*c_objs) * tcnt);
for (i = 0; i < tcnt; i++) {
    NewTopic *newt = (NewTopic *)PyList_GET_ITEM(topics, i);   // borrowed, no incref
    r = PyObject_IsInstance((PyObject *)newt, ...);            // reads newt->ob_type
    c_objs[i] = rd_kafka_NewTopic_new(
        newt->topic, newt->num_partitions, newt->replication_factor, ...); // 623-624
}
```

`newt` is borrowed, never increfed; it is cast to `(NewTopic *)` and its `char *topic` and two `int` fields are read straight out of the object. Two failure modes under `Py_GIL_DISABLED`, both driven by another thread mutating `topics`:

1. **stale `tcnt`** - the length is read once. If the list shrinks, `PyList_GET_ITEM(topics, i)` indexes past `ob_item` and the out-of-bounds slot is dereferenced as a `NewTopic`.
2. **freed element** - if element `i` is replaced, the old `NewTopic` drops to zero references and is freed; `newt` still points into freed memory and `newt->topic` is handed to librdkafka.

The whole parse-and-convert loop runs synchronously before `rd_kafka_CreateTopics` (line 680), so **no broker is required** - the `rd_kafka_t` handle is created at `AdminClient` construction and a bogus `bootstrap.servers` is enough.

The same borrow-then-dereference shape is at `Admin.c:864` (`create_partitions`, `NewPartitions *`) and the topic-name loops at 751 / 2513; I reproduced only `create_topics` and list the others as the same shape, not separately verified.

### Measured

confluent-kafka `2.15.0` built from source against librdkafka on a `python3.14.0rc1t` venv. 6 caller threads call `create_topics` on a bogus AdminClient while 3 threads mutate the list; 10 rounds per arm:

| arm | result |
|---|---|
| mutate, GIL off (`PYTHON_GIL=0`) | **10/10 SIGSEGV** |
| control - no mutator thread | clean 10/10 |
| control - mutate a different list `create_topics` never reads | clean 10/10 |
| control - same mutation, GIL left on | clean 10/10 |

The GIL-on control is the one that matters most: it is the current shipped behaviour, and it is clean - that is the whole "latent" point in one line.

<details><summary>Reproducer (ft_ckafka.py)</summary>

```python
"""confluent-kafka `Admin.c:609` — a borrowed NewTopic dereferenced as a C struct.

The exact shape, from confluent-kafka 2.15.0 `src/confluent_kafka/src/Admin.c`:

    tcnt = (int)PyList_Size(topics);            // 576, captured once
    c_objs = malloc(sizeof(*c_objs) * tcnt);    // 607
    for (i = 0; i < tcnt; i++) {
        NewTopic *newt = (NewTopic *)PyList_GET_ITEM(topics, i);   // 609 — BORROWED
        r = PyObject_IsInstance((PyObject *)newt, ...);            // reads newt->ob_type
        ...
        c_objs[i] = rd_kafka_NewTopic_new(
            newt->topic, newt->num_partitions, newt->replication_factor, ...);  // 623-624
                     // ^ char* and two ints read straight out of the C struct

`newt` is never increfed. Two ways this is a use-after-free under
`Py_GIL_DISABLED`, both exercised here by a second thread mutating `topics`
while `create_topics` runs:

  (A) stale tcnt.  tcnt is read once. If the list shrinks before the loop reaches
      index i, PyList_GET_ITEM(topics, i) indexes past ob_item — an OOB read of a
      freed slot — and the garbage is dereferenced as `newt->ob_type` /
      `newt->topic`.

  (B) freed element.  If the list keeps its length but element i is replaced, the
      old NewTopic loses its last reference and is freed. `newt` still points at
      it, and `newt->topic` (a `char *` inside freed memory) is handed to
      librdkafka.

None of this needs a broker. AdminClient creates its `rd_kafka_t` handle at
construction without connecting; the whole parse-and-convert loop runs
synchronously in the calling thread, before `rd_kafka_CreateTopics` is called at
line 680. A bogus bootstrap.servers is enough.

Run under the free-threaded build; a SIGSEGV / bus error / malloc abort is the
positive result. Compare against `ctrl_*` — same code, one property removed each.
"""
import os
import sys
import threading
import time

from confluent_kafka.admin import AdminClient, NewTopic

# a bogus broker: the handle is created, nothing connects
ADMIN_CONF = {"bootstrap.servers": "127.0.0.1:9", "socket.timeout.ms": 10}

N_CALLERS = int(os.environ.get("N_CALLERS", 6))
N_MUTATORS = int(os.environ.get("N_MUTATORS", 3))
LIST_LEN = int(os.environ.get("LIST_LEN", 64))
SECONDS = float(os.environ.get("SECONDS", 8))
# mutate=True  -> the concurrent hazard (this file)
# mutate=False -> ctrl_A_no_mutator
MUTATE = os.environ.get("MUTATE", "1") == "1"
# decoy=True   -> mutate a *different* list create_topics never reads
DECOY = os.environ.get("DECOY", "0") == "1"


def fresh_topics(n):
    return [NewTopic(f"t{i}", num_partitions=1, replication_factor=1)
            for i in range(n)]


def main():
    gil = getattr(sys, "_is_gil_enabled", lambda: True)()
    print(f"py={sys.version.split()[0]} gil={gil} "
          f"callers={N_CALLERS} mutators={N_MUTATORS if MUTATE else 0} "
          f"len={LIST_LEN} decoy={DECOY}", flush=True)

    admin = AdminClient(ADMIN_CONF)
    topics = fresh_topics(LIST_LEN)
    decoy = fresh_topics(LIST_LEN)
    stop = threading.Event()
    start = threading.Barrier(N_CALLERS + (N_MUTATORS if MUTATE else 0) + 1)
    calls = [0] * N_CALLERS

    def caller(t):
        start.wait()
        n = 0
        while not stop.is_set():
            try:
                admin.create_topics(topics, request_timeout=0.001)
            except Exception:
                pass          # broker errors, validation errors — irrelevant
            n += 1
        calls[t] = n

    def mutator():
        start.wait()
        target = decoy if DECOY else topics
        i = 0
        while not stop.is_set():
            # (B) replace an element: old NewTopic drops to zero refs
            target[i % len(target)] = NewTopic(
                f"m{i}", num_partitions=1, replication_factor=1)
            # (A) shrink then regrow: makes the captured tcnt stale mid-loop
            if i % 7 == 0:
                del target[:]
                target.extend(fresh_topics(LIST_LEN))
            i += 1

    threads = [threading.Thread(target=caller, args=(t,))
               for t in range(N_CALLERS)]
    if MUTATE:
        threads += [threading.Thread(target=mutator) for _ in range(N_MUTATORS)]

    for th in threads:
        th.start()
    start.wait()
    time.sleep(SECONDS)
    stop.set()
    for th in threads:
        th.join()

    print(f"clean: {sum(calls)} create_topics calls survived", flush=True)
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

Run with `PYTHON_GIL=0`. The controls are env-var variants of the same script: `MUTATE=0` (no mutator), `DECOY=1` (mutate a different list), and running without `PYTHON_GIL=0` (GIL on).
</details>

### A possible fix

When free-threading is on the roadmap: incref `newt` for the body of the loop (or hold the list under a critical section), and re-read the list length rather than trusting the captured `tcnt`.

### Not claimed

No severity - reachable only with the GIL disabled, which is not the default for this module today. No exploitability - the observed faults are consistent with the mechanism; I did not build a primitive.

ojasvajain:
1. Fixes #2319 and related issues across Producer/Consumer/Admin clients.
2. Adds unit and integration test cases to validate the fixes.

What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -434,9 +434,7 @@ static int Admin_incremental_config_to_c(PyObject *incremental_configs,
         char *name                      = NULL;
         char *value                     = NULL;
         PyObject *incremental_operation = NULL;
-#ifdef Py_GIL_DISABLED
-        PyObject *owned_configs = NULL;
-#endif
+        PyObject *owned_configs         = NULL;
 
         if (!PyList_Check(incremental_configs)) {
                 PyErr_Format(PyExc_TypeError,
@@ -445,12 +443,16 @@ static int Admin_incremental_config_to_c(PyObject *incremental_configs,
                 goto err;
         }
 
-#ifdef Py_GIL_DISABLED
+        /* Snapshot the list before iterating it even on GIL-based builds.
+         * Fetching ConfigEntry.incremental_operation's .value below re-enters
+         * the interpreter, which is a legal GIL yield point. Without this
+         * snapshot, a concurrent mutation of the caller's list can cause
+         * PyList_GET_ITEM(incremental_configs, i) calls to read out of bounds.
+         */
         owned_configs = PyList_GetSlice(incremental_configs, 0, PY_SSIZE_T_MAX);
         if (!owned_configs)
                 goto err;
         incremental_configs = owned_configs;
-#endif
 
         if ((config_entry_count = (int)PyList_Size(incremental_configs)) < 1) {
                 PyErr_Format(PyExc_ValueError,
@@ -518,9 +520,7 @@ static int Admin_incremental_config_to_c(PyObject *incremental_configs,
                 value                 = NULL;
                 incremental_operation = NULL;
         }
-#ifdef Py_GIL_DISABLED
         Py_XDECREF(owned_configs);
-#endif
         return 1;
 err:
         Py_XDECREF(incremental_operation);
@@ -528,9 +528,7 @@ err:
                 free(name);
         if (value)
                 free(value);
-#ifdef Py_GIL_DISABLED
         Py_XDECREF(owned_configs);
-#endif
         return 0;
 }
 
@@ -1202,9 +1200,7 @@ static PyObject *Admin_incremental_alter_configs(Handle *self,
                                                  PyObject *kwargs) {
         PyObject *resources, *future;
         PyObject *validate_only_obj = NULL;
-#ifdef Py_GIL_DISABLED
-        PyObject *owned_resources = NULL;
-#endif
+        PyObject *owned_resources   = NULL;
         static char *kws[] = {"resources", "future",
                               /* options */
                               "validate_only", "request_timeout", "broker",
@@ -1247,13 +1243,17 @@ static PyObject *Admin_incremental_alter_configs(Handle *self,
         if (!c_options)
                 goto done; /* Exception raised by options_to_c() */
 
-#ifdef Py_GIL_DISABLED
+        /* Snapshot the list before iterating it even on GIL-based builds.
+         * Fetching ConfigEntry.incremental_operation's .value below re-enters
+         * the interpreter, which is a legal GIL yield point. Without this
+         * snapshot, a concurrent mutation of the caller's list can cause
+         * PyList_GET_ITEM(incremental_configs, i) calls to read out of bounds.
+         */
         owned_resources = PyList_GetSlice(resources, 0, PY_SSIZE_T_MAX);
         if (!owned_resources)
                 goto done;
         resources = owned_resources;
         cnt       = (int)PyList_Size(resources);
-#endif
 
         /* Look up the ConfigResource class so we can check if the provided
          * topics are of correct type.
@@ -1359,9 +1359,7 @@ done:
         free(c_objs);
         if (rkqu)
                 rd_kafka_queue_destroy(rkqu); /* drop ref from get_background */
-#ifdef Py_GIL_DISABLED
         Py_XDECREF(owned_resources);
-#endif
         Py_XDECREF(ConfigResource_type);      /* from lookup() */
         Py_XDECREF(ConfigEntry_type);         /* from lookup() */
         /* Release our extra ref only on failure; on success the opaque keeps
--- a/src/confluent_kafka/src/Admin.c
+++ b/src/confluent_kafka/src/Admin.c
@@ -568,8 +568,10 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
         int tcnt;
         int i;
         int topic_partition_count;
-        rd_kafka_NewTopic_t **c_objs;
-        rd_kafka_queue_t *rkqu;
+        rd_kafka_NewTopic_t **c_objs = NULL;
+        rd_kafka_queue_t *rkqu       = NULL;
+        PyObject *result             = NULL;
+        int future_incremented       = 0;
         CallState cs;
 
         /* topics is a list of NewTopic objects. */
@@ -595,15 +597,14 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
 
         c_options = Admin_options_to_c(self, RD_KAFKA_ADMIN_OP_CREATETOPICS,
                                        &options, future);
-        if (!c_options) {
-                Handle_exit_rk_use(self);
-                return NULL; /* Exception raised by options_to_c() */
-        }
+        if (!c_options)
+                goto done; /* Exception raised by options_to_c() */
 
         /* options_to_c() sets future as the opaque, which is used in the
          * background_event_cb to set the results on the future as the
          * admin operation is finished, so we need to keep our own refcount. */
         Py_INCREF(future);
+        future_incremented = 1;
 
         /*
          * Parse the list of NewTopics and convert to corresponding C types.
@@ -618,11 +619,11 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
                 r = PyObject_IsInstance((PyObject *)newt,
                                         (PyObject *)&NewTopicType);
                 if (r == -1)
-                        goto err; /* Exception raised by IsInstance() */
+                        goto done; /* Exception raised by IsInstance() */
                 else if (r == 0) {
                         PyErr_SetString(PyExc_ValueError,
                                         "Expected list of NewTopic objects");
-                        goto err;
+                        goto done;
                 }
 
                 c_objs[i] = rd_kafka_NewTopic_new(
@@ -632,7 +633,7 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
                         PyErr_Format(PyExc_ValueError,
                                      "Invalid NewTopic(%s): %s", newt->topic,
                                      errstr);
-                        goto err;
+                        goto done;
                 }
 
                 if (newt->replica_assignment) {
@@ -642,7 +643,7 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
                                                 "replica_assignment are "
                                                 "mutually exclusive");
                                 i++;
-                                goto err;
+                                goto done;
                         }
 
                         if (newt->num_partitions == -1) {
@@ -656,7 +657,7 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
                                 newt->replica_assignment, topic_partition_count,
                                 topic_partition_count, "num_partitions")) {
                                 i++;
-                                goto err;
+                                goto done;
                         }
                 }
 
@@ -665,7 +666,7 @@ Admin_create_topics(Handle *self, PyObject *args, PyObject *kwargs) {
                                                     newt->config,
                                                     "newtopic_set_config")) {
                                 i++;
-                                goto err;
+                                goto done;
                         }
                 }
         }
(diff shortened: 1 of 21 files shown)


ojasvajain:
@espressolee Feel free to run your harness against 980398c04e2c97bc471585fb0e92b475db31139c

espressolee:
Tested exact head `980398c04e2c97bc471585fb0e92b475db31139c` against exact base `5ccf0ab2bf16f990525f8bb286b995b04b5ba9ae`.

Environment: macOS/arm64, CPython 3.14.0rc1t, `sys._is_gil_enabled() == False`, C extension built from each exact tree, Homebrew librdkafka 2.15.0. The fake broker is unreachable; the reported parse path is synchronous and does not require a broker.

| arm | result |
|---|---:|
| base, constant-length `topics[i]` replacement | 10/10 SIGSEGV |
| base, shrink/regrow stale-count race | 10/10 SIGSEGV |
| base, same replacement with runtime GIL enabled | 0/5 faults |
| candidate, constant-length replacement | 0/20 faults |
| candidate, shrink/regrow | 0/20 faults |
| candidate, no mutator | 0/5 faults |
| candidate, decoy-list mutation | 0/5 faults |
| candidate, runtime GIL enabled | 0/5 faults |

There were no timeouts or other nonzero exits in the matrix. The constant-length arm isolates the borrowed-element lifetime window independently of the captured count; the shrink/regrow arm isolates the stale-count window.

I also ran the new broker-free memory-safety file on the candidate:

```
PYTHON_GIL=0 python -m pytest -q tests/concurrency/test_memory_safety.py -vv
21 passed in 120.44s
```

Bounded conclusion: on this measured host, exact head `980398c` closes both #2319 mechanisms. I did not test other platforms or allocators, and I did not run the broker-backed integration tests. Semaphore CI is green; the current SonarQube failure is the separate new-code coverage gate.

ojasvajain:
Thanks.

> base, constant-length `topics[i]` replacement	10/10 SIGSEGV
> base, shrink/regrow stale-count race	10/10 SIGSEGV

Did you observe failures again or am I interpreting the results incorrectly?

espressolee:
No new failures were observed on the candidate — those two `10/10 SIGSEGV` rows are the exact base (`5ccf0ab2bf16f990525f8bb286b995b04b5ba9ae`) positive controls, included to confirm that the harness still reproduces the original bugs.

For exact candidate `980398c04e2c97bc471585fb0e92b475db31139c`:

- constant-length replacement: **0/20 faults**
- shrink/regrow: **0/20 faults**
- candidate memory-safety test file: **21/21 passed**

So on this host: **base fails; candidate passes both arms**. I could have made that split more prominent in the table.

espressolee:
Refreshed the focused validation after the bot-updated PR head.

Exact pair now checked: base `74860f00ac2ca9b63cf814099032fd179609452f`, head `7db086ee8abfdc521737ffb3244dd1dc37b614c7`.

The four changed C source files and `tests/concurrency/test_memory_safety.py` relevant to #2319 are blob-identical between the previously tested `980398c` and current `7db086e`; the previous/current base C blobs are likewise identical. I nevertheless rebuilt the extension from exact `7db086e` and reran the two candidate arms on the same macOS/arm64 CPython 3.14.0rc1t host with the GIL disabled and librdkafka 2.15.0:

- constant-length replacement: **20/20 clean**
- shrink/regrow: **20/20 clean**
- signals, timeouts, other nonzero exits: **0**
- `PYTHON_GIL=0 python -m pytest -q tests/concurrency/test_memory_safety.py -vv`: **21 passed, 1 warning in 119.97s**

No new candidate failure was observed on this host. I did not rerun the base positive controls because the relevant base C blobs are byte-identical to the previously reproduced base. The current Semaphore check is failing, and its external job log is not available to me, so this focused result does not claim overall CI or merge readiness.

espressolee (commented):
Reviewed exact PR head `ca340b81627f23ad5a4c3deebbb854677a630f2d`
against exact base `143a219bf69cb0c310db7c0d37190e654da71515`.

I rebuilt both trees independently on macOS/arm64 with CPython 3.14.6t,
`sys._is_gil_enabled() == False`, and librdkafka 2.15.0.

Focused #2319 matrix:

- base, constant-length `topics[i]` replacement: **5/5 SIGSEGV**
- base, shrink/regrow stale-count race: **5/5 SIGSEGV**
- base, same replacement with the GIL enabled: **5/5 clean**
- head, constant-length replacement: **10/10 clean**
- head, shrink/regrow: **10/10 clean**
- head, no-mutator / decoy-list / GIL controls: **5/5 clean each**

The PR's broker-free memory-safety file passed twice on the exact head:
**21/21 in 119.62 s** and **21/21 in 120.39 s**. `tests/test_Admin.py`
passed 33/33 and `tests/test_Consumer.py` passed 21/21 with the GIL disabled.

I also checked the current diff and cleanup paths for the shallow snapshots
covering the #2319 list-count and borrowed-element lifetime windows. I found no
blocking issue in that measured scope.

Limits: I did not run broker-backed integration tests, other platforms or
allocators, or the full project suite. The free-threaded snapshot allocation
cost is unmeasured. Current SonarQube and Semaphore statuses are still failing,
so this review does **not** claim overall CI or merge readiness.

espressolee:
Retested exact head `00d97b063b7f2ee9ff0ced5cf4e513414a135f16` against exact base `2dd1397525fef9c02ec4bffabed503b5c5b024e3` on macOS arm64 / CPython 3.14.0rc1t / librdkafka 2.15.0.

Issue #2319 matrix:

| revision | constant-length replacement | shrink/regrow stale count |
| --- | ---: | ---: |
| base `2dd1397` | 9/10 SIGSEGV, 1 clean | 10/10 SIGSEGV |
| candidate `00d97b0` | 10/10 clean | 10/10 clean |

Controls were clean: base GIL-on 5/5; candidate no-mutator 5/5, decoy-list mutation 5/5, and GIL-on 5/5. There were no timeouts or unclassified non-signal exits.

I also exercised the changes added after the previously tested `a9994be` head:

- incremental-config mutation: 10/10 clean with the GIL disabled and 10/10 with it enabled;
- Producer/Admin close-versus-`len()` stress: 250 total close transactions with eight readers, all clean;
- upstream suites: memory-safety 21/21, Admin 33/33, Consumer 21/21, Producer 49/49, and the two close-race files 22/22.

The installed extension kept the GIL disabled, and the exact worktree was clean with `git diff --check` passing. I did not find a new blocker in the `a9994be..00d97b0` delta.

This is scoped local evidence for the named list-lifetime, incremental-config, and Producer/Admin close/len paths. It does not supersede the separate SonarQube gate, which remains at 12.0% new-code coverage versus the required 80%, and it is not a broker-backed or cross-platform claim.

src/confluent_kafka/src/Admin.c
```diff
@@ -1508,6 +1666,9 @@ Admin_create_acls(Handle *self, PyObject *args, PyObject *kwargs) {
                 rd_kafka_AclBinding_destroy_array(c_objs, i);
                 free(c_objs);
         }
+#ifdef Py_GIL_DISABLED
```
k-raina, line 1672: Cleanup is added to error path but missed in success path. This might cause memory leak. 

**Application Impact:**
A long-running service on a free-threaded build that manages ACLs in a loop  e.g. a multi-tenant platform granting/revoking topic permissions as tenants come and go, or a reconciliation loop that periodically re-applies desired ACLs. 

Process's memory grows without bound, over days it gets **OOM-killed**

tests/concurrency/test_memory_safety.py
```diff
@@ -0,0 +1,517 @@
...
+def test_incremental_alter_configs_races_list_mutation():
+    """AdminClient.incremental_alter_configs() -- borrows each ConfigResource +
+    nested incremental_configs (Admin.c:1153)."""
+    _admin_list_crash(lambda a, s: a.incremental_alter_configs(s), _incremental_config_resources)
+
+
+@subprocess_isolated
+def test_create_acls_races_list_mutation():
+    """AdminClient.create_acls() -- borrows each AclBinding (Admin.c:1463)."""
+    _admin_list_crash(lambda a, s: a.create_acls(s), _acl_bindings)
```
k-raina, line 364: In these tests we are checking "did the process die?". Maybe we should be checking ""did it clean up"

tests/test_Consumer.py
```diff
@@ -502,7 +502,7 @@ def error_cb_that_raises(error):
 
     # Test error_cb callback
     with pytest.raises(RuntimeError) as exc_info:
-        consumer1.consume(timeout=0.1)
+        consumer1.consume(timeout=0.5)
```
k-raina: Timeout based assertion might be flaky, should we poll for the "callback fired" flag in a bounded loop instead of sleeping a fixed time?

k-raina (commented):
Thanks for PR @ojasvajain . Thanks @espressolee for validations
Couple of minor comments on memory leak and test hardening.

src/confluent_kafka/src/Admin.c
ojasvajain: Fixed here and in delete_acl as well.

ojasvajain: Added some ref counting test cases for this

tests/test_Consumer.py
ojasvajain: Yes, fixed it.

tests/concurrency/test_memory_safety.py
ojasvajain: By "did it clean up", do you mean that checking if ref counts are balanced? If yes, I have added some tests for the create and delete acls. We can add it for the rest of the methods as well but it is a big list so I have added a TODO for it and add them in a later PR.

src/confluent_kafka/src/Admin.c
```diff
@@ -314,6 +314,11 @@ static int Admin_set_replica_assignment(const char *forApi,
                                         int max_count,
                                         const char *err_count_desc) {
         int pi;
+        int ret = 0;
+#ifdef Py_GIL_DISABLED
```
k-raina, line 318: There are 80+ calls to "def Py_GIL_DISABLED" . Could we use  macros to reduce this or any better way to clean up?

k-raina (approved):
LGTM! Overall, with one comment on improving code quality

src/confluent_kafka/src/Admin.c
ojasvajain: The reason they are added is that we want to create a copy only on GIL free builds because there is no GIL to serialize. We could remove all `ifdef`s but that would mean that GIL based builds will also create a copy of the list, which is not wrong in any way, but it is an extra thing that GIL based builds will do even though they don't need to. Since most of changes are in Admin clients, I don't think performance might not be a factor so  I can discuss this with Emanuele and reach a conclusion.

PR #2335, opened 2026-08-27
[NOGIL] Exclude Avro tests from running on free threaded python builds

ojasvajain:
`fastavro` provides a wheel for Python 3.14t but does not declare itself to be safe without GIL (https://github.com/fastavro/fastavro/issues/873). So on importing fastavro, the GIL gets re-enabled. We need to exclude the test cases that use this dependency from running on free threaded Python in the same way as we did for `rules` related test cases.

What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


rayokota (approved):
Thanks @ojasvajain , LGTM

PR #2344, opened 2026-09-03
Invoke isort as a binary, not via python -m, to support isort>=9.0.0

ojasvajain:
isort released a new major version a few days back which broke our style checks as we do not pin an isort version. In CI, we invoked isort using `python -m` which is no more supported. This PR now invokes isort using the command "isort" instead.

What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


Ankith-Confluent (approved):
LTGM! @ojasvajain

PR #2350, opened 2026-09-07
[NOGIL] Make Message class methods thread safe

ojasvajain:
- Message class exposes getters and setters to five fields. On free threaded Python builds, calling setters and getters concurrently can cause use-after-free conditions and incorrect reference counts. 

- This PR fixes this by acquiring a lock around code snippets where we want only one thread to run at a time (critical sections). The getters and setters have been refactored as part of this process. In Message_headers, an unrelated bug related to error not being handled is also fixed.

- For the five fields that have getters and setters, we have introduced five lock fields within the Message struct.

- Also contains a minor fix in Consumer_consume method.

----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/confluent_kafka/src/Consumer.c
+++ b/src/confluent_kafka/src/Consumer.c
@@ -1300,7 +1300,7 @@ Consumer_consume(Handle *self, PyObject *args, PyObject *kwargs) {
                               NULL};
         rd_kafka_message_t **rkmessages;
         PyObject *msglist;
-        rd_kafka_queue_t *rkqu = self->u.Consumer.rkqu;
+        rd_kafka_queue_t *rkqu;
         CallState cs;
         Py_ssize_t i, n;
 
@@ -1317,6 +1317,8 @@ Consumer_consume(Handle *self, PyObject *args, PyObject *kwargs) {
                 Handle_serialize_exit(self);
                 return NULL;
         }
+
+        rkqu = self->u.Consumer.rkqu;
 
         if (num_messages > 1000000) {
                 PyErr_SetString(
--- a/src/confluent_kafka/src/confluent_kafka.c
+++ b/src/confluent_kafka/src/confluent_kafka.c
@@ -477,40 +477,69 @@ static void cfl_PyErr_Fatal(rd_kafka_resp_err_t err, const char *reason) {
 
 
 /**
+ * @brief Return a new reference to one of the Message's PyObject fields,
+ *        or None if it is unset.
+ *
+ *        The pointer is read and INCREF'd under that field's own lock so
+ *        that a concurrent set_*() of the same field on another thread
+ *        (free-threaded builds) cannot drop the last reference between our
+ *        read and our INCREF. No-op on GIL builds.
+ */
+static PyObject *
+Message_get_field(PyObject **field, cfl_lock_t *lock) {
+        PyObject *obj;
+
+        cfl_lock(lock);
+        obj = *field;
+        Py_XINCREF(obj);
+        cfl_unlock(lock);
+
+        if (!obj)
+                Py_RETURN_NONE;
+        return obj;
+}
+
+/**
+ * @brief Replace one of the Message's PyObject fields with @p new_obj.
+ *
+ *        The swap happens under that field's own lock so it is atomic with
+ *        respect to Message_get_field() and other setters of the same
+ *        field; the old value is released only after unlocking, since a
+ *        DECREF can run arbitrary Python code (__del__).
+ */
+static PyObject *
+Message_set_field(PyObject **field, cfl_lock_t *lock, PyObject *new_obj) {
+        PyObject *old;
+
+        Py_INCREF(new_obj);
+        cfl_lock(lock);
+        old    = *field;
+        *field = new_obj;
+        cfl_unlock(lock);
+        Py_XDECREF(old);
+
+        Py_RETURN_NONE;
+}
+
+/**
  * @returns a Message's error object, if any, else None.
  * @remark The error object refcount is increased by this function.
  */
 PyObject *Message_error(Message *self, PyObject *ignore) {
-        if (self->error) {
-                Py_INCREF(self->error);
-                return self->error;
-        } else
-                Py_RETURN_NONE;
+        return Message_get_field(&self->error, &self->error_lock);
 }
 
 static PyObject *Message_value(Message *self, PyObject *ignore) {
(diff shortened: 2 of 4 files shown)


src/confluent_kafka/src/confluent_kafka.c
k-raina, line 1: Why are we not checking lock for  "==" , "len()" etc?

tests/concurrency/test_message_thread_safety.py
k-raina, line 1: Can we expand more on test coverage, considering its change to message class? One thing i can thing of is memory leak test by counting refcount before and after?

k-raina (commented):
Thanks for PR! Provided couple of comments

src/confluent_kafka/src/confluent_kafka.c
ojasvajain: This got missed. I have added the checks.

tests/concurrency/test_message_thread_safety.py
ojasvajain: Added a few test cases related to reference counts.

src/confluent_kafka/src/Consumer.c
k-raina: What happens when The standard consume loop,  read a message, then commit it. While another thread touches that same message object?

k-raina: Same.  What happens when manually storing an offset for a message while another thread touches it?

src/confluent_kafka/src/confluent_kafka.h
```diff
@@ -700,6 +715,14 @@ typedef struct {
         int16_t delivery_count; /**< Share consumer: number of times this
                                  *   record has been delivered. 1 on first
                                  *   delivery. -1 if unavailable. */
+
+        /* One lock per mutable PyObject field above: serializes concurrent
+         * get/set of the same field on a shared Message (free-threaded). */
+        cfl_lock_t topic_lock;
```
k-raina: Do we need 5 locks accross message file? Is there scope to reduce number of locks?

k-raina (commented):
Thanks for PR @ojasvajain 
Left couple of comments on correctness

Also. Are we making share consumer free threading safe?

src/confluent_kafka/src/confluent_kafka.h
ojasvajain: We could have one global lock for the entire object but that would cause more contention. Let's say, if one thread wants to update topic and another thread wants to update key, they would have to wait for each other. Having 5 locks allows such operations to run concurrently.

src/confluent_kafka/src/Consumer.c
ojasvajain: It is very rare for an application code to be modifying messages but I can see that here message fields are being accessed without the getters, which gets flagged in TSAN runs. I have modified it to use the getters.

ojasvajain: Fixed this too.

ojasvajain:
> Also. Are we making share consumer free threading safe?

No it is out of scope.

src/confluent_kafka/src/Consumer.c
```diff
@@ -569,10 +569,12 @@ Consumer_commit(Handle *self, PyObject *args, PyObject *kwargs) {
...
                 }
 
+                PyObject *error;
+                PyObject *topic;
                 m = (Message *)msg;
-
-                if (m->error && m->error != Py_None) {
-                        PyObject *error = Message_error(m, NULL);
+                error = Message_error(m, NULL);
+                topic = Message_topic(m, NULL);
```
k-raina: Is there a memory leak due to topic reference not cleared in `f (error != Py_None) {` path?

ojasvajain: My bad, fixed.

src/confluent_kafka/src/confluent_kafka.h
k-raina: Is it possible that in case of  `__reduce__`, pickling a message at the same moment another thread calls set_value() on it causes race condition. `__reduce__` reads topic, key, value, headers, error one field at a time where each read is individually safe, but a setter can slip in between two of the reads. 

Can we check reduce for case of 5 locks? Will keeping 5 locks cause similar maintainability issues ?

k-raina (commented):
Comment to evaluate current design which keeps 5 different locks for message object

src/confluent_kafka/src/confluent_kafka.h
ojasvajain: Thought more about it - I have changed it to using 1 lock instead of 5. It causes more contention but we only acquire lock for a very short duration so it should be fine.

ojasvajain: > Is it possible that in case of __reduce__, pickling a message at the same moment another thread calls set_value() on it causes race condition.

Yes, I fixed it. I am now taking a local copy of the Message and using it for pickling.

src/confluent_kafka/src/Consumer.c
```diff
@@ -569,10 +569,11 @@ Consumer_commit(Handle *self, PyObject *args, PyObject *kwargs) {
...
+                PyObject *error;
+                PyObject *topic;
                 m = (Message *)msg;
-
-                if (m->error && m->error != Py_None) {
-                        PyObject *error = Message_error(m, NULL);
+                error = Message_error(m, NULL);
+                if (error != Py_None) {
                         PyObject *errstr =
                             PyObject_CallMethod(error, "str", NULL);
```
k-raina, line 578: If App does msg.set_error("string") then calls consumer.commit(message=msg), commit will be calling .str() on error object. Will app crash ?

k-raina (commented):
Overall LGTM! One comment

src/confluent_kafka/src/Consumer.c
ojasvajain: Yes app will crash. It is a pre-existing issue though so not considering it in the scope of this PR.

k-raina: Maybe lets add TODO for fix?

ojasvajain: I am raising another PR for adding CHANGELOG. I can add as part of that PR.

k-raina (approved):
Thanks for PR @ojasvajain .

PR #2360, opened 2026-09-16
Add s390x wheels to the build and verification matrix

Devarsh010:
What
----
Adds `linux-s390x` (IBM Z) to the Python wheel build and verification matrix, at
parity with the existing Linux arm64/x64 lanes, so `pip install confluent-kafka`
works on s390x with a prebuilt manylinux wheel.

- `tools/wheels/install-librdkafka.sh`: on s390x, use the plain `librdkafka.so`
  (same as arm64). The s390x `librdkafka.redist` runtime ships only the plain
  build, not a `centos8-` variant.
- `tools/wheels/build-wheels.sh`: set `CIBW_MANYLINUX_S390X_IMAGE="manylinux_2_28"`
  so cibuildwheel builds s390x wheels in the same manylinux image as the other
  Linux arches.
- `.semaphore/semaphore.yml`: add `Wheels: Linux s390x` and
  `Wheel Verification: Linux s390x` blocks (native `s1-ubuntu-24-s390x-4` agent,
  mirroring the arm64 blocks) and wire the verification into the
  `Source Distribution` dependencies.

The Schema Registry serdes are pure Python and already ship in the wheel, so no
serdes code changes are required.

Checklist
------------------
  - New s390x wheels published on release; no API or behavior changes. Existing
    arches are unaffected (the s390x branch/env var/blocks are additive).
  - This is CI/packaging . It is exercised by the existing
    `Wheel Verification` blocks (`tools/test-wheels.sh`), which run the built
    wheel's smoke tests in clean distro containers — the s390x block runs the
    same verification. 

Test & Review
------------
Built the s390x wheel on a native IBM Z host via this exact path
(`export ARCH=s390x` → `tools/wheels/build-wheels.sh` → cibuildwheel →
`confluent_kafka-<ver>-cp312-cp312-manylinux_2_28_s390x.whl`) and verified:

- Portability: installs + imports + loads librdkafka in clean s390x containers
  across the glibc range manylinux_2_28 targets — AlmaLinux 8 (glibc 2.28, the
  floor), RHEL/Rocky 9 (2.34), Ubuntu 24.04 (2.39).
- Functional: produce/consume round-trip against a single-node broker


tools/verify-free-threaded-job.sh
```diff
@@ -5,28 +5,40 @@
...
 
+# s390x: the schema-registry stack's deps (cryptography, via authlib and trivup's
+# jwcrypto) publish no s390x wheels, so install the reduced set and skip the
+# schema-registry tests, as the s390x smoke test does (see
+# requirements/requirements-tests-install-nogil-s390x.txt).
+tests_install_reqs=requirements/requirements-tests-install-nogil.txt
+ignore_sr_tests=
+if [[ $ARCH == s390x ]]; then
+    tests_install_reqs=requirements/requirements-tests-install-nogil-s390x.txt
+    ignore_sr_tests=--ignore=tests/schema_registry
```
ojasvajain: My suggestion would be to mimic what's done for free threading. The exclusions are defined in `tests/schema_registry/conftest.py`. You can also read the arch there and exclude all test cases. Add a warning too.

Once you do that, you can get rid of `ignore_sr_tests`.

PR #2361, opened 2026-09-17
Prefer to use httpx2 on py version >=3.10 for schema-registry

### LINKED ISSUE #2351: Dependence on httpx results in deprecation warnings
Authlib has [recently dropped httpx in favour of httpx2](https://github.com/authlib/authlib/issues/904), resulting in complaints about deprecation warnings on every invocation.

If you're going to depend on authlib, it doesn't make sense to also depend on `httpx` for SchemaRegistry support as it commits your users to being spammed with this warning on every invocation.

As a work-around, users can install httpx2 *as well*, though this isn't ideal if the project doesn't use it directly.

tobiogunbi:
What
----
Fixes #2351 

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


rayokota (approved):
Thanks @tobiogunbi , LGTM

